PR #3, opened 2018-08-29
Changes after moving monitoring app to monitoring namespace

UnderGreen:
+ fix some issues from linter output

Closes https://github.com/gravitational/gravity.e/issues/3748.

### DIFF
--- a/lib/clients/clients.go
+++ b/lib/clients/clients.go
@@ -224,7 +224,7 @@ func (r *ClusterClients) newKubeClient(operator ops.Operator, user ops.UserInfo,
 			transport := t.(*http.Transport)
 			transport.DialContext = func(ctx context.Context, network, address string) (net.Conn, error) {
 				return remoteCluster.Dial(
-					&httplib.Addr{Net: "tcp", Addr: "127.0.0.1:3024"},
+					&httplib.Addr{Net: "tcp", Addr: defaults.RemoteClusterDialAddr},
 					&httplib.Addr{Net: "tcp", Addr: defaults.KubernetesAPIAddress},
 					nil)
 			}
--- a/lib/ops/monitoring/monitoring.go
+++ b/lib/ops/monitoring/monitoring.go
@@ -24,7 +24,7 @@ import (
 	"github.com/gravitational/trace"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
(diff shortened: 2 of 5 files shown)


r0mant:
@UnderGreen @a-palchikov I've updated this PR to determine monitoring namespace before forwarding request to remote cluster to preserve compatibility b/w updated Ops Center and older clusters. I've tested it in the following scenario: install 5.0 Ops Center, install 5.0 cluster via it, upgrade Ops Center to 5.2 -> monitoring tab for the cluster works.

Also addressed Dima's previous comments, so this should be good to go.

lib/clients/clients.go
```diff
@@ -191,7 +196,51 @@ func (r *ClusterClients) newAppsClient(clusterName string) (app.Applications, er
...
+		TLSClientConfig: rest.TLSClientConfig{
+			CAData:   signResponse.CACert,
+			CertData: signResponse.Cert,
+			KeyData:  key,
+		},
+		WrapTransport: func(t http.RoundTripper) http.RoundTripper {
+			transport := t.(*http.Transport)
+			transport.DialContext = func(ctx context.Context, network, address string) (net.Conn, error) {
+				return remoteCluster.Dial(
+					&httplib.Addr{Net: "tcp", Addr: "127.0.0.1:3024"},
```
a-palchikov: nit: Should this be in defines? I believe it's being used [here](https://github.com/gravitational/gravity/blob/sergei/monitoring_namespace/lib/webapi/forwarder.go#L198) and [here](https://github.com/gravitational/gravity/blob/sergei/monitoring_namespace/lib/webapi/forwarder.go#L316) as a literal as well

lib/ops/monitoring/monitoring.go
```diff
@@ -33,3 +42,19 @@ type RetentionPolicy struct {
 	// Duration is the policy duration
 	Duration time.Duration `json:"duration"`
 }
+
+// GetNamespace uses the provided Kubernetes client to determine namespace
+// where monitoring resources reside
+func GetNamespace(client *kubernetes.Clientset) (string, error) {
```
a-palchikov: I would use the interface as minimal as possible (`CoreV1`):
```go
GetNamespace(client kubernetes.CoreV1Interafce) {...}
```

lib/clients/clients.go
r0mant: Yeah, good point. Wanted to make it a default but forgot.

PR #8, opened 2018-08-30
 Do not ignore initial RPC connect error during expand.

r0mant:
Previously, when expand agent was starting and connecting to the cluster, any RPC connect error would be effectively ignored. This would lead to a situation where an expand operation has been created, but RPC agent failed to connect (for various reasons, e.g. time skew) and the progress would be hanging indefinitely on "connected to cluster".

This PR adds an explicit check that initial connection to the cluster can be established successfully and returns an error and cleans up the unstarted operation otherwise.

Updates https://github.com/gravitational/gravity.e/issues/3667.


r0mant:
Gonna close this for the time being, some of this will be obsolete after expand FSM changes.

PR #10, opened 2018-09-10
Kevin/tls cert update

knisbet:
Requires: https://github.com/gravitational/planet/pull/312
Requires: https://github.com/gravitational/monitoring-app/pull/104

Updates: gravitational/gravity.e#3726

Create specific keypair for apiserver client, and a new keypair for polling planet. Update planet clients to use client certificates when contact planet ports.


lib/httplib/client.go
```diff
@@ -105,6 +107,24 @@ func WithDialTimeout(t time.Duration) ClientOption {
 	}
 }
 
+func WithClientCert(cert tls.Certificate) ClientOption {
```
a-palchikov: These are missing Godocs

lib/httplib/client.go
```diff
@@ -118,14 +138,47 @@ func GetClient(insecure bool, options ...ClientOption) *http.Client {
...
-		client = &http.Client{Transport: &http.Transport{}}
+		client = &http.Client{Transport: &http.Transport{TLSClientConfig: &tls.Config{}}}
 	}
 	for _, o := range options {
 		o(client)
 	}
 	return client
 }
 
+func GetPlanetClient(options ...ClientOption) (*http.Client, error) {
```
a-palchikov: Missing Godoc

lib/httplib/client.go
```diff
@@ -118,14 +138,47 @@ func GetClient(insecure bool, options ...ClientOption) *http.Client {
...
+	}
+
+	caFile := state.Secret(stateDir, defaults.RootCertFilename)
+	clientCertFile := state.Secret(stateDir, fmt.Sprint(constants.PlanetRpcKeyPair, ".", utils.CertSuffix))
+	clientKeyFile := state.Secret(stateDir, fmt.Sprint(constants.PlanetRpcKeyPair, ".", utils.KeySuffix))
+
+	// Load the CA of the server
+	ca, err := ioutil.ReadFile(caFile)
+	if err != nil {
+		return nil, trace.Wrap(err)
```
a-palchikov: `trace.ConvertSystemError`

lib/httplib/client.go
```diff
@@ -118,14 +138,47 @@ func GetClient(insecure bool, options ...ClientOption) *http.Client {
...
+	}
+	options = append(options, WithCa(ca))
+
+	// For backwards compatability, only add the client key file if it exists on disk
+	// TODO(knisbet) this fallback can be removed when we no longer suppor upgrades from 5.0
```
a-palchikov: suppor**t**

lib/constants/constants.go
```diff
@@ -246,6 +248,9 @@ const (
 	ETCDKeyPair = "etcd"
 	// OpsCenterKeyPair is a name of key pair for OpsCenter
 	OpsCenterKeyPair = "ops"
+	// PlanetRPCKeyPair is a keypair for planet's RPC client for
+	// satellite monitoring and exchange
+	PlanetRpcKeyPair = "planet-rpc-client"
```
a-palchikov: PlanetRpcKeyPair -> PlanetRPCKeyPair

lib/httplib/client.go
```diff
@@ -105,6 +107,24 @@ func WithDialTimeout(t time.Duration) ClientOption {
...
 }
 
+func WithClientCert(cert tls.Certificate) ClientOption {
+	return func(c *http.Client) {
+		transport := c.Transport.(*http.Transport)
+		transport.TLSClientConfig.Certificates = append(transport.TLSClientConfig.Certificates, cert)
+	}
+}
+
+func WithCa(cert []byte) ClientOption {
```
r0mant: WithCA

lib/constants/constants.go
r0mant: This will also be the cert/key filename, won't it? How about we just call it "planet"? So it'll be planet.cert / planet.key. Or "satellite".

knisbet: Well, the difficulty is, this doesn't really distinguish if this is a client key or a server key if we just call it planet or satellite.

Right now we don't generate these key's separately, so it's not a big deal, but I think something we'll need to tackle at some point is actually setting ClientKeyUsage/ServerKeyUsage based on the key usage, which means this key isn't really the planet server key, it's the client key. This is why I was headed the direction of marking the client key in it's name.

PR #13, opened 2018-09-13
Dmitri/3515 dnsmasq addr

a-palchikov:
This PR adds support for configuring DNS on a different ip:port.
The default DNS configuration is `127.0.0.2:53`.
The upgrade persists the previous DNS configuration (i.e. `127.0.0.1:53`).

DNS configuration is saved into 3 places:
  1. host-local boltdb as the initial source of truth to be able to bootstrap other services
  1. install operation state used during installation
  1. cluster record

Some code shuffles resulted from searching for alternatives and even though some of the sub-solutions were rejected, I left the code movements as improvements.

:warning: Uses a development planet tag (`5.2.9-11102`). Planet should be republished once this is approved. 

Fixes https://github.com/gravitational/gravity.e/issues/3515.
Requires https://github.com/gravitational/satellite/pull/80, https://github.com/gravitational/gravity.e/pull/3739 and https://github.com/gravitational/planet/pull/313.


lib/checks/checks.go
```diff
@@ -846,25 +846,47 @@ func defaultPortChecker(options *validationpb.ValidateOptions) health.Checker {
...
-		monitoring.PortRange{"tcp", 3080, 3080, "teleport Web UI"},
-		monitoring.PortRange{"tcp", 3008, 3011, "internal Telekube services"},
-		monitoring.PortRange{"tcp", 32009, 32009, "telekube OpsCenter control panel"},
-		monitoring.PortRange{"tcp", 7575, 7575, "telekube RPC agent"},
-		monitoring.PortRange{"udp", 53, 53, "internal cluster DNS"},
-		monitoring.PortRange{"udp", vxlanPort, vxlanPort, "overlay network"},
-	)
+
+	var portRanges = []monitoring.PortRange{
+		// FIXME: we don't configure dnsmasq to listen on UDP port 53
```
r0mant: Do we need to? In any case, I would delete the commented line below.

lib/localenv/clusterenv.go
```diff
@@ -54,7 +68,12 @@ type ClusterEnvironment struct {
 
 // NewClusterEnvironment initializes local cluster services and
 // returns a new instance of cluster environment
+// FIXME: this is deprecated. Use localenv.LocalEnvronment.NewClusterEnvironment instead
```
r0mant: Why is this deprecated? Can there be no scenario where you want to create a cluster environment without having a local env?

PR #15, opened 2018-09-15
Make join operation use fsm approach

r0mant:
This PR refactors join operation to use fsm approach similar to install/upgrade.

Here are a few implementation/design points:

* Join operation stores all data in the machine-local bolt backend (I called it "join backend") and operation progress is synced back to the cluster when possible.
* This also means that all phases are executing exclusively on the machine that is joining and in general a running agent is not needed.
* The only case when a remote agent is needed is when joining a 2nd master. In this case the agent is started on the 1st master and is used to take etcd backup and restore it during rollback should anything go wrong.
* There is no separate "checks" phase during join because install and join commands now run local checks during startup.

Here are a few other improvements worth mentioning that I made along the way:

* Tweak `gravity plan` output a bit. It now shows where the phase is supposed to execute if this information is available in the plan.
* Tweak `gravity status` output a bit. It now also displays active operations (may be multiple since we support parallel joins for regular nodes), separates nodes into masters/regulars and shows node profiles.
* Introduce a special directory (which I called "ephemeral"), `/usr/local/share/gravity`, where certain short-lived operation data is stored. This is where rpc secrets, wizard backend and join backend go so we don't clutter working dir anymore. This dir is also removed when gravity is uninstalled via `gravity leave` so we don't leave garbage anymore.

Requires https://github.com/gravitational/planet/pull/315. Closes https://github.com/gravitational/gravity.e/issues/3690.


a-palchikov:
Question about no explicit checks for install/join as these run them locally - what about distributed checks? We have at least two types:
  * port occupancy
  * time sync

lib/expand/phases/wait.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// PostCheck is no-op for this phase
+func (*waitPlanetExecutor) PostCheck(ctx context.Context) error {
+	return nil
+}
+
+// NewWaitK8s returns executor that waits for Kubernetes node to register
+func NewWaitK8s(p fsm.ExecutorParams, operator ops.Operator) (*waitK8sExecutor, error) {
+	client, err := httplib.GetUnprivilegedKubeClient() // Kubernetes should be up by now
```
a-palchikov: You should retry to avoid/minimize the "should" part.

lib/expand/plan_test.go
```diff
@@ -0,0 +1,404 @@
...
+			Progress: &ops.ProgressEntry{
+				SiteDomain:  s.cluster.Domain,
+				OperationID: s.installOpKey.OperationID,
+				State:       ops.ProgressStateCompleted,
+				Completion:  constants.Completed,
+				Created:     time.Now(),
+			},
+		})
+	c.Assert(err, check.IsNil)
+	fmt.Println(s.services.Operator.GetSiteOperations(s.cluster.Key()))
```
a-palchikov: Do you need this for tests? It makes it impossible to turn off verbosity.

lib/install/phases/postsystem.go
```diff
@@ -150,29 +151,28 @@ type nodesExecutor struct {
...
-		}
+	p.Progress.NextStep("Updating node %v with labels and taints", p.Node.Hostname)
+	// find this node's profile
+	profile, err := p.Application.Manifest.NodeProfiles.ByName(p.Node.Role)
+	if err != nil {
+		return trace.Wrap(err, "could not find node profile for %#v", p.Node)
+	}
+	// update the node with labels and taints, try a few times to
+	// account for possible transient errors
+	err = utils.Retry(defaults.RetryInterval, defaults.RetryLessAttempts,
```
a-palchikov: I think it's better to use exponential backoff when communicating with the API server

lib/ops/ops.go
```diff
@@ -1353,6 +1353,16 @@ func (s *Site) IsAWS() bool {
...
 }
 
+// Masters returns a list of master nodes from the cluster's state
+func (s *Site) Masters() (masters []storage.Server) {
+	for _, node := range s.ClusterState.Servers {
+		if node.ClusterRole == string(schema.ServiceRoleMaster) {
+			masters = append(masters, node)
+		}
+	}
+	return
```
a-palchikov:
```go
return masters
```

lib/ops/utils.go
```diff
@@ -248,3 +281,21 @@ func VerifyLicense(packages pack.PackageService, license string) error {
 	}
 	return parsed.Verify(ca.CertPEM)
 }
+
+// GetExpandOperation returns the expand operation from the provided backend
```
a-palchikov: ... returns the first available expand operation ...

lib/storage/plan.go
```diff
@@ -93,6 +93,8 @@ type OperationPhaseData struct {
 	// ExecServer is an optional server the phase is supposed to be executed on.
 	// If unspecified, the Server is used
 	ExecServer *Server `json:"exec_server,omitempty" yaml:"exec_server,omitempty"`
+	// Master is a selected master node the phase needs access to
```
a-palchikov: ... is the selected ...

lib/storage/storage.go
```diff
@@ -1498,6 +1498,11 @@ func (s *Server) KubeNodeID() string {
 	return s.AdvertiseIP
 }
 
+// Master returns true if the server has a master role
+func (s *Server) Master() bool {
```
a-palchikov: IsMaster

lib/utils/exec.go
```diff
@@ -80,6 +80,11 @@ func RunPlanetCommand(ctx context.Context, log log.FieldLogger, args ...string)
 	return RunCommand(ctx, log, args...)
 }
 
+// RunInPlanetCommand executes the command specified with args inside planet container
+func RunInPlanetCommand(ctx context.Context, log log.FieldLogger, args ...string) ([]byte, error) {
```
a-palchikov: But you already introduced `PlanetEnterCommand`?

a-palchikov:
Also, seems like the resize does not work for `devicemapper`:
```
Sep 17 14:03:55 node-1 dockerd[21502]: Error starting daemon: error initializing graphdriver: devicemapper: Non existing device docker-thinpool
```

lib/expand/phases/etcd.go
```diff
@@ -0,0 +1,285 @@
...
+		defaults.StatBin, backupPath)...)
+	if err != nil {
+		return trace.Wrap(err, "failed to check backup file %v: %s", backupPath, out.String())
+	}
+	return nil
+}
+
+func (p *etcdExecutor) stopEtcd(ctx context.Context, agent rpcclient.Client) error {
+	var out bytes.Buffer
+	err := agent.Command(ctx, p.FieldLogger, &out, utils.PlanetEnterCommand(
```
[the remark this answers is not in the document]
r0mant: `RunPlanetCommand` executes the command locally while I only need to construct its arguments since it's run on a remote node via an agent. And I couldn't use `PlanetCommandSlice` because it uses local gravity binary path.

lib/expand/join.go
```diff
@@ -477,19 +658,96 @@ func (p *Peer) Done() <-chan struct{} {
...
+		p.Silent.Println(`Operation was started in manual mode
+Inspect the operation plan using "gravity plan" and execute plan phases manually using "gravity join --phase=<phase-id>"
+After all phases have completed successfully, shutdown this process using Ctrl-C`)
+		return nil
+	}
+	fsm, err := p.getFSM(ctx)
 	if err != nil {
 		return trace.Wrap(err)
 	}
+	go func() {
```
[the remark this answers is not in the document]
r0mant: No, there's a return above inside `if p.Manual` block.

lib/clients/etcd.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/gravitational/gravity/lib/defaults"
+
+	etcd "github.com/coreos/etcd/client"
+	"github.com/coreos/etcd/pkg/transport"
+	"github.com/gravitational/trace"
+)
+
+// EtcdConfig is the etcd client configuration
+type EtcdConfig struct {
```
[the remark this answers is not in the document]
r0mant: No specific reason, I just like `clients.NewEtcd()` / `clients.NewTeleport()` a little better than having a package per client, and also package names such as "etcd" and "teleport" are too generic and would potentially steal good names from the user and/or collide with existing packages.

lib/expand/phases/wait.go
r0mant: There's nothing really to retry here - this just creates the client. The phase execution itself it wrapped into retry.

lib/ops/ops.go
r0mant: I was trying to be fancy :)

lib/utils/exec.go
r0mant: `PlanetEnterCommand` builds arguments for the command. This thing runs it.

lib/expand/phases/etcd.go
```diff
@@ -0,0 +1,285 @@
...
+
+	etcd "github.com/coreos/etcd/client"
+	"github.com/gravitational/trace"
+	"github.com/sirupsen/logrus"
+)
+
+// NewEtcd returns executor that adds a new etcd member to the cluster
+func NewEtcd(p fsm.ExecutorParams, operator ops.Operator, runner fsm.AgentRepository) (*etcdExecutor, error) {
+	// create etcd client that's talking to members running on master nodes
+	var masters []storage.Server
```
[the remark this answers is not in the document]
r0mant: Well, the operation plan contains a list of all cluster servers so it should not be a problem for executor to gather this list itself.

r0mant:
@a-palchikov About prechecks: not sure if you misunderstood my comment, or I misunderstood yours :) Prechecks are of course still a phase of the install operation plan, I haven't touched this part. But they are not a phase of the join operation plan because I thought we can only run local checks before join? I.e. we can't test ports/bandwidth b/w nodes during join. Are there any non-local checks that can be run during join? Only time skew maybe. In any case, I would address this as a separate improvement after this gets merged.

a-palchikov:
I think it is reasonable to run time test so that a node with an out of sync clock will not break during etcd setup. Reachability checks are likewise reasonable - no?

PR #18, opened 2018-09-17
Security: tiller listening on localhost only

knisbet:
Harden tiller to only listen on localhost as per: https://engineering.bitnami.com/articles/helm-security.html


klizhentas:
we've been told that we are shipping super old version of tiller and everyone upgrades it anyways, so we should consider upgrading to some latest stable version at least in master.

knisbet:
I think we can update helm/tiller version as a separate PR/Issue

PR #23, opened 2018-09-19
Dmitri/3733 switch storage driver

a-palchikov:
Add ability to switch Docker storage driver during update. The driver is a command line option to `gravity upgrade` atm (i.e. no UI changes were introduced):

```
# given a cluster with devicemapper as storage driver, change it to overlay2:
./gravity upgrade --docker-storage-driver=overlay2
```

Changing the storage driver for Docker is straightforward - the Docker daemon's command line is updated with the new driver. This also means that all containers will be lost and need to be repopulated when kubernetes starts. This is usually not an issue - even if it is, the containers with ephemeral storage can be committed and exported prior to update and loaded back after as additional preparation steps (out of scope in this PR).

Requires https://github.com/gravitational/gravity.e/pull/3749.
Fixes https://github.com/gravitational/gravity.e/issues/3733


a-palchikov:
Just occurred to me that I don't support -> `devicemapper` path which would require the specification of device.
Even though the main use-case is to go **from** `devicemapper` to `overlay` (or any other supported driver), I think the other way should be supported too.

r0mant:
@a-palchikov Have a question just reading the PR description: if you change storage driver in manifest, will it be detected too? Or is only upgrade command flag supported?

klizhentas:
@r0mant question is important one - it **should** be supported. Also, is this flag a new flag or it always existed?

lib/checks/checks.go
```diff
@@ -157,29 +168,45 @@ func ValidateLocal(req LocalChecksRequest) (*LocalChecksResult, error) {
...
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
+
+	failedProbes, err := ValidateManifest(req.Manifest, *profile, stateDir)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	failedProbes, err = ValidateDocker(req.Manifest.Docker(*profile), req.Docker, stateDir)
```
r0mant: This will override `failedProbes` above - don't you need to append them instead?

lib/ops/utils.go
```diff
@@ -300,3 +320,48 @@ func GetExpandOperation(backend storage.Backend) (*storage.SiteOperation, error)
...
+	if installDocker.StorageDriver != "" {
+		config.StorageDriver = installDocker.StorageDriver
+	}
+	config.Args = append(config.Args, installDocker.Args...)
+
+	updateOperation, err := GetLastCompletedUpdateOperation(clusterKey, operator)
+	if err != nil && !trace.IsNotFound(err) {
+		return nil, trace.Wrap(err)
+	}
+	if updateOperation != nil &&
```
r0mant: Hm, I wonder if that's a reliable approach - to dynamically construct "docker" config by querying info from past operations every time. As opposed to, say, saving up-to-date info in ClusterState or something.

tool/gravity/cli/update.go
```diff
@@ -173,22 +173,42 @@ will be returned to the "active" state.
...
+		return trace.Wrap(err)
 	}
+
+	storageDriver := manifestDocker.StorageDriver
+	if docker.StorageDriver != "" {
+		storageDriver = docker.StorageDriver
+	}
+
+	if docker.StorageDriver != existingDocker.StorageDriver &&
+		!utils.StringInSlice(constants.DockerSupportedTargetDrivers, storageDriver) {
```
r0mant: Why is this check performed in CLI only? Don't we have something similar to "canUpdate" on the server side?

lib/ops/utils.go
a-palchikov: Yes, I already changed this. Bigger reason to keep the state is that after the update the Docker configuration of the previous application is replaced (as it is part of the application manifest) and it will not be possible to deduce the current Docker config otherwise.

lib/checks/checks.go
a-palchikov: This is gone as well.

a-palchikov:
I removed the command line option to simplify the implementation (which also led to discovering new corner cases), so the update will only use the value from the manifest.

tool/gravity/cli/update.go
```diff
@@ -73,21 +85,12 @@ func updateTrigger(
...
-		return trace.BadParameter("this operation can only be executed on one of the master nodes")
-	}
-
 	app, err := checkForUpdate(localEnv, operator, cluster, appPackage)
 	if err != nil {
 		return trace.Wrap(err)
 	}
 
-	err = checkCanUpdate(*cluster)
+	err = checkCanUpdate(*cluster, operator, app.Manifest)
```
r0mant: I wonder - if you change storage driver, how does operation creation below work? Won't `operator.CreateSiteAppUpdateOperation` fail because the server still does not allow switching drivers?

tool/gravity/cli/update.go
```diff
@@ -181,22 +183,51 @@ will be returned to the "active" state.
...
+	}
+
+	docker := manifest.SystemOptions.DockerConfig()
+	if docker == nil {
+		// No changes
+		return nil
 	}
+
+	existingDocker := cluster.ClusterState.Docker
+	if existingDocker.IsEmpty() {
```
r0mant: Should not similar check also be added on the server side? In `checkUpdateParameters`. That's where the check that didn't allow switching drivers used to be.

a-palchikov:
There's actually - there're tests that install with `devicemapper` and then upgrade which switches to `overlay2`. What's missing is the validation that the storage driver has been actually changed.

PR #25, opened 2018-09-19
Better handle initial RPC connect error during join.

r0mant:
This is the reiteration of https://github.com/gravitational/gravity/pull/8 which I postponed until after expand refactoring. Includes addressed review comments from that PR, where applicable.

To address another comment from that PR: I think it makes sense to delete _unstarted_ expand operation in  case of CLI join instead of failing it, because from user's (and ours too, actually) perspective it hasn't started yet. This is also the same behavior as it used to be in case operation failed to start.

Closes https://github.com/gravitational/gravity.e/issues/3667.

### DIFF
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -502,7 +502,7 @@ func (p *Peer) getAgent(opCtx operationContext) (*rpcserver.PeerServer, error) {
 	// make sure that connection to the RPC server can be established
 	ctx, cancel := context.WithTimeout(p.Context, defaults.PeerConnectTimeout)
 	defer cancel()
-	err = agent.Check(ctx)
+	err = agent.ValidateConnection(ctx)
 	if err != nil {
 		return agent, trace.Wrap(err)
 	}
--- a/lib/rpc/server/peer.go
+++ b/lib/rpc/server/peer.go
@@ -173,8 +173,8 @@ func (r *PeerServer) Serve() error {
 	return r.agentServer.Serve()
 }
 
-// Check makes sure that connection to the control server can be established
-func (r *PeerServer) Check(ctx context.Context) error {
+// ValidateConnection makes sure that connection to the control server can be established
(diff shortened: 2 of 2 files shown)


lib/expand/join.go
```diff
@@ -469,66 +469,107 @@ func (p *Peer) agentURL(ctx operationContext) (string, error) {
...
+		if err := p.Stop(stopCtx); err != nil {
+			p.Errorf("Failed to stop peer: %v.", trace.DebugReport(err))
+		}
+		// in case of join via CLI the operation has already been created
+		// above but the agent failed to connect so we're deleting the
+		// operation because from user's perspective it hasn't started
+		//
+		// in case of join via UI the peer is joining to the existing
+		// operation created via UI so we're not touching it and the
+		// user can cancel it in the UI
```
a-palchikov: This means inconsistent handling:
  - if started (or created?) in UI, user is responsible for cleaning up the operation (cancelling, which I suppose, will mark it as failed)
  - if created in CLI flow, the operation vanishes completely even though it was created in the db and there was a failure to progress it

whereas if it were always moved to a failed state (for UI flow as well) uniformly, you need not bother with this special handling.

lib/rpc/server/peer.go
```diff
@@ -168,13 +168,18 @@ type WatchEvent struct {
...
 
 // Serve starts this peer
-func (r *peerServer) Serve() error {
+func (r *PeerServer) Serve() error {
 	r.peers.start()
 	return r.agentServer.Serve()
 }
 
+// Check makes sure that connection to the control server can be established
+func (r *PeerServer) Check(ctx context.Context) error {
```
a-palchikov: `ValidateConnection`?

lib/expand/join.go
r0mant: On the contrary, it is consistent with the user expectations. If they created the operation in the UI they likely don't want it to suddenly disappear or fail if the agent failed to connect.

PR #26, opened 2018-09-19
Write system log during upgrade.

r0mant:
This PR makes sure that `telekube-system` log file is written during automatic upgrade on the node where master agent is running. Also includes the following changes/improvements:

* Capture update hook output and write it to the system log as well, similar to what install does.
* Move default location of our log files to `/var/log` instead of the "current working dir". This gives us a stable well-known location where these files can be found, as opposed to currently having to search for them everywhere because the working dir may be completely different depending on how the command was run (e.g. if it's run via an agent, then the file will end up in / and so on).
* Do not truncate telekube-system log file every time. I think it doesn't make sense to truncate it since it oftentimes useful to look at the past logs.

Closes https://github.com/gravitational/gravity.e/issues/3596.

### DIFF
--- a/lib/defaults/defaults.go
+++ b/lib/defaults/defaults.go
@@ -210,11 +210,14 @@ const (
 	// GravityPackagePrefix defines base prefix of gravity package
 	GravityPackagePrefix = "gravitational.io/gravity"
 
-	// TelekubeSystemLog defines the default location for the system log
-	TelekubeSystemLog = "/var/log/telekube-system.log"
+	// TelekubeSystemLogFile is the system log file name
+	TelekubeSystemLogFile = "telekube-system.log"
 
-	// TelekubeUserLog the default location for user-facing log file
-	TelekubeUserLog = "/var/log/telekube-install.log"
+	// TelekubeUserLogFile is the user log file name
+	TelekubeUserLogFile = "telekube-install.log"
+
+	// SystemLogDir is the directory where gravity logs go
+	SystemLogDir = "/var/log"
 
 	// TelekubePackage is the Telekube application package name
 	TelekubePackage = "telekube"
@@ -1023,6 +1026,12 @@ var (
 	// CertificateExpiry is the validity period of certificates generated
 	// during cluster installation (such as apiserver, etcd, kubelet, etc.)
 	CertificateExpiry time.Duration = 10 * 365 * 24 * time.Hour // 10 years
+
+	// TelekubeSystemLog defines the default location for the system log
+	TelekubeSystemLog = filepath.Join(SystemLogDir, TelekubeSystemLogFile)
+
+	// TelekubeUserLog the default location for user-facing log file
+	TelekubeUserLog = filepath.Join(SystemLogDir, TelekubeUserLogFile)
 )
 
 // HookSecurityContext returns default securityContext for hook pods
--- a/lib/defaults/defaults.go
+++ b/lib/defaults/defaults.go
@@ -211,10 +211,10 @@ const (
 	GravityPackagePrefix = "gravitational.io/gravity"
 
 	// TelekubeSystemLog defines the default location for the system log
-	TelekubeSystemLog = "./telekube-system.log"
+	TelekubeSystemLog = "/var/log/telekube-system.log"
 
 	// TelekubeUserLog the default location for user-facing log file
-	TelekubeUserLog = "./telekube-install.log"
+	TelekubeUserLog = "/var/log/telekube-install.log"
 
 	// TelekubePackage is the Telekube application package name
 	TelekubePackage = "telekube"
--- a/lib/update/phase_app.go
+++ b/lib/update/phase_app.go
@@ -193,24 +193,28 @@ func (p *phaseApp) runHooks(ctx context.Context, hooks ...schema.HookType) error
 			return trace.Wrap(err)
 		}
 		reader, writer := io.Pipe()
-		go func() {
-			defer reader.Close()
-			scanner := bufio.NewScanner(reader)
-			for scanner.Scan() {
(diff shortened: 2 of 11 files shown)


lib/update/phase_app.go
```diff
@@ -190,11 +192,25 @@ func (p *phaseApp) runHooks(ctx context.Context, hooks ...schema.HookType) error
...
+			for scanner.Scan() {
+				log.WithField(trace.Component, "hook").Info(scanner.Text())
+			}
+			if err := scanner.Err(); err != nil {
+				log.Warnf("Failed to stream hook logs: %v.", err)
+			}
+		}()
+		_, err = app.StreamAppHook(ctx, p.Apps, req, writer)
+		if err != nil {
+			return trace.Wrap(err, "%v %s hook failed", p.Package, hook)
```
a-palchikov: `writer.Close` missing

lib/update/phase_app.go
```diff
@@ -190,11 +192,25 @@ func (p *phaseApp) runHooks(ctx context.Context, hooks ...schema.HookType) error
 			}
 			return trace.Wrap(err)
 		}
-		_, out, err := app.RunAppHook(ctx, p.Apps, req)
+		reader, writer := io.Pipe()
+		go func() {
+			defer reader.Close()
+			scanner := bufio.NewScanner(reader)
+			for scanner.Scan() {
+				log.WithField(trace.Component, "hook").Info(scanner.Text())
```
a-palchikov: I think the streaming part should be factored out and use the name of the hook in the log at the very least:
```go
func (p *phaseApp) runHooks(ctx context.Context, hooks ...schema.HookType) error {
	// ...
	reader, writer := io.Pipe()
	go streamHook(hook, reader)
	// ...
}

func streamHook(hook string, rd io.ReadCloser) error {
	defer rc.Close()
	scanner := bufio.NewScanner(rc)
	log := log.WithFields(log.Fields{trace.Component: "hook", "hook": hook})
	for scanner.Scan() {
		log.Info(scanner.Text())
	}
	if err := scanner.Err(); err != nil {
		return trace.Wrap(err)
	}
	return nil
}
```

tool/gravity/cli/update.go
```diff
@@ -51,7 +52,14 @@ func updateTrigger(
 	localEnv *localenv.LocalEnvironment,
 	upgradeEnv *localenv.LocalEnvironment,
 	appPackage string,
-	manual bool) (err error) {
+	manual bool,
+	systemLogFile string,
+) error {
+	err := install.InitLogging(systemLogFile)
```
a-palchikov: This could be done outside - prior to calling the function. Same for `rpcAgentRun`

knisbet:
By moving the log, I'm curious about the implications for existing users. Have you considered symlinking where logs from their existing location to the /var/log directory. Or dropping a local file, that indicates the new location? It's just one of those small things, where I think lots of our users have muscle memory about where to find these logs.

Also, related to upgrades, I think there is a second issue that may not have been captured in gravitational/gravity.e#3596 or was captured somewhere else. The gravity agent that is coordinating the upgrade may be on a different node then where it was triggered. So I thought part of this issue was to try and address that, by ensuring the rpcAgent coordinating the upgrade would be on the same node that triggered it, so a complete log would be available. Just wanted to make sure that component of this problem didn't get lost somewhere.

lib/update/phase_app.go
r0mant: I actually did that initially (added "hook" field) but it was always hook=update so I doubted its usefulness and removed. I can bring it back though.

lib/defaults/defaults.go
```diff
@@ -211,10 +211,10 @@ const (
 	GravityPackagePrefix = "gravitational.io/gravity"
 
 	// TelekubeSystemLog defines the default location for the system log
-	TelekubeSystemLog = "./telekube-system.log"
+	TelekubeSystemLog = "/var/log/telekube-system.log"
```
klizhentas: why this change?

r0mant: I explained it in the PR description.

> This gives us a stable well-known location where these files can be found, as opposed to currently having to search for them everywhere because the working dir may be completely different depending on how the command was run (e.g. if it's run via an agent, then the file will end up in / and so on).

klizhentas: ok, makes sense

r0mant:
@knisbet Honestly, I'd be surprised if anyone had muscle memory of finding this log file because I can't find it myself a lot of times :) I guess I could still write to /var/log and do a symlink from cwd to it. I think it was an unfortunate design decision to put it in cwd to start with. We end up with garbage files laying around which even gravity leave can't cleanup, which was also part of the reason I did this (I already moved rpcsecrets and other .db files we used to create in cwd elsewhere).

I will see if it's easy to make sure master agent is started on the same node and if so will implement. Otherwise, will make a ticket to address separately.

knisbet:
I agree /var/log is a better location for it, it's just I think from what I see at customers, is they do go to this log when an installation fails. So, all I'm getting at is moving it while better, we just need to be careful we're not making confusing changes.

tool/gravity/cli/update.go
```diff
@@ -132,6 +125,13 @@ func updateTrigger(
 
 	if !manual {
 		req.leaderParams = []string{constants.RpcAgentUpgradeFunction}
+		// attempt to schedule the master agent on this node but do not
+		// treat the failure to do so as critical
+		req.leader, err = findLocalServer(*cluster)
+		if err != nil {
+			log.Errorf("Failed to determine local node: %v.",
```
a-palchikov: `log.Warnf` if it's not critical

r0mant: Will change in the next PR so I don't have to wait for CI again.

PR #28, opened 2018-09-24
Tweak disk space status output.

r0mant:
Requires https://github.com/gravitational/planet/pull/316.

Updates https://github.com/gravitational/gravity.e/issues/3440.


lib/status/status.go
```diff
@@ -357,7 +362,31 @@ func probeErrorDetail(p pb.Probe) string {
 	if p.Detail == "" {
 		detail = p.Checker
 	}
-	return fmt.Sprintf("%v failed", detail)
+	// state directory disk space checker always checks /var/lib/gravity
+	// which is default path inside planet but may be different on host
+	// so determine the real state directory if needed
+	if strings.HasPrefix(p.Checker, monitoring.DiskSpaceCheckerID) {
```
a-palchikov: Please see my comment about embedding path into the checker name and the resulting need to parse it here.
Also, it looks like `probeFromErr` has blown up and turned into something hard to read - can you factor this new code out?
On the other hand, this code looks completely out of place here, inside a simple `probeErrorDetail` which used to extract the detail information from a probe - this should be done somewhere up the call chain.

lib/status/status.go
```diff
@@ -357,7 +362,31 @@ func probeErrorDetail(p pb.Probe) string {
...
+		var data monitoring.HighWatermarkCheckerData
+		err := json.Unmarshal(p.CheckerData, &data)
+		if err != nil {
+			logrus.Error(trace.DebugReport(err))
```
a-palchikov: Meh, the function should return error and let the caller decide

lib/status/status.go
r0mant: It can't be done up the chain because the "Probe" object (along with its details) is not available there anymore. This is the place the object basically gets converted to a string so I think this is a suitable place to construct a proper error message. I can factor this logic out though so it does not look so intimidating.

lib/status/status.go
r0mant: The error would then propagate all the way up to a bunch of methods that also don't return it. I'd rather just log it here, it's only used in this module.

PR #32, opened 2018-09-27
Move cluster to degraded based on planet status.

r0mant:
This PR introduces the following changes:

* Cluster moves to degraded state if any of the nodes in the planet agent report are degraded. I've extended the existing status checker for this (the one that runs app status hook).
* Prohibit new app version upload when cluster is degraded.
* Allow shrinking degraded clusters (to be able to remove offline nodes).

This PR concludes the work started by adding disk space checkers as it makes sure that if high watermark is reached and cluster becomes degraded, new apps can't be uploaded and upgrade can't be started.

Closes https://github.com/gravitational/gravity.e/issues/3440.

Also updates https://github.com/gravitational/gravity.e/issues/3206 because with this change cluster UI now automatically shows red "with issues" state icon when a node becomes unhealthy.

### DIFF
--- a/lib/ops/opsservice/status.go
+++ b/lib/ops/opsservice/status.go
@@ -25,6 +25,7 @@ import (
 	"github.com/gravitational/gravity/lib/status"
 	"github.com/gravitational/gravity/lib/storage"
 
+	"github.com/gravitational/satellite/agent/proto/agentpb"
 	"github.com/gravitational/trace"
 )
 
@@ -45,7 +46,7 @@ func (o *Operator) CheckSiteStatus(key ops.SiteKey) error {
 	}
 
 	statusErr := cluster.checkPlanetStatus(context.TODO())
-	reason := storage.ReasonNodeDegraded
+	reason := storage.ReasonClusterDegraded
 	if statusErr == nil {
 		statusErr = cluster.checkStatusHook(context.TODO())
 		reason = storage.ReasonStatusCheckFailed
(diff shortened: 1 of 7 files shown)


lib/ops/opsservice/status.go
```diff
@@ -22,61 +22,101 @@ import (
...
+}
+
+// checkPlanetStatus checks the cluster health using planet agents
+func (s *site) checkPlanetStatus(ctx context.Context) error {
+	planetStatus, err := status.FromPlanetAgent(ctx, nil)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	for _, node := range planetStatus.Nodes {
+		if node.AdvertiseIP != "" && node.Status != status.NodeHealthy {
```
a-palchikov: That would be very surprising to find a node with en empty `AdvertiseIP` and consider it an `ok` outcome.

lib/ops/opsservice/status.go
```diff
@@ -22,61 +22,101 @@ import (
...
+		s.Debugf("Application %s does not have status hook.", s.app)
+		return nil
+	}
+	ref, out, err := app.RunAppHook(ctx, s.service.cfg.Apps, app.HookRunRequest{
+		Application: s.backendSite.App.Locator(),
+		Hook:        schema.HookStatus,
+		ServiceUser: s.serviceUser(),
+	})
+	if ref != nil {
+		defer func() {
```
a-palchikov: Why `defer` if the hook has already completed?

tool/gravity/cli/ops.go
```diff
@@ -106,6 +107,13 @@ func uploadUpdate(env *localenv.LocalEnvironment, opsURL string) error {
 		return trace.Wrap(err)
 	}
 
+	if cluster.State == ops.SiteStateDegraded {
+		return trace.BadParameter("The cluster is in degraded state so " +
+			"uploading new applications is prohibited. Please check " +
+			"gravity status output and correct the situation before " +
+			"attempting again.")
```
a-palchikov:
```
It is not possible to upload new applications while the cluster is not healthy.
Please make sure the cluster is healthy before reattempting the update.
```
I think `prohibited` is a too strong here.

r0mant: I think "not possible" is actually stronger than "prohibited" (and not entirely true: it is possible but not allowed). Also my wording hints them at the specific action ("check gravity status") rather than just saying "make sure cluster is healthy".

PR #36, opened 2018-10-02
Dmitri/3755 node runtime

a-palchikov:
Move logic of labeling to the update bootstrap phase packages so that `pull-updates` can pull the correct runtime package update as the phase contains the exact information to make the decision.

Also, add an upgrade test on a 6-node cluster for latest 5 LTS.

Fixes https://github.com/gravitational/gravity.e/issues/3755.


lib/update/utils.go
```diff
@@ -96,3 +90,24 @@ func planetNeedsUpdate(profile string, installed, update appservice.Application)
...
+	}
+	// Look for legacy package
+	packageName := loc.LegacyPlanetMaster.Name
+	if clusterRole == schema.ServiceRoleNode {
+		packageName = loc.LegacyPlanetNode.Name
+	}
+	runtimePackage, err = manifest.Dependencies.ByName(packageName)
+	if err != nil {
+		logrus.Warnf("Failed to find the legacy runtime package in manifest: %v.", err)
+		return nil, trace.NotFound("runtime package not found")
```
r0mant: I would make error message a bit more verbose, at least add profile/role information. Same for the log entry above.

lib/update/phase_bootstrap.go
```diff
@@ -253,6 +270,69 @@ func (p *updatePhaseBootstrap) syncPlan() error {
...
+}
+
+func getInstalledRuntime(apps app.Applications, installedApp loc.Locator, serverRole, clusterRole string) (*loc.Locator, error) {
+	installed, err := apps.GetApp(installedApp)
+	if err != nil {
+		return nil, trace.Wrap(err, "failed to query installed application")
+	}
+	installedProfile, err := installed.Manifest.NodeProfiles.ByName(serverRole)
+	if err != nil {
+		return nil, trace.NotFound("failed to find node profile %q", serverRole)
```
r0mant: Why not `trace.Wrap(err, ...)`?

lib/update/automatic.go
```diff
@@ -67,7 +68,8 @@ func AutomaticUpgrade(ctx context.Context, localEnv, updateEnv *localenv.LocalEn
 	force := false
 	fsmErr := fsm.ExecutePlan(ctx, progress, force)
 	if fsmErr != nil {
-		return trace.Wrap(err)
+		log.Warnf("Failed to execute plan: %v.", fsmErr)
```
knisbet: Hmm... I'm a bit confused by the logic of the fallthrough here. 

If I understand this, if Executing the plan has an error of some sort, we keep that error and call fsm.Complete. In fsm.Complete, it checks if the plan has been completed, and if not it tell's the ops.Service to Fail the Operation and mark it as complete.

Won't this basically the leave the cluster in a partially upgraded state, but also having released the cluster lock that prevents other operations? If someone retries the upgrade, will it try and continue the existing upgrade, or start a new plan?

Sorry, not sure if I'm just missing something here, without spending the time to dig through the operation code.

a-palchikov: Yes, you're right. I was thinking about the automatic update observed by the robotest. But it does not make sense to remove the cluster lock for upgrades - I'll have this fixed in the robotest code.

a-palchikov: On the other hand, I think I did not implement the `Complete` operation correctly - in case of error, it should mark only the operation as `failed` and leave the cluster in the state it was during the operation. This will prevent another upgrade operation.

PR #37, opened 2018-10-03
coredns

knisbet:
Note: Do not merge until 5.2 maintenance branch is created.

dns-app is deprecated, as CoreDNS is now embedded within planet. I left the app, so that during upgrades it cleans up the no longer used resources.

RBAC updates, allow CoreDNS on all nodes read only access to monitor service discovery info required for DNS mappings, and the configmap kube-system/coredns for customer supplied configuration. All nodes get a new certificate to the API to be used by CoreDNS to access this info.

During upgrade of an existing cluster, I had to add a phase to partially create the RBAC configuration, because as the cluster upgrades it will begin using CoreDNS before the RBAC phase runs.

Updates https://github.com/gravitational/gravity.e/issues/3742
Updates https://github.com/gravitational/gravity.e/issues/3542


docs/5.x/cluster.md
```diff
@@ -2069,3 +2069,50 @@ With `promiscuous-bridge`, the behavior is similar to that of the kubenet networ
...
+        health_check 0
+      }
+    }
+```
+
+!!! tip "Templates":
+  The CoreDNS configuration in this example is a go template, which will generate configuration based on the node and install time configuration. 
+
+The following template variables are supported:
+* `ListenAddrs` - (Required) The overlay network address to bind CoreDNS to.
```
a-palchikov: The list of overlay network addresses ...

docs/5.x/cluster.md
```diff
@@ -2069,3 +2069,50 @@ With `promiscuous-bridge`, the behavior is similar to that of the kubenet networ
...
+* `Port` - The port to use for DNS binding ([see Standalone Offline CLI Installation](/installation/#standalone-offline-cli-installation))
+* `UpstreamNameservers` - A list of upstream resolvers to use for general DNS reoslution. ([see Standalone Offline CLI Installation](/installation/#standalone-offline-cli-installation))
```
a-palchikov: re**so**lution

lib/fsm/kubernetes.go
```diff
@@ -62,6 +62,34 @@ func GetUpsertBootstrapResourceFunc(client *kubernetes.Clientset) resources.Reso
 				return trace.Wrap(rigging.ConvertError(err))
 			}
 			log.Debugf("Updated ClusterRoleBinding %q.", resource.Name)
+		case *rbacv1.Role:
+			_, err = client.Rbac().Roles(resource.Namespace).Create(resource)
```
a-palchikov: I think we should refrain from deprecated APIs (at least in new code):
```go
_, err = client.RbacV1().Roles(resource.Namespace).Create(resource)
```

lib/update/phase_coredns.go
```diff
@@ -0,0 +1,131 @@
...
+limitations under the License.
+*/
+
+package update
+
+import (
+	"context"
+
+	"github.com/gravitational/gravity/lib/storage"
+	"github.com/gravitational/rigging"
```
a-palchikov: nit: missing empty line to separate local imports from 3rd party

lib/update/phase_coredns.go
```diff
@@ -0,0 +1,131 @@
...
+// HACK: This phase needs to run before updating planet, as coredns is embedded in planet
+// and will come online as planet restarts
+func NewPhaseCoreDNS(c FSMConfig, plan storage.OperationPlan, phase storage.OperationPhase) (*updatePhaseCoreDNS, error) {
+	op, err := newKubernetesOperation(c, plan, phase)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	return &updatePhaseCoreDNS{
+		kubernetesOperation: *op,
+		FieldLogger:         log.NewEntry(log.New()),
```
a-palchikov: I think you should at least inject a phase ID into the logger context.

lib/update/phase_coredns.go
```diff
@@ -0,0 +1,131 @@
...
+	err = rigging.ConvertError(err)
+	if err != nil && !trace.IsAlreadyExists(err) {
+		return trace.Wrap(err)
+	}
+
+	_, err = p.kubernetesOperation.Client.RbacV1().ClusterRoleBindings().Create(&rbacv1.ClusterRoleBinding{
+		ObjectMeta: metav1.ObjectMeta{Name: corednsResourceName},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
```
a-palchikov: We have lib/constants#`KubernetesKindXXX` for that.

lib/update/plan.go
```diff
@@ -342,6 +350,12 @@ func newOperationPlan(p newPlanParams) (*storage.OperationPlan, error) {
 	if len(runtimeUpdates) > 0 {
 		// if there are no runtime updates, then these phases are not needed
 		// as we're not going to update system software
+		if p.updateCoreDNS {
+			corednsPhase := *builder.corednsPhase(leadMaster)
```
a-palchikov: My guess is that this will require changes to plan tests as well.

lib/update/phase_coredns.go
knisbet: I'm going to ask this every time... why do we care how automatically generated code get's generated. I get that someone has written a style guideline for this, I just don't understand why we care if goimports doesn't.

lib/update/phase_coredns.go
knisbet: Not really... there is constants.KubernetesKindUser and nothing else. Rigging has more... but also isn't an exhaustive list (it has KindClusterRole, but not KindUser below).

lib/update/phase_coredns.go
a-palchikov: Hmm, I don't know if that's automatically generated code. I guess you mean the difference between running goimports on a file w/o imports vs running it on a file with imports in some (alternative - i.e. grouped) format?
I personally like the grouping, but it's, well, a personal opinion. Besides, there was a precedent for goimports to adopt a custom grouping with the `-local` flag: https://github.com/golang/go/issues/12420, which will probably end this debate for both sides.
I don't feel too strong about caring whether we continue to group the imports as we do now (manually), so it's a big deal for me.

lib/update/phase_coredns.go
a-palchikov: I meant, we have a place for these constants.

knisbet:
This PR requires the planet changes, so right now it's running the tests against the wrong planet. I'm not worry about the tests, since this PR will be sitting until the 5.2 maintenance branch is created, and I'll likely have to update it then anyway's.

tool/gravity/cli/register.go
```diff
@@ -72,8 +72,8 @@ func RegisterCommands(app *kingpin.Application) *Application {
 	g.InstallCmd.PodCIDR = g.InstallCmd.Flag("pod-network-cidr", "Subnet range for pods. Must be a minimum of /16").Default(defaults.PodSubnet).String()
 	g.InstallCmd.ServiceCIDR = g.InstallCmd.Flag("service-cidr", "Subnet range for services").Default(defaults.ServiceSubnet).String()
 	g.InstallCmd.VxlanPort = g.InstallCmd.Flag("vxlan-port", "Custom overlay network port").Default(strconv.Itoa(defaults.VxlanPort)).Int()
-	g.InstallCmd.DNSListenAddrs = g.InstallCmd.Flag("dns-listen-addr", "Custom listen address for dnsmasq").Default(defaults.DNSListenAddr).IPList()
-	g.InstallCmd.DNSPort = g.InstallCmd.Flag("dns-port", "Custom DNS port for dnsmasq").Default(strconv.Itoa(defaults.DNSPort)).Int()
+	g.InstallCmd.DNSListenAddrs = g.InstallCmd.Flag("dns-listen-addr", "Custom listen address for planet dns.").Default(defaults.DNSListenAddr).IPList()
```
r0mant: dns -> DNS

tool/gravity/cli/register.go
```diff
@@ -72,8 +72,8 @@ func RegisterCommands(app *kingpin.Application) *Application {
...
+	g.InstallCmd.DNSPort = g.InstallCmd.Flag("dns-port", "Custom DNS port for planet dns.").Default(strconv.Itoa(defaults.DNSPort)).Int()
```
r0mant: dns -> DNS

tool/gravity/cli/register.go
r0mant: Also, I don't think "planet" should appear anywhere in the help strings as it doesn't say anything to a user, maybe "for in-cluster DNS" or something like this. Same below.

lib/update/phase_coredns.go
```diff
@@ -0,0 +1,133 @@
...
+	_, err := client.RbacV1().ClusterRoles().Get(corednsResourceName, metav1.GetOptions{})
+	err = rigging.ConvertError(err)
+	if err != nil {
+		if trace.IsNotFound(err) {
+			return true, nil
+		}
+		return false, trace.Wrap(err)
+	}
+
+	_, err = client.RbacV1().ClusterRoleBindings().Get(corednsResourceName, metav1.GetOptions{})
```
r0mant: Is the above check not enough?

lib/update/phase_coredns.go
```diff
@@ -0,0 +1,133 @@
...
+		Rules: []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{""},
+				Verbs:     []string{"list", "watch"},
+				Resources: []string{"endpoints", "services", "namespaces"},
+			},
+		},
+	})
+	err = rigging.ConvertError(err)
+	if err != nil && !trace.IsAlreadyExists(err) {
```
r0mant: For informational/debugging purposes I would log if this resource already exists, otherwise there's 0 feedback in this case.

lib/update/phase_coredns.go
```diff
@@ -0,0 +1,133 @@
...
+	})
+	return &phase
+}
+
+type updatePhaseCoreDNS struct {
+	log.FieldLogger
+	kubernetesOperation
+}
+
+// NewPhaseCoreDNS create an upgrade phase to add coredns rbac permissions
```
r0mant: create**s**

lib/update/phase_coredns.go
```diff
@@ -0,0 +1,133 @@
...
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:     "User",
+				Name:     "coredns",
+				APIGroup: "rbac.authorization.k8s.io",
+			},
+		},
+	})
+	err = rigging.ConvertError(err)
+	if err != nil && !trace.IsAlreadyExists(err) {
```
r0mant: Same as above - can you log the already exists case? And the "created" case.

lib/update/phase_coredns.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+const (
+	corednsResourceName = "gravity:coredns"
+)
+
+func (r phaseBuilder) corednsPhase(leadMaster storage.Server) *phase {
+	phase := root(phase{
+		ID:          "coredns",
+		Description: "Provision coredns resources",
```
r0mant: Let's at least make it properly-cased (is that a word? :smile:) in the user visible string. "coredns" -> "CoreDNS"

lib/update/phase_coredns.go
```diff
@@ -0,0 +1,133 @@
...
+	return &phase
+}
+
+type updatePhaseCoreDNS struct {
+	log.FieldLogger
+	kubernetesOperation
+}
+
+// NewPhaseCoreDNS create an upgrade phase to add coredns rbac permissions
+// HACK: This phase needs to run before updating planet, as coredns is embedded in planet
```
r0mant: "HACK" is not too useful, I would either just get rid of this word, or replace it with a proper TODO (or FIXME) that explains what needs to be done to unhack it.

docs/5.x/cluster.md
```diff
@@ -2069,3 +2069,50 @@ With `promiscuous-bridge`, the behavior is similar to that of the kubenet networ
...
+      }{{end}}
+      forward . {{range $server := .UpstreamNameservers}}{{$server}} {{end}}{
+        policy sequential
+        health_check 0
+      }
+    }
+```
+
+!!! tip "Templates":
+  The CoreDNS configuration in this example is a go template, which will generate configuration based on the node and install time configuration.
```
r0mant: go -> Go

docs/5.x/cluster.md
```diff
@@ -2069,3 +2069,50 @@ With `promiscuous-bridge`, the behavior is similar to that of the kubenet networ
...
 [promiscuous-mode]: https://en.wikipedia.org/wiki/Promiscuous_mode
+
+## Customizing Cluster DNS
+
+Gravity uses [CoreDNS](https://coredns.io) for DNS resolution and service discovery within the cluster. 
+
+The default configuration can be customized, by creating a kubernetes configmap kube-system/coredns.
+
+```
+kind: ConfigMap
```
r0mant: Does this config map represent default configuration we ship, or just some example configuration? It is not clear and I think should be made clear.

docs/5.x/cluster.md
```diff
@@ -2069,3 +2069,50 @@ With `promiscuous-bridge`, the behavior is similar to that of the kubenet networ
...
+      }
+    }
+```
+
+!!! tip "Templates":
+  The CoreDNS configuration in this example is a go template, which will generate configuration based on the node and install time configuration. 
+
+The following template variables are supported:
+* `ListenAddrs` - (Required) The list of overlay network address to bind CoreDNS to.
+* `Port` - The port to use for DNS binding ([see Standalone Offline CLI Installation](/installation/#standalone-offline-cli-installation))
```
r0mant: Dot's missing from this sentence, unlike others here (I know, I'm a dot nerd :) )

docs/5.x/cluster.md
```diff
@@ -2069,3 +2069,50 @@ With `promiscuous-bridge`, the behavior is similar to that of the kubenet networ
...
+* `UpstreamNameservers` - A list of upstream resolvers to use for general DNS resolution. ([see Standalone Offline CLI Installation](/installation/#standalone-offline-cli-installation))
+* `Zones` - A map of zones to servers for upstream DNS resolution. ([see Standalone Offline CLI Installation](/installation/#standalone-offline-cli-installation))
```
r0mant: Is "zones to servers" correct or a typo? Reads weird.

lib/update/phase_coredns.go
r0mant: Also, `p.Infof("Created ClusterRole %q", corednsResourceName)` would be nice.

lib/update/phase_coredns.go
knisbet: Yes, only one of the checks itself should be required to see if we're on a pre-coredns version and need to insert the configuration.

I'm just more following the philosophy, of if the cluster appears to be missing configuration that's required for coredns to function, regardless of the reason, the upgrade should detect that and correct the state. Assuming if one object is missing the other will be too is reasonable, it's just this is more likely to work if an unpredictable scenario happens (Something like Partial Upgrade, Rollback, Re-Upgrade scenarios can be difficult to reason about and predict).

PR #40, opened 2018-10-04
Do not require merged docker configuration for local checks

a-palchikov:
Merge docker configuration at the point of use - in `RunLocalChecks` instead of requiring the merged configuration to be provided as function configuration.

Fixes the 'docker configuration is required' during installation with defaults.

I had to move `OverrideDockerConfigXXX` and Co to a more appropriate location (lib/checks) to avoid import cycles.
I also set `overlay2` as the default docker storage driver for `opscenter` application and removed the `storagedriver` CLI option for robotest to have this tested.

### DIFF
--- a/lib/checks/checks.go
+++ b/lib/checks/checks.go
@@ -69,7 +69,7 @@ func New(remote Remote, servers []Server, manifest schema.Manifest, requirements
 func ValidateManifest(
 	manifest schema.Manifest,
 	profile schema.NodeProfile,
-	dockerSchema schema.Docker,
+	dockerConfig storage.DockerConfig,
 	stateDir string,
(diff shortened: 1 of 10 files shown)


lib/checks/checks.go
```diff
@@ -256,6 +255,32 @@ func FormatFailedChecks(failed []*agentpb.Probe) string {
 	return buf.String()
 }
 
+// OverrideDockerConfig updates given config with values from overrideConfig where necessary
```
r0mant: Feels like something's missing b/w "updates" and "given".

a-palchikov: What's missing? The method **updates** given **config** (hence passed by reference)

PR #42, opened 2018-10-05
Dmitri/3754 gc

a-palchikov:
- Enable 'gravity gc' when the cluster is degraded
 - Fix cleanup ordering so that dependent packages are properly removed
 - Prune obsolete runtime application packages

Updates https://github.com/gravitational/gravity.e/issues/3734.

### DIFF
--- a/lib/ops/opsservice/operationgroup.go
+++ b/lib/ops/opsservice/operationgroup.go
@@ -121,6 +121,13 @@ func (g *operationGroup) canCreateOperation(operation ops.SiteOperation) error {
 		default:
 			return trace.CompareFailed("the cluster is %v", cluster.State)
 		}
+	case ops.OperationGarbageCollect:
+		// garbage collection is allowed for degraded clusters (to be able to free space)
+		switch cluster.State {
+		case ops.SiteStateActive, ops.SiteStateDegraded:
+		default:
+			return trace.CompareFailed("the cluster is %v", cluster.State)
+		}
 	default:
 		// other operation can be performed by active clusters only
 		switch cluster.State {
--- a/lib/ops/opsservice/operationgroup.go
+++ b/lib/ops/opsservice/operationgroup.go
@@ -122,7 +122,7 @@ func (g *operationGroup) canCreateOperation(operation ops.SiteOperation) error {
 			return trace.CompareFailed("the cluster is %v", cluster.State)
 		}
 	case ops.OperationGarbageCollect:
-		// garbage collection is allowed for degraded clusters (to be able to free space)
+		// garbage collection is allowed for degraded clusters
 		switch cluster.State {
 		case ops.SiteStateActive, ops.SiteStateDegraded:
 		default:
--- a/lib/ops/opsservice/operationgroup.go
+++ b/lib/ops/opsservice/operationgroup.go
@@ -114,15 +114,9 @@ func (g *operationGroup) canCreateOperation(operation ops.SiteOperation) error {
 		if err != nil {
 			return trace.Wrap(err)
 		}
-	case ops.OperationShrink:
-		// shrink is allowed for degraded clusters (to be able to remove offline nodes)
-		switch cluster.State {
-		case ops.SiteStateActive, ops.SiteStateDegraded:
-		default:
-			return trace.CompareFailed("the cluster is %v", cluster.State)
-		}
-	case ops.OperationGarbageCollect:
-		// garbage collection is allowed for degraded clusters
+	case ops.OperationShrink, ops.OperationGarbageCollect:
+		// shrink and gc are allowed for degraded clusters
+		// shrink is allowed to be able to remove failed/offline nodes
 		switch cluster.State {
 		case ops.SiteStateActive, ops.SiteStateDegraded:
 		default:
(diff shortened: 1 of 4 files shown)


lib/ops/opsservice/operationgroup.go
```diff
@@ -121,6 +121,13 @@ func (g *operationGroup) canCreateOperation(operation ops.SiteOperation) error {
 		default:
 			return trace.CompareFailed("the cluster is %v", cluster.State)
 		}
+	case ops.OperationGarbageCollect:
```
r0mant: Nit: you could've just added it to the previous case clause since the conditions seem to be the same:

```go
case ops.OperationShrink, ops.OperationGarbageCollect:
    // shrink and gc are allowed for degraded clusters
```

lib/vacuum/prune/pack/pack.go
```diff
@@ -96,47 +96,37 @@ type Application struct {
...
 // unless it can tell if a package is safe to remove.
 func (r *cleanup) Prune(context.Context) error {
-	repositories, err := r.Packages.GetRepositories()
+	required, err := r.mark()
 	if err != nil {
 		return trace.Wrap(err)
 	}
 
-	required, err := r.mark()
+	state, err := r.build(required)
```
r0mant: I just got a thought: cleanup is an upgrade phase now, right? Is it enabled for Ops Center app too? When Ops Center is upgraded, won't it nuke all apps in it? Maybe the gc command should just not work for Ops Centers too.

lib/vacuum/prune/pack/pack.go
```diff
@@ -150,124 +140,219 @@ func (r *cleanup) mark() (required packageMap, err error) {
...
 
 	return false, nil
 }
 
-func packageForConfig(envelope pack.PackageEnvelope, packages packageMap) (*requiredPackage, error) {
-	parentPackageName := envelope.RuntimeLabels[pack.ConfigLabel]
-	parentPackageFilter, err := loc.ParseLocator(parentPackageName)
-	if err != nil {
-		return nil, trace.Wrap(err, "invalid package locator")
+// withDependencies computes depender packages for the specified package pkg using
```
r0mant: depender -> dependent?

lib/vacuum/prune/pack/pack.go
```diff
@@ -150,124 +140,219 @@ func (r *cleanup) mark() (required packageMap, err error) {
...
+			return nil, trace.NotFound("no depender package %v found for package %v",
+				depender, envelope.Locator)
 		}
+		return nil, trace.Wrap(err)
 	}
-	return nil, trace.NotFound("no package found for resources package %v", envelope.Locator)
+	return []pack.PackageEnvelope{*dependerEnv}, nil
+}
+
+func getPackagesByFilter(service packageService, filter loc.Locator) (result []pack.PackageEnvelope, err error) {
```
r0mant: Is this used anywhere? Can't find it. If yes, we also have `pack.FindPackage`.

lib/vacuum/prune/pack/pack.go
```diff
@@ -150,124 +140,219 @@ func (r *cleanup) mark() (required packageMap, err error) {
...
-		Version:    loc.ZeroVersion,
+		Version:    envelope.Locator.Version,
 	}
-	for packageFilter, pkg := range packages {
-		if appPackageFilter.IsEqualTo(packageFilter) {
-			return &pkg, nil
+	return dependerForPackage(envelope, service, appPackage)
+}
+
+func dependerForPackage(envelope pack.PackageEnvelope, service packageService, depender loc.Locator) ([]pack.PackageEnvelope, error) {
```
r0mant: Does this function need both `envelope` and `depender`? Isn't `depender` the same locator as `envelope.Package`?

lib/vacuum/prune/pack/pack.go
a-palchikov: Upgrade only cleans up obsolete `journald` directories.
`gravity gc` will not touch additional applications (and their dependencies) as it works off of a set of required dependencies based on the cluster application.
So it would only prune older versions of all packages directly or indirectly related to the cluster application and leave the rest.

lib/vacuum/prune/pack/pack.go
a-palchikov: No, it's depender package - i.e. that's the owner of this package, so to speak. The depender package would be the application package for `<app-name>-resources:<app-version>` or `planet-master:<version>` for the `planet-config-<some-suffix>:<version>`

lib/vacuum/prune/pack/pack.go
a-palchikov: See above: depender is the owning package for the package described by `envelope`. It feels that a better comment is in order.

lib/vacuum/prune/pack/pack.go
a-palchikov: Oh, yes, that it will.

r0mant: Got it, if upgrade doesn't cleanup much else beside logs, it should be fine.

The fact that it doesn't make much sense doesn't mean nobody will try it :) Honestly, I'm not sure what kind of behavior would be best when you run it on Ops Center. I would think that we should not implicitly remove any apps from it. In any case, that's out of the scope of this PR.

PR #43, opened 2018-10-05
Use default system log for shrink agent.

r0mant:
Fixes the `Failed to fire hook: open /var/lib/gravity/var/log/telekube-system.log: no such file or directory` issue for shrink agent. Related to https://github.com/gravitational/gravity.e/issues/3761.

Also add "wizard" command to the list of commands that write to /var/log/telekube-system.log too (forgot to add it originally), otherwise it only writes to journald.

Also closes https://github.com/gravitational/gravity.e/issues/3763.

### DIFF
--- a/lib/install/hook.go
+++ b/lib/install/hook.go
@@ -22,7 +22,6 @@ import (
 
 	"github.com/gravitational/gravity/lib/defaults"
 
-	"github.com/gravitational/trace"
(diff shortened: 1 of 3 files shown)


a-palchikov:
Maybe [this](https://github.com/gravitational/gravity.e/issues/3763) is also worth a look here.

PR #54, opened 2018-10-10
Fix a couple of issues.

r0mant:
* Fix an issue with installing AWS cluster via Ops Center.
* Fix an issue with runtime package not being marked as such during pull phase.
* Add a bit more debug info to the operation group to confirm the expand failure theory from https://github.com/gravitational/gravity.e/issues/3773 if it happens again.

Update: Despite robotest succeeding, it actually failed resize scenario in the same fashion as described in https://github.com/gravitational/gravity.e/issues/3773. My last commit here added more debug info for this scenario so let's see if it fails again.

Update 2: And of course with debug info turned on, the test succeeded :)

### DIFF
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -407,17 +407,16 @@ func (p *Peer) tryConnect() (op *operationContext, err error) {
 			p.sendMessage("Connected to installer at %v", addr)
 			return op, nil
 		}
+		p.Infof("Failed connecting to wizard: %v.", err)
 		if utils.IsAbortError(err) {
 			return nil, trace.Wrap(err)
 		}
-
-		// already exists error is returned when there's an ongoing install operation,
-		// do not attempt to dial the site until it fully completes
+		// already exists error is returned when there's an ongoing install
+		// operation, do not attempt to dial the site until it completes
 		if trace.IsAlreadyExists(err) {
 			p.sendMessage("Waiting for the install operation to finish")
 			return nil, trace.Wrap(err)
 		}
-		p.Infof("Failed connecting to wizard: %v.", err)
 
 		op, err = p.dialSite(addr)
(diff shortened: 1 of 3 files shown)


lib/expand/join.go
```diff
@@ -407,29 +407,27 @@ func (p *Peer) tryConnect() (op *operationContext, err error) {
...
 		}
+		p.Infof("Failed connecting to wizard: %v.", err)
 		if utils.IsAbortError(err) {
 			return nil, trace.Wrap(err)
 		}
-
-		// already exists error is returned when there's an ongoing install operation,
-		// do not attempt to dial the site until it fully completes
+		// already exists error is returned when there's an ongoing install
+		// operation, do not attempt to dial the site until it completes
```
a-palchikov: nit: ~site~ cluster

lib/expand/join.go
```diff
@@ -821,12 +819,12 @@ func (p *Peer) validateWizardState(operator ops.Operator) (*ops.Site, *ops.SiteO
...
 
 	switch operation.State {
-	case ops.OperationStateInstallInitiated, ops.OperationStateFailed:
+	case ops.OperationStateInstallInitiated, ops.OperationStateInstallProvisioning, ops.OperationStateFailed:
 		// Consider these states for resuming the installation
 		// (including failed that puts the operation into manual mode)
 	default:
-		return nil, nil,
-			trace.AlreadyExists("operation %v is in progress", operation)
+		return nil, nil, trace.AlreadyExists("operation %#v is in progress",
```
a-palchikov: Why do you want to dump the object directly in the error message? Why not use `log.Warnf()`

r0mant: It was already being dumped here, I just added detailed format.

PR #55, opened 2018-10-11
Use cluster state to determine ability to prune packages for the low-level command

a-palchikov:
Account for older clusters with not explicitly completed update operation when determining if package pruning can be performed using `system gc package`.
This also matches the behavior of `lib/ops/opsservice#operationGroup.canCreateOperation`.

### DIFF
--- a/tool/gravity/cli/gc.go
+++ b/tool/gravity/cli/gc.go
@@ -344,7 +344,7 @@ func removeUnusedPackages(env *localenv.LocalEnvironment, dryRun, pruneClusterPa
 		return trace.Wrap(err)
 	}
 
-	if err = validateCanPrunePackages(operator, *cluster); err != nil {
+	if err = validateCanPrunePackages(*cluster); err != nil {
 		return trace.Wrap(err)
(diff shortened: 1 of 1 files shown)


tool/gravity/cli/gc.go
```diff
@@ -425,13 +425,16 @@ func collectRemoteApplications(operator ops.Operator, clusterKey ops.SiteKey) ([
...
 	if err != nil {
 		return trace.Wrap(err)
 	}
 
-	if !progress.IsCompleted() {
+	// Use the cluster state to determine the operation progress to account
+	// for older clusters where update operation was not explicitly completed.
+	// TODO(dmitri): remove when there's no more need to support this legacy case
+	if cluster.State != ops.SiteStateActive {
 		return trace.CompareFailed("%v is still in progress. "+
```
r0mant: There may be no operation in progress and cluster can still be not active, e.g. degraded. Then the "xxx is still in progress" message will be confusing. I would just remove parts about active operation from this message.

PR #56, opened 2018-10-11
Dmitri/3770 dnsconfig

a-palchikov:
This PR updates the DNS configuration in the cluster record copy when a cluster is installed via an Ops Center and the Ops Center is older than the version of gravity used to create the cluster.

I removed all default DNS configuration settings besides those in `CreateSite` (and now one in `RequestClusterCopy`).

Fixes https://github.com/gravitational/gravity.e/issues/3770.


a-palchikov:
Yes, I teste on Azure/AWS going from 1->2

PR #57, opened 2018-10-12
Add changelog for rc.2 and rc.3.

r0mant:


docs/5.x/changelog.md
```diff
@@ -21,6 +21,22 @@ LTS starts with `3.51.0` with minor backwards compatible changes added over time
...
 ## 5.x Releases
 
+### 5.2.0-rc.3
+
+#### Bugfixes
+
+* Fix an issue with deploying AWS clusters using provisioner.
+* Fix an issue with installers downloaded from the distribution portal.
+* Fix an issue with expanding clusters installed via an Ops Center.
+* Fix an issue with garbage collection not working on older clusters sometimes.
```
a-palchikov: This gives an impression of garbage collection being in GA for older versions while all this is to provide limited support in terms of a custom script file that uses low-level undocumented commands.
I would not mention it.

docs/5.x/changelog.md
```diff
@@ -21,6 +21,22 @@ LTS starts with `3.51.0` with minor backwards compatible changes added over time
...
+
+### 5.2.0-rc.2
+
+#### Bugfixes
+
+* Fix `tele build` failure when encountering unrecognized resources.
+* Fix an issue with garbage collector removing used packages sometimes.
```
a-palchikov: ditto here: it was kind of related to only failing on older clusters (where it was not supposed to be used in the first place).

docs/5.x/changelog.md
r0mant: Ok, I'll remove.

PR #67, opened 2018-10-18
terraform provider publishing

knisbet:
I'm going to be honest... I don't think I really understand these build scripts all that well. I think this is the required configuration to add support for publishing the terraform provider modules.


build.assets/Makefile
```diff
@@ -202,9 +203,10 @@ build-in-container: buildbox
 		   -e "GRAVITY_VERSION=$(GRAVITY_VERSION)" \
 		   -e "GRAVITY_TAG=$(GRAVITY_TAG)" \
 		   -e 'GRAVITY_LINKFLAGS=$(GRAVITY_LINKFLAGS)' \
+		   -e 'TF_BINARIES=$(TF_BINARIES)' \
```
a-palchikov: We should do the same with `$(BINARIES) ` and do away with literal paths (i.e. `$(LOCAL_GRAVITY_BUILDDIR)/tele`) in the `make` invocation.

build.assets/Makefile
```diff
@@ -241,13 +243,12 @@ publish-telekube: build
...
-#	TODO Terraform modules publishing is disabled for now because it's not working:
-#	@echo -e "\n----> Publishing terraform-provider-gravity to $(DISTRIBUTION_OPSCENTER)...\n"
-#	$(GRAVITY_OUT) package delete --ops-url=$(DISTRIBUTION_OPSCENTER) $(TELEKUBE_PUBLISH_FLAGS) --force $(TF_PROVIDER_GRAVITY_PKG) && \
-#	$(GRAVITY_OUT) package import --ops-url=$(DISTRIBUTION_OPSCENTER) $(TELEKUBE_PUBLISH_FLAGS) $(TF_PROVIDER_GRAVITY_OUT) $(TF_PROVIDER_GRAVITY_PKG)
-#	@echo -e "\n----> Publishing terraform-provider-gravityenterprise to $(DISTRIBUTION_OPSCENTER)...\n"
-#	$(GRAVITY_OUT) package delete --ops-url=$(DISTRIBUTION_OPSCENTER) $(TELEKUBE_PUBLISH_FLAGS) --force $(TF_PROVIDER_GRAVITYENTERPRISE_PKG) && \
-#	$(GRAVITY_OUT) package import --ops-url=$(DISTRIBUTION_OPSCENTER) $(TELEKUBE_PUBLISH_FLAGS) $(TF_PROVIDER_TELEKUBE_OUT) $(TF_PROVIDER_GRAVITYENTERPRISE_PKG)
+	@echo -e "\n----> Publishing terraform-provider-gravity to $(DISTRIBUTION_OPSCENTER)...\n"
+	$(GRAVITY_OUT) package delete --ops-url=$(DISTRIBUTION_OPSCENTER) $(TELEKUBE_PUBLISH_FLAGS) --force $(TF_PROVIDER_GRAVITY_PKG) && \
```
a-palchikov: this is using some undefined variables (`TF_PROVIDER_GRAVITY_OUT` and `TF_PROVIDER_TELEKUBE_OUT`) so it won't work.

build.assets/Makefile
```diff
@@ -295,8 +296,8 @@ $(GRAVITY_BUILDDIR):
 	mkdir -p $@
 
 # Builds binaries inside a docker container
-.PHONY: $(BINARIES)
-$(BINARIES):
+.PHONY: $(BINARIES) $(TF_BINARIES)
```
a-palchikov: This target is not `PHONY`, not sure how this came to be.

PR #68, opened 2018-10-18
Use system APIs to search for user/group.

r0mant:
This PR removes manual /etc/passwd and /etc/group parsing instead relying on "os/user" package methods from stdlib which use regular syscalls to retrieve users/groups. I have tested this on a machine connected to an LDAP directory where specified system user was created (i.e. it wasn't present in passwd file locally).

Requires https://github.com/gravitational/planet/pull/321.

Closes https://github.com/gravitational/gravity.e/issues/3706.


lib/systeminfo/user.go
```diff
@@ -76,43 +73,76 @@ func GetRealUser() (*User, error) {
...
-		Name: u.Name,
-		UID:  u.Uid,
-		GID:  u.Gid,
+		Name: usr.Username,
+		UID:  uid,
+		GID:  gid,
 	}, nil
 }
 
+func checkGroup(groupName, groupID string) error {
```
a-palchikov: nit: I would place these in the newspaper order - i.e. after the exported `NewUser` that uses them. I know it's not a widespread style convention in the Go community but I feel it adheres the readability.

r0mant: I agree... I forgot to address it and CI just finished. I'll fix this in another PR.

PR #78, opened 2018-10-29
pass taints to planet instead of applying them later without permissions

knisbet:
Updates https://github.com/gravitational/gravity.e/issues/3789

Pass node taints to planet / kubelet, to be created automatically when the node object is registered.


assets/telekube/resources/app.yaml
```diff
@@ -138,6 +138,9 @@ nodeProfiles:
           - i2.2xlarge
   - name: master
     description: "Telekube Master"
+    taints:
```
r0mant: Did you add this temporarily just for testing, or do you wanna keep it in the telekube app?

lib/ops/opsservice/configure.go
```diff
@@ -995,6 +995,10 @@ func (s *site) getPlanetConfigPackage(
 		args = append(args, fmt.Sprintf("--device=%v", device.Format()))
 	}
 
+	for _, taint := range profile.Taints {
+		args = append(args, fmt.Sprintf("--taint=%v=%v:%v", taint.Key, taint.Value, taint.Effect))
```
r0mant: Is planet gonna parse this flag okay? (I mean, the fact that there is "=" in the middle.)

knisbet: It seems to work when I tested it with a local build.

assets/telekube/resources/app.yaml
knisbet: I did add it for testing, however, I left it in the PR since I do think it makes sense to keep, when someone is specifically using a node profile of master, that we do include the taints.

PR #79, opened 2018-10-30
Wolfe/docs

alexwolfe:
@alex-kovoy I've updated the gravity docs with the new them as well. Please review and merge thanks.


docs/5.x/cluster.md
```diff
@@ -1656,7 +1656,7 @@ With this configuration, the Ops Center cluster will provide a single Kubernetes
 service called `gravity-public` configured to serve both user and cluster
 traffic:
 
-```bash
+```bsh
```
r0mant: Are you sure `bsh` is the correct snippet type? Not `bash`? Cause all these snippets do not display correctly (I'm looking at the staging):

screenshot from 2018-10-30 15-01-19

Note "$" sign is missing. And here "#" is treated like a comment:

screenshot from 2018-10-30 15-01-56

alexwolfe: Yes, 100% sure. It's a different code formatter now and the it looks for "lang-bsh" instead of the old "bash".

r0mant: Okay. Well, somehow snippets are still not rendered correctly... (see my comment above about missing $ character)

docs/theme/base.html
```diff
@@ -1,168 +1,26 @@
...
-  {% if config.site_author %}<meta name="author" content="{{ config.site_author }}">{% endif %}
-  {% block htmltitle %}
-  <title>{% if page.title %}{{ page.title }} - {% endif %}{{ config.site_name }}</title>
-  {% endblock %}
-
-  <link rel="shortcut icon" href="https://gravitational.com/favicon.ico" type="image/x-icon" />
-
-  {# CSS #}
-  
-  <link rel="stylesheet" href="{{ base_url }}/css/theme.css" type="text/css" />
```
alex-kovoy: Lets delete these files then
`theme.css`
`highlight.css`

docs/theme/js/theme.js
```diff
@@ -1,10 +1,188 @@
+
+(function ($) {
```
alex-kovoy: Since `theme.js` looks identical to teleport docs, it seems that it makes sense to move this code to `web` repo and expose it via grvLib.

docs/theme/src/index.scss
```diff
@@ -1,92 +1,94 @@
...
-  border-radius: 5px;
-}
+$unit: 8px;
+
+/*
+* Breakpoints
+*
+*/
+
+$phone-breakpoint: 480px !default;
```
alex-kovoy: do we need this variables here? If not, lets delete

PR #81, opened 2018-10-30
[5.3.x] Add support for recursive volume mounts.

r0mant:
Forward-port of https://github.com/gravitational/gravity/pull/80. Closes https://github.com/gravitational/gravity.e/issues/3794.


docs/5.x/pack.md
```diff
@@ -457,8 +457,11 @@ nodeProfiles:
           # UID and GID set linux UID and GID on the directory if specified
           uid: 114
           gid: 114
-          # mode is a linux mode to set on the directory
+          # Linux mode to set on the directory
```
a-palchikov: `Unix file permissions` as it is not really Linux-specific. Also would be nice to add whether it's numeric or also accepts symbolic values.

PR #88, opened 2018-11-01
Use On-Premises instead of On-Premise

yuvalherziger:
This whole PR is the embodiment of nitpicking, but it is actually **wrong** to use the term _on-premise_ in English, in the context that we usually use it in.

**Pro links:**
- https://en.wikipedia.org/wiki/On-premises_software
- https://brians.wsu.edu/2016/05/30/premise-premises/
- https://www.brianmadden.com/opinion/So-apparently-we-lost-the-grammar-war-and-on-premises-is-just-called-on-premise-now
- https://unscramblegk.tumblr.com/post/77067365896/on-premises-vs-on-premise-removing-some

**Con link:**
- https://www.adamfowlerit.com/2017/04/premise-vs-premises-cares/

Feel free to close this PR if you don't believe that it should be changed 🤓


knisbet:
@yuvalherziger Thanks for you're contribution. 

I don't really have a strong opinion on which form is more correct, but I do agree premises seems a bit more appropriate. Let's see what @kontsevoy thinks.

web/src/app/modules/site/components/navTopBar.jsx
```diff
@@ -74,7 +74,7 @@ const LocationIndicator = ({ provider, location }) => {
       providerLabel = `AWS${location}`;
       break;
     case ProviderEnum.ONPREM:
-      providerLabel = 'On premise';
+      providerLabel = 'On premises';
```
r0mant: @alex-kovoy Just making sure - this will not break anything, will it? I.e. we don't search anything anywhere in the UI by this label?

alex-kovoy: looks good.

PR #91, opened 2018-11-02
Dmitri/3770 missing dnsconfig

a-palchikov:
This PR fixes a regression with DNS configuration not propagating during install/expand by moving the configuration to the operation plan (for the `bootstrap` step).

Also included are various improvements to issues I discovered while working on the original issue.

I tested wizard installs/expands (w/ and w/o custom DNS configuration), CLI installs/expands and Ops Center-based install/expand.

Requires https://github.com/gravitational/gravity.e/pull/3799.
Fixes https://github.com/gravitational/gravity.e/issues/3770.

### DIFF
--- a/lib/expand/fsm.go
+++ b/lib/expand/fsm.go
@@ -146,7 +146,17 @@ type fsmEngine struct {
 
 // GetExecutor returns a new executor based on the provided parameters
 func (e *fsmEngine) GetExecutor(p fsm.ExecutorParams, remote fsm.Remote) (fsm.PhaseExecutor, error) {
-	return e.Spec(p, remote)
+	logger := &fsm.Logger{
+		FieldLogger: logrus.WithField(constants.FieldPhase, p.Phase.ID),
+		Key:         p.Key(),
+		Operator:    e.Operator,
+	}
+	executor, err := e.Spec(p, remote)
+	if err != nil {
+		logger.Warnf("Failed to initialize phase: %v.", err)
+		return nil, trace.Wrap(err)
+	}
+	return executor, nil
 }
 
 // ChangePhaseState updates the phase state based on the provided parameters
--- a/lib/fsm/fsm.go
+++ b/lib/fsm/fsm.go
@@ -399,7 +399,7 @@ func (f *FSM) executeOnePhase(ctx context.Context, p Params, phase storage.Opera
 
 	err = executor.PreCheck(ctx)
 	if err != nil {
-		executor.Errorf("Phase %v precheck failed: %v.", phase.ID, err)
+		executor.Errorf("Phase precheck failed: %v.", err)
 		return trace.Wrap(err)
 	}
 
@@ -416,7 +416,7 @@ func (f *FSM) executeOnePhase(ctx context.Context, p Params, phase storage.Opera
 
 	err = executor.Execute(ctx)
 	if err != nil {
-		executor.Errorf("Phase %v execution failed: %v.", phase.ID, err)
+		executor.Errorf("Phase execution failed: %v.", err)
 		if err := f.ChangePhaseState(ctx,
 			StateChange{
 				Phase: phase.ID,
@@ -430,7 +430,7 @@ func (f *FSM) executeOnePhase(ctx context.Context, p Params, phase storage.Opera
 
 	err = executor.PostCheck(ctx)
 	if err != nil {
-		executor.Errorf("Phase %v postcheck failed: %v.", phase.ID, err)
+		executor.Errorf("Phase postcheck failed: %v.", err)
 		return trace.Wrap(err)
 	}
 
--- a/lib/install/flow.go
+++ b/lib/install/flow.go
@@ -58,16 +58,16 @@ func (i *Installer) StartInteractiveInstall() error {
 			if tm.IsZero() {
 				return trace.ConnectionProblem(nil, "timeout")
 			}
-			sites, err := i.Operator.GetSites(i.AccountID)
+			clusters, err := i.Operator.GetSites(i.AccountID)
 			if err != nil {
 				i.Warnf("Failed to get sites: %v.", trace.DebugReport(err))
 				continue
 			}
-			if len(sites) == 0 {
-				i.Info("No sites created yet.")
+			if len(clusters) == 0 {
+				i.Info("No clusters created yet.")
 				continue
 			}
-			i.Cluster = &sites[0]
+			i.Cluster = &clusters[0]
 			// set site domain set by user, otherwise we will attempt
 			// to generate new cluster
 			i.SiteDomain = i.Cluster.Domain
@@ -409,7 +409,7 @@ func (i *Installer) waitForAgents() error {
 	for {
 		select {
 		case <-i.Context.Done():
-			return trace.ConnectionProblem(nil, "context is closing")
+			return trace.ConnectionProblem(i.Context.Err(), "context is closing")
 		case tm := <-ticker.C:
 			if tm.IsZero() {
 				return trace.ConnectionProblem(nil, "timed out waiting for agents to join")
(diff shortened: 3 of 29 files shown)


lib/expand/fsm.go
```diff
@@ -146,7 +146,17 @@ type fsmEngine struct {
 
 // GetExecutor returns a new executor based on the provided parameters
 func (e *fsmEngine) GetExecutor(p fsm.ExecutorParams, remote fsm.Remote) (fsm.PhaseExecutor, error) {
-	return e.Spec(p, remote)
+	logger := &fsm.Logger{
```
a-palchikov: This will hopefully include more error information in the operator logs if the phase fails to initialize.

lib/fsm/fsm.go
```diff
@@ -399,7 +399,7 @@ func (f *FSM) executeOnePhase(ctx context.Context, p Params, phase storage.Opera
 
 	err = executor.PreCheck(ctx)
 	if err != nil {
-		executor.Errorf("Phase %v precheck failed: %v.", phase.ID, err)
+		executor.Errorf("Phase precheck failed: %v.", err)
```
a-palchikov: The phase ID should already be included in the executor's logger.

lib/install/flow.go
```diff
@@ -409,7 +409,7 @@ func (i *Installer) waitForAgents() error {
 	for {
 		select {
 		case <-i.Context.Done():
-			return trace.ConnectionProblem(nil, "context is closing")
+			return trace.ConnectionProblem(i.Context.Err(), "context is closing")
```
a-palchikov: Persist the original error so it can be acted upon.

lib/install/install.go
```diff
@@ -258,6 +258,9 @@ func (c *Config) CheckAndSetDefaults() (err error) {
 	if c.NewProcess == nil {
 		c.NewProcess = process.NewProcess
 	}
+	if c.DNSConfig.IsEmpty() {
+		c.DNSConfig = storage.DefaultDNSConfig
```
a-palchikov: Set the configuration default for installs.

lib/install/install.go
```diff
@@ -403,6 +406,9 @@ func (i *Installer) Wait() error {
 			return nil
 		case event := <-i.EventsC:
 			if event.Error != nil {
+				if utils.IsContextCancelledError(event.Error) {
```
a-palchikov: Ignore the `context is closing` errors as spam.

lib/localenv/clusterenv.go
```diff
@@ -135,7 +135,6 @@ func newClusterEnvironment(args clusterEnvironmentArgs) (*ClusterEnvironment, er
 		Apps:     apps,
 		Users:    users,
 		StateDir: siteDir,
-		Local:    true,
```
a-palchikov: I factored operation status tailing into a separate function so this is not strictly needed anymore.

lib/ops/opsservice/shrink.go
```diff
@@ -146,7 +146,7 @@ func (s *site) validateShrinkRequest(req ops.CreateSiteShrinkOperationRequest, c
 	teleserver := servers.getWithLabels(labels{ops.Hostname: server.Hostname})
 	if len(teleserver) == 0 {
 		if !req.Force {
-			return nil, trace.Wrap(err,
+			return nil, trace.BadParameter(
```
a-palchikov: This is the bug we fixed in 5.0.x but not in master. Basically, when removing an offline server w/o `--force`, the code would panic instead of displaying this error message.

lib/status/status.go
```diff
@@ -212,10 +212,15 @@ type ClusterOperationProgress struct {
...
+// GetSystemStatus returns the status of the system
+func (r Agent) GetSystemStatus() pb.SystemStatus_Type {
+	return pb.SystemStatus_Type(r.SystemStatus)
+}
+
 // Agent specifies the status of the system and individual nodes
 type Agent struct {
 	// SystemStatus defines the health status of the whole cluster
-	SystemStatus pb.SystemStatus_Type `json:"system_status"`
+	SystemStatus SystemStatus `json:"system_status"`
```
a-palchikov: I moved this to a variable of an alias type to be able to format it for display (see `String` and `GoString` overrides below), so that system status comes out as `degraded` instead of `2`.

tool/gravity/cli/ops.go
```diff
@@ -101,7 +101,9 @@ func uploadUpdate(env *localenv.LocalEnvironment, opsURL string) error {
 
 	clusterOperator, err := defaultEnv.SiteOperator()
 	if err != nil {
-		return trace.Wrap(err)
+		return trace.Wrap(err, "unable to access cluster.\n"+
```
a-palchikov: This is likely more user-friendly than the original `open /var/lib/gravity/local/.../kubectl.kubeconfig: file or directory not found`

tool/gravity/cli/run.go
```diff
@@ -369,11 +369,13 @@ func Execute(g *Application, cmd string, extraArgs []string) error {
 	case g.StatusCmd.FullCommand():
 		printOptions := printOptions{
 			token:       *g.StatusCmd.Token,
-			tail:        *g.StatusCmd.Tail,
 			operationID: *g.StatusCmd.OperationID,
 			quiet:       *g.Silent,
 			format:      *g.StatusCmd.Output,
 		}
+		if *g.StatusCmd.Tail {
```
a-palchikov: I moved operation status tailing into a separate function as it needs to work with the actual cluster controller (not the 'local' operator from localenv/clusterenv) so it properly receives the operation log.

It also works around the issue with cluster state other than healthy when tailing operation logs:
```
[vagrant@node-1 ~]$ gravity status --operation-id=11c333ad-2bf4-412b-981f-c55d7e7f054f --tail
[ERROR]: Cluster is degraded
```

tool/gravity/cli/status.go
```diff
@@ -88,6 +88,33 @@ func status(env *localenv.LocalEnvironment, printOptions printOptions) error {
...
+	}
+
+	operator, err := env.SiteOperator()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	status, err := acquireClusterStatus(context.TODO(), env, operator, operationID)
+	if err == nil {
+		err = tailOperationLogs(operator, status.Operation.Key())
```
r0mant: Actually, during one of recent refactorings (I think it was expand-fsm) I made changes so the `Operation` field in the collected status object refers to the _last completed_ operation. There is now also `ActiveOperations` field on the status object which contains the operation(-s) that is currently in progress. So you probably want to use that one here. Also, there may be several operations in progress at the same time (joins), so in this case we should probably fail and ask to provide a specified operation-id flag to the command.

Also, as an idea, if there are no operations in progress, we could just fall back to this `Operation` field to display logs for the last completed operation, would be pretty convenient.

a-palchikov: I think it depends on whether the `operationID` parameter specifies a valid operation ID, in which case it will always query info for the specified operation.

lib/install/install.go
knisbet: Should this be suppressed here, or should it be up to the caller on whether to suppress logging of a context cancellation?

lib/status/status.go
```diff
@@ -402,6 +407,26 @@ func diskSpaceProbeErrorDetail(p pb.Probe) (string, error) {
...
 
+// String returns a textual representation of this system status
+func (r SystemStatus) String() string {
+	switch pb.SystemStatus_Type(r) {
+	case pb.SystemStatus_Running:
+		return "running"
+	case pb.SystemStatus_Degraded:
+		return "degraded"
+	default: // pb.SystemStatus_Unknown:
+		return "unknown"
```
knisbet: May not be necessary here, but I would consider separating the default -> "unknown" and SystemStatus_Unknown -> "unknown". The basic idea is, if a new SystemStatus code get's added, but this switch statement isn't updated, that the bug becomes visible and easier to locate.

PR #94, opened 2018-11-07
better handling of redirects to install complete step

alex-kovoy:
Instead of redirecting to last installation step (bandwagon app), redirect a user back to installer if last installation step has not been completed.

The installer in this case will show that installation has been completed and ask a user to click a button to complete the installation.

Also fixes https://github.com/gravitational/gravity.e/issues/3781

### DIFF
--- a/lib/webapi/webhandler.go
+++ b/lib/webapi/webhandler.go
@@ -88,8 +88,6 @@ func NewHandler(cfg WebHandlerConfig) *WebHandler {
 		FieldLogger: log.WithField(trace.Component, "webhandler"),
 	}
 
-	wh.Handle("GET", "/web/config.js", wh.noLogin(wh.configHandler))
-
 	noLoginRoutes := []string{
 		"/web/login",
 		"/web/recover/*rest",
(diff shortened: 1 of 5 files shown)


lib/webapi/webhandler.go
```diff
@@ -101,40 +99,29 @@ func NewHandler(cfg WebHandlerConfig) *WebHandler {
...
-		"/web/site/:site_domain",
-		"/web/site/:site_domain/*rest",
-	}
-	for _, route := range siteRoutes {
-		for _, m := range []string{"GET", "POST"} {
-			wh.Handle(m, route, wh.needsLogin(wh.siteHandler))
-		}
-	}
+	// installer
+	wh.Handle("GET", "/web/installer/new/*rest", wh.needsLogin(wh.installerHandler))
```
r0mant: What about post, put and other types of requests that were handled before, here and below?

lib/webapi/webhandler.go
```diff
@@ -196,6 +183,21 @@ func (h *WebHandler) configHandler(w http.ResponseWriter, r *http.Request, p htt
 	fmt.Fprintf(w, "var GRV_CONFIG = %v;", string(out))
 }
 
+func (h *WebHandler) rootHandler(w http.ResponseWriter, r *http.Request, p httprouter.Params, s session) {
+	site, err := h.cfg.Operator.GetLocalSite()
```
r0mant: site -> cluster

lib/webapi/webhandler.go
```diff
@@ -196,6 +183,21 @@ func (h *WebHandler) configHandler(w http.ResponseWriter, r *http.Request, p htt
...
+	if err != nil && !trace.IsNotFound(err) {
+		log.Errorf("failed to get local site: %v", trace.DebugReport(err))
```
r0mant: Capitalize "Failed" plz.

lib/webapi/webhandler.go
```diff
@@ -307,8 +309,8 @@ func (h *WebHandler) siteHandler(w http.ResponseWriter, r *http.Request, p httpr
 	}
 
 	// if the app defines custom install step and the site hasn't completed it yet,
-	// always redirect the site to it
-	http.Redirect(w, r, "/web/installer/site/"+siteDomain+"/complete/", http.StatusFound)
+	// always redirect the site back to installer
+	http.Redirect(w, r, "/web/installer/site/"+siteDomain, http.StatusFound)
```
r0mant: This means the user will see "installation complete" page again and will have to click "continue", right? Would it be complicated to fix this session storage issue you told me about and redirect directly to bandwagon instead?

lib/webapi/webhandler.go
alex-kovoy: there are no need to have them as there is no exposed APIs using these verbs

lib/webapi/webhandler.go
alex-kovoy: This approach is simpler as we do not have much control over customized bandwagon step

lib/webapi/webhandler.go
r0mant: So when bandwagon app does a POST request - how is this handled then?

alex-kovoy: good catch

PR #96, opened 2018-11-08
[Will be merged in 5.4] Changed gravity-site to be as unprivileged as possible

eldios:
Changed gravity-site to be as unprivileged as possible.

Created dedicated service account (gravity-site) in kube-system namespace and at cluster level,
then gave permission as needed.


eldios:
I tried fiddling around with all the options I could find and it all worked.. but I'll test other ones if needed.

eldios:
All meaningful tests worked. I took this list as a starting point:

https://github.com/gravitational/gravity.e/issues/3769

Couldn't find any broken behavior with current permissions.

r0mant:
@eldios @a-palchikov Okay, I did a quick check and it looks like everything works because gravity-site explicitly uses privileged kube client (using scheduler cert) for certain things so this service account doesn't have any effect on them. For other things it uses in-cluster config though and I still thing some permissions are missing. E.g. have you tried installing an Ops Center and then running `gravity resource get endpoints` on it?

eldios:
@r0mant that's fixed now.. anything else you'd suggest trying?

BTW, I didn't try that operation cause it was not mentioned in the test list mentioned above.

knisbet:
Maybe not specifically related to this PR, but it would also be good to set / test setting readOnlyFs, and dropping all unneeded capabilities.

assets/site-app/resources/site.yaml
```diff
@@ -1,3 +1,119 @@
...
+  - ""
+  resources:
+  - services
+  - configmaps
+  verbs:
+  - get
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
```
r0mant: Isn't configmaps already included in the above rule?

eldios:
@r0mant definitely a duplicate. Fixed.

r0mant:
I believe the following permissions are also required:

```
secrets: update/delete
configmaps: update
service: create/update/delete
```

eldios:
> secrets: update/delete configmaps: update service: create/update/delete

do you think those would be needed only for the kube-system namespace or also for the monitoring one?

eldios:
I'm guessing we should keep treat this as a major change and thus not introduce it in branches that are already considered "stable" (and thus potentially break them). I'd merge it in 5.4 and only if needed, evaluate backporting it in 5.3.

a-palchikov:
I'm fine with merging this and following up on the missing permissions should they come up.

PR #104, opened 2018-11-15
Revendor Teleport v3.0.1.

r0mant:


lib/process/proxy.go
```diff
@@ -190,6 +205,18 @@ func (t *teleportProxyService) GetCertAuthorities(caType services.CertAuthType)
...
 }
 
+// GetCertAuthority returns the requested certificate authority
+func (t *teleportProxyService) GetCertAuthority(id services.CertAuthID, loadSigningKeys bool) (*authority.TLSKeyPair, error) {
+	ca, err := t.authClient.GetCertAuthority(id, loadSigningKeys)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	return &authority.TLSKeyPair{
+		KeyPEM:  ca.GetTLSKeyPairs()[0].Key,
```
a-palchikov: Does `GetTLSLeyPairs` guarantee that the key pairs slice is not empty?

lib/storage/keyval/authorities.go
```diff
@@ -162,3 +161,29 @@ func (b *backend) DeactivateCertAuthority(id teleservices.CertAuthID) error {
...
+	newData, err := teleservices.GetCertAuthorityMarshaler().MarshalCertAuthority(new)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	existingData, err := teleservices.GetCertAuthorityMarshaler().MarshalCertAuthority(existing)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	ttl := b.ttl(new.Expiry())
+	var outData []byte
```
a-palchikov: Is `outData` used anywhere?

lib/storage/keyval/remoteclusters.go
```diff
@@ -0,0 +1,89 @@
...
+package keyval
+
+import (
+	"encoding/json"
+
+	"github.com/gravitational/teleport/lib/services"
+	"github.com/gravitational/trace"
+)
+
+// CreateRemoteCluster creates remote cluster
```
a-palchikov: `adds a new remote cluster entry under remoteClusters tree`?

lib/update/phase_bootstrap.go
```diff
@@ -247,9 +247,54 @@ func (p *updatePhaseBootstrap) pullSystemUpdates() error {
 		return trace.Wrap(err, "failed to pull system updates: %s", out)
 	}
 	log.Debugf("Pulled system updates: %s.", out)
+	// the low-level pull-updates call above does not pull teleport config
+	// package updates, so do it here: it's easier since we have access to
+	// plan, cluster state, etc.
+	updates, err := p.collectTeleportUpdates()
```
a-palchikov: Should we instead make the system part (i.e. pulling updates) an API so it stays in one place? This can lead to code fragmentation where we'd have pieces to do one thing spread over multiple places.

lib/update/phase_init.go
```diff
@@ -369,7 +386,30 @@ func removeLegacyUpdateDirectory(log log.FieldLogger) error {
...
+// Rollback rolls back the init phase
+func (p *updatePhaseInit) Rollback(context.Context) error {
+	if err := p.removeConfiguredPackages(); err != nil {
+		return trace.Wrap(err)
+	}
+	return nil
+}
+
+// removeConfiguredPackages removes packages configured during init phase
+func (p *updatePhaseInit) removeConfiguredPackages() error {
```
a-palchikov: I wonder if this would need to be done after the update regardless - i.e. removing the auxiliary labels?

lib/update/phase_migration.go
```diff
@@ -194,3 +202,201 @@ func (p *phaseUpdateLabels) PreCheck(context.Context) error {
...
+			}
+			allowKubeGroups = append(allowKubeGroups, groups...)
+		}
+		rule.Actions = actions
+		allowRules = append(allowRules, rule)
+	}
+
+	var denyKubeGroups []string
+	var denyRules []teleservices.Rule
+	for _, rule := range role.GetRules(teleservices.Deny) {
```
a-palchikov: that looks pretty similar to the loop above - are there any differences?

lib/update/phase_migration.go
```diff
@@ -194,3 +202,201 @@ func (p *phaseUpdateLabels) PreCheck(context.Context) error {
...
+func (*phaseMigrateRoles) PreCheck(context.Context) error {
+	return nil
+}
+
+// PostCheck is no-op for this phase
+func (*phaseMigrateRoles) PostCheck(context.Context) error {
+	return nil
+}
+
+// getBackupBackend returns a new local backend specific for current operation
```
a-palchikov: nit: `specific to`

lib/update/phase_migration_test.go
```diff
@@ -18,45 +18,103 @@ package update
...
 func (s *PhaseMigrationSuite) SetUpSuite(c *check.C) {
 	services := opsservice.SetupTestServices(c)
 	s.backend = services.Backend
-}
-
-func (s *PhaseMigrationSuite) TestMigrateLinks(c *check.C) {
 	cluster, err := s.backend.CreateSite(storage.Site{
 		AccountID: uuid.New(),
 		Domain:    "example.com",
 		Created:   time.Now(),
```
a-palchikov: Tests should avoid variables as much as possible - better to use some predefined time constant.

lib/update/phase_migration_test.go
```diff
@@ -18,45 +18,103 @@ package update
...
 	c.Assert(err, check.IsNil)
+	s.cluster = *cluster
+	s.backupDir = c.MkDir()
+}
 
+func (s *PhaseMigrationSuite) TestMigrateRoles(c *check.C) {
+	phase, err := NewPhaseMigrateRoles(
+		FSMConfig{Backend: s.backend},
+		storage.OperationPlan{
+			OperationID: uuid.New(),
```
a-palchikov: nit: same about the uuids

lib/update/phase_system.go
```diff
@@ -104,6 +109,102 @@ func (p *updatePhaseSystem) Rollback(context.Context) error {
...
+		p.Packages, p.Plan.ClusterName, map[string]string{
+			pack.AdvertiseIPLabel: p.Phase.Data.Server.AdvertiseIP,
+			pack.OperationIDLabel: p.Plan.OperationID,
+			pack.PurposeLabel:     pack.PurposeTeleportMasterConfig,
+		})
+	if err != nil && !trace.IsNotFound(err) {
+		return trace.Wrap(err)
+	}
+	if trace.IsNotFound(err) {
+		p.Infof("No teleport master config update found.")
```
a-palchikov: `Info`

lib/utils/parse.go
```diff
@@ -419,3 +419,29 @@ func ToUnknownResource(resource teleservices.Resource) (*teleservices.UnknownRes
...
 	return &unknown, nil
 }
+
+// ParseProxyAddr parses proxy address in the format "host:webPort,sshPort"
+//
+// If web/SSH ports are missing the provided defaults are used.
+func ParseProxyAddr(proxyAddr, defaultWebPort, defaultSSHPort string) (host string, webPort string, sshPort string, err error) {
+	host, port, err := net.SplitHostPort(proxyAddr)
+	if err != nil {
+		return proxyAddr, defaultWebPort, defaultSSHPort, nil
```
a-palchikov: isn't this kind of dangerous - since the error is suppressed and not evaluated? Or is the condition flipped?

lib/process/process.go
```diff
@@ -1779,6 +1812,17 @@ func (p *Process) initAccount() error {
 	return nil
 }
 
+// removeLegacyIdentities removes legacy admin/proxy identities so that new
+// ones can be generated new ones upon the first start of new teleport
```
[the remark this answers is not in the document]
r0mant: @a-palchikov Yes, it won't start, it'll fail on some internal teleport migration cause 3.0 is not fully backward-compatible with 2.4 (we skipped 4 releases). I've added this to the comment.

@knisbet It will regenerate identities in the old format just like on fresh install.

lib/clients/teleport.go
```diff
@@ -46,44 +58,85 @@ func Teleport(operator ops.Operator, proxyHost string) (*client.TeleportClient,
...
+		CN:    constants.OpsCenterUser,
+		Names: []csr.Name{{O: defaults.SystemAccountOrg}},
+	}, private)
+	if err != nil {
+		return nil, nil, trace.Wrap(err)
+	}
 	response, err := operator.SignSSHKey(ops.SSHSignRequest{
 		User:          constants.OpsCenterUser,
-		AccountID:     cluster.AccountID,
+		AccountID:     defaults.SystemAccountID,
```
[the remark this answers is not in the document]
r0mant: IIRC I took advantage of the fact that our account ID is always the same so I don't have to pass the full cluster object here.

lib/install/phases/connect.go
```diff
@@ -0,0 +1,165 @@
...
+		return trace.Wrap(err)
+	}
+	installerAuthorities, err := p.getAuthorities(installerClient, trustedCluster.GetName())
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	// exchange cluster/installer authorities
+	for _, ca := range clusterAuthorities {
+		err := installerClient.UpsertCertAuthority(ca)
```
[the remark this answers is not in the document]
r0mant: This is calling teleport API so it would need to be extended if we wanted to do this.

lib/install/planbuilder.go
```diff
@@ -525,10 +550,27 @@ func (i *Installer) GetPlanBuilder(cluster ops.Site, op ops.SiteOperation) (*Pla
...
+		DNSConfig:               cluster.DNSConfig,
+		InstallerTrustedCluster: trustedCluster,
 	}, nil
 }
 
+// getInstallerTrustedCluster returns trusted cluster representing installer process
+func (i *Installer) getInstallerTrustedCluster() (storage.TrustedCluster, error) {
+	seedConfig := i.Process.Config().OpsCenter.SeedConfig
+	if seedConfig == nil {
+		return nil, trace.NotFound("process config is missing seed config: %#v",
```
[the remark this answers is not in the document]
r0mant: Yeah, these are tricky because it would effectively mean a programming error since seed config always contains trusted clusters for the installer process. I've updated the wording in attempt to make it clearer.

lib/ops/opsclient/opsclient.go
```diff
@@ -870,10 +870,15 @@ func (c *Client) RotateSecrets(req ops.RotateSecretsRequest) (*ops.RotatePackage
 }
 
 // RotatePlanetConfig rotates planet configuration package for the server specified in the request
-func (c *Client) RotatePlanetConfig(req ops.RotatePlanetConfigRequest) (*ops.RotatePackageResponse, error) {
+func (c *Client) RotatePlanetConfig(req ops.RotateConfigPackageRequest) (*ops.RotatePackageResponse, error) {
 	return nil, trace.NotImplemented("this method is only supported by local operator")
 }
 
+// RotateTeleportConfig rotates teleport configuration package for the server specified in the request
+func (c *Client) RotateTeleportConfig(req ops.RotateConfigPackageRequest) (*ops.RotatePackageResponse, *ops.RotatePackageResponse, error) {
```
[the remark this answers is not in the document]
r0mant: I imagine this would require a lot of refactoring e.g. because Operator we use in upgrade FSM has to call methods of both "local" and "not local" ops service.

lib/ops/opsservice/configure.go
```diff
@@ -1172,26 +1080,38 @@ func toObject(in interface{}) (map[string]interface{}, error) {
 
 func (s *site) teleportMasterConfigPackage(master remoteServer) (*loc.Locator, error) {
 	configPackage, err := loc.ParseLocator(
-		fmt.Sprintf("%v/%v:0.0.1-%v", s.siteRepoName(), constants.TeleportMasterConfigPackage,
-			PackageSuffix(master, s.domainName)))
+		fmt.Sprintf("%v/%v:0.0.%v-%v", s.siteRepoName(), constants.TeleportMasterConfigPackage,
+			time.Now().UTC().Unix(), PackageSuffix(master, s.domainName)))
```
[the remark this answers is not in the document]
r0mant: I thought about this actually. I can see a use-case where we might need to update a teleport config w/o updating the package version, for example to update a value of some field (idk, like a timeout or something). Although the way it's implemented right now, the code does not really support this use-case (it will not rotate config unless teleport package version changed).

lib/storage/keyval/authorities.go
r0mant: No, but I don't think I can pass `nil` instead of it and I wanted to keep the interface of `compareAndSwapBytes` similar to existing `compareAndSwap`.

lib/update/phase_bootstrap.go
r0mant: We definitely should get rid of `system pull-updates` and use similar approach as here (or in "pull" install phase)... but not as a part of this PR...

lib/update/phase_init.go
r0mant: Hm, why? I think it's okay to keep configured packages in the cluster db (e.g. if you need to re-pull it locally for whatever reason or what not).

lib/ops/opsservice/service.go
```diff
@@ -379,12 +381,13 @@ func (o *Operator) GetExpandToken(key ops.SiteKey) (*storage.ProvisioningToken,
 	}
 
 	for _, token := range tokens {
-		if token.Type == storage.ProvisioningTokenTypeExpand {
+		// return long-lived join token
+		if token.Type == storage.ProvisioningTokenTypeExpand && token.Expires.IsZero() {
```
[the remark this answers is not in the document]
r0mant: We would need to implement this feature specifically. This function basically returns token displayed by "gravity status" so I've just updated it to be 100% sure it will return long-lived token.

r0mant:
@a-palchikov @knisbet I believe I have addressed all your comments.

r0mant:
Will merge this PR and its enterprise part after @a-palchikov has merged his outstanding PRs to master to avoid extra backporting work (because current master will become `version/5.3.x` and this PR will be `5.4.0-alpha.1`).

lib/process/process.go
```diff
@@ -1813,7 +1813,10 @@ func (p *Process) initAccount() error {
 }
 
 // removeLegacyIdentities removes legacy admin/proxy identities so that new
-// ones can be generated new ones upon the first start of new teleport
+// ones can be generated upon the first start of new teleport
+//
+// If they are not removed, this process that includes teleport 3.0 will not
+// be able to start after upgrade from older gravity that used teleport 2.4.
```
a-palchikov: So, as @knisbet suggested, this should get a TODO to remove once 2.4 upgrades are no longer supported (?)

PR #112, opened 2018-11-21
Fix terraform providers publishing.

r0mant:


build.assets/Makefile
```diff
@@ -300,10 +294,10 @@ $(GRAVITY_BUILDDIR):
 	mkdir -p $@
 
 # Builds binaries inside a docker container
-$(BINARIES) $(TF_BINARIES):
-	@echo -e "\n----> Building $@ binary...\n"
-	go build -o $@ -ldflags $(GRAVITY_LINKFLAGS) $(GRAVITY_PKG_PATH)/tool/$(notdir $@)
-	@echo "Done --> $@"
+$(BINARIES):
```
a-palchikov: This no longer a file target (i.e. not a phony target), this either needs a `.PHONY` or why was it changed to just a filename again?

r0mant: Good catch, I've added .phony.

I had to change it because it was not otherwise possible to alter `LOCAL_GRAVITY_BUILDDIR` for terraform binaries passed from the enterprise Makefile. Besides, I actually think this made the binaries build logic clearer.

PR #113, opened 2018-11-21
coredns v2

knisbet:
Updates https://github.com/gravitational/gravity.e/issues/3805

Update CoreDNS to v2 deployment, which is closer to a traditional coredns installation. The main difference, is we split coredns into two daemonsets / services. This allows us to prevent a retry for DNS resolution from being load balanced to the same node multiple times in a row. 

Also updates the way coredns configuration is generation to be within gravity, and fixes a few glitches in the upgrade path.


lib/install/phases/coredns_test.go
```diff
@@ -0,0 +1,119 @@
...
+    policy sequential
+    health_check 0
+  }
+}
+`,
+		},
+		{
+			coreDNSConfig{
+				UpstreamNameservers: []string{"1.1.1.1"},
+				Rotate:              true,
```
a-palchikov: Does it make sense to rotate when `len(UpstreamNameservers) < 2`?

lib/systeminfo/dns.go
```diff
@@ -0,0 +1,131 @@
...
+		line = strings.Split(line, "#")[0]
+
+		f := strings.Fields(line)
+		if len(f) < 1 {
+			continue
+		}
+
+		switch f[0] {
+		case "nameserver":
+			if len(f) > 1 && len(resolv.Servers) < 3 { // limit number of servers
```
a-palchikov: Does it make sense to enforce the same restriction as the [resolv.conf](http://man7.org/linux/man-pages/man5/resolv.conf.5.html) itself or can it use more than 3?

lib/systeminfo/dns_test.go
```diff
@@ -0,0 +1,111 @@
...
+				Ndots:    1,
+				Timeout:  5,
+				Attempts: 2,
+				Servers:  []string{"1.1.1.1"},
+			},
+		},
+		{
+			`
+			# domain domain.com
+			nameserver 1.1.1.1`, // <- missing newline at EOF
```
a-palchikov: the comment could be a literal to use with `assert.Nil(t, err, tt.comment)`:
```go
	`
	# domain domain.com
	nameserver 1.1.1.1`,
	comment: "missing newline at EOF",
```

lib/systeminfo/dns_test.go
```diff
@@ -0,0 +1,111 @@
...
+				Search:   []string{"a.com", "b.com", "c.com"},
+				Rotate:   true,
+			},
+		},
+		{
+			`
+			options ndots:-1
+			options timeout:-5
+			options attempts:-5
+			options blahblah
```
a-palchikov: nit: `unknownoption` instead of `blahblah`? :-)

lib/systeminfo/dns_test.go
```diff
@@ -0,0 +1,111 @@
...
+				Timeout:    1,
+				Attempts:   1,
+				UnknownOpt: true,
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		r, err := ResolvFromReader(strings.NewReader(tt.in))
+		fmt.Print(err)
```
a-palchikov: Isn't `Nil` going to dump the error if the assertion has failed?

lib/update/phase_coredns.go
```diff
@@ -143,5 +162,70 @@ func shouldUpdateCoreDNS(client *kubernetes.Clientset) (bool, error) {
...
+	}
+	return false, nil
+}
+
+// generateCorefile will generate a coredns corefile, only if not already present on the system
+// with settings from the cluster configuration and local system. It should not overwrite an existing
+// Corefile, as that may have been modified by a user.
+func (p *updatePhaseCoreDNS) generateCorefile(context.Context) error {
+	p.Info("Generating coredns corefile.")
+	// Read the resolv.conf from the host doing installation
```
a-palchikov: `during update`

lib/systeminfo/dns_test.go
knisbet: It could... I don't think I would remove the comment pointing at the character indicating it should be missing a newline at EOF though, as it makes it very clear that it shouldn't be reformatted to match the other blocks, and if someone does reformat it by accident, it won't cause the test to begin failing.

lib/update/phase_kubernetes.go
```diff
@@ -392,7 +392,7 @@ func supportsTaints(gravityPackage loc.Locator) (supports bool, err error) {
 
 func waitForEndpoints(ctx context.Context, client corev1.CoreV1Interface, server storage.Server) error {
 	clusterLabels := labels.Set{"app": defaults.GravityClusterLabel}
-	kubednsLegacyLabels := labels.Set{"k8s-app": "kube-dns"}
+	kubednsLegacyLabels := labels.Set{"k8s-app": "kubedns"}
```
a-palchikov: Now both legacy and [current](https://github.com/gravitational/gravity/pull/113/files#diff-f9e91517c31865c93494fba9442eeff9R807) DNS labels are `kubedns`

lib/systeminfo/dns.go
knisbet: It's a good point... this is the behaviour we had in planet, and I simple maintained it.

I'll update it though, as I think the argument can be made that this should parse and represent the file as it exists on the system, and not the limitations of that system.

lib/defaults/defaults.go
```diff
@@ -806,6 +806,9 @@ const (
 	// KubeDNSLabel defines the label to select cluster DNS service Pods
 	KubeDNSLabel = "kubedns"
```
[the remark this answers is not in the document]
knisbet: I've reversed it... and you're right, it was incorrect previously, and i was originally just going to maintain that.

lib/install/phases/coredns_test.go
knisbet: It's an interesting topic. I haven't explicitly tested or reviewed the CoreDNS code to see if it'll accept this miss-configuration. My basic thought would be, if CoreDNS accepts this, it doesn't do us any harm to just pass along the setting, as it's a truer reflection of the original system configuration. On the other hand, if we rewrite configuration we believe is incorrect, it's less likely to fail now, or in the future as the software is patched.

I think I naturally lean towards passing along this type of miss-configuration as is when mapping the resolv.conf to coredns configuration, vs. say rewriting it. If this is problematic configuration downstream though, I think it's an interesting question on where it should be handled, and bubble up to the user.

lib/systeminfo/dns_test.go
knisbet: Sorry, was some debugging that got left in place... I removed it.

lib/update/phase_kubernetes.go
knisbet: You're watching too quickly, I did update both, just missed committing the update to defaults in the same push.

lib/install/phases/coredns.go
```diff
@@ -0,0 +1,176 @@
...
+func (r *corednsExecutor) PreCheck(context.Context) error {
+	return nil
+}
+
+// PostCheck is no-op for this phase
+func (r *corednsExecutor) PostCheck(context.Context) error {
+	return nil
+}
+
+// Execute generated coredns configuration
```
r0mant: generate**s**

lib/install/phases/coredns.go
```diff
@@ -0,0 +1,176 @@
...
+func (r *corednsExecutor) Execute(ctx context.Context) error {
+	r.Progress.NextStep("Configuring Coredns")
```
r0mant: Coredns -> CoreDNS here and below

lib/install/phases/coredns.go
```diff
@@ -0,0 +1,176 @@
...
+		Hosts:               r.DNSOverrides.Hosts,
+		Zones:               r.DNSOverrides.Zones,
+	})
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	_, err = r.Client.CoreV1().ConfigMaps(constants.KubeSystemNamespace).Create(&v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "coredns",
```
r0mant: Should the name of the CoreDNS ConfigMap also move to constants/defaults? It is being used in multiple places as far as I can see.

lib/install/phases/coredns.go
```diff
@@ -0,0 +1,176 @@
...
+	err := coreDNSTemplate.Execute(&coredns, config)
+	if err != nil {
+		return "", trace.Wrap(err)
+	}
+	return coredns.String(), nil
+}
+
+// CoreDNSConfig represents the CoreDNS configuration options to apply to our template
+type CorednsConfig struct {
+	Zones               map[string][]string
```
r0mant: Can you add Godocs on these exported fields pls?

lib/install/phases/coredns.go
```diff
@@ -0,0 +1,176 @@
...
+	Hosts               map[string]string
+	UpstreamNameservers []string
+	Rotate              bool
+}
+
+var coreDNSTemplate = template.Must(template.New("coredns").Parse(coreDNSTemplateText))
+var coreDNSTemplateText = `
```
r0mant: Can be `const`.

lib/systeminfo/dns.go
```diff
@@ -0,0 +1,131 @@
...
+			}
+
+		case "lookup":
+			// OpenBSD option:
+			// http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man5/resolv.conf.5
+			// "the legal space-separated values are: bind, file, yp"
+			resolv.Lookup = f[1:]
+
+		default:
+			resolv.UnknownOpt = true
```
r0mant: What's the purpose of this? Wouldn't it be more useful if it contained a list of unknown options rather than just a bool value?

lib/update/phase_coredns.go
```diff
@@ -143,5 +162,70 @@ func shouldUpdateCoreDNS(client *kubernetes.Clientset) (bool, error) {
...
+		return true, trace.Wrap(err)
+	}
+	return false, nil
+}
+
+// generateCorefile will generate a coredns corefile, only if not already present on the system
+// with settings from the cluster configuration and local system. It should not overwrite an existing
+// Corefile, as that may have been modified by a user.
+func (p *updatePhaseCoreDNS) generateCorefile(context.Context) error {
+	p.Info("Generating coredns corefile.")
```
r0mant: CoreDNS

lib/systeminfo/dns.go
```diff
@@ -0,0 +1,131 @@
...
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/gravitational/gravity/lib/storage"
+	"github.com/gravitational/trace"
+)
+
+func ResolvFromFile(filename string) (*storage.ResolvConf, error) {
+	path, err := filepath.EvalSymlinks(filename)
```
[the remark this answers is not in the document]
knisbet: This was the way it was originally done in planet, and I wasn't sure if it was written this way for a specific reason or not:
https://github.com/gravitational/planet/blob/version/5.0.x/tool/planet/start.go#L515-L517

lib/systeminfo/dns.go
knisbet: I agree it would be more useful, however, this is a port of how it was implemented within planet: https://github.com/gravitational/planet/blob/39b6100646fa7727057aafacd62b1f9b9cecc56d/lib/utils/dns.go#L143-L145

PR #116, opened 2018-11-27
Exit after operation failure

a-palchikov:
This PR adds a knob to exit a failed installation/join operation necessary in automated environments (i.e. robotest or script-driven installation).

Also fixes an issue with teleport S3 location for `pull-teleport`.

Requires https://github.com/gravitational/gravity.e/pull/3829.
Fixes https://github.com/gravitational/gravity.e/issues/3397.
Fixes https://github.com/gravitational/gravity.e/issues/3790.

### DIFF
--- a/build.assets/Makefile
+++ b/build.assets/Makefile
@@ -12,7 +12,9 @@ LOCAL_BUILDDIR ?= /gopath/src/github.com/gravitational/gravity/build
 # Path to the local gravity build assets directory used inside the container
 LOCAL_GRAVITY_BUILDDIR ?= /gopath/src/github.com/gravitational/gravity/build/$(GRAVITY_VERSION)
 
+DOCKER_ARGS ?= --pull
 BBOX = gravity-buildbox:latest
+GOCACHE ?= $(HOME)/.cache/go-build
 
 GRAVITY_WEB_APP_DIR ?= $(abspath $(LOCALDIR)/../web)/
 
@@ -196,6 +198,8 @@ validate-deps: buildbox
 build-in-container: buildbox
 	docker run --rm=true -u $$(id -u) \
 		   -v $(TOP):$(SRCDIR) \
+		   -v $(GOCACHE):/gocache \
+		   -e "GOCACHE=/gocache" \
 		   -e "LOCAL_BUILDDIR=$(LOCAL_BUILDDIR)" \
 		   -e "LOCAL_GRAVITY_BUILDDIR=$(LOCAL_GRAVITY_BUILDDIR)" \
 		   -e "GRAVITY_PKG_PATH=$(GRAVITY_PKG_PATH)" \
@@ -458,7 +462,7 @@ clone-teleport:
 	cd $(TELEPORT_SRCDIR)/teleport && git fetch --all --tags && git checkout $(TELEPORT_REPOTAG)
 
 pull-teleport:
-	-$(AWS) s3 cp $(S3_OPTS) $(BUILD_BUCKET_URL)/teleport/$(TELEPORT_TAG) $(TELEPORT_OUT)
+	-$(AWS) s3 cp $(S3_OPTS) $(BUILD_BUCKET_URL)/teleport/$(TELEPORT_TAG)/$(TELEPORT_S3_BUCKET) $(TELEPORT_OUT)
 
 push-teleport: $(TELEPORT_OUT)
 	$(AWS) s3 cp $(S3_OPTS) $(TELEPORT_OUT) $(BUILD_BUCKET_URL)/teleport/$(TELEPORT_TAG)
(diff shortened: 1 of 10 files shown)


build.assets/Makefile
```diff
@@ -196,6 +198,8 @@ validate-deps: buildbox
 build-in-container: buildbox
 	docker run --rm=true -u $$(id -u) \
 		   -v $(TOP):$(SRCDIR) \
+		   -v $(GOCACHE):/gocache \
```
a-palchikov: This shares the go build cache directory with the container speeding up rebuilds. Otherwise, it takes 1.5min to 2min for subsequent `make build` invocations (i.e. w/o changes to source code).

build.assets/Makefile
```diff
@@ -554,4 +558,4 @@ buildbox:
 		--build-arg GOGO_PROTO_TAG=$(GOGO_PROTO_TAG) \
 		--build-arg GRPC_GATEWAY_TAG=$(GRPC_GATEWAY_TAG) \
 		--build-arg VERSION_TAG=$(VERSION_TAG) \
-		--pull --tag $(BBOX) .
+		$(DOCKER_ARGS) --tag $(BBOX) .
```
a-palchikov: I always end up disabling the pull flag in the makefile as it takes me forever.

lib/expand/join.go
```diff
@@ -656,12 +660,8 @@ func (p *Peer) send(e install.Event) {
 	select {
 	case p.EventsC <- e:
 	case <-p.Context.Done():
-		select {
-		case p.EventsC <- install.Event{Error: trace.ConnectionProblem(p.Context.Err(), "context is closing")}:
```
a-palchikov: I don't think this is necessary as all critical loops are already listening on context expiration/cancellation.

tool/gravity/cli/install.go
```diff
@@ -87,7 +87,7 @@ func startInstall(env *localenv.LocalEnvironment, i InstallConfig) error {
 
 	err = installer.Wait()
 	if utils.IsContextCancelledError(err) {
-		return nil
+		return trace.BadParameter("cancelled")
```
a-palchikov: Simple `cancelled` looks better than default `context cancelled` which reads super technical.

a-palchikov:
@r0mant, @kontsevoy I've updated the flag to be `--exit-after-error`, this indeed is a better name.

tool/gravity/cli/register.go
```diff
@@ -84,6 +84,9 @@ func RegisterCommands(app *kingpin.Application) *Application {
 	g.InstallCmd.Force = g.InstallCmd.Flag("force", "Force phase execution").Bool()
 	g.InstallCmd.Resume = g.InstallCmd.Flag("resume", "Resume installation from last failed step").Bool()
 	g.InstallCmd.Manual = g.InstallCmd.Flag("manual", "Manually execute install operation phases").Bool()
+	g.InstallCmd.ExitAfterError = g.InstallCmd.Flag("exit-after-error", "Exit after operation failure without waiting for input").
```
r0mant: Sorry to be pedantic, but I think `--exit-on-failure` sounds better :) I'm fine with either one though.

klizhentas:
why do we need a flag? can we just detect if the terminal (pty) is allocated, and if not, exit on failure.

r0mant:
What if you want to launch it in unattended mode (w/o tty) but then still have a chance to see the plan / resume in case of failure?

a-palchikov:
I'd be happier with an explicit flag instead of implicit logic and having no idea why it did not work.

klizhentas:
why would you want to do that? the unattended mode but with pause?

a-palchikov:
My primary use-case is robotest - how would I be able to start the installation/expand remotely, fetch output from the command and detect that I need to run in actual unattended mode vs half-way unattended (the default) when it pauses during CLI operation?

a-palchikov:
I agree - the only reason we pause the installer is the otherwise volatile state. If there's a way to keep state and restart to known state - there's no reason for pausing.
I'll have a look at what can be done to retain state between restarts (i.e. in the simplest case just storing the path to state directory upon exit and reading from the same directory upon restart).
It should be possible to do the same w.r.t plan viewing.

kontsevoy:
I was under impression that `gravity install` is "unattended" and should basically exit with error by default. Then we have `gravity wizard` which is a Web UI driven flow. No?

a-palchikov:
I'm going to split this PR into multiple steps:
  * major part is true unattended mode which deserves more design and investment and will likely follow the path suggested by @knisbet - i.e. install/expand process will always exit regardless of the operation outcome. It should be possible to resume a failed operation though.
  * this PR will only include changes necessary to return an error (vs success) if the paused (i.e. failed) process is interrupted + plus non-mandatory Makefile changes

PR #126, opened 2018-11-29
Fix export of planet journald logs

a-palchikov:
Now, when the debug report is generated, depending on the options the host `systemd` was built with, the `journalctl` tool might not be compatible with the journal format used inside the planet container resulting in its inability to work with planet's `journald` logs.

To fix this, export planet `journald` logs by `chroot`ing into planet package `rootfs` and using the compatible binary.

Tested on all supported distributions: `ubuntu:16`, `redhat:7.5`, `centos:7.5`, `debian:9`, `suse:12`

Fixes https://github.com/gravitational/gravity.e/issues/2994.


tool/gravity/cli/register.go
```diff
@@ -576,6 +576,13 @@ func RegisterCommands(app *kingpin.Application) *Application {
 	g.SystemDisablePromiscModeCmd.CmdClause = g.SystemCmd.Command("disable-promisc-mode", "Remove promiscuous mode flag and deduplication rules").Hidden()
 	g.SystemDisablePromiscModeCmd.Iface = g.SystemDisablePromiscModeCmd.Arg("name", "Name of the interface (i.e. docker0)").Default(defaults.DockerBridge).String()
 
+	g.SystemExportRuntimeJournalCmd.CmdClause = g.SystemCmd.Command("export-runtime-journal", "Export runtime journal logs to a file").Hidden()
+	g.SystemExportRuntimeJournalCmd.Package = Locator(g.SystemExportRuntimeJournalCmd.Arg("pkg", "Runtime package").Required())
```
r0mant: Runtime package can probably be auto-detected if not specified, no? By installed/runtime labels. Wouldn't it make this command easier to use?

tool/gravity/cli/journal.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	if err := system.Chroot(rootDir); err != nil {
+		return trace.Wrap(err)
+	}
+
+	args := []string{
+		"--since=yesterday",
```
r0mant: Why not everything?

tool/gravity/cli/journal.go
```diff
@@ -0,0 +1,120 @@
...
+		"--output=export",
+		"-D",
+		journalDir,
+	}
+	cmd := "/bin/journalctl"
```
r0mant: We have `defaults.JournalctlBin` (although, it points to `/usr/bin/journalctl` for some reason).

tool/gravity/cli/journal.go
```diff
@@ -0,0 +1,120 @@
...
+
+	runtimePath, err := env.Packages.UnpackedPath(runtimePackage)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	rootDir := filepath.Join(runtimePath, "rootfs")
+	logDir := filepath.Join(state.LogDir(stateDir), "journal")
+	m := mount.NewMounter(rootDir)
+	if err := m.RoBindMount(logDir, journalDir); err != nil {
```
r0mant: What is this mount used for? Wouldn't just chrooting, doing export and then grabbing it from planet's rootfs work?

tool/gravity/cli/journal.go
UnderGreen: in almost every modern linux distributions `/bin` and `/sbin` are symlinks to `/usr/bin` and `/usr/sbin`.

tool/gravity/cli/journal.go
a-palchikov: It might megabytes of data if the cluster has been active for some time. I'd be cautious to fetch everything.

tool/gravity/cli/journal.go
a-palchikov: After chroot, there's no way to access `<gravity-state-dir>/planet/log/journal` which lies outside of planet's rootfs, hence the need to mount it inside.

tool/gravity/cli/journal.go
UnderGreen: IMO, anyway, it should cover a situation, when on Monday you asked someone to get a report about something that happened last Friday. What is a common in support pipeline.

a-palchikov:
@knisbet If the mount is left unmounted, it will be mounted (as already existing) on the next run.
But, in case the collector _does not_ run, and then the gravity cluster is removed, then it will not be able to remove the rootfs directory.
Even with the signal handler, the mount might be left behind (i.e. in the unlikely case of an unexpected system shutdown or something similar). I guess for this to be 100% working, the clean up code needs to remove any lingering mounts.

tool/gravity/cli/journal.go
r0mant: Oh, it's cause /var/lib/gravity/planet is mounted inside planet. I see.

a-palchikov:
@knisbet I've addressed your comment w.r.t signal handling - it should be good now with interrupting the operation half-way.

PR #128, opened 2018-11-30
Nightly builds

a-palchikov:
This PR adds nightly builds with notifications to slack if the build has failed.

The test matrix is still in a bit of a flux as I'm trying to figure out which combination does not lead to quota violations, so this is still WIP.

Requires https://github.com/gravitational/gravity.e/pull/3845.

### DIFF
--- a/Jenkinsfile.nightly
+++ b/Jenkinsfile.nightly
@@ -7,6 +7,26 @@ def propagateParamsToEnv() {
   }
 }
 
+/*
+* sendBuildNotification sends a notification to slack if the build
+* has recovered from a failure (i.e. had failed previously)
+* or failed.
+*/
+def sendBuildNotification(String currentBuildResult, String prevousBuildResult) {
+  currentBuildResult = currentBuildResult ?: 'SUCCESS'
+  prevousBuildResult = prevousBuildResult ?: 'SUCCESS'
+
+  def color = "warning"
+  def message = "${env.JOB_NAME} ${env.BUILD_NUMBER} (<${env.BUILD_URL}|Open>)"
+  if currentBuildResult == 'SUCCESS' {
+    color = "good"
+  }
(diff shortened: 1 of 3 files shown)


Jenkinsfile.nightly
```diff
@@ -7,16 +7,50 @@ def propagateParamsToEnv() {
...
+  if (shouldNotify) {
+    slackSend(color: color, message: message)
+  }
 }
 
 properties([
   disableConcurrentBuilds(),
-  pipelineTriggers(triggers),
+  // Schedule a daily build between 6:00am and 6:30am UTC (10:00am-10:30pm PST)
+  pipelineTriggers([cron('H(0-30) 06 * * 1-5')]),
```
knisbet: 10:00pm PST

a-palchikov:
Yes, I could not increase limits myself so I guess someone with "admin" privileges should be able to bump that. But that's not the issue - I've not been able to get a single success run using preemptible instances which means that we either need to switch to regular instances or handle preemption differently (i.e. by not handling them as failures and retrying then over the current failed limit of 3).
I'm currently experimenting with second option to see if it has merit.

Jenkinsfile.nightly
```diff
@@ -39,79 +74,90 @@ properties([
...
     string(name: 'PARALLEL_TESTS',
-           defaultValue: '1',
+           defaultValue: '4',
            description: 'Number of parallel tests to run.'),
     string(name: 'REPEAT_TESTS',
            defaultValue: '1',
            description: 'How many times to repeat each test.'),
     string(name: 'ROBOTEST_VERSION',
-           defaultValue: 'stable-gce',
+           defaultValue: 'dima',
```
a-palchikov: Changing back to `stable-gce` requires a couple more changes in robotest.

PR #135, opened 2018-12-07
[5.3] environment resource

a-palchikov:
This PR introduces a new Environment resource as described [here](https://github.com/gravitational/gravity.e/issues/3737#issuecomment-443849432).

The update operation has been implemented as a state machine and a rolling restart of nodes with new configuration.

Since the consensus was to keep `gravity resource` command as the trigger I've made a couple of changes I'm not yet happy with - this is due to the fact that spawning the update operation requires (as with other fsm-based operations) root privileges to deploy agents. This might be something I can do away with but requires additional time to improve.
The general fsm operations have been tucked under `gravity plan`:

 * `plan execute` to execute a particular phase
 * `plan rollback` to rollback a particular phase
 * `plan resume` to resume aborted operation
 * `plan display` or just `plan` to display the active or last completed operation

Requires https://github.com/gravitational/planet/pull/338 and https://github.com/gravitational/gravity.e/pull/3875.


lib/ops/opsclient/opsclient.go
```diff
@@ -1085,6 +1085,37 @@ func (c *Client) DeleteAlertTarget(key ops.SiteKey) error {
 	return trace.Wrap(err)
 }
 
+// GetClusterEnvironment retrieves the cluster environment
```
knisbet: So calling this "the cluster environment" is used fairly consistently throughout the PR. However, if I were to come across this not knowing this was related to Environment variables I suspect I would be pretty confused.

lib/ops/resources/gravity/gravity.go
```diff
@@ -335,6 +351,13 @@ func (r *Resources) GetCollection(req resources.ListRequest) (resources.Collecti
 			return nil, trace.Wrap(err)
 		}
 		return alertTargetCollection(alertTargets), nil
+	case storage.KindEnvironment, "environments", "env":
+		// always ignore name parameter for tls key pairs, because there is only one
```
knisbet: copy-paste error (tls key pairs)?

lib/kubernetes/nodes.go
```diff
@@ -239,14 +239,14 @@ func addTaints(oldTaints []v1.Taint, newTaints *[]v1.Taint) bool {
...
+// Retry retries the specified function fn using classify to determine
+// whether to Retry a particular error.
 // Returns the first permanent error
-func retry(ctx context.Context, fn func() error) error {
+func Retry(ctx context.Context, fn func() error) error {
 	interval := backoff.NewExponentialBackOff()
 	err := utils.RetryWithInterval(ctx, interval, func() error {
 		err := RetryOnUpdateConflict(fn())
-		return trace.Wrap(err)
+		return err
```
r0mant: Nit: just `return RetryOnUpdateConflict(fn())` then?

assets/rbac-app/resources/resources.yaml
```diff
@@ -278,4 +278,29 @@ subjects:
...
   apiGroup: rbac.authorization.k8s.io
   kind: Role
-  name: gravity:coredns
\ No newline at end of file
+  name: gravity:coredns
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: gravity:agent
```
r0mant: Which "agent" is being meant here?

lib/storage/environment.go
```diff
@@ -0,0 +1,174 @@
...
+
+	teleservices "github.com/gravitational/teleport/lib/services"
+	teleutils "github.com/gravitational/teleport/lib/utils"
+	"github.com/gravitational/trace"
+	"github.com/jonboulle/clockwork"
+)
+
+// Environment defines the environment variables resource.
+// It allows to override environment variables on each node in the cluster.
+// There is only a single instance of the resource in a cluster
```
knisbet: So I've honestly always found this a bit weird about how gravity emulates kubernetes resource type behaviour. We defined the object like a kubernetes object, with a name+namespace, but then the management of that is just a flat configuration that ignores those fields.

lib/storage/environment.go
```diff
@@ -0,0 +1,174 @@
...
+			Name:      constants.ClusterEnvironmentMap,
+			Namespace: defaults.Namespace,
+		},
+		Spec: EnvironmentSpec{
+			KeyValues: kvs,
+		},
+	}
+}
+
+// EnvironmentV2 describes the environment variable resource
```
knisbet: Why V2 if this is the first time environment has been added as an object??

lib/ops/operatoracl.go
```diff
@@ -768,6 +768,23 @@ func (o *OperatorACL) DeleteAlertTarget(key SiteKey) error {
...
+// GetClusterEnvironment retrieves the cluster environment
+func (o *OperatorACL) GetClusterEnvironment(key SiteKey) (storage.Environment, error) {
+	if err := o.ClusterAction(key.SiteDomain, storage.KindEnvironment, teleservices.VerbList); err != nil {
+		return nil, trace.Wrap(err)
+	}
+	return o.operator.GetClusterEnvironment(key)
+}
+
+// UpdateClusterEnvironment updates the cluster with the specified environment.
+// Returns the updated environment
```
r0mant: Doesn't look like it returns anything except for error.

lib/ops/resources/gravity/gravity.go
r0mant: The comment seems to have been copied from elsewhere.

lib/ops/resources/gravity/collection.go
```diff
@@ -458,8 +458,8 @@ func (r alertCollection) ToMarshal() interface{} {
 }
 
 // Resources returns the resources collection in the generic format
-func (c alertCollection) Resources() (resources []teleservices.UnknownResource, err error) {
-	for _, item := range c {
+func (r alertCollection) Resources() (resources []teleservices.UnknownResource, err error) {
```
r0mant: lol, I remember renaming these to 'c' some time ago because 'c' stands for "collection"; what does `r` stand for I wonder? :smile:

lib/ops/opsservice/environment.go
```diff
@@ -0,0 +1,92 @@
...
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+// GetClusterEnvironment retrieves the cluster environment
+func (o *Operator) GetClusterEnvironment(key ops.SiteKey) (env storage.Environment, err error) {
+	client, err := o.GetKubeClient()
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	configmap, err := client.CoreV1().ConfigMaps(defaults.KubeSystemNamespace).
```
knisbet: Why represent the object in kubernetes as a configmap instead of a crd? I think it's easier to implement as a configmap, just seems a bit odd that you get a resource implementation when using gravity, and a configmap representation when using kubectl.

lib/storage/environment.go
a-palchikov: This has to do with the teleport resource schema - i.e. it is the `v2` of the teleport schema.

lib/ops/opsservice/environment.go
a-palchikov: Yes, fair point. I did not think I'd bring much with turning it into a CRD other than the implementation would be more complex as the primary way of managing these resources are via the `gravity` binary.

lib/storage/environment.go
knisbet: This doesn't make sense to me... we bump all resource versions when the teleport schema get's bumped?

tool/gravity/cli/rpcagent.go
```diff
@@ -348,6 +344,10 @@ func rpcAgentShutdown(env *localenv.LocalEnvironment) error {
 	return trace.Wrap(err)
 }
 
+func executeSyncOperationPlan(ctx context.Context, localEnv, updateEnv *localenv.LocalEnvironment, args []string) error {
```
a-palchikov: Since I'm reusing some of the update operation infrastructure and behavior for updating the environment variables, 
instead of `SyncOperationPlanToCluster`, agents are now started to sync the operation plan as the first thing to do.
except on the initiator node that creates the plan and runs the update operation. This is also relevant for update envars operation.

lib/environ/environ.go
```diff
@@ -0,0 +1,258 @@
...
+	defer progress.Stop()
+
+	return trace.Wrap(machine.ExecutePhase(ctx, libfsm.Params{
+		PhaseID:  phase,
+		Progress: progress,
+		Force:    force,
+	}))
+}
+
+// Rollbackhase rolls back the specified phase.
```
knisbet: Rollbackhase -> Rollback**P**hase

lib/environ/environ.go
```diff
@@ -0,0 +1,258 @@
...
+	Runner libfsm.AgentRepository
+	// FieldLogger is the logger to use
+	log.FieldLogger
+	// Silent controls whether the process outputs messages to stdout
+	localenv.Silent
+	// Emitter outputs progress messages to stdout
+	utils.Emitter
+}
+
+// Updater executes a cobntrolled update of cluster environment variables
```
knisbet: cobntrolled -> controlled

a-palchikov: Yes, it seems that I'd need to update the planet's configuration package to persist the environment update.
Also, besides the fact that we had had bad experiences with container restarts previously on older kernels which hopefully would be less and less of a problem, there's nothing preventing the operation from doing the rolling restart.
I've spent enough time procrastinating over a set of generalizations for the fsm procedure and factoring out the rolling container restart and making it reusable is definitely a good idea.
I would prefer to do the restart manually at this point though as I don't have enough time capacity for the refactoring work, but I will create a ticket for me to return to this problem some time later.

a-palchikov:
Yes, it seems that I'd need to update the planet's configuration package to persist the environment update.
Also, besides the fact that we had had bad experiences with container restarts previously on older kernels which hopefully would be less and less of a problem, there's nothing preventing the operation from doing the rolling restart.
I've spent enough time thinking over a set of generalizations for the fsm procedure and factoring out the rolling container restart and making it reusable is definitely a step in the right direction.
I would prefer to do the restart manually at this point though as I don't have enough time capacity for the refactoring work, but I will create a ticket for me to return to this problem some time later.

lib/update/phase_kubernetes.go
```diff
@@ -406,26 +406,40 @@ func waitForEndpoints(ctx context.Context, client corev1.CoreV1Interface, server
 }
 
 func hasEndpoints(client corev1.CoreV1Interface, labels labels.Set, fn endpointMatchFn) error {
+	services, err := client.Services(metav1.NamespaceSystem).List(
```
a-palchikov: Had to hack around [this issue](https://github.com/gravitational/gravity.e/issues/3866)

tool/gravity/cli/register.go
```diff
@@ -133,17 +133,29 @@ func RegisterCommands(app *kingpin.Application) *Application {
...
-	g.PlanCmd.Output = common.Format(g.PlanCmd.Flag("output", "Output format for the plan, text, json or yaml").Short('o').Default(string(constants.EncodingText)))
-	g.PlanCmd.OperationID = g.PlanCmd.Flag("operation-id", "ID of the operation to display the plan for. It not specified, the last operation plan will be displayed").String()
-
-	g.RollbackCmd.CmdClause = g.Command("rollback", "Rollback actions")
-	g.RollbackCmd.Phase = g.RollbackCmd.Flag("phase", "Operation phase to rollback").Required().String()
-	g.RollbackCmd.PhaseTimeout = g.RollbackCmd.Flag("timeout", "Phase rollback timeout").Default(defaults.PhaseTimeout).Hidden().Duration()
-	g.RollbackCmd.Force = g.RollbackCmd.Flag("force", "Force phase rollback").Bool()
-	g.RollbackCmd.SkipVersionCheck = g.RollbackCmd.Flag("skip-version-check", "Bypass version compatibility check").Hidden().Bool()
+	g.PlanCmd.CmdClause = g.Command("plan", "Manage ongoing operation")
+	g.PlanCmd.OperationID = g.PlanCmd.Flag("operation-id", "ID of the active operation. It not specified, the last operation will be used").String()
```
knisbet: UX Thought. How does someone find the operation-id to pass if they don't want the active operation?

lib/environ/internal/fsm/engine.go
```diff
@@ -0,0 +1,294 @@
...
+
+// GetExecutor returns the appropriate phase executor based on the
+// provided parameters
+func (r *engine) GetExecutor(params libfsm.ExecutorParams, remote libfsm.Remote) (libfsm.PhaseExecutor, error) {
+	return r.spec(params, remote)
+}
+
+// RunCommand executes the phase specified by params on the specified server
+// using the provided runner
+func (r *engine) RunCommand(ctx context.Context, runner libfsm.RemoteRunner, server storage.Server, params libfsm.Params) error {
```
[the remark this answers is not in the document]
a-palchikov: Yes, it's an interface method.

lib/ops/operatoracl.go
```diff
@@ -494,6 +494,14 @@ func (o *OperatorACL) CreateClusterGarbageCollectOperation(req CreateClusterGarb
 	return o.operator.CreateClusterGarbageCollectOperation(req)
 }
 
+// CreateUpdateEnvarsOperation creates a new operation to update cluster environment variables
+func (o *OperatorACL) CreateUpdateEnvarsOperation(req CreateUpdateEnvarsOperationRequest) (*SiteOperationKey, error) {
```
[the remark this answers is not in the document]
a-palchikov: It's a create for `updateenvars` operation.

lib/ops/opsservice/environment.go
```diff
@@ -0,0 +1,147 @@
...
+	op := ops.SiteOperation{
+		ID:         uuid.New(),
+		AccountID:  s.key.AccountID,
+		SiteDomain: s.key.SiteDomain,
+		Type:       ops.OperationUpdateEnvars,
+		Created:    s.clock().UtcNow(),
+		Updated:    s.clock().UtcNow(),
+		State:      ops.OperationUpdateEnvarsInProgress,
+		UpdateEnvars: &storage.UpdateEnvarsOperationState{
+			PreviousEnv: configmap.Data,
```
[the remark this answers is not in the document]
a-palchikov: It used to be but not anymore - I'll remove the attribute.

tool/gravity/cli/register.go
a-palchikov: True, it's currently more of a low-level configuration. `gravity status` lists active/last operations, could be extended to output all operation with an additional flag I suppose.

lib/environ/internal/phases/restart.go
```diff
@@ -0,0 +1,123 @@
...
+// Execute restarts the runtime container with the new configuration package
+func (r *restart) Execute(ctx context.Context) error {
+	commands := [][]string{
+		{"system", "pull-updates",
+			"--uid", r.serviceUser.UID,
+			"--gid", r.serviceUser.GID,
+			"--runtime-package", r.runtimePackage.String(),
+			"--ops-url", defaults.GravityServiceURL,
+			"--insecure", "--quiet"},
+		{"system", "update",
```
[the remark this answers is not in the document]
a-palchikov: Yes, that's the idea - restart if either (or both) changes. Since we're rolling out a new configuration, the container needs to restart to pick it up.

lib/install/phases/pull.go
```diff
@@ -165,10 +165,15 @@ func (p *pullExecutor) applyPackageLabels() error {
 		constants.TeleportPackage,
 		constants.GravityPackage,
 	}
+	purposeLabels := []string{
+		pack.PurposePlanetConfig,
+		pack.PurposePlanetSecrets,
+	}
 	var locators []loc.Locator
-	err := pack.ForeachPackageInRepo(p.LocalPackages, defaults.SystemAccountOrg,
+	err := pack.ForeachPackage(p.LocalPackages,
```
a-palchikov: This is to be able to update runtime (planet) configuration and secret package labels as these are not in the default repository but in cluster's.

PR #136, opened 2018-12-11
Wireguard support

knisbet:
Updates https://github.com/gravitational/gravity.e/issues/3797

Add support for custom CNI plugins / wormhole CNI plugin
- Add new application hooks, for install/upgrade/rollback of additional CNI plugins, and insert into install/upgrade plans
Note: Using a CNI plugin is only supported at installation time, migrating between flannel and a CNI plugin on a live cluster isn't supported
- Separate /wait in installation plan into two phases, /wait and /health.
  - /wait only waits for the kubernetes API server to become available, before allowing kubernetes configuration/installation steps to proceed
  - /health occurs later in the installation, which is the full satellite health checks, specifically after CNI installation has been completed
- Add new phase for running CNI installation job

- The app manifest can be configured with a network type of wormhole, which will cause tele to automatically generate the hooks matching the tele version.


lib/install/phases/postsystem.go
```diff
@@ -67,10 +70,91 @@ type waitExecutor struct {
...
 	// ExecutorParams is common executor params
 	fsm.ExecutorParams
+	// Client is the Kubernetes client
+	Client *kubernetes.Clientset
 }
 
 // Execute executes the wait phase
+// This waits for critical components to start within planet
 func (p *waitExecutor) Execute(ctx context.Context) error {
+	ctx, _ = context.WithTimeout(ctx, 5*time.Minute)
```
a-palchikov: nit: I don't think `cancel` should be suppressed but used to release whatever resources (goroutines or similar) if `waitForAPI` completes before the timeout:
```go
ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
defer cancel()
```

lib/install/phases/postsystem.go
```diff
@@ -67,10 +70,91 @@ type waitExecutor struct {
...
+	defer timer.Stop()
+	for {
+		select {
+		case <-timer.C:
+			_, err := p.Client.CoreV1().ComponentStatuses().Get("scheduler", metav1.GetOptions{})
+			if err != nil {
+				p.Info("Waiting for kubernetes API to start: ", err)
+				continue
+			}
+			done <- true
```
a-palchikov: You can just `close(done)` with the same result.

lib/install/fsmspec.go
```diff
@@ -59,7 +60,16 @@ func FSMSpec(config FSMConfig) fsm.FSMSpecFunc {
 				config.Operator, remote)
 
 		case p.Phase.ID == phases.WaitPhase:
+			client, err := httplib.GetClusterKubeClient(config.DNSConfig.Addr())
```
a-palchikov: Not sure if @r0mant was going to revert the API change but this now returns 3 values

lib/schema/hooks.go
```diff
@@ -220,6 +232,8 @@ func AllHooks() []HookType {
 		HookDump,
 		HookBackup,
 		HookRestore,
+		HookOverlayInstall,
+		HookOverlayUpdate,
```
a-palchikov: `HookOverlayRollback`?

lib/schema/hooks.go
```diff
@@ -281,6 +295,10 @@ func HookFromString(hookType HookType, manifest Manifest) (*Hook, error) {
 		hook = manifest.Hooks.Backup
 	case HookRestore:
 		hook = manifest.Hooks.Restore
+	case HookOverlayInstall:
+		hook = manifest.Hooks.OverlayInstall
+	case HookOverlayUpdate:
+		hook = manifest.Hooks.OverlayUpdate
```
a-palchikov: Also, `HookOverlayRollback` handling missing?

assets/wormhole/resources/app.yaml
```diff
@@ -0,0 +1,154 @@
+apiVersion: bundle.gravitational.io/v2
+kind: Bundle
+metadata:
+  name: telekube
```
r0mant: Why do we need a separate app which seems to be exactly like "telekube", but with a different network provided? Wouldn't it be better to add a flag to gravity install to allow selecting network type (like we do with some other fields, e.g. storage driver)?

Or will it come later and this app is temporary?

lib/schema/hooks.go
```diff
@@ -77,6 +77,12 @@ type Hooks struct {
 	Backup *Hook `json:"backup,omitempty"`
 	// Restore restores application state from a backup
 	Restore *Hook `json:"restore,omitempty"`
+	// OverlayInstall is a hook for installing a custom overlay network
+	OverlayInstall *Hook `json:"overlayInstall,omitempty"`
```
r0mant: I wonder if `networkInstall` would be a better choice of word (since "overlay" can refer not just to network but e.g. to storage driver too so may lead to confusion). Same for other hooks.

lib/defaults/defaults.go
```diff
@@ -1066,6 +1066,10 @@ var (
 	// TransientErrorTimeout specifies the maximum amount of time to attempt
 	// an operation experiencing transient errors
 	TransientErrorTimeout = 15 * time.Minute
+
+	// WormholeImg is the docker image reference to use when embedding wormhole
+	// Note: This is a build parameter, and the build scripts will replace this with an image reference
+	WormholeImg = "<build param>"
```
r0mant: Maybe add some kind of init-time assertion that this has been replaced during build time, e.g. something like:

```go
func init() {
  if WormholeImg == "<build param>" {
    panic("defaults.WormholeImg wasn't set during build time. Did you build gravity using Makefile?")
  }
}
```

assets/wormhole/resources/app.yaml
knisbet: @klizhentas Didn't want to add more installation flags, and the way it's currently built wormhole only get's vendored / included if configured by the app maintainer.

So with that limit in place, the question is now whether we should publish this app to get.gravitational/hub as well.

lib/defaults/defaults.go
knisbet: it's a good point, but would this trigger when running things like test's, where we may not care if the parameter has been set...

PR #138, opened 2018-12-11
[5.2] Revert removal of update triggering from UI

a-palchikov:
This PR reverts the removal of code that triggers update from UI. I accidentally removed it when working on another issue thinking it would not be needed.

I also found a couple of issues in update when it was triggered from UI:
  - set of remote command executed to prepare and spawn the update engine would not create the agent directory (this was synced from lib/rpc/deploy.go)
  - `agent deploy upgrade` command would erroneously use the last (unfinished) update operation from local backend as the backend sync was missing

Additionally, this adds proper logging to update operation (like install) using the `fms.Logger`.

### DIFF
--- a/lib/localenv/clusterenv.go
+++ b/lib/localenv/clusterenv.go
@@ -135,6 +135,7 @@ func newClusterEnvironment(args clusterEnvironmentArgs) (*ClusterEnvironment, er
 		Apps:     apps,
 		Users:    users,
 		StateDir: siteDir,
+		Local:    true,
 	})
 	if err != nil {
 		return nil, trace.Wrap(err)
--- a/lib/ops/opsservice/update.go
+++ b/lib/ops/opsservice/update.go
@@ -292,7 +292,7 @@ func (s *site) createUpdateOperation(req ops.CreateSiteAppUpdateOperationRequest
 		Message:    "initializing the operation",
 	})
 
-	if req.Manual {
+	if !req.StartAgents {
 		return key, nil
 	}
 
--- a/lib/ops/opsservice/update.go
+++ b/lib/ops/opsservice/update.go
@@ -17,10 +17,13 @@ limitations under the License.
 package opsservice
 
 import (
+	"context"
 	"path/filepath"
 
+	"github.com/gravitational/gravity/lib/app"
 	"github.com/gravitational/gravity/lib/checks"
 	"github.com/gravitational/gravity/lib/constants"
+	"github.com/gravitational/gravity/lib/defaults"
 	"github.com/gravitational/gravity/lib/loc"
 	"github.com/gravitational/gravity/lib/ops"
 	"github.com/gravitational/gravity/lib/pack"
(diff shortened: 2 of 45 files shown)


lib/localenv/clusterenv.go
```diff
@@ -135,6 +135,7 @@ func newClusterEnvironment(args clusterEnvironmentArgs) (*ClusterEnvironment, er
 		Apps:     apps,
 		Users:    users,
 		StateDir: siteDir,
+		Local:    true,
```
a-palchikov: This is needed (again) as using `fsm.Logger` with an operator from clusterenv would not log to the correct location.

lib/ops/opsservice/update.go
```diff
@@ -260,23 +263,53 @@ func (s *site) createUpdateOperation(req ops.CreateSiteAppUpdateOperationRequest
...
+	resetClusterState := func() {
 		if err == nil {
 			return
 		}
-		errReset := s.setSiteState(ops.SiteStateActive)
+
+		// Fail the operation and reset cluster state.
+		// It is important to complete the operation as subsequent same type operations
+		// will not be able to complete if there's an existing incomplete one
+		errReset := ops.FailOperation(op.Key(), s.service, trace.Unwrap(err).Error())
```
a-palchikov: I'm not removing the operation as I want to persist whatever operation logs that have been written to this point.

lib/update/phase_app.go
```diff
@@ -187,34 +186,31 @@ func (p *phaseApp) runHooks(ctx context.Context, hooks ...schema.HookType) error
 		_, err := app.CheckHasAppHook(p.Apps, req)
 		if err != nil {
 			if trace.IsNotFound(err) {
-				log.Debugf("%v does not have %v hook.", p.Package, hook)
+				p.Debugf("%v does not have %v hook.", p.Package, hook)
```
a-palchikov: This cannot use `WithFields` as `fsm.Logger` cannot fully implement `logrus.FieldLogger` - namely, `WithXXX` methods are supposed to return `*logrus.Entry` instead of expected `logrus.FieldLogger`

tool/gravity/cli/run.go
```diff
@@ -231,13 +231,13 @@ func Execute(g *Application, cmd string, extraArgs []string) error {
 	defer localEnv.Close()
 
 	// create an environment used during upgrades
-	var upgradeEnv *localenv.LocalEnvironment
+	var updateEnv *localenv.LocalEnvironment
```
a-palchikov: nit: this is for naming consistency

lib/ops/opsservice/update.go
```diff
@@ -260,23 +263,53 @@ func (s *site) createUpdateOperation(req ops.CreateSiteAppUpdateOperationRequest
...
-	defer resetSiteState()
+	defer resetClusterState()
 
 	s.reportProgress(ctx, ops.ProgressEntry{
 		State:      ops.ProgressStateInProgress,
 		Completion: 0,
 		Message:    "initializing the operation",
 	})
 
+	if req.Manual {
```
r0mant: I think "Manual" is a bit misleading here as it can be confused with manual upgrade. Maybe "StartAgents" would be more appropriate field. Or is it used for anything else as well?

PR #159, opened 2018-12-19
Propagate node labels via kubelet.

r0mant:
Requires https://github.com/gravitational/planet/pull/352.


lib/ops/opsservice/configure.go
```diff
@@ -1021,6 +1025,22 @@ func (s *site) getPlanetConfigPackage(
 	}, nil
 }
 
+// getNodeLabels returns labels a Kubernetes node should register with
+func getNodeLabels(node *ProvisionedServer, profile *schema.NodeProfile) map[string]string {
+	labels := profile.Labels
+	for k := range labels {
```
a-palchikov:
```go
// getNodeLabels returns labels a Kubernetes node should register with
func getNodeLabels(node *ProvisionedServer, profile *schema.NodeProfile) map[string]string {
	labels := profile.Labels
	if labels == nil {
		labels = make(map[string]string)
	}
	if _, exists := labels[defaults.KubernetesRoleLabel]; exists {
		role := schema.ServiceRoleNode
		if node.IsMaster() {
			role = schema.ServiceRoleMaster
		}
		labels[defaults.KubernetesRoleLabel] = string(role)
	}
	labels[defaults.KubernetesAdvertiseIPLabel] = node.AdvertiseIP
	return labels
}
```

PR #164, opened 2018-12-20
fixed old link in quickstart

twakes:
We were linking to an old URL for the list of linux distros supported.


a-palchikov:
I believe this has already been taken care of [here](https://github.com/gravitational/gravity/commit/ff3ef2e019048ca3bf9714e1a9a0a8c30ab0fc88#diff-9491b56a238266d127151dd657483fc9R19).

PR #172, opened 2019-01-07
Update trusted cluster documentation

UnderGreen:


docs/5.x/cluster.md
```diff
@@ -1567,10 +1567,10 @@ $ gravity resource create trustedcluster.yaml
 View the currently configured trusted cluster:
 
 ```bsh
-$ gravity resource get trustedclusters
-Name                       Enabled     Pull Updates
-----                       -------     ------------
-opscenter.example.com      true        true
+$ gravity resource get trusted_cluster
```
a-palchikov: I believe `trustedcluster` (just as `trustedclusters`) is a supported [alias](https://github.com/gravitational/gravity.e/blob/master/lib/ops/resources/gravity/gravity.go#L294-L295)

UnderGreen: Yes, but our `gravity resource --help` says `Supported resources are: [cluster_auth_preference github user token logforwarder smtp alert alerttarget tlskeypair
    role oidc saml trusted_cluster endpoints]`. We have to update help or use what help says in documentation.

PR #176, opened 2019-01-10
restrict teleport ciphers

knisbet:
This PR restricts the ciphers for the embedded teleport instances within gravity.

@russjones Including you, so you can double check me on which ciphers are currently considered secure defaults.

If it looks good, this PR will be backported to all 5.x gravity releases.


lib/process/wizard.go
```diff
@@ -94,6 +94,22 @@ func WizardTeleportConfig(clusterName, stateDir string) *telecfg.FileConfig {
 				Params: make(backend.Params),
 			},
 			AuthServers: []string{fmt.Sprintf("localhost:%v", defaults.WizardAuthServerPort)},
+			CipherSuites: []string{
```
a-palchikov: Why not move these to consts? Used in at least 3 places

a-palchikov:
One of the ciphers is not supported:
```
ERROR REPORT:\nOriginal Error: *trace.BadParameterError cipher suite not supported: aes128-gcm@openssh.com
```

russjones:
Why exclude `chacha20-poly1305@openssh.com` from the list of ciphers?

knisbet:
@russjones It wasn't listed in https://gravitational.com/teleport/docs/admin-guide/#configuration-file so I wasn't sure if teleport supports chacha20-poly1305 / if it's supported in teleport 2.4 since I need to backport this change.

PR #179, opened 2019-01-10
Update README.md

klizhentas:


knisbet:
Hmm, I get the following error when following the link, when using my gravitational account: 

``
You can not invite members to this group.
Contact a group owner to request permission.
``

PR #180, opened 2019-01-10
Add security updates mailing list

klizhentas:


klizhentas:
@knisbet better mailing list link

PR #191, opened 2019-01-14
[5.3] gc: avoid deleting a package if another package has the same checksum

a-palchikov:
Cherry pick changes to avoid deleting a package if another existing package has the same checksum.
Forgot to include the change when I was working on [this](https://github.com/gravitational/gravity/pull/171) gc fix.


lib/pack/localpack/packageserver.go
```diff
@@ -434,6 +436,32 @@ func (p *PackageServer) ConfigurePackage(loc loc.Locator, confLoc loc.Locator, a
 	return pack.ConfigurePackage(p, loc, confLoc, args, nil)
 }
 
+func (p *PackageServer) tryDeleteBlob(toDelete storage.Package) error {
```
a-palchikov: Simplest possible implementation based on the fact that deletes are rare - check for collisions on delete.

a-palchikov: Simplest possible implementation based on the fact that deletes are rare - check for collisions on delete.

PR #193, opened 2019-01-15
[5.2] add ability to override connection idle timeout

a-palchikov:
Add a `WithIdleConnTimeout` functional option to override a connection idle timeout on the client's transport.

Updates https://github.com/gravitational/gravity.e/issues/3894.


a-palchikov:
I was not sure we needed the timeout for all cases.

PR #200, opened 2019-01-16
Update docs on `runtimeenvironment` resource, 'gravity plan'

a-palchikov:
Also ran everything through a spellchecker, hence multiple typos fixed.

I will update the upgrade section (remove mostly as it now duplicates information from the new `Managing an ongoing Operation` section that describes the updated `gravity plan`) once I've forward ported the `runtime_environment` implementation.

I also feel that all operation-related content should probably go into a separate file, like `operations.md`.

### DIFF
--- a/docs/5.x/changelog.md
+++ b/docs/5.x/changelog.md
@@ -9,7 +9,7 @@ LTS starts with `3.51.0` with minor backwards compatible changes added over time
 | --------------|-----| -------------------- | -------------------- | ------------------ |------------------|
 | 5.5.0-alpha.6 | No  | January, 15th 2019   | -                    | 1.12.3             | 3.0.1            |
 | 5.4.2         | No  | January, 8th 2019    | -                    | 1.13.0             | 2.4.10           |
-| 5.3.7         | No  | January, 17th 2019    | -                    | 1.12.3             | 2.4.10           |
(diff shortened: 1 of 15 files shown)


docs/5.x/changelog.md
```diff
@@ -9,6 +9,7 @@ LTS starts with `3.51.0` with minor backwards compatible changes added over time
 | --------------|-----| -------------------- | -------------------- | ------------------ |------------------|
 | 5.5.0-alpha.5 | No  | January, 8th 2019  | -                    | 1.12.3             | 3.0.1            |
 | 5.4.2         | No  | January, 8th 2019  | -                    | 1.13.0             | 2.4.10
+| 5.3.7-alpha.1 | No  | January, 16th 2019  | -                    | 1.12.3             | 2.4.10            |
```
r0mant: Why does it appear as a separate release? It's just 5.3.x, no?

docs/5.x/instance.md
```diff
@@ -4,7 +4,7 @@
 
 ### Create instance
 
-Choose debian image from community images:
+Choose a debian image from community images:
```
knisbet: This entire file probably shouldn't be in the OSS repo as it appears to be mostly internal, and looks to be largely out of date?

a-palchikov: It looks like it but I'd do this separately.

PR #205, opened 2019-01-17
autofix during planet start

knisbet:
Fixes https://github.com/gravitational/gravity.e/issues/3897

Link in the kernel module / sysctl checks when starting planet, and try and autofix any failures. This should address the install->uninstall->re-install issues described in https://github.com/gravitational/gravity.e/issues/3897. 

Also, remove the alias for br_netfilter / bridge, as on most distributions these are separate kernel modules. My understanding is this alias was created for rhel/centos 7.2, which we no longer need to support.


tool/gravity/cli/package.go
```diff
@@ -267,6 +269,11 @@ func executePackageCommand(s *localenv.LocalEnvironment, cmd string, loc loc.Loc
 		}
 	}
 
+	// Try and autofix sysctls / kernel modules during startup
+	if strings.ToLower(cmd) == "start" {
```
a-palchikov: `teleport` will also be `start`ed, so more context is needed I believe.

a-palchikov: I'm not sure about the full impact, but thinking about `teleport` being constantly restarted by systemd as it fails to connect as long as `gravity-site` is not up makes me uneasy.

knisbet:
@a-palchikov @r0mant I took a different approach with this after discussing with Sasha yesterday. Basically, always configure the complete list of modules in the systemd moduled-load.d, regardless of whether they're loaded at install time.

lib/checks/autofix/autofix.go
```diff
@@ -53,6 +55,25 @@ func Fix(ctx context.Context, probes []*agentpb.Probe, progress utils.Progress)
...
+// modules-load.d is smart enough to skip builtin modules, so we just add all modules we require to the list,
+// even if they are added or already running for other reasons.
+func AutoloadModules(ctx context.Context, modules []monitoring.ModuleRequest, progress utils.Progress) {
+	modList := make([]string, 0, len(modules))
+	for _, module := range modules {
+		modList = append(modList, module.Name)
+		if err := utils.EnsureLineInFile(defaults.ModulesPath, module.Name); err != nil && !trace.IsAlreadyExists(err) {
+			progress.PrintWarn(err, "Could not set up kernel module %v to load on boot", module)
+		}
+		for _, altName := range module.Names {
```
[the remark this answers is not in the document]
knisbet: Yea, it's just until someone goes and strips it from satellite, I don't want it to accidentally get re-used and then have some code somewhere that inconsistently doesn't work.

Issue #211, opened 2019-01-22
devicemapper default


knisbet:
Encountered while providing support in the gravity community slack, it looks like a minimal app.yaml set devicemapper as the default docker graph driver. Since we're no longer supporting devicemapper, we should check to make sure we don't have any defaults still pointing to devicemapper. 

```
apiVersion: bundle.gravitational.io/v2
kind: Bundle
metadata:
 name: test
 resourceVersion: "1.0.0"
```

klizhentas:
Duplicate of https://github.com/gravitational/gravity.e/issues/3913

PR #212, opened 2019-01-22
Set overlay driver as default in application manifest

knisbet:
Use overlay as the default storage driver. I don't know if overlay2 would be a better default, or if overlay is more likely to support a wider range of systems.

Updates https://github.com/gravitational/gravity/issues/211
Updates https://github.com/gravitational/gravity.e/issues/3913

Requires backport to 5.3.x/5.4.x.

### DIFF
--- a/docs/5.x/pack.md
+++ b/docs/5.x/pack.md
@@ -525,7 +525,8 @@ systemOptions:
 
   # Docker section allows to customize docker
   docker:
-    # Storage backend used, supported: "devicemapper" (default), "overlay", "overlay2"
+    # Storage backend used, supported: "devicemapper", "overlay" (default), "overlay2"
+    # Note: devicemapper is no longer available from gravity 5.3.4 or later.
     storageDriver: overlay
     # List of additional command line args to provide to docker daemon
     args: ["--log-level=DEBUG"]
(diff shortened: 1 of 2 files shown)


klizhentas:
I would prefer overlay2 - it has less issues (e.g no hard links exhaustion)

knisbet:
@klizhentas alright I switched it to overlay2

docs/5.x/pack.md
```diff
@@ -525,7 +525,8 @@ systemOptions:
 
   # Docker section allows to customize docker
   docker:
-    # Storage backend used, supported: "devicemapper" (default), "overlay", "overlay2"
+    # Storage backend used, supported: "devicemapper", "overlay", "overlay2" (default)
+    # Note: devicemapper is no longer available from gravity 5.3.4 or later.
```
r0mant: I would then just remove the reference to it from here and above. IIRC we have a note about it on the System Requirements page.

knisbet:
@r0mant I agree we should remove all related code at some point, I was just being a bit cautious, as if someone comes to us and really wants device mapper on a more recent release we might need to make it work. But then again, if that were to come up, we could try and back out the code removal.

PR #213, opened 2019-01-22
[5.3] Create environment variables during installation

a-palchikov:
This PR complements the last changes to add the new `RuntimeEnvironment` resource and adds the ability to setup runtime environment during install. It does this by extending the handling of resources specified with `gravity install --config=/path/to/resources.yaml`. The resources file can mix Kubernetes/Gravity resources.
Gravity resources found in the resources file will be created as the final step during install. Additionally, the runtime configuration package is created with specified environment if the resources contained the `RuntimeEnvironment` resource.

I removed the resources package as created by previous versions as now it is implemented by the install operation step.


lib/install/phases/resources.go
```diff
@@ -0,0 +1,187 @@
...
+	return &gravityExecutor{
+		FieldLogger: logger,
+		factory:     factory,
+		resources:   p.Phase.Data.Install.Resources,
+	}, nil
+}
+
+// Execute creates the Gravity resources from the configured list
+func (r *gravityExecutor) Execute(context.Context) (err error) {
+	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(r.resources), defaults.DecoderBufferSize)
```
r0mant: There is a `lib/app/resources/resourcefiles.go:ForEachObject()` method that nicely abstracts this resource decoding logic and invokes a method for each decoded resource. Can we use something similar here, for gravity resources?

lib/install/planbuilder.go
```diff
@@ -555,19 +625,70 @@ func splitServers(servers []storage.Server, app app.Application) (masters []stor
...
+		err = resources.NewResource(kubernetesResources...).Encode(&buf)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		builder.resources = buf.Bytes()
+	}
+	return nil
+}
+
+func splitResources(resourceBytes []byte) (kubernetesResources []runtime.Object, gravityResources []storage.UnknownResource, err error) {
```
r0mant: I wonder if this function should rather live somewhere in `lib/ops/resources`? Seems like a resources utility method rather than plan builder.

lib/install/planbuilder.go
```diff
@@ -555,19 +625,70 @@ func splitServers(servers []storage.Server, app app.Application) (masters []stor
...
+	reader := bytes.NewReader(resourceBytes)
+	decoder := yaml.NewYAMLOrJSONDecoder(reader, defaults.DecoderBufferSize)
+	for err == nil {
+		var resource storage.UnknownResource
+		err = decoder.Decode(&resource)
+		if err != nil {
+			break
+		}
+		kind := modules.Get().CanonicalKind(resource.Kind)
```
r0mant: You can just do

```go
resource.Kind = modules.Get().CanonicalKind(resource.Kind)
```
here, then you won't need `resource.Kind = kind` in the "else" (and it won't affect the first "if" branch).

lib/install/planbuilder.go
```diff
@@ -555,19 +625,70 @@ func splitServers(servers []storage.Server, app app.Application) (masters []stor
...
+		if resource.Version == "" && kind == "" {
```
r0mant: Why is this checking `resource.Version`? If "kind" returned empty, it should be enough indicator that this is not a gravity resource, no?

lib/install/planbuilder.go
```diff
@@ -555,19 +625,70 @@ func splitServers(servers []storage.Server, app app.Application) (masters []stor
...
+			if err := json.Unmarshal(resource.Raw, &kResource); err != nil {
+				return nil, nil, trace.Wrap(err)
+			}
+			kubernetesResources = append(kubernetesResources, &kResource)
+		} else {
+			resource.Kind = kind
+			gravityResources = append(gravityResources, resource)
+		}
+	}
+	if err == io.EOF {
```
r0mant: Do you need to unwrap the error?

lib/install/planbuilder.go
```diff
@@ -555,19 +625,70 @@ func splitServers(servers []storage.Server, app app.Application) (masters []stor
...
+		err = nil
+	}
+	if err != nil {
```
r0mant: Can just do

```go
if err != nil && !trace.IsEOF(err) {
}
```

and remove the if above.

lib/install/planbuilder.go
```diff
@@ -555,19 +625,70 @@ func splitServers(servers []storage.Server, app app.Application) (masters []stor
...
-		setup := b.Application.Manifest.SetupEndpoint()
-		if setup == nil || setup.ServiceName != defaults.BandwagonServiceName {
-			return true
+	if len(gravityResources) != 0 {
+		var buf bytes.Buffer
+		err = storage.Encode(gravityResources, &buf)
+		if err != nil {
+			return trace.Wrap(err)
 		}
+		builder.gravityResources = buf.Bytes()
```
r0mant: Do we not need to exclude RuntimeEnvironment resource from here as well? Won't it launch env update operation right after install otherwise?

lib/install/planbuilder.go
```diff
@@ -555,19 +625,70 @@ func splitServers(servers []storage.Server, app app.Application) (masters []stor
...
 	}
-	return false
+	for _, res := range gravityResources {
+		if res.Kind != storage.KindRuntimeEnvironment {
```
r0mant: I would clarify with a comment as to why this is being skipped.

lib/localenv/localenv.go
```diff
@@ -463,6 +464,11 @@ func (env *LocalEnvironment) SiteOperator() (*opsclient.Client, error) {
 	return operator, trace.Wrap(err)
 }
 
+// LocalClusterOperator returns Operator for the local gravity cluster
+func (env *LocalEnvironment) LocalClusterOperator() (ops.Operator, error) {
```
r0mant: Why not just rename `SiteOperator()` to `LocalClusterOperator()` then? :) Too many changes?

lib/storage/plan.go
```diff
@@ -141,6 +141,14 @@ type GarbageCollectOperationData struct {
 	RemoteApps []Application `json:"remote_apps,omitempty" yaml:"remote_apps,omitempty"`
 }
 
+// InstallOperationData describes configuration for the install operation
+type InstallOperationData struct {
+	// Env specifies optional cluster environment variables to add
+	Env map[string]string `json:"env,omitempty"`
+	// Resources specifies optional Kubernetes resources to create upon successful installation
```
r0mant: Kubernetes or Gravity resources?

lib/storage/resources.go
```diff
@@ -82,12 +91,54 @@ func CanonicalKind(kind string) string {
...
-	case KindRuntimeEnvironment, "environments", "env":
+	case KindRuntimeEnvironment, "environment", "env":
 		return KindRuntimeEnvironment
 	}
 	return ""
 }
 
+// UnknownResource represents an unparsed Gravity resource with an interpreted ResourceHeader
+type UnknownResource struct {
+	teleservices.ResourceHeader
```
r0mant: Could you add godocs on these two fields?

lib/install/planbuilder.go
a-palchikov: `io.EOF` is not wrapped as trace error type.

lib/install/planbuilder.go
a-palchikov: It won't but I will.

PR #240, opened 2019-01-29
Add 5.5.0-alpha.7 notes.

r0mant:


docs/5.x/pack.md
```diff
@@ -834,7 +834,30 @@ installer:
 !!! Note:
 	Currently, only one setup endpoint per application is supported.
 
+## Excluding System Applications
 
+By default, Gravity cluster installs with a number of system applications
+that provide logging, monitoring and application catalog functionality. You
+may want to disable any of these components, for example if you prefer to
+replace them with your own preferred solution. To do that, define the following
```
a-palchikov: with your own ~~preferred~~ solution.
or
`with a solution of your choice`
There's `prefer` before that and it should be sufficient.

r0mant: Good call, I'll update.

PR #243, opened 2019-01-30
update gravity-site healthz endpoint to more reliably indicate problems

knisbet:
Updates healthz processing, so that:
- Generate a timeout if the backend is slow/unresponsive instead of waiting until the client has timed out
- Log slow backend requests
- Removed sending raw backend error messages to clients (this endpoint allows unauthenticated access)

Extra:
- Set server cipher selection, to avoid a bug that's tripped me up a couple times, where the curl in planet can't communicate using http2 to the gravity-site server
- Update cipher selection to be stricter recommendations, using teleport defaults

### DIFF
--- a/lib/process/process.go
+++ b/lib/process/process.go
@@ -870,37 +870,34 @@ func (p *Process) ReportHealth(w http.ResponseWriter, r *http.Request) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
 
-	health := make(chan bool)
+	healthCh := make(chan error, 1)
 	go func() {
 		started := time.Now()
 		_, err := p.backend.GetAccounts()
-		if err == nil {
-			health <- true
-		} else {
-			log.Error(trace.DebugReport(err))
-			health <- false
-		}
 		elapsed := time.Now().Sub(started)
 		if elapsed > 25*time.Millisecond {
 			log.WithField("elapsed", elapsed).Error("Backend is slow.")
 		}
+		healthCh <- err
 	}()
 
 	select {
-	case healthy := <-health:
-		if healthy {
-			roundtrip.ReplyJSON(w, http.StatusOK,
+	case err := <-healthCh:
+		if err != nil {
+			log.Error(trace.DebugReport(err))
+			roundtrip.ReplyJSON(w, http.StatusServiceUnavailable,
 				map[string]string{
-					"status": "ok",
-					"info":   "service is up and running",
+					"status": "degraded",
+					"info":   "backend is in error state",
 				})
 			return
 		}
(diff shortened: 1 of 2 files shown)


lib/process/process.go
```diff
@@ -866,22 +866,48 @@ func (p *Process) ReportReadiness(w http.ResponseWriter, r *http.Request) {
...
+	defer cancel()
+
+	health := make(chan bool)
+	go func() {
+		started := time.Now()
+		_, err := p.backend.GetAccounts()
+		if err == nil {
+			health <- true
+		} else {
+			log.Error(trace.DebugReport(err))
```
a-palchikov: nit: I'd just send the error and log on the receiver.
```go
healthCh := make(chan error, 1)  // Also, note the buffer - for the case when the receiver is not fast enough
...
go func() {
	...
	_, err := p.backend,GetAccounts()
	healthCh <- err
	...
}()
```

vendor/github.com/gravitational/teleport/lib/utils/tls.go
```diff
@@ -122,7 +122,7 @@ func GenerateSelfSignedCert(hostNames []string) (*TLSCredentials, error) {
 		NotAfter:              notAfter,
 		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
 		BasicConstraintsValid: true,
-		IsCA: true,
+		IsCA:                  true,
```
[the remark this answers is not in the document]
knisbet: I think the vendoring was fine, just as I was debugging I apparently ran gofmt which produced a whitespace change.

lib/process/process.go
```diff
@@ -866,22 +866,45 @@ func (p *Process) ReportReadiness(w http.ResponseWriter, r *http.Request) {
...
+	log := p.WithField(trace.Component, "healthz")
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	healthCh := make(chan error, 1)
+	go func() {
+		started := time.Now()
+		_, err := p.backend.GetAccounts()
+		elapsed := time.Now().Sub(started)
+		if elapsed > 25*time.Millisecond {
```
r0mant: I'm curious how you landed on this number.

knisbet: I ran a perf test on one of my clusters, where the slowest request took about 15 ms, so I configured this with a small margin of head room, since it's only used in logging, and I'd like to see it trigger the log if things are only a little slow.

Ideally, we would have embedded metrics, and I would be able to scrape the histogram over a prometheus endpoint for all etcd queries, but I think wiring in metrics should be a bigger undertaking.

I'll add a comment before merging, so this is clearer when it comes back up.

PR #249, opened 2019-02-04
[5.4] Forward port

a-palchikov:
This PR forward-ports the following from 5.3:
  * https://github.com/gravitational/gravity/pull/226 (bugfix to use runtime environment in regular node configuration)
  * https://github.com/gravitational/gravity/pull/213 (create runtime environment during installation)
  * https://github.com/gravitational/gravity/pull/135 (runtime environment resource)
  * https://github.com/gravitational/gravity/pull/139 (fix triggering of upgrade from UI)

Requires https://github.com/gravitational/gravity.e/pull/3940.

:warning: Some changes were found to contain bugs which I will be fixing in the subsequent PR. This is to minimize the amount of changes for the next review.


lib/environ/internal/fsm/builder.go
```diff
@@ -0,0 +1,285 @@
...
+		ID:          id,
+		Description: fmt.Sprintf(format, server.Hostname),
+	}
+}
+
+// setLeaderElection creates a phase that will change the leader election state in the cluster
+// enable - the list of servers to enable election on
+// disable - the list of servers to disable election on
+// server - The server the phase should be executed on, and used to name the phase
+// key - is the identifier of the phase (combined with server.Hostname)
```
r0mant: `key` and `msg` don't match the actual function argument names.

lib/environ/internal/phases/kubernetes.go
```diff
@@ -0,0 +1,275 @@
...
+	"github.com/cenkalti/backoff"
+	"github.com/gravitational/trace"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/api/core/v1"
+	kubeapi "k8s.io/client-go/kubernetes"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+// NewTaint returns an executor for adding a taint to a node
+func NewTaint(params libfsm.ExecutorParams, client *kubeapi.Clientset, logger log.FieldLogger) (*tainter, error) {
```
r0mant: Is this executor still being used anywhere? Didn't we move to supplying taints via kubelet flags because this approach doesn't work on regular nodes in the new Kubernetes?

lib/install/install.go
```diff
@@ -214,6 +222,9 @@ func (c *Config) CheckAndSetDefaults() (err error) {
 	if c.AdvertiseAddr == "" {
 		return trace.BadParameter("missing AdvertiseAddr")
 	}
+	if c.LocalClusterClient == nil {
+		return trace.BadParameter("missing LocalClient")
```
r0mant: `LocalClusterClient`

lib/ops/ops.go
```diff
@@ -117,6 +117,7 @@ type Operator interface {
 	Install
 	Updates
 	Identity
+	EnvironmentVariables
```
r0mant: Nit: I think just `Environment` would be fine.

lib/ops/ops.go
```diff
@@ -455,6 +456,14 @@ type Certificates interface {
 	DeleteClusterCertificate(SiteKey) error
 }
 
+// EnvironmentVariables manages runtime environment variables in cluster
+type EnvironmentVariables interface {
+	// CreateUpdateEnvarsOperation creates a new operation to update cluster runtime environment variables
+	CreateUpdateEnvarsOperation(CreateUpdateEnvarsOperationRequest) (*SiteOperationKey, error)
```
r0mant: I would either expand `Envars` to `EnvironmentVariables` here, or vice versa in the method below, for naming consistency.

lib/ops/resources/gravity/gravity.go
```diff
@@ -207,10 +210,8 @@ func (r *Resources) Create(req resources.CreateRequest) error {
 			return trace.Wrap(err)
 		}
 		r.Printf("Updated monitoring alert target %q\n", target.GetName())
-	case "":
-		return trace.BadParameter("missing resource kind")
 	default:
-		return trace.BadParameter("unsupported resource %q, supported are: %v",
+		return trace.NotImplemented("unsupported resource %q, supported are: %v",
```
r0mant: Nit: Not sure if "not implemented" is semantically better suited here than "bad parameter" tbh, as it usually implies that something is not implemented _yet_ while this is supposed to filter out garbage/typos.

lib/ops/ops.go
```diff
@@ -905,6 +952,8 @@ func (s *SiteOperation) String() string {
 		typeS = "uninstall"
 	case OperationGarbageCollect:
 		typeS = "garbage collect"
+	case OperationUpdateEnvars:
+		typeS = "update cluster envars"
```
r0mant: Nit: Maybe "env vars" at least? I don't think "envars" is even a word (or a valid shorthand).

PR #250, opened 2019-02-04
AuthGateway resource.

r0mant:
This PR adds support for the `authgateway` resource which basically allows to tweak various configuration parameters of embedded Teleport.

I have removed documentation for the `cluster_auth_preference` resource because it is obsoleted by auth gateway. The actual resource support is still there for backward compatibility (and prints a warning when used).

Closes https://github.com/gravitational/gravity.e/issues/3918.

### DIFF
--- a/docs/5.x/cluster.md
+++ b/docs/5.x/cluster.md
@@ -1230,6 +1230,7 @@ Resource Name             | Resource Description
 `logforwarder`            | forwarding logs to a remote rsyslog server
 `trusted_cluster`         | managing access to remote Ops Centers
 `endpoints`               | Ops Center endpoints for user and cluster traffic
+`cluster_auth_preference` | cluster authentication settings such as second-factor
 `alert`                   | cluster monitoring alert
 `alerttarget`             | cluster monitoring alert target
 `smtp`                    | cluster monitoring SMTP configuration
@@ -1904,6 +1905,9 @@ gravity-agents   LoadBalancer   10.100.91.204   <pending>     4443:30873/TCP,302
 
 ### Configuring Cluster Authentication Gateway
 
+!!! note:
+    Authentication gateway resource is supported starting Gravity version `5.5.0`.
+
 Cluster authentication gateway handles authentication/authorization and allows
 users to remotely access the cluster nodes via SSH or Kubernetes API.
 
@@ -1911,7 +1915,7 @@ To tweak authentication gateway configuration use the following resource:
 
 ```yaml
 kind: authgateway
-version: v2
+version: v1
 spec:
   # Connection throttling settings
   connection_limits:
@@ -1932,7 +1936,7 @@ spec:
   client_idle_timeout: never
   # Determines if the clients will be forcefully disconnected when their
   # certificates expire in the middle of an active SSH session
-  disconnect_expired_cert: no
+  disconnect_expired_cert: false
   # DNS name that applies to all SSH, Kubernetes and web proxy endpoints
   public_addr:
     - example.com
@@ -1960,12 +1964,12 @@ $ gravity resource create gateway.yaml
 
 When authentication gateway resource is created, only settings that were
 explicitly set are applied to the current configuration. For example, to
-update only the maximum connections limit, you can create the following
+only limit the maximum number of connections, you can create the following
 resource:
 
 ```yaml
 kind: authgateway
-version: v2
+version: v1
 spec:
   connection_limits:
     max_conections: 1500
@@ -1975,6 +1979,71 @@ The following command will display current authentication gateway configuration:
 
 ```bash
 $ gravity resource get authgateway
+```
+
+### Configuring Cluster Authentication Preference
+
+!!! warning "Deprecation warning":
+    Cluster authentication preference resource is obsolete starting Gravity
+    version `5.5.0` and will be removed in a future version. Please use
+    [Authentication Gateway](/cluster/#configuring-cluster-authentication-gateway)
+    resource instead.
+
+Cluster authentication preference resource allows to configure method of
+authentication users will use when logging into a Gravity cluster.
+
+The resource has the following format:
+
+```yaml
+kind: cluster_auth_preference
+version: v2
+metadata:
+  name: auth-oidc
+spec:
+  # preferred auth type, can be "local" (to authenticate against
+  # local users database) or "oidc"
+  type: oidc
+  # second-factor auth type, can be "off" or "otp"
+  second_factor: otp
+  # default authentication connector to use for tele login
+  connector_name: google
+```
+
+By default the following authentication method is configured:
+
+* For Ops Centers: OIDC or local with second-factor authentication.
+* For regular clusters: local without second-factor authentication.
+
+To update authentication preference, for example to allow local users to log
+into an Ops Center without second-factor, define the following resource:
+
+```yaml
+kind: cluster_auth_preference
+version: v2
+metadata:
+  name: auth-local
+spec:
+  type: local
+  second_factor: "off"
+```
+
+Create it:
+
+```bsh
+$ gravity resource create auth.yaml
+```
+
+!!! note:
+    Make sure to configure a proper [OIDC connector](/cluster/#configuring-openid-connect)
+    when using "oidc" authentication type.
+
+To view the currently configured authentication preference:
+
+```bsh
+$ gravity resource get cluster_auth_preference
+Type      ConnectorName     SecondFactor
+----      -------------     ------------
+local                       off
 ```
 
 ### Configuring Monitoring
--- a/docs/5.x/cluster.md
+++ b/docs/5.x/cluster.md
@@ -1230,11 +1230,11 @@ Resource Name             | Resource Description
 `logforwarder`            | forwarding logs to a remote rsyslog server
 `trusted_cluster`         | managing access to remote Ops Centers
 `endpoints`               | Ops Center endpoints for user and cluster traffic
-`cluster_auth_preference` | cluster authentication settings such as second-factor
 `alert`                   | cluster monitoring alert
 `alerttarget`             | cluster monitoring alert target
 `smtp`                    | cluster monitoring SMTP configuration
-`runtimeenvironment`     | cluster runtime environment variables
+`runtimeenvironment`      | cluster runtime environment variables
+`authgateway`             | authentication gateway configuration
 
 ### Configuring OpenID Connect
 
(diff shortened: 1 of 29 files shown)


lib/storage/authgateway.go
```diff
@@ -0,0 +1,561 @@
...
+	}
+	if gw.Spec.WebPublicAddr != nil {
+		other.SetWebPublicAddrs(*gw.Spec.WebPublicAddr)
+	}
+}
+
+// Apply applies auth gateway settings to the provided config.
+func (gw *AuthGatewayV2) ApplyToTeleportConfig(config *teleconfig.FileConfig) {
+	if gw.Spec.ConnectionLimits != nil {
+		config.Global.Limits.MaxConnections = *gw.Spec.ConnectionLimits.MaxConnections
```
a-palchikov: `ConnectionLimits.MaxConnections` might still be `nil` (ditto for `MaxUsers`).
I'd just use `GetMaxConnections` / `GetMaxUsers`

lib/storage/authgateway.go
```diff
@@ -0,0 +1,561 @@
...
+		// It never returns an error.
+		authPreference, _ := teleservices.NewAuthPreference(*gw.Spec.Authentication)
+		return authPreference
+	}
+	return nil
+}
+
+// SetAuthPreference sets the authentication preference settings on the resource.
+func (gw *AuthGatewayV2) SetAuthPreference(authPreference teleservices.AuthPreference) {
+	// It only returns trace.NotFound() error.
```
a-palchikov: What if this is changed and it starts returning a different error?

lib/storage/authgateway.go
```diff
@@ -0,0 +1,561 @@
...
+// SetTTL sets the resource TTL.
+func (gw *AuthGatewayV2) SetTTL(clock clockwork.Clock, ttl time.Duration) {
+	gw.Metadata.SetTTL(clock, ttl)
+}
+
+// String returns the object's string representation.
+func (gw AuthGatewayV2) String() string {
+	var parts []string
+	if gw.Spec.ConnectionLimits != nil {
+		parts = append(parts, fmt.Sprintf("%s", *gw.Spec.ConnectionLimits))
```
a-palchikov:
```go
parts = append(parts, gw.Spec.ConnectionLimits.String())
```

lib/storage/authgateway_test.go
```diff
@@ -0,0 +1,244 @@
...
+			MaxUsers:       intp(20),
+		},
+		Authentication: &teleservices.AuthPreferenceSpecV2{
+			Type:          "oidc",
+			SecondFactor:  "off",
+			ConnectorName: "google",
+		},
+		ClientIdleTimeout:     durp(60 * time.Second),
+		DisconnectExpiredCert: teleservices.NewBoolOption(true),
+		PublicAddr:            &[]string{"example.com"},
```
a-palchikov: To be honest, I don't the benefit to using pointers to slice.

lib/storage/authgateway_test.go
```diff
@@ -0,0 +1,244 @@
...
+			MaxConnections: int64p(1000),
+			MaxUsers:       intp(5),
+		},
+		SSHPublicAddr:        &[]string{"example.com"},
+		KubernetesPublicAddr: &[]string{"example.com"},
+		WebPublicAddr:        &[]string{"example.com"},
+	}))
+}
+
+func (s *AuthGatewaySuite) TestApplyToTeleportConfig(c *check.C) {
```
a-palchikov: Frankly, I would've replaced all these tests with a single integration test that executes the event loops and validates that the configuration is properly received/updated.

lib/storage/authgateway_test.go
```diff
@@ -0,0 +1,244 @@
...
+			PublicAddr:    teleutils.Strings([]string{"example.com"}),
+			SSHPublicAddr: teleutils.Strings([]string{"example.com"}),
+			Kube: teleconfig.Kube{
+				PublicAddr: teleutils.Strings([]string{"k8s.example.com"}),
+			},
+		},
+	})
+}
+
+func int64p(i int64) *int64 {
```
a-palchikov: nit: these are useful outside. Besides we've got a bunch of these in lib/utils:
```go
func Int64Ptr(int64) *int64 {}
func BoolValue(*bool) bool {}
func BoolPtr(bool) *bool {}
```

lib/storage/resources.go
```diff
@@ -82,3 +85,25 @@ var SupportedGravityResourcesToRemove = []string{
...
+
+// MetadataSchema is a copy of teleport/lib/services.MetadataSchema but with
+// optional 'name' property because some Gravity resources do not require it
+const MetadataSchema = `{
+  "type": "object",
+  "additionalProperties": false,
+  "default": {},
+  "properties": {
+    "name": {"type": "string"},
+    "namespace": {"type": "string", "default": "default"},
```
a-palchikov: `namespace` attribute is not serialized in `teleservices.Metadata` so setting the default does not have the desired effect.

docs/5.x/cluster.md
```diff
@@ -1902,74 +1902,85 @@ gravity-public   LoadBalancer   10.100.20.71    <pending>     443:31792/TCP,3023
...
+To update authentication gateway configuration, run:
+
+```bash
+$ gravity resource create gateway.yaml
+```
 
-* For Ops Centers: OIDC or local with second-factor authentication.
-* For regular clusters: local without second-factor authentication.
+!!! note:
+    The `gravity-site` pods will be restarted upon resource creation in order
```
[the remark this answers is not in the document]
r0mant: Hmm, there's a pretty extensive resource validation, both schema and values, so it should be impossible to set invalid values using this resource. In my testing, I tried pretty hard to "break" it so the process wouldn't start, but couldn't.

lib/ops/opsservice/authgateway.go
```diff
@@ -0,0 +1,108 @@
...
+	gw, err := storage.UnmarshalAuthGateway([]byte(data))
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	// Use auth preference value from the database in case it was set
+	// independently of auth gateway resource via separate auth preference
+	// resource, which is obsolete. Once dedicated auth preference resource
+	// is removed, auth gateway will become a sole source of truth and this
+	// will no longer be needed.
+	authPreference, err := identity.GetAuthPreference()
```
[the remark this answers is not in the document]
r0mant: Auth preference is always initialized during the process startup (more precisely, during embedded teleport process startup).

lib/process/watch.go
```diff
@@ -97,3 +99,150 @@ func (p *Process) watchCertificate(ctx context.Context, client *kubernetes.Clien
...
+	defer watcher.Stop()
+	for {
+		select {
+		case event, ok := <-watcher.ResultChan():
+			if !ok {
+				p.Debugf("Watcher channel closed: %v.", event)
+				return nil
+			}
+			if event.Type != watch.Modified && event.Type != watch.Deleted {
+				p.Debugf("Ignoring event: %v.", event.Type)
```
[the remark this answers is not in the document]
r0mant: The only 2 other types of events are "added" and "error" and I don't think they're happening that often.

lib/process/watch.go
```diff
@@ -97,3 +99,150 @@ func (p *Process) watchCertificate(ctx context.Context, client *kubernetes.Clien
...
+				continue
+			}
+			configMap, ok := event.Object.(*v1.ConfigMap)
+			if !ok {
+				p.Warningf("Expected ConfigMap, got: %T %v.", event.Object, event.Object)
+				continue
+			}
+			if configMap.Name != constants.AuthGatewayConfigMap {
+				p.Debugf("Ignoring ConfigMap change: %v.", configMap.Name)
```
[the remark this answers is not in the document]
r0mant: I don't think it will, the watch only watches "auth-gateway" config maps so technically this should never be hit.

lib/process/watch.go
```diff
@@ -97,3 +99,150 @@ func (p *Process) watchCertificate(ctx context.Context, client *kubernetes.Clien
...
+		p.Infof("Started watching %v events.", service.TeleportReloadEvent)
+		for {
+			select {
+			case event := <-eventsCh:
+				if event.Name != service.TeleportReloadEvent {
+					p.Warnf("Expected %v event, got: %#v.", service.TeleportReloadEvent, event)
+					continue
+				}
+				p.Infof("Received event: %#v.", event)
+				err := libkube.DeleteSelf(client, p.FieldLogger)
```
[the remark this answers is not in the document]
r0mant: Well, teleport has a hot-reload feature but I'm not sure how hard would be to integrate it into gravity (or even if possible at all with it running inside k8s and all) or if that's worth the time. I have also originally attempted to perform a graceful shutdown instead of deleting the pod, but that just leads to the pod going into crashloop as k8s does not distinguish b/w 0 and non-0 exit codes.

lib/processconfig/merge.go
```diff
@@ -100,3 +101,12 @@ func MergeTeleConfig(into, from *telecfg.FileConfig) error {
 	}
 	return nil
 }
+
+// ReplacePublicAddrs replaces public addresses of all services in the
+// specified config from the provided config.
+func ReplacePublicAddrs(into, from *service.Config) {
+	into.Auth.PublicAddrs = from.Auth.PublicAddrs
+	into.Proxy.SSHPublicAddrs = from.Proxy.SSHPublicAddrs
```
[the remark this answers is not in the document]
r0mant: Yes, the point of the function is to replace addresses from the new config, whatever they are.

lib/storage/authgateway.go
```diff
@@ -0,0 +1,561 @@
...
+	// Resource provides common resource methods.
+	teleservices.Resource
+	// CheckAndSetDefaults validates the resource and fills in some defaults.
+	CheckAndSetDefaults() error
+	// GetMaxConnections returns maximum allowed connections number.
+	GetMaxConnections() int64
+	// GetMaxUsers returns maximum allowed users number.
+	GetMaxUsers() int
+	// GetConnectionLimits returns all configured connection limits.
+	GetConnectionLimits() *ConnectionLimits
```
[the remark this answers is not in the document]
r0mant: I used pointers to be able to distinguish b/w unset and zero values.

lib/storage/authgateway.go
```diff
@@ -0,0 +1,561 @@
...
+	// GetKubernetesPublicAddrs returns Kubernetes public addresses.
+	GetKubernetesPublicAddrs() []string
+	// SetKubernetesPublicAddrs sets Kubernetes public addresses on the resource.
+	SetKubernetesPublicAddrs([]string)
+	// GetWebPublicAddrs returns web service public addresses.
+	GetWebPublicAddrs() []string
+	// SetWebPublicAddrs sets web service public addresses on the resource.
+	SetWebPublicAddrs([]string)
+	// GetPublicAddrs returns public addresses set for all services.
+	GetPublicAddrs() []string
```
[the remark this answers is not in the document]
r0mant: Not sure what you mean?

lib/storage/authgateway.go
```diff
@@ -0,0 +1,561 @@
...
+	// KubernetesPublicAddr sets public addresses for Kubernetes proxy service.
+	KubernetesPublicAddr *[]string `json:"kubernetes_public_addr,omitempty"`
+	// WebPublicAddr sets public addresses for web service.
+	WebPublicAddr *[]string `json:"web_public_addr,omitempty"`
+}
+
+// ConnectionLimits defines connection limits setting on auth gateway resource.
+type ConnectionLimits struct {
+	// MaxConnections is the maximum number of connections to auth/proxy services.
+	MaxConnections *int64 `json:"max_connections,omitempty"`
```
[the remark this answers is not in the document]
r0mant: Yes.

lib/storage/authgateway.go
```diff
@@ -0,0 +1,561 @@
...
+	// MaxUsers is the maximum number of simultaneously connected users.
+	MaxUsers *int `json:"max_users,omitempty"`
+}
+
+// String returns the object's string representation.
+func (l ConnectionLimits) String() string {
+	var parts []string
+	if l.MaxConnections != nil {
```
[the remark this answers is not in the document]
r0mant: The gateway's `String()` function will skip it if it'll be nil.

lib/storage/authgateway_test.go
r0mant: It's for the same reason as I mentioned above: to be able to tell whether it's an empty slice or the field wasn't set.

lib/storage/authgateway_test.go
r0mant: I think unit tests have their own merit. I've added one more higher-level test to test configuration update though (in process_test.go).

lib/storage/authgateway.go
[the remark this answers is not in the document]
a-palchikov: I thought this was returning all kinds of (public) addresses (i.e. web and SSH) - so it read from the comment and was puzzled about how to figure which one is which given just a string slice.

lib/storage/authgateway_test.go
a-palchikov: Well, you can use `nil` for this - empty slice and `nil` slice are different.

lib/process/watch.go
```diff
@@ -101,16 +104,19 @@ func (p *Process) watchCertificate(ctx context.Context, client *kubernetes.Clien
...
 // startAuthGatewayWatch launches watcher that monitors config map with
-// auth gateway configuration.
+// auth gateway configuration and updates Teleport configuration
+// appropriately.
 func (p *Process) startAuthGatewayWatch(ctx context.Context, client *kubernetes.Clientset) error {
 	go func() {
+		ticker := time.NewTicker(time.Second)
+		defer ticker.Stop()
 		for {
 			err := p.watchAuthGateway(ctx, client)
```
a-palchikov: From what I can tell, `watchAuthGateway` is also a loop exiting only when `ctx` has expired. Why no have it run inside of this one - or why is this loop required if you can do the same inside of `watchAuthGateway` (but instead of returning on error - you can loop forever until context expiration?

lib/storage/authgateway.go
```diff
@@ -422,80 +451,128 @@ func (gw *AuthGatewayV2) GetWebPublicAddrs() (addrs []string) {
...
-func (gw *AuthGatewayV2) CheckAndSetDefaults() error {
+func (gw *AuthGatewayV1) CheckAndSetDefaults() error {
 	if gw.Metadata.Name == "" {
 		gw.Metadata.Name = KindAuthGateway
 	}
 	err := gw.Metadata.CheckAndSetDefaults()
 	if err != nil {
 		return trace.Wrap(err)
 	}
+	if gw.Spec.ConnectionLimits != nil {
```
a-palchikov: nit: If you implemented Check on the pointer receiver you could check for nil inside the implementation.

lib/storage/authgateway.go
```diff
@@ -422,80 +451,128 @@ func (gw *AuthGatewayV2) GetWebPublicAddrs() (addrs []string) {
...
+		}
+		if auth != nil {
+			err := auth.CheckAndSetDefaults()
+			if err != nil {
+				return trace.Wrap(err)
+			}
+		}
+	}
+	var addresses []string
+	if gw.Spec.PublicAddr != nil {
```
a-palchikov: nit: If you had `Spec.PublicAddr []string`, you would not need this check and you'd still be able to tell between unset (`nil`) and empty.

lib/storage/authgateway_test.go
```diff
@@ -127,31 +130,64 @@ func (s *AuthGatewaySuite) TestPrincipalsChanged(c *check.C) {
 	}
 }
 
+func (s *AuthGatewaySuite) TestResourceValidation(c *check.C) {
+	specs := []string{
+		// Invalid connections limit.
```
a-palchikov: nit:
```go
comment: "Invalid connections limit",
...
c.Assert(err, IsNil, Commentf(tc.comment))
``

File: lib/storage/authgateway_test.go [hunk 22]
```diff
@@ -127,31 +130,64 @@ func (s *AuthGatewaySuite) TestPrincipalsChanged(c *check.C) {
...
+    second_factor: "off"`,
+		// Invalid principal (empty address).
+		`kind: authgateway
+version: v1
+spec:
+  ssh_public_addr: [""]`,
+	}
+	for _, spec := range specs {
+		_, err := UnmarshalAuthGateway([]byte(spec))
+		c.Assert(err, check.NotNil, check.Commentf("Test case %q failed.", spec))
```
a-palchikov: nit: this can fail for reasons not directly related to the test case

lib/process/watch.go
r0mant: Because IIRC watch can expire so this loop is used to restart the watch loop. Certificate update (in the same file) is implemented the same way.

lib/storage/authgateway.go
r0mant: Not during marshaling because omitempty skips both empty and nil slices.

PR #255, opened 2019-02-07
Tele CLI/UX improvements.

r0mant:
This PR implements most of UX improvements for tele from https://github.com/gravitational/gravity.e/issues/3911. The next big chunk of work that will have to be done before that ticket could be closed will be migrating OSS hub to become a chart repository and teaching tele to work with it, but I wanted to merge these smaller fixes/improvements now. Here's the gist:

* When building telekube-app, add description that includes Kubernetes version (propagated from makefile).
* Add support for `baseImage` manifest field. This is basically the same as `systemOptions.runtime` but the future docs will be referring to base image instead.
* Multiple `tele ls` behavior and UX tweaks:
  - Move common parts of code to an interface shared by OSS/enterprise so we don't have to reimplement stuff anymore.
  - Replace "telekube" with "gravity". This is a purely cosmetic change, again b/c future docs will be using "gravity" as base image.
  - Show only latest stable versions by default.
  - Add `--all` flag to show all.
  - Sort output by image type / name / version.
* Move `--quiet` flag from being shared by all command only to those commands that actually use it.

### DIFF
--- a/lib/builder/builder.go
+++ b/lib/builder/builder.go
@@ -232,8 +232,8 @@ func (b *Builder) SelectRuntime() (*semver.Version, error) {
 	switch runtime.Name {
 	case constants.BaseImageName, defaults.Runtime:
 	default:
-		return nil, trace.BadParameter("unsupported base image %q, only "+
-			"\"gravity\" is supported as a base image at the moment", runtime.Name)
+		return nil, trace.BadParameter("unsupported base image %q, only %q is "+
+			"supported as a base image", runtime.Name, constants.BaseImageName)
 	}
 	if runtime.Version != loc.LatestVersion {
 		b.Infof("Using pinned runtime version: %s.", runtime.Version)
--- a/lib/builder/builder.go
+++ b/lib/builder/builder.go
@@ -229,6 +229,10 @@ func (b *Builder) SelectRuntime() (*semver.Version, error) {
 	if runtime == nil {
 		return nil, trace.NotFound("failed to determine application runtime")
 	}
+	if runtime.Name != constants.BaseImageName {
+		return nil, trace.BadParameter("unsupported base image %q, only "+
+			"\"gravity\" is supported as a base image at the moment", runtime.Name)
+	}
 	if runtime.Version != loc.LatestVersion {
 		b.Infof("Using pinned runtime version: %s.", runtime.Version)
 		b.PrintSubStep("Will use runtime version %s from manifest", runtime.Version)
--- a/lib/builder/builder.go
+++ b/lib/builder/builder.go
@@ -229,7 +229,9 @@ func (b *Builder) SelectRuntime() (*semver.Version, error) {
 	if runtime == nil {
 		return nil, trace.NotFound("failed to determine application runtime")
 	}
-	if runtime.Name != constants.BaseImageName {
+	switch runtime.Name {
+	case constants.BaseImageName, defaults.Runtime:
+	default:
 		return nil, trace.BadParameter("unsupported base image %q, only "+
 			"\"gravity\" is supported as a base image at the moment", runtime.Name)
 	}
--- a/lib/catalog/lister.go
+++ b/lib/catalog/lister.go
@@ -27,6 +27,7 @@ import (
 	"github.com/gravitational/gravity/lib/app"
 	"github.com/gravitational/gravity/lib/constants"
 	"github.com/gravitational/gravity/lib/hub"
+	"github.com/gravitational/gravity/lib/schema"
 
 	"github.com/coreos/go-semver/semver"
 	"github.com/ghodss/yaml"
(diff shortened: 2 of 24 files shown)


lib/builder/builder.go
```diff
@@ -229,6 +229,12 @@ func (b *Builder) SelectRuntime() (*semver.Version, error) {
 	if runtime == nil {
 		return nil, trace.NotFound("failed to determine application runtime")
 	}
+	switch runtime.Name {
+	case constants.BaseImageName, defaults.Runtime:
+	default:
+		return nil, trace.BadParameter("unsupported base image %q, only "+
+			"\"gravity\" is supported as a base image at the moment", runtime.Name)
```
a-palchikov: ~~at the moment~~
and leave a TODO to remove or update the text.

lib/catalog/lister.go
```diff
@@ -0,0 +1,258 @@
...
+func NewListItemFromHubApp(app hub.App) (*listItem, error) {
+	semver, err := semver.NewVersion(app.Version)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	return &listItem{
+		Name:    app.Name,
+		Version: *semver,
+		Created: app.Created,
+		Type:    "Cluster", // We currently publish only cluster images to the hub.
```
a-palchikov: That would also deserve a dedicated TODO with explanation when that might change.

lib/catalog/lister.go
```diff
@@ -0,0 +1,258 @@
...
+	m := make(map[string]ListItem)
+	for _, item := range l {
+		// Skip pre-releases.
+		if item.GetVersion().PreRelease != "" {
+			continue
+		}
+		if _, ok := m[item.GetName()]; !ok {
+			m[item.GetName()] = item
+		} else {
+			if m[item.GetName()].GetVersion().LessThan(item.GetVersion()) {
```
a-palchikov: nit: shorter:
```go
existing, ok := m[item.GetName()]
if !ok || existing.GetVersion().LessThan(item.GetVersion()) { 
	m[item.GetName()] = item
}
```

lib/catalog/lister.go
```diff
@@ -0,0 +1,258 @@
...
+func (l ListItems) Swap(i, j int) {
+	l[i], l[j] = l[j], l[i]
+}
+
+// Less implements sort.Interace.
+//
+// The items are sorted first by type (cluster images appear before application
+// images), then by name (lexicographically) and finally by semantic version.
+func (l ListItems) Less(i, j int) bool {
+	if l[i].GetType() == "Cluster" && l[j].GetType() != "Cluster" {
```
a-palchikov: "Cluster" -> const

lib/catalog/lister.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+		result = append(result, i)
+	}
+	return result, nil
+}
+
+// List uses the provided lister to obtain a list of application and
+// cluster images and displays them in the specified format.
+func List(lister Lister, all bool, format constants.Format) error {
+	items, err := lister.List(all)
```
a-palchikov: Does not look like `all` is applicable to the lister.

lib/catalog/lister_test.go
```diff
@@ -0,0 +1,57 @@
...
+		listItem{Name: "nginx", Version: v("2.0.0-alpha.1"), Type: "Application"},
+		listItem{Name: "kafka", Version: v("1.0.1"), Type: "Application"},
+		listItem{Name: "zookeeper", Version: v("1.0.0-beta.2"), Type: "Application"},
+		listItem{Name: "kubernetes", Version: v("13.0.0"), Type: "Cluster"},
+		listItem{Name: "kubernetes", Version: v("14.0.0-beta.1"), Type: "Cluster"},
+	}
+	latest, err := items.Latest()
+	c.Assert(err, check.IsNil)
+	sort.Sort(latest)
+	c.Assert(latest, compare.DeepEquals, ListItems{
```
a-palchikov: nit: `compare.SortedSliceEquals` (then `sort.Sort` will not be required)

lib/constants/constants.go
```diff
@@ -710,6 +710,11 @@ var (
 		"kubernetes.default.svc.cluster",
 		"kubernetes.default.svc.cluster.local",
 	}
+
+	// LegacyBaseImageName is the legacy base cluster image name
```
a-palchikov:
```go
// LegacyBaseImageName is the name of the base cluster image used in earlier versions
```

lib/schema/constants.go
```diff
@@ -117,8 +121,14 @@ const (
 	ApplicationDefaultNamespace = "default"
 )
 
-// APIVersionV2 specifies the current API version
-var APIVersionV2 = fmt.Sprintf("%v/%v", GroupName, Version)
+var (
+	// APIVersionV2 specifies the current API version
+	APIVersionV2 = fmt.Sprintf("%v/%v", GroupName, Version)
+	// APIVersionV2Cluster is the apiVersion for cluster images
```
a-palchikov: `apiVersion` -> API version

lib/catalog/lister_test.go
r0mant: This is actually on purpose as this way it also makes sure that sorting order is correct.

PR #270, opened 2019-02-14
Update test plan & some fixes.

r0mant:
* Add Application Catalog section to the test plan. I've added a couple of sample apps to the assets.
* Remove one-time install link use-case from the plan. We're removing this feature.
* Fix an issue with tele build: I've discovered that `tele build alpine` works, but `tele build alpine/` because the trailing slash somehow messes up our infamous `utils.CopyDirContents()` method.


assets/charts/alpine-0.1.0/templates/alpine-pod.yaml
```diff
@@ -0,0 +1,23 @@
...
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    # This makes it easy to audit chart usage.
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    app.kubernetes.io/name: {{ template "alpine.name" . }}
+spec:
+  # This shows how to use a simple value. This will look for a passed-in value called restartPolicy.
+  restartPolicy: {{ .Values.restartPolicy }}
+  containers:
+  - name: waiter
+    image: "{{ .Values.image.registry }}{{ .Values.image.repository }}:{{ .Values.image.tag }}"
```
a-palchikov: I'm not too familiar with template behavior - will this work fine even if `image.registry` is undefined?

docs/5.x/testplan.md
```diff
@@ -199,6 +195,107 @@ $ tele login -o example.gravitational.io
...
+
+- [ ] Verify an application instance has been deployed.
+```bash
+$ gravity app ls
+$ kubectl get pods
+```
+
+- [ ] Install the application directly from the Ops Center.
+```bash
+$ gravity app install <opscenter-name>/alpine:0.1.0
```
a-palchikov: What does it mean to install the same application twice (if this is installing the same application)?

assets/charts/alpine-0.1.0/templates/alpine-pod.yaml
r0mant: Yeah, it'll just default to an empty string.

docs/5.x/testplan.md
r0mant: Yeah, so we're using Helm for all application lifecycle and the way it works is when you install a chart (or "app" in our terminology) it creates a "release" to track this instance of the app. If you install the same app again, it'll create another "release", so it'll be another instance of the same app running. And so on. So you can have multiple instances of the same application running in the same cluster, upgrade/rollback them individually, etc. Here's some info on it in Helm docs: https://github.com/helm/helm/blob/master/docs/glossary.md#release.

PR #281, opened 2019-02-22
When there is 1 IP address, use it as a default value for IP address selector

alex-kovoy:
addresses https://github.com/gravitational/gravity.e/issues/3924


web/src/app/flux/apps/actions.js
```diff
@@ -20,9 +20,12 @@ import cfg from 'app/config';
 import { APPS_INIT, APPS_ADD } from './actionTypes';
 import { RepositoryEnum } from 'app/services/enums';
 
-function handleServerResponse(json){
+function handleAppsResponse(json){
   json = json || [];
   json = Array.isArray(json) ? json : [json];
+
+  // quick fix: keep only cluster bundles
+  json = json.filter(item => item.manifest.kind === "Bundle");
```
r0mant: We should include `kind == "Cluster"` too here.

PR #282, opened 2019-02-22
configure tiller to use absolute query for localhost

knisbet:
I've tested this, and it seems to work without issue. This change configures tiller to do an absolute query when resolving its bind address. This avoids issues, where localhost may exist within the search path of the host, such as localhost.example.com resolving to 1.1.1.1.


aranair:
> I'm curious, why not just do `--listen=127.0.0.1:44134` to exclude DNS from the picture?

Same question actually^

knisbet:
I've updated it to 127.0.0.1.

PR #295, opened 2019-03-04
Document ClusterConfiguration resource

a-palchikov:


docs/5.x/cluster.md
```diff
@@ -2124,6 +2026,89 @@ $ gravity resource rm runtimeenvironment
     of runtime containers on each cluster node. Take this into account and plan each update accordingly.
 
 
+### Configuring Cluster
+
+It is possible to perform last-minute configuration before the installation or update some aspects of the cluster
```
knisbet: I would consider re-wording this, not as last-minute configuration, but maybe as customizing the cluster per environment?

docs/5.x/cluster.md
```diff
@@ -2124,6 +2026,89 @@ $ gravity resource rm runtimeenvironment
...
 
 
+### Configuring Cluster
+
+It is possible to customize the cluster per environment before the installation or update some aspects of the cluster
+using the `ClusterConfiguration` resource:
+
+[cluster-config.yaml]
+```yaml
+kind: ClusterConfiguration
```
r0mant: I think you also forgot to add this resource to the table of all resources at the beginning of https://gravitational.com/gravity/docs/cluster/#configuring-a-cluster.

docs/5.x/cluster.md
```diff
@@ -2124,6 +2026,89 @@ $ gravity resource rm runtimeenvironment
...
+version: v1
+spec:
+  global:
+    cloudProvider: gce    # configures the cloud provider
+    cloudConfig: |        # free-form cloud configuration
+      multizone=true
+      gce-node-tags=demo-cluster
+    serviceCIDR:  "10.0.0.0/24"          # represents the IP range from which to assign service cluster IPs
```
r0mant: Also, nit: I would apply comments on top of YAML fields like other YAML examples instead of sticking them to the right. Otherwise, there is a horizontal scroll bar cause the lines are long and it's not very convenient to look at.

PR #300, opened 2019-03-05
App list command tweaks.

r0mant:
Made a couple of adjustments to the `gravity app ls` command while playing with the application catalog:

* Add `list` alias for the command.
* Include additional release statuses in default output (to be able to see releases that are being installed/uninstalled/upgraded/etc. - will be needed for the dashboard too).
* Add `--all` flag to the command to display releases with all possible statuses (including "deleted", "superseded", etc. which you normally don't need to see).

### DIFF
--- a/tool/gravity/cli/register.go
+++ b/tool/gravity/cli/register.go
@@ -242,7 +242,7 @@ func RegisterCommands(app *kingpin.Application) *Application {
 	g.AppInstallCmd.RegistryKey = g.AppInstallCmd.Flag("registry-key", "Docker registry client private key path.").String()
 
 	g.AppListCmd.CmdClause = g.AppCmd.Command("ls", "Show all application releases.").Alias("list")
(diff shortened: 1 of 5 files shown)


tool/gravity/cli/register.go
```diff
@@ -241,7 +241,8 @@ func RegisterCommands(app *kingpin.Application) *Application {
 	g.AppInstallCmd.RegistryCert = g.AppInstallCmd.Flag("registry-cert", "Docker registry client certificate path.").String()
 	g.AppInstallCmd.RegistryKey = g.AppInstallCmd.Flag("registry-key", "Docker registry client private key path.").String()
 
-	g.AppListCmd.CmdClause = g.AppCmd.Command("ls", "Show all application releases.")
+	g.AppListCmd.CmdClause = g.AppCmd.Command("ls", "Show all application releases.").Alias("list")
+	g.AppListCmd.All = g.AppListCmd.Flag("all", "Display releases with all statuses.").Short('a').Bool()
```
a-palchikov:
```
Do not filter out releases
```

PR #301, opened 2019-03-06
Fixes for shrink and process.

r0mant:
This PR improves the way the "master" server runner is chosen during shrink operation. It makes sure that if the node that's being removed is master, it can't be chosen as a master runner. Previously, the code would choose any master node which had a decent chance of choosing the node that's being removed which could lead to all sorts of issues. Closes https://github.com/gravitational/gravity.e/issues/3740.

While testing, I have also observed that when installing a 3-node cluster, the two non-leader gravity-site processes would not initialize properly with error `failed to find any peer with hash(964ae66380e9b7cfadb9752193aafe8aea7596819bfad0ae302315a9fac8f763), failed to load RPC credentials` so I added retry to RPC credentials initialization and made sure it fixes the issue.

### DIFF
--- a/lib/ops/opsservice/shrink.go
+++ b/lib/ops/opsservice/shrink.go
@@ -423,8 +423,8 @@ func (s *site) pickShrinkMasterRunner(ctx *operationContext, removedServer stora
 			}, nil
 		}
 	}
-	return nil, trace.NotFound("could not find appropriate master server among %v",
-		masters)
+	return nil, trace.NotFound("%v is being removed and no more master nodes are available to execute the operation",
+		removedServer)
 }
 
 func (s *site) waitForServerToDisappear(hostname string) error {
--- a/lib/ops/opsservice/shrink.go
+++ b/lib/ops/opsservice/shrink.go
@@ -121,7 +121,7 @@ func (s *site) validateShrinkRequest(req ops.CreateSiteShrinkOperationRequest, c
 	serverName := req.Servers[0]
 	if len(cluster.ClusterState.Servers) == 1 {
 		return nil, trace.BadParameter(
-			"cannot shrink 1-node cluster, please uninstall it via Ops Center")
+			"cannot shrink 1-node cluster, use --force flag to uninstall")
 	}
 
 	server, err := cluster.ClusterState.FindServer(serverName)
@@ -130,7 +130,7 @@ func (s *site) validateShrinkRequest(req ops.CreateSiteShrinkOperationRequest, c
 	}
 
 	// check to make sure the server exists and can be found
-	servers, err := s.getTeleportServers()
+	servers, err := s.getAllTeleportServers()
 	if err != nil {
 		return nil, trace.Wrap(err, "failed to query teleport servers")
 	}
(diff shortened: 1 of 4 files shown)


lib/ops/opsservice/teleport.go
```diff
@@ -124,28 +124,47 @@ func (s *site) getTeleportServerNoRetry(labelName, labelValue string) (server *t
 	return newTeleportServer(servers[0])
 }
 
-// getTeleportServer queries all teleport servers in a retry loop
-func (s *site) getTeleportServers() (teleservers, error) {
+// getAllTeleportServer queries all teleport servers in a retry loop
```
a-palchikov: getAllTeleportServer**s**

lib/process/process.go
```diff
@@ -2091,7 +2091,15 @@ func (p *Process) loginWithToken(tokenID string, w http.ResponseWriter, r *http.
 	}
 }
 
-func (p *Process) loadRPCCredentials() (*rpcserver.Credentials, utils.TLSArchive, error) {
+func (p *Process) loadRPCCredentials() (creds *rpcserver.Credentials, archive utils.TLSArchive, err error) {
+	err = utils.Retry(defaults.RetryInterval, defaults.RetryAttempts, func() error {
```
a-palchikov: I'd be cautious about running anything in a loop for **any** error - is there a specific set of errors this can be retried for?

lib/ops/opsservice/shrink.go
```diff
@@ -406,6 +410,23 @@ func (s *site) shrinkOperationStart(ctx *operationContext) (err error) {
...
+	}
+	// Pick any master server except the one that's being removed.
+	for _, master := range masters {
+		if master.IP != removedServer.AdvertiseIP {
+			return &serverRunner{
+				&master, &teleportRunner{ctx, s.domainName, s.teleport()},
+			}, nil
+		}
+	}
+	return nil, trace.NotFound("could not find appropriate master server among %v",
```
a-palchikov: The message could be more specific:
```
%v is being removed and no more master nodes are available to execute the operation
```
which is kind of weird since there's already a check elsewhere that verifies that the last master node cannot be shrunk - it only means teleport can be flaky and not return all master nodes when queried (?)

lib/ops/opsservice/shrink.go
```diff
@@ -406,6 +410,23 @@ func (s *site) shrinkOperationStart(ctx *operationContext) (err error) {
 	return nil
 }
 
+func (s *site) pickShrinkMasterRunner(ctx *operationContext, removedServer storage.Server) (*serverRunner, error) {
+	masters, err := s.getTeleportServers(schema.ServiceLabelRole, string(schema.ServiceRoleMaster))
```
a-palchikov: This is already done in `validateShrinkRequest` - I wonder if you can reuse the results (it is also better to init in `validateShrinkRequest` **before** starting the operation) and set the masterRunner as well as initialize teleport server for the server being removed.

r0mant: Server can't be initialized there b/c "validate" and "start" functions can be executed by different gravity-sites (if the removed node is active master) so operation context is not passed directly from "validate" to "start". I guess I could determine master server in "validate" and save it in the operation data so "start" would later fetch and initialize it, but I'm not sure how that's better?

a-palchikov: ok, good point.

PR #305, opened 2019-03-07
Update README.md

swemarx:


r0mant:
@swemarx Thank you for the fix!

PR #310, opened 2019-03-07
Stricter validation of cloud configuration changes

a-palchikov:
* Do not allow to change/set cloud provider on the installed cluster
 * Keep cloud provider CLI parameter and cloud config block from cluster configuration in sync
 * Only allow to update cloud configuration if the cluster has cloud integration

Requires https://github.com/gravitational/gravity.e/pull/3994.


lib/install/install.go
```diff
@@ -162,12 +162,13 @@ type Config struct {
 	Role string
 	// AppPackage is the application being installed
 	AppPackage *loc.Locator
-	// Resources specifies optional resources to create.
-	// This is assumed to be either JSON- or YAML-encoded list of resources
-	Resources []byte
```
a-palchikov: No longer needed as I opted to keep the resources structured.

lib/install/planbuilder.go
```diff
@@ -602,7 +602,7 @@ func (i *Installer) GetPlanBuilder(cluster ops.Site, op ops.SiteOperation) (*Pla
 		},
 		InstallerTrustedCluster: trustedCluster,
 	}
-	err = addResources(builder, cluster.Resources, i.Config.RuntimeResources)
+	err = addResources(builder, cluster.Resources, i.Config.RuntimeResources, i.Config.ClusterResources)
```
a-palchikov: `cluster.Resources` is the only case of []byte-typed resource left - it is used for (and with) the Gravity `cluster` resource.

tool/gravity/cli/clusterconfig.go
```diff
@@ -218,6 +221,42 @@ type configInitializer struct {
...
+	}
+	globalConfig := clusterConfig.GetGlobalConfig()
+	if globalConfig == nil || isCloudConfigEmpty(globalConfig) {
+		if newGlobalConfig := config.GetGlobalConfig(); newGlobalConfig != nil && !isCloudConfigEmpty(newGlobalConfig) {
+			return trace.BadParameter("cannot change cloud configuration: cluster does not have cloud provider configured")
+		}
+	}
+	if globalConfig != nil {
+		if newGlobalConfig := config.GetGlobalConfig(); newGlobalConfig != nil {
+			if newGlobalConfig.CloudProvider != "" && globalConfig.CloudProvider != newGlobalConfig.CloudProvider {
```
r0mant: Should this also check the case when `newGlobalConfig.CloudProvider` is empty but `globalConfig.CloudProvider` is not?

tool/gravity/cli/config.go
```diff
@@ -346,56 +347,78 @@ func (i *InstallConfig) ToInstallerConfig(env *localenv.LocalEnvironment) (*inst
...
-	}
-	err := resources.ForEach(bytes.NewReader(resourceBytes), func(res storage.UnknownResource) error {
-		if res.Kind != storage.KindClusterConfiguration {
-			return nil
+func (i *InstallConfig) updateClusterConfig(resources []storage.UnknownResource) (updated []storage.UnknownResource, err error) {
+	if len(resources) == 0 {
+		return nil, nil
+	}
+	var clusterConfig *storage.UnknownResource
+	updated = resources[:0]
```
r0mant: Weird way to declare an empty slice, wouldn't `updated = make([]storage.UnknownResource, 0, len(resources))` do?

tool/gravity/cli/config.go
```diff
@@ -346,56 +347,78 @@ func (i *InstallConfig) ToInstallerConfig(env *localenv.LocalEnvironment) (*inst
...
+		ClusterResources:   gravityResources,
 	}, nil
 }
 
-// ValidateResources validates the resources specified in ResourcePath
-// using the given validator
-func (i *InstallConfig) ValidateResources(validator resources.Validator) error {
+// splitResources validates the resources specified in ResourcePath
+// using the given validator and splits them into Kubernetes and Gravity-specific
+func (i *InstallConfig) splitResources(validator resources.Validator) ([]runtime.Object, []storage.UnknownResource, error) {
```
r0mant: Use named returns so it's clear which list contains what?

tool/gravity/cli/clusterconfig.go
```diff
@@ -218,6 +221,42 @@ type configInitializer struct {
...
+			if globalConfig.CloudProvider == "" && newGlobalConfig.CloudConfig != "" {
+				return trace.BadParameter("cannot set cloud configuration: cluster does not have cloud provider configured")
+			}
+		}
+	}
+	return nil
+}
+
+func isCloudConfigEmpty(global *libclusterconfig.Global) bool {
+	return global.CloudProvider == "" && global.CloudConfig != ""
```
r0mant: Hmm, if it tests for "empty", shouldn't it be `global.CloudConfig == ""`?

tool/gravity/cli/clusterconfig.go
```diff
@@ -218,6 +221,42 @@ type configInitializer struct {
 	config   libclusterconfig.Interface
 }
 
+func validateUpdateConfig(localEnv *localenv.LocalEnvironment, config libclusterconfig.Interface) error {
```
r0mant: Nit: I would call this function `validateCloudConfig` or something to avoid confusion with regular update operation.

lib/storage/clusterconfig/clusterconfig.go
```diff
@@ -145,6 +146,23 @@ func Marshal(config Interface, opts ...teleservices.MarshalOption) ([]byte, erro
 	return json.Marshal(config)
 }
 
+// ToUnknown returns this resource as a storage.UnknownResource
+func ToUnknown(config Interface) (*storage.UnknownResource, error) {
```
r0mant: I'm just curious: we already have `utils.ToUnknownResource` but it returns `teleservices.UnknownResource`. Is our `storage.UnknownResource` different?

tool/gravity/cli/config.go
a-palchikov: It's a way of reusing a throw-away space

lib/storage/clusterconfig/clusterconfig.go
a-palchikov: `teleservices.UnknownResource` does not properly serialize (it has a `Raw []byte` w/o json tags) and I need them to be serializable.

tool/gravity/cli/clusterconfig.go
a-palchikov: This is what the if block above is about.

PR #316, opened 2019-03-12
Emit gravity events to audit log.

r0mant:
This PR implements gravity audit log events as described in https://github.com/gravitational/gravity.e/issues/3980.

I've tried to encapsulate all code around emission of events and their names/fields in a single package, `lib/ops/events`, so all our events are defined in a single place and its easy to emit it from any part of the code.

Closes https://github.com/gravitational/gravity.e/issues/3980.


lib/ops/events/events.go
```diff
@@ -0,0 +1,94 @@
...
+	AppInstalled = "application.installed"
+	// AppUpgraded is emitted when an application release is upgraded.
+	AppUpgraded = "application.upgraded"
+	// AppRolledBack is emitted when an application release is rolled back.
+	AppRolledBack = "application.rolledback"
+	// AppUninstalled is emitted when an application release is uninstalled.
+	AppUninstalled = "application.uninstalled"
+
+	// ResourceCreated is emitted when a Gravity resource is created or updated.
+	ResourceCreated = "resource.created"
```
a-palchikov: nit: `resource.updated` would be more appropriate

lib/ops/events/events.go
```diff
@@ -0,0 +1,94 @@
...
+	UpdatesDisabled = "periodicupdates.disabled"
+	// UpdatesDownloaded is emitted when periodic updates download an update package.
+	UpdatesDownloaded = "periodicupdates.downloaded"
+
+	// ClusterDegraded is emitted when cluster health check fails.
+	ClusterDegraded = "cluster.degraded"
+	// ClusterActivated is emitted when cluster becomes healthy again.
+	ClusterActivated = "cluster.activated"
+
+	// LicenseExpiref is emitted when cluster license expires.
```
a-palchikov: LicenseExpire**d**

lib/helm/release.go
```diff
@@ -31,6 +31,10 @@ type Release struct {
 	Status string
 	// Chart is a deployed chart name and version.
 	Chart string
+	// ChartName is a chart name.
```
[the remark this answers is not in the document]
r0mant: Yeah, I have removed the "Chart" field.

lib/ops/events/events.go
r0mant: I was going for an analogy with "gravity resource" subcommands. I think ideally we'd have both created and updated for that's for later (which likely means "never" if we're honest :) )

PR #324, opened 2019-03-17
Separate default and kube-system role bindings

novas0x2a:
This makes it easier to disable cluster-admin in the default namespace.


knisbet:
@novas0x2a Thanks, I think I'll accept this as is and backport it to 5.5.

Please be aware, that I'm currently considering hardening this, so that software deployed in the default namespace will only have access to the default namespace in a default gravity install (unless separate RBAC configuration is used). If I go this route, it means deleting the object in hooks might change behaviour on you abruptly.

novas0x2a:
Works for me, just poke us if it changes (although, depending on how you do it, you might land it before we ship the disable in the first place, so... lol)

PR #337, opened 2019-03-21
[5.5] tele: unkown resource kind error

a-palchikov:
Factor resource-specific handling from the `Modules` interface to be able to support resource handler request validation for `tele`.

Fixes https://github.com/gravitational/gravity.e/issues/4014.

### DIFF
--- a/lib/modules/modules.go
+++ b/lib/modules/modules.go
@@ -157,7 +157,7 @@ func (*defaultResources) SupportedResources() []string {
 	return storage.SupportedGravityResources
 }
 
-// SupportedResourcesToRemoves returns a list of resources that can be removed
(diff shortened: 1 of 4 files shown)


lib/modules/modules.go
```diff
@@ -148,6 +150,24 @@ The cluster is up and running. Please take a look at "cluster management" sectio
...
 }
 
+type defaultResources struct{}
+
+// SupportedResources returns a list of resources that can be created/viewed
+func (*defaultResources) SupportedResources() []string {
+	return storage.SupportedGravityResources
+}
+
+// SupportedResourcesToRemoves returns a list of resources that can be removed
```
r0mant: ToRemove

PR #342, opened 2019-03-22
Add API for retrieving releases.

r0mant:
This pull request adds API (including a web API method) for retrieving application releases currently deployed in a cluster. This will be used by new dashboard.

I've also made a "release" object into a resource along the way so it is consistent with how we treat our other objects and what operator methods operate on.

Closes https://github.com/gravitational/gravity.e/issues/4000.


lib/helm/testclient.go
```diff
@@ -0,0 +1,116 @@
...
+	if !ok {
+		return nil, trace.NotFound("release %v not found", name)
+	}
+	delete(c.releases, name)
+	return release, nil
+}
+
+// Closer allows to cleanup the client.
+func (c *testClient) Close() error {
+	c.closed = true
```
a-palchikov: Not sure what this is so far useful for but I'd assert this in methods to catch early release or lack thereof?

lib/ops/opsservice/testhelpers.go
```diff
@@ -54,6 +55,8 @@ type TestServices struct {
 	AgentServer rpcserver.Server
 	// Operator is the ops service
 	Operator *Operator
+	// HelmClient is the mock Helm client
```
a-palchikov: nit: I'd move these either into a file_test.go or mark it with `test` tag.

lib/storage/release.go
```diff
@@ -0,0 +1,255 @@
...
+type ReleaseSpecV1 struct {
+	// ChartName is the name of the deployed chart.
+	ChartName string `json:"chart_name"`
+	// ChartVersion is the deployed chart version.
+	ChartVersion string `json:"chart_version"`
+	// AppVersion is the application version (may be empty).
+	AppVersion string `json:"app_version"`
+	// Namespace is the namespace where release is deployed.
+	//
+	// It is a part of spec rather than metadata because Teleport resources
```
a-palchikov: nit: add a TODO so it's more visible and can be easily found for refactoring.

lib/storage/release.go
```diff
@@ -0,0 +1,255 @@
...
+func (r *ReleaseV1) SetTTL(clock clockwork.Clock, ttl time.Duration) {
+	r.Metadata.SetTTL(clock, ttl)
+}
+
+// MarshalRelease marshals provided release resource to JSON.
+func MarshalRelease(release Release, opts ...services.MarshalOption) ([]byte, error) {
+	return json.Marshal(release)
+}
+
+// UnmarshalRelease unmarshals release resource from the provided JSON data.
```
a-palchikov: nit: it can be YAML too

lib/storage/release.go
```diff
@@ -0,0 +1,255 @@
...
+    "app_version": {"type": "string"},
+    "namespace": {"type": "string"}
+  }
+},
+"status": {
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "status": {"type": "string"},
+    "revision": {"type": "number"},
```
a-palchikov: Can this be float?

lib/ops/operatoracl.go
```diff
@@ -114,12 +119,18 @@ func (o *OperatorACL) ClusterAction(clusterName, resourceKind, action string) er
...
 func (o *OperatorACL) repoContext(repoName string) *users.Context {
-	return &users.Context{
-		Context: teleservices.Context{
-			User:     o.user,
-			Resource: storage.NewRepository(repoName),
-		},
+	return o.resourceContext(storage.NewRepository(repoName))
+}
+
+func (o *OperatorACL) resourceActions(resource teleservices.Resource, kind string, actions ...string) error {
```
a-palchikov: Does not seem to be used anywhere.

lib/ops/opsservice/testhelpers.go
r0mant: Not following, move what?

lib/storage/release.go
r0mant: No, why? Can `number` be a float? We use it everywhere AFAICS.

PR #365, opened 2019-04-10
Add a note about inotify watch.

r0mant:
We should also consider it auto-increasing it as a part of the auto-fix if it's low, there may even be a ticket about it already.

### DIFF
--- a/docs/5.x/requirements.md
+++ b/docs/5.x/requirements.md
@@ -235,9 +235,9 @@ Gravity requires that these modules are loaded prior to installation.
 ### Inotify watches
 
 Kubelet configures multiple inotify watches per container so it's recommended
-to increase the `max_user_watches` kernel parameter to avoid running out of
-limits. The actual number is application-specific, but we recommended setting
-it to some large value:
+to increase the `max_user_watches` kernel parameter. Gravity's built-in
+monitoring system checks for inotify watches exhaustion but we recommended setting
+it to some large value to avoid running out of limits:
 
 ```bsh
 $ sysctl -w fs.inotify.max_user_watches=1048576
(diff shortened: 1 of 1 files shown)


docs/5.x/requirements.md
```diff
@@ -232,6 +232,24 @@ Gravity requires that these modules are loaded prior to installation.
...
+
+```bsh
+$ sysctl -w fs.inotify.max_user_watches=1048576
+```
+
+To make the change persistent so it survives the node reboots, set the setting
+in a file inside `/etc/sysctl.d` directory, for example:
+
+```bsh
+$ cat /etc/sysctl.d/42-inotify.conf
```
a-palchikov: nit: `42` seems like an arbitrary unexplained prefix. I'd keep the example simple (i.e. `/etc/sysctl.d/inotify.conf`) and maybe link to the `sysctl.d` article for details: https://www.freedesktop.org/software/systemd/man/sysctl.d.html

docs/5.x/requirements.md
```diff
@@ -232,6 +232,24 @@ Gravity requires that these modules are loaded prior to installation.
 | Ubuntu-Core | 16.04 | br_netfilter, ebtable_filter, iptables, overlay |
 | Suse Linux (openSUSE and Enterprise) | 12 SP2, 12 SP3 | br_netfilter, ebtable_filter, iptables, overlay |
 
+### Inotify watches
+
+Kubelet configures multiple inotify watches per container so it's recommended
+to increase the `max_user_watches` kernel parameter to avoid running out of
+limits. The actual number is application-specific, but we recommended setting
+it to some large value:
```
a-palchikov: nit: I would hint at the existence of the inotify checker and the need to bump the value in that case.

PR #372, opened 2019-04-16
Update to planet with docker 18.09.

r0mant:


r0mant:
Looks like the tests on this PR fail when installing on RHEL/CentOS due to planet not being able to start with the `cgroup namespaces aren't enabled in the kernel` error. This is likely caused by the "resource starvation prevention" feature that recently landed in planet - and not by the actual Docker/Kubernetes upgrade.

I did some research and it looks like cgroup namespaces (which the latest planet now requires) are supported starting from kernel version 4.6 while latest RHEL/CentOS are using 3.10.xxx. Not sure if there's a way to enable them this in 3.10 or if this feature has been backported to 3.10 at all.

Will need to talk to @knisbet about it when he gets back. As it is, it seems like the latest planet is not able to start on 3.10 kernels anymore.

PR #373, opened 2019-04-16
[5.5] Delete completed status hook pods.

r0mant:
By default deleting a job does not delete its dependent pod which in our case leads to accumulation of completed pods of status hooks. This PR specifies appropriate propagation policy so status hook pods are also deleted. Refs https://github.com/gravitational/gravity.e/issues/4047.

### DIFF
--- a/lib/app/apps.go
+++ b/lib/app/apps.go
@@ -123,12 +123,12 @@ func StreamAppHookLogs(ctx context.Context, client *kubernetes.Clientset, ref Ho
 }
 
 // DeleteAppHookJob deletes app hook job
-func DeleteAppHookJob(ctx context.Context, client *kubernetes.Clientset, ref HookRef) error {
+func DeleteAppHookJob(ctx context.Context, client *kubernetes.Clientset, req DeleteAppHookJobRequest) error {
 	runner, err := hooks.NewRunner(client)
 	if err != nil {
 		return trace.Wrap(err)
 	}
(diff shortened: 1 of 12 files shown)


lib/app/apps.go
```diff
@@ -123,12 +123,12 @@ func StreamAppHookLogs(ctx context.Context, client *kubernetes.Clientset, ref Ho
...
 
 // DeleteAppHookJob deletes app hook job
-func DeleteAppHookJob(ctx context.Context, client *kubernetes.Clientset, ref HookRef) error {
+func DeleteAppHookJob(ctx context.Context, client *kubernetes.Clientset, req DeleteAppHookJobRequest) error {
 	runner, err := hooks.NewRunner(client)
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	return runner.DeleteJob(ctx, hooks.JobRef{Name: ref.Name, Namespace: ref.Namespace})
+	return runner.DeleteJob(ctx, hooks.JobRef{Name: req.Name, Namespace: req.Namespace}, req.Cascade)
```
a-palchikov: Have `Runner.DeleteJob` accept an equivalent of `DeleteAppHookJobRequest` instead of dealing with a bunch of attributes (especially a boolean that does not look self-explanatory):
```go
err = runner.DeleteJob(context.TODO(), *ref, false)
```
vs
```go
err = runner.DeleteJob(context.TODO(), DeleteJobRequest{Ref: *ref})
```

PR #376, opened 2019-04-18
Update webapi for new dashboard.

r0mant:
This PR implements a few changes in web API needed to power the new dashboard:

* Add a new "cluster info" web API method that returns basic cluster information such as public/internal endpoints.
* The bulk of the changes is making handlers that previously always used local cluster services (such as update user, create invite, etc.) to be routed to remote clusters when necessary. This required adding those methods to the operator service so router can be used. Note: I kept old web handlers for compatibility so current UI continues to work, will remove them once we're fully migrated to the new UI.
* Other minor things, such as including user's "bundle" in a list of application releases so it is displayed in the table with other apps too.

Closes https://github.com/gravitational/gravity.e/issues/4039, closes https://github.com/gravitational/gravity.e/issues/4040.

### DIFF
--- a/lib/ops/opshandler/opshandler.go
+++ b/lib/ops/opshandler/opshandler.go
@@ -42,7 +42,6 @@ import (
 	"github.com/gravitational/roundtrip"
 	telehttplib "github.com/gravitational/teleport/lib/httplib"
 	teleservices "github.com/gravitational/teleport/lib/services"
-	"github.com/gravitational/teleport/lib/utils"
 	"github.com/gravitational/trace"
 	"github.com/jonboulle/clockwork"
 	"github.com/julienschmidt/httprouter"
@@ -65,8 +64,6 @@ type WebHandlerConfig struct {
 	Authenticator httplib.Authenticator
 	// Devmode is whether the process is started in dev mode
 	Devmode bool
-	// PublicAdvertiseAddr is the process public advertise address
-	PublicAdvertiseAddr utils.NetAddr
 }
 
 type WebHandler struct {
@@ -103,6 +100,7 @@ func NewWebHandler(cfg WebHandlerConfig) (*WebHandler, error) {
 
 	// Applications API
 	h.GET("/portal/v1/apps", h.needsAuth(h.getApps))
+	h.GET("/portal/v1/gravity", h.needsAuth(h.getGravity))
 
 	// Accounts API
 	h.POST("/portal/v1/accounts", h.needsAuth(h.createAccount))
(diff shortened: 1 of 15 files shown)


lib/ops/opshandler/opshandler.go
```diff
@@ -114,15 +112,20 @@ func NewWebHandler(cfg WebHandlerConfig) (*WebHandler, error) {
...
+	h.PUT("/portal/v1/accounts/:account_id/sites/:site_domain/users/:user_email", h.needsAuth(h.updateUser))
 
 	// API keys API
 	h.POST("/portal/v1/apikeys/user/:user_email", h.needsAuth(h.createAPIKey))
 	h.GET("/portal/v1/apikeys/user/:user_email", h.needsAuth(h.getAPIKeys))
 	h.DELETE("/portal/v1/apikeys/user/:user_email/:api_key", h.needsAuth(h.deleteAPIKey))
 
-	// Tokens
+	// Invites API
+	h.POST("/portal/v1/accounts/:account_id/sites/:site_domain/tokens/userinvites", h.needsAuth(h.createUserInvite))
```
alex-kovoy: can we get rid of account_id in these modified APIs? I have been passing dummy values for some time now.

r0mant: No, we can't. Keep in mind this is ops handler, not web handler. A pair of accountID/clusterID is used to route request to appropriate cluster. In web handlers there's no account ID IIRC.

alex-kovoy: You are right, for these endpoints we do not require accountID on webapi level but we do for some others which is out of scope of this PR.

lib/ops/opshandler/opshandler.go
```diff
@@ -367,6 +370,30 @@ func (h *WebHandler) getApps(w http.ResponseWriter, r *http.Request, p httproute
 	return nil
 }
 
+/* getGravity exports the cluster's gravity binary.
+
+   GET /portal/v1/gravity
+*/
+func (h *WebHandler) getGravity(w http.ResponseWriter, r *http.Request, p httprouter.Params, ctx *HandlerContext) error {
```
a-palchikov: `getGravityBinary` would be a more descriptive name

lib/utils/parse.go
```diff
@@ -101,6 +101,16 @@ func SplitHostPort(in, defaultPort string) (host string, port string) {
...
 }
 
+// EnsurePort makes sure that the provided address includes a port and adds
+// the specified default one if it does not.
+func EnsurePort(address, defaultPort string) string {
+	if _, _, err := net.SplitHostPort(address); err == nil {
+		return address
+	} else {
+		return net.JoinHostPort(address, defaultPort)
+	}
```
a-palchikov:
```go
if _, _, err := net.SplitHostPort(address); err == nil {
	return address
}
return net.JoinHostPort(address, defaultPort)

```

lib/webapi/clusterinfo.go
```diff
@@ -0,0 +1,155 @@
...
+// webClusterInfo encapsulates basic information about cluster such as
+// management endpoints and status used by the control panel.
+type webClusterInfo struct {
+	// ClusterState is the current cluster state.
+	ClusterState string `json:"clusterState"`
+	// PublicURLs is the advertised public cluster URLs set via auth gateway resource.
+	PublicURLs []string `json:"publicURL"`
+	// InternalURLs is a list of internal cluster management URLs.
+	InternalURLs []string `json:"internalURLs"`
+	// Commans contains various commands that can be run on the cluster.
```
a-palchikov: Comman**d**s

PR #386, opened 2019-04-24
Update changelog for 5.0.33

a-palchikov:


docs/5.x/changelog.md
```diff
@@ -60,6 +60,12 @@ LTS starts with `3.51.0` with minor backwards compatible changes added over time
 
 * Upgrade Kubernetes to `v1.14.0`.
 
+### 5.0.33 LTS
+
+#### Bugfixes
+
+* Fix a regression with 'gravity upgrade --complete'
```
r0mant: Nit: I would use backticks (`) instead of apostrophes so the command gets formatted nicer and add a dot.

PR #389, opened 2019-04-25
Update audit events.

r0mant:
This PR follows a similar PR from Teleport that augments all audit events emitted by Gravity with additional fields such as "code" that uniquely defines each individual event type, "severity" and user-friendly "message". Each event is now represented as a struct with common fields, defined in lib/ops/events package.

Among other notable changes, I've refactored how "resource" related events are emitted. They are no longer emitted by the resource controller because not everything goes through it (e.g. certificate update via UI, user deletion, etc.). Instead, I've moved these events to the backend (operator service) which makes sure everything is captured.

Also, updated Teleport to 3.2 which includes the latest event changes.

Closes https://github.com/gravitational/gravity.e/issues/4031.


lib/ops/events/codes.go
```diff
@@ -0,0 +1,430 @@
...
+	// OperationEnvCompleteCode is the runtime environment update operation complete event code.
+	OperationEnvCompleteCode = "G0014I"
+	// OperationEnvFailureCode is the runtime environment update operation failure event code.
+	OperationEnvFailureCode = "G0014E"
+	// OperationConfigStartCode is the cluster configuration update operation start event code.
+	OperationConfigStartCode = "G0015I"
+	// OperationConfigCompleteCode is the cluster configuration update operation complete event code.
+	OperationConfigCompleteCode = "G0016I"
+	// OperationConfigFailureCode is the cluster configuration update operation failure event code.
+	OperationConfigFailureCode = "G0016E"
```
a-palchikov: I assume there's logic behind the numbering scheme - can you elaborate? Is there a reason not to keep the sub-groups in the same number cluster or why are certain values skipped?

lib/ops/opsservice/authgateway.go
```diff
@@ -45,7 +48,12 @@ func (o *Operator) UpsertAuthGateway(key ops.SiteKey, gw storage.AuthGateway) er
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	return UpsertAuthGateway(client, o.users(), gw)
+	err = UpsertAuthGateway(client, o.users(), gw)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	events.Emit(ctx, o, events.AuthGatewayUpdated, events.Fields{})
```
a-palchikov: nit: Maybe you should have added another API - `EmitWithFields` so that the empty `Fields` would not be necessary?
Another thought is why not keep event and its fields together? This way it would be natural both to construct new events:
```go
// this one has the benefit of allowing an empty constructor:
events.Emit(ctx, o, events.NewTLSKeyPairCreated())
events.Emit(ctx, o, events.NewUserCreated(events.FieldName, user.GetName())
// or
events.Emit(ctx, o, events.NewUserCreated(events.Fields{events.FieldName: user.GetName()})
```
then:
[lib/ops/events/events.go]
```go
type Event struct {
	events.Event
	events.Fields
}
```
and emit them.

lib/process/process.go
```diff
@@ -1706,7 +1706,10 @@ func (p *Process) initClusterCertificate(client *kubernetes.Clientset) error {
 		return trace.Wrap(err)
 	}
 
-	_, err = p.operator.UpdateClusterCertificate(ops.UpdateCertificateRequest{
+	localCtx := context.WithValue(ctx, constants.UserContext,
```
a-palchikov: Why is this not a direct dependency for `UpdateClusterCertificate` but instead is being passed on via context?

lib/ops/events/codes.go
```diff
@@ -0,0 +1,430 @@
...
+	// ApplicationUninstall is emitted when an application release is uninstalled.
+	ApplicationUninstall = events.Event{
+		Name: AppUninstalledEvent,
+		Code: ApplicationUninstallCode,
+	}
+)
+
+const (
+	// OpereationInstallStartCode is the install operation start event code.
+	OperationInstallStartCode = "G0001I"
```
alex-kovoy: Should we also remove the severity suffixes from the code?

lib/ops/events/codes.go
r0mant: Yeah, I tried to maintain some logic just for consistency for these events, e.g.

Operation events are in `00xx` group,
Resource-related events (or things that are created/deleted in general) are in `1xxx` (for "create") and `2xxx` (for "delete") groups,
Cluster-related events (remote support, status ,etc.) are in `3xxx`,
Application-related events are in `4xxx`.

I'll add a comment about it - there's no strict rules about choosing codes (e.g. for future events), I just tried to setup some basic rules for these ones.

lib/ops/events/codes.go
r0mant: I think it's fine to keep it in the code. Makes it convenient to use the same number code for different outcomes of the same thing (e.g. operation succeeded/failed, etc.).

lib/ops/opsservice/authgateway.go
r0mant: I'd really like to avoid having "constructors" for each event type. That was the approach I took originally and it didn't bring any benefits, only 500+ lines of paperweight code b/c we have so many events. It should be fine to emit just the event struct + fields. I'll see about having "Emit" and "EmitWithFields" though, that sounds reasonable.

lib/process/process.go
r0mant: UpdateClusterCertificate can be called by a user (e.g. when a user updates cert via UI) and in that case the context will contain the appropriate authenticated user. In this case, there is no external authenticated "user" per se - the process itself is a user, so it passes itself (`@system`) in the context.

PR #395, opened 2019-05-03
kubeaudit updates

knisbet:
Running integration tests for now.


knisbet:
Looks like it passed tests, so what do you think about including in 6.0?

PR #396, opened 2019-05-06
Web API updates.

r0mant:
A few updates to web API for the new dashboard:

* Return information about installed app releases in `/sites` and `/sites/<clustername>` handlers.
* Add `shallow` flag to the aforementioned methods which, if set to true, makes them return stripped down cluster objects which UI does not always need (saving traffic / increasing performance during polling).
* Remove standalone `/releases` web API method cause they are available as a part of web cluster object now.

### DIFF
--- a/lib/ops/opshandler/resources.go
+++ b/lib/ops/opshandler/resources.go
@@ -20,12 +20,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"strconv"
 
 	"github.com/gravitational/gravity/lib/constants"
 	"github.com/gravitational/gravity/lib/ops"
 	"github.com/gravitational/gravity/lib/ops/opsclient"
 	"github.com/gravitational/gravity/lib/storage"
+	"github.com/gravitational/gravity/lib/utils"
 
 	"github.com/gravitational/roundtrip"
 	telehttplib "github.com/gravitational/teleport/lib/httplib"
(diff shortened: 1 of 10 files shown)


lib/storage/release.go
```diff
@@ -129,6 +135,16 @@ func (r *ReleaseV1) GetChartVersion() string {
 	return r.Spec.ChartVersion
 }
 
+// GetChartIcon returns the chart application icon.
+func (r *ReleaseV1) GetChartIcon() string {
+	return r.Spec.ChartIcon
+}
+
+// SetChartIcon returns the chart application icon.
```
a-palchikov: ~~returns~~ sets the icon to the specified value.
Btw, is ChartIcon the actual icon (base64-encoded?) or a reference to the external resource?

lib/webapi/webapi.go
```diff
@@ -1298,33 +1295,36 @@ func (m *Handler) shrinkSite(w http.ResponseWriter, r *http.Request, p httproute
...
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	cluster, err := context.Operator.GetSite(*key)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	releases, err := getReleases(context.Operator, *cluster)
+	if err != nil {
+		m.Errorf("Failed to retrieve releases information for cluster %v: %v.",
```
a-palchikov: Does this need a return?

lib/ops/opshandler/resources.go
```diff
@@ -143,7 +145,20 @@ func (h *WebHandler) getAuthGateway(w http.ResponseWriter, r *http.Request, p ht
      []storage.Release
 */
 func (h *WebHandler) getReleases(w http.ResponseWriter, r *http.Request, p httprouter.Params, ctx *HandlerContext) error {
-	releases, err := ctx.Operator.ListReleases(siteKey(p))
+	err := r.ParseForm()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	var includeIcons bool
+	if i := r.Form.Get("include_icons"); i != "" {
```
a-palchikov: I've seen a handy `getBoolFlag(r, "include_icons", false)` elsewhere.

lib/storage/release.go
r0mant: It can be either, but in our use-case it is most often the actual b64-encoded icon data that comes from the manifest's "logo" field.

lib/webapi/webapi.go
r0mant: I was intending for this to be a non-critical failure (so cluster UI would still load if releases retrieval fails).

r0mant:
CI failed due to flaky agent test, robotest passed, verified tests manually, merging.

PR #398, opened 2019-05-07
Freeze changeset after installation

UnderGreen:


assets/dns-app/hooks/entrypoint.sh
```diff
@@ -28,6 +28,8 @@ elif [ $1 = "rollback" ]; then
 elif [ $1 = "install" ]; then
     echo "Creating new resources"
     rig upsert -f /var/lib/gravity/resources/dns.yaml
+    echo "Freezing"
```
a-palchikov: nit: I would suggest turning on verbose mode for hook scripts.

PR #399, opened 2019-05-08
Monitoring updates.

r0mant:
* Update monitoring app to a new tag that includes Prometheus-based watcher from https://github.com/gravitational/monitoring-app/pull/123.
* Also update Tiller to the latest version.
* Add missing permissions for gravity-site in monitoring namespace. Without them gravity can't manage resources like SMTP configuration, alerts and alert targets.
* Augment Alert resource to include additional fields that Prometheus alerting rules expose.

Closes https://github.com/gravitational/gravity.e/issues/4062.


lib/ops/resources/gravity/gravity.go
```diff
@@ -444,6 +456,19 @@ func (r *Resources) Remove(ctx context.Context, req resources.RemoveRequest) err
...
 	case storage.KindSMTPConfig:
+		// SMTP configuration can be deleted only if there is no
+		// alert recipient configured, otherwise it will result
+		// into invalid Alertmanager configuration.
+		alertTargets, err := r.Operator.GetAlertTargets(r.cluster.Key())
+		if err != nil && !trace.IsNotFound(err) {
+			return trace.Wrap(err)
+		}
+		if len(alertTargets) != 0 {
+			return trace.BadParameter("SMTP configuration can " +
```
a-palchikov: Does this need to make user do something or can we remove the alert target automatically? This might include a warning to the user when deleting the SMTP configuration that the alert target will be also removed?

r0mant: I was considering this but eventually decided against it as it would be an unexpected side-effect (even if there's a warning). In situations like this I prefer to let the user make decision explicitly.

assets/site-app/resources/monitoring.yaml
```diff
@@ -10,6 +10,9 @@ rules:
   - secrets
   verbs:
   - get
+  - create
+  - update
+  - delete
```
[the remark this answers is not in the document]
r0mant: Yes, for SMTP configuration resource - Gravity manages it via a Secret. We can restrict permission just for this particular resource name though.

PR #401, opened 2019-05-13
Add cluster metrics API.

r0mant:
This is not integrated into the new dashboard yet, but it's been sitting in my branch for too long now (at least couple of weeks) so I'd like to merge this in.

This PR adds a new API for fetching cluster metrics (such as CPU/RAM usage) from in-cluster Prometheus service. This API will be used by the new dashboard to display usage charts on the cluster page.

Also, this PR includes an experimental command `gravity top` that uses "termui" library to draw charts and display cluster metrics right in the terminal. We'll keep iterating on it, adding more stuff and making it more and more useful so you'd be able to monitoring cluster performance right from your CLI.

Closes https://github.com/gravitational/gravity.e/issues/4001.


lib/ops/ops.go
```diff
@@ -1887,6 +1887,54 @@ type Monitoring interface {
...
+	}
+	return nil
+}
+
+// ClusterMetricsResponse is the response containing cluster CPU/RAM metrics.
+type ClusterMetricsResponse struct {
+	// TotalCPUCores is the total number of CPU cores in the cluster.
+	TotalCPUCores int `json:"total_cpu_cores"`
+	// TotalMemoryBytes is the total amount of memory in the cluster.
+	TotalMemoryBytes int64 `json:"total_memory_bytes"`
```
a-palchikov: nit: why not `uint64`?

lib/ops/opsservice/monitoring.go
```diff
@@ -34,6 +35,65 @@ import (
...
+	}
+	currentCPURate, err := metrics.GetCurrentCPURate(ctx)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	maxCPURate, err := metrics.GetMaxCPURate(ctx, req.Interval)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	historicCPURate, err := metrics.GetCPURate(ctx, time.Now().Add(-req.Interval), time.Now(), req.Step)
```
a-palchikov: `GetCPURate` returns results based on the last time interval (similar to `GetMaxMemoryRate`) but expects `start, stop` - is this distinction useful, or you could simply pass just the `interval` and `step` instead?

tool/gravity/cli/top.go
```diff
@@ -0,0 +1,287 @@
...
+		termui.StopLoop()
+	})
+	termui.Loop()
+
+	return nil
+}
+
+// render continuously spins in a loop retrieving cluster metrics and rendering
+// terminal widgets at a certain interval.
+func render(env *localenv.LocalEnvironment, ctx context.Context, client monitoring.Metrics, interval, step time.Duration) {
```
a-palchikov: I wonder if all this belongs in `tool/gravity/cli` and not in a separate package?

lib/ops/monitoring/metrics.go
```diff
@@ -0,0 +1,265 @@
...
+func (p *prometheus) GetMaxMemoryRate(ctx context.Context, interval time.Duration) (int, error) {
+	var query bytes.Buffer
+	if err := queryMaxMemory.Execute(&query, map[string]string{"interval": fmt.Sprintf("%vh", interval.Hours())}); err != nil {
+		return 0, trace.Wrap(err)
+	}
+	vector, err := p.getVector(ctx, query.String())
+	if err != nil {
+		return 0, trace.Wrap(err)
+	}
+	if len(vector) != 1 {
```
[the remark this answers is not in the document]
r0mant: In a general case vector can have more than a single element.

lib/ops/ops.go
```diff
@@ -1887,6 +1887,54 @@ type Monitoring interface {
...
+	GetClusterMetrics(context.Context, ClusterMetricsRequest) (*ClusterMetricsResponse, error)
+}
+
+// ClusterMetricsRequest is a request for cluster metrics.
+type ClusterMetricsRequest struct {
+	// SiteKey is the cluster routing key.
+	SiteKey
+	// Interval is the requested metrics interval.
+	Interval time.Duration `json:"interval"`
+	// Step is the optional maximum time b/w two datapoints.
```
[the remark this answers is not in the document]
r0mant: Default is used. Added the comment.

lib/ops/ops.go
r0mant: Is there a real benefit to using unsigned types other than signifying that these values can't be negative (or if you need those 2x larger numbers)? I think uints are mostly needed for bitwise operations. Otherwise they just introduce a bunch of extra type conversions because most of the code uses signed types. Even if you take standard Go library, e.g. builtin `len` method returns int.

lib/ops/opsservice/monitoring.go
r0mant: We could but the Prometheus API allows retrieving metrics for arbitrary time intervals so I wanted to have this possibility exposed in our API too.

tool/gravity/cli/top.go
r0mant: Well, this draws in the terminal so IMO it fits well in the CLI package for now. We can move this out elsewhere if we keep developing / adding more stuff to it.

lib/ops/monitoring/metrics.go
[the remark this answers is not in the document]
a-palchikov: Yeah - I just haven't seen one such use and I thought it's too early to make this generalization.

PR #404, opened 2019-05-15
Wait for readyness of one gravity pod

UnderGreen:


assets/site-app/images/hook/entrypoint.sh
```diff
@@ -35,6 +35,12 @@ if [ $1 = "update" ]; then
     echo "Checking status"
     rig status $RIG_CHANGESET --retry-attempts=120 --retry-period=1s --debug
 
+    while ! kubectl get pods --namespace=kube-system -l app=gravity-site | grep -q '1/1'
```
a-palchikov: Hm, we have a dedicated command to check whether the service reports healthy (`gravity site status`).
Also, kubectl supports waiting for specific resource conditions:
```sh
$ kubectl wait --for=condition=Ready -nkube-system pod -l app=gravity-site
```

r0mant: Yeah, why not add post-update hook that polls `gravity status`, same as post-install hook does?

PR #407, opened 2019-05-16
Kubernetes proxy fixes.

r0mant:
This PR addresses a few issues with Kubernetes proxy that appeared after upgrading to Teleport 3.2. Teleport starting from 3.2 uses Kubernetes Impersonation API instead of CSR API.

* Revendor latest Teleport to address a bug related to sending empty auth information to API server.
* Update gravity-site RBAC to allow impersonation requests.
* Fix a couple of issues with `tsh login` - e.g. it now expects Teleport proxy to returns "proxy settings" which gravity-site wasn't returning before.

Plus a couple of general fixes:

* Factor out the logic for assembling "cluster endpoints" (auth gateway address and UI) to a common method which is now used by both gravity status and new UI so they display consistent info.
* A couple of tweaks to gravity status output.

### DIFF
--- a/lib/ops/endpoints.go
+++ b/lib/ops/endpoints.go
@@ -1,0 +1,100 @@
+/*
+Copyright 2019 Gravitational, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ops
+
+import (
+	"fmt"
+
+	"github.com/gravitational/gravity/lib/defaults"
+	"github.com/gravitational/gravity/lib/utils"
+
+	"github.com/gravitational/trace"
+)
+
+// ClusterEndpoints contains system cluster endpoints such as Teleport
+// proxy address or cluster control panel URL.
+type ClusterEndpoints struct {
+	// Internal contains internal cluster endpoints.
+	Internal clusterEndpoints
+	// Public contains public cluster endpoints.
+	Public clusterEndpoints
+}
+
+// AuthGateways returns all auth gateway endpoints.
+func (e ClusterEndpoints) AuthGateways() []string {
+	if len(e.Public.AuthGateways) > 0 {
+		return e.Public.AuthGateways
+	}
+	return e.Internal.AuthGateways
+}
+
+// AuthGateway returns a single auth gateway endpoint.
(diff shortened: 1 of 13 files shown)


lib/ops/endpoints.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func getClusterEndpoints(cluster *Site, gateway storage.AuthGateway) (*ClusterEndpoints, error) {
+	// Internal endpoints point directly to master nodes.
+	var internal clusterEndpoints
+	for _, master := range cluster.Masters() {
+		internal.AuthGateways = append(internal.AuthGateways,
+			fmt.Sprintf("%v:%v", master.AdvertiseIP, defaults.GravitySiteNodePort))
+		internal.ManagementURLs = append(internal.ManagementURLs,
+			fmt.Sprintf("https://%v:%v", master.AdvertiseIP, defaults.GravitySiteNodePort))
```
a-palchikov:
```go
internal.ManagementURLs = append(internal.ManagementURLs,
	defaults.GravityClusterNodePort(master.AdvertiseIP))
```
would be more readable and stay at the same level of abstraction (not having to do Sprintf-style formatting).
If you needed to distinguish between internal and public, I'd also make this more explicit by having to queries:
```go
defaults.GravityInternalClusterNodePort(master.AdvertiseIP)
defaults.GravityClusterNodePort(master.AdvertiseIP)
```
I think the tests below will especially benefit from that.

lib/ops/endpoints.go
```diff
@@ -0,0 +1,105 @@
...
+// AuthGateways returns all auth gateway endpoints.
+func (e ClusterEndpoints) AuthGateways() []string {
+	if len(e.Public.AuthGateways) > 0 {
+		return e.Public.AuthGateways
+	}
+	return e.Internal.AuthGateways
+}
+
+// AuthGateway returns a single auth gateway endpoint.
+func (e ClusterEndpoints) AuthGateway() string {
```
a-palchikov: I'd name this `FirstAuthGateway` to make it real clear which one is returned.

lib/process/teleport.go
```diff
@@ -139,3 +144,19 @@ func (p *Process) getAuthGatewayConfig() (storage.AuthGateway, error) {
...
+	settings := client.ProxySettings{
+		Kube: client.KubeProxySettings{
+			Enabled: p.teleportConfig.Proxy.Kube.Enabled,
+		},
+		SSH: client.SSHProxySettings{
+			ListenAddr: p.teleportConfig.Proxy.SSHAddr.String(),
+		},
+	}
+	if len(p.teleportConfig.Proxy.Kube.PublicAddrs) > 0 {
+		settings.Kube.PublicAddr = p.teleportConfig.Proxy.Kube.PublicAddrs[0].String()
```
a-palchikov: Have you considered shortening this to:
```go
settings.Kube.PublicAddr = p.teleportConfig.FirstProxyPublicKubeAddr()
```
?

r0mant: This is teleport struct, it would be nice if it had this method but not sure it's worth the trouble adding it specifically for this PR.

PR #416, opened 2019-05-21
http no_proxy

knisbet:
This changes gravity based processes to append no_proxy settings for .local and 0.0.0.0/0 (all ipv4 addresses) which may not be present in the users shell ahead of time.

Setting 0.0.0.0/0 is unfortunately a bit overly broad and restrictive, but knowing all the correct IP addresses within each context gravity runs is a bit tricky, and running processes would also need to be updated in real-time as nodes are added (ie a Join operation introduces a new peer). I suspect all outbound connections can be made towards a DNS name, so it feels like a reasonable restriction and trade off to no_proxy 0.0.0.0/0. Let me know what you think.

Updates https://github.com/gravitational/gravity.e/issues/3998


tool/gravity/cli/utils.go
```diff
@@ -181,3 +182,32 @@ func (g *Application) isExpandCommand(cmd string) bool {
 	}
 	return false
 }
+
+// ConfigureNoProxy configures the current process to not use any configured HTTP process when connecting to any
```
a-palchikov: HTTP process -> HTTP proxy

tool/gravity/cli/utils.go
```diff
@@ -181,3 +182,32 @@ func (g *Application) isExpandCommand(cmd string) bool {
...
+	proxy := map[string]string{
+		"NO_PROXY": os.Getenv("NO_PROXY"),
+		"no_proxy": os.Getenv("no_proxy"),
+	}
+
+	found := false
+	for k, v := range proxy {
+		if len(v) != 0 {
+			os.Setenv(k, strings.Join([]string{v, "0.0.0.0/0", ".local"}, ","))
+			found = true
```
a-palchikov: this is effectively an early exit:
```go
proxy := map[string]string{
	"NO_PROXY": os.Getenv("NO_PROXY"),
	"no_proxy": os.Getenv("no_proxy"),
}
for k, v := range proxy {
	if len(v) != 0 {
		os.Setenv(k, strings.Join([]string{v, "0.0.0.0/0", ".local"}, ","))
		return
	}
}
os.Setenv("NO_PROXY", strings.Join([]string{"0.0.0.0/0", ".local"}, ","))
```

knisbet:
@r0mant I did consider exposing this through runtime environment, but sort of ended up on this solution for a couple of reasons:

- If a user submits a runtime environment configuration, it didn't seem right to me to append the gravity sessions, and then when a user retrieves the resource seeing that we made changes to it. And then not having a way for a user to make those changes.
- On the other hand, what seems to happen in the field, is a user when configuring runtime environment ports their own settings, and doesn't know what settings gravity may need internally. So if they have their own no_proxy settings that we allow overrides for, they can sort of shoot their self in the foot and gravity doesn't break cleanly when these settings are incorrect.

So that's why I ended up in this direction, within gravity processes we don't need to talk to any customer internal networks afaik, so just default the no_proxy settings, and in planet, I append the default no_proxy settings to any customer provided settings within the RuntimeEnvironment. It's just gravity as a cli will be running with the shell sessions environment variables, and may not have consistent access to load the RuntimeEnvironment settings from the kubernetes API.

knisbet:
Sorry, 1 tiny correction. In gravity processes it does append any no_proxy settings to user shell sessions if they're present, although it shouldn't need to.

PR #428, opened 2019-06-04
Quote environment values if necessary

a-palchikov:
Updates https://github.com/gravitational/gravity.e/issues/4095.
Requires https://github.com/gravitational/planet/pull/425.


lib/ops/opsservice/configure.go
```diff
@@ -932,7 +933,7 @@ func (s *site) getPlanetConfig(config planetConfig) (args []string, err error) {
 	}
 
 	for k, v := range config.env {
-		args = append(args, fmt.Sprintf("--env=%v=%v", k, v))
+		args = append(args, fmt.Sprintf("--env=%v=%v", k, maybeEnvValueQuoted(v)))
```
knisbet: Why not just always quote the argument?

a-palchikov: That's very pragmatic.

PR #433, opened 2019-06-07
Dmitri/4096 install unattended

a-palchikov:
This PR updates the behavior of installer and install agent by moving event dispatching and operation concurrency control out of the gRPC server (lib/install/server).

This enables to execute both automatic operation and individual phase steps from the service (as opposed to implementing these directly in `gravity/cli`) which, in turn, enables better control over which steps can be executed in parallel and when.

Event dispatching has been factored out of `lib/install/server` into a dedicated dispatcher package which enabled testing.

Updates http://github.com/gravitational/gravity.e/issues/4096.


build.assets/Dockerfile
```diff
@@ -42,7 +42,7 @@ RUN (mkdir -p /opt/protoc && \
 RUN cd ${GOPATH}/src/github.com/gogo/protobuf && git reset --hard ${GOGO_PROTO_TAG} && make install
 RUN cd ${GOPATH}/src/github.com/grpc-ecosystem/grpc-gateway && git reset --hard ${GRPC_GATEWAY_TAG} && go install ./protoc-gen-grpc-gateway
 
-ENV PROTO_INCLUDE "/usr/local/include":"${GOPATH}/src":"${GOPATH}/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis":"${GOPATH}/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis":"${GOGOPROTO_ROOT}/protobuf"
+ENV PROTO_INCLUDE "/usr/local/include":"${GOPATH}/src":"${GOPATH}/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis":"${GOPATH}/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis":"${GOGOPROTO_ROOT}/protobuf":"${GOGOPROTO_ROOT}/gogoproto"
```
a-palchikov: I forgot to update the Makefile to generate proto stubs for installer gRPC server.

lib/expand/join.go
```diff
@@ -557,17 +667,14 @@ func (p *Peer) newAgent(opCtx operationContext) (*rpcserver.PeerServer, error) {
 	return agent, nil
 }
 
-func (p *Peer) run() error {
+func (p *Peer) run(ctx operationContext) error {
+	// FIXME: install gravity binary before any plan steps are executed
```
a-palchikov: I will address this in a later PR.

lib/install/install.go
```diff
@@ -101,6 +97,32 @@ func (i *Installer) Stop(ctx context.Context) error {
 	return nil
 }
 
+// Execute executes the install operation using the configured engine.
+// Implements server.Executor
+func (i *Installer) Execute(req *installpb.ExecuteRequest, stream installpb.Agent_ExecuteServer) error {
+	if !i.submit(req) && !req.IsResume() {
+		return status.Error(codes.AlreadyExists, "operation is already active")
```
a-palchikov: Installer does not allow to run multiple plan steps while the agent does.
The agent's main operation is fetching and displaying progress events (as far as the client is concerned) and running remote commands (for the installer) - this means that only the main operation is resumable if the client drops off - from the architecture point of view. If the client reconnects, it needs to continue fetching and displaying events from the ongoing operation.

lib/install/install.go
```diff
@@ -330,181 +270,46 @@ func (i *Installer) stopWithContext(ctx context.Context) error {
...
+	if i.agent != nil {
+		i.agent.Stop(ctx)
 	}
+	err := i.stopAborters(ctx)
 	i.config.Process.Shutdown(ctx)
 	i.server.Stop(ctx)
-	return trace.NewAggregate(errors...)
-}
-
-func (i *Installer) sendElapsedTime(timeStarted time.Time) {
```
a-palchikov: I moved these to another file (`lib/install/operation.go`).

tool/gravity/cli/config.go
```diff
@@ -383,6 +387,11 @@ func (i *InstallConfig) NewInstallerConfig(
 
 }
 
+func (i *InstallConfig) validateApplicationDir() error {
```
a-palchikov: This is an additional test in attempt to move as much validation as possible to as early as possible.
The issue here is that the way client/service interaction works - client needs to connect to the service in order to display anything (incl. errors). If the service has an initialization time failure, it is currently difficult to catch and respond to. I'm still considering options as to how best solve this.

tool/gravity/cli/register.go
```diff
@@ -93,7 +93,7 @@ func RegisterCommands(app *kingpin.Application) *Application {
 	g.InstallCmd.GCENodeTags = g.InstallCmd.Flag("gce-node-tag", "Override node tag on the instance in GCE required for load balanacing. Defaults to cluster name.").Strings()
 	g.InstallCmd.DNSHosts = g.InstallCmd.Flag("dns-host", "Specify an IP address that will be returned for the given domain within the cluster. Accepts <domain>/<ip> format. Can be specified multiple times.").Hidden().Strings()
 	g.InstallCmd.DNSZones = g.InstallCmd.Flag("dns-zone", "Specify an upstream server for the given zone within the cluster. Accepts <zone>/<nameserver> format where <nameserver> can be either <ip> or <ip>:<port>. Can be specified multiple times.").Strings()
-	g.InstallCmd.ExcludeHostFromCluster = g.InstallCmd.Flag("exclude-from-cluster", "Do not use this node in the cluster").Bool()
+	g.InstallCmd.ExcludeHostFromCluster = g.InstallCmd.Flag("remote", "Do not use this node in the cluster").Bool()
```
a-palchikov: This change is as per https://github.com/gravitational/gravity.e/issues/3859#issuecomment-498850080

lib/install/server/dispatcher/buffered/dispatcher.go
```diff
@@ -0,0 +1,121 @@
...
+	case r.respC <- event.AsProgressResponse():
+	case <-r.ctx.Done():
+	}
+}
+
+// Close stops the dispatcher internal processes
+// Implements EventDispatcher
+func (r *Dispatcher) Close() {
+	r.cancel()
+	r.wg.Wait()
```
r0mant: Nit: I wonder if that's a good semantics for "close" to be potentially blocking.

lib/install/server/dispatcher/direct/dispatcher.go
```diff
@@ -0,0 +1,90 @@
...
+				close(r.notifyC)
+				return
+			}
+		}
+	}()
+}
+
+// Dispatcher implements event dispatcher that dispatches events
+// without intermediate buffering.
+// If requires that the receiving side is servicing the notification channel
```
r0mant: If -> It

lib/install/server/dispatcher/buffered/dispatcher.go
a-palchikov: It is only blocking if there's a bug in the implementation - it does not spawn goroutines it does not control that might be the cause of the block in a non-deterministic way.
I could add context to the method but I don't think it will make the API simpler or more robust.

lib/expand/join.go
```diff
@@ -812,47 +944,50 @@ func (p *Peer) validateWizardState(operator ops.Operator) (*ops.Site, *ops.SiteO
...
 
 // eventDispatcher implements eventDispatcher for the agent
 // and maps completed progress event to client completed event
 type eventDispatcher struct {
-	server *server.Server
+	w *dispatcher.EventWriter
+}
+
+func (r operationContext) supportsResume() bool {
+	return r.Operation.Type == ops.OperationExpand
```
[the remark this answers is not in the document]
a-palchikov: I'll try to come up with a name that describes the context better.

PR #437, opened 2019-06-11
Add helper method.

r0mant:
That's for https://github.com/gravitational/gravity.e/pull/4114.

### DIFF
--- a/tool/common/print.go
+++ b/tool/common/print.go
@@ -31,8 +31,8 @@ func PrintError(err error) {
 }
 
 // PrintWarn outputs a warning message to stdout.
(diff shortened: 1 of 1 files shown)


tool/common/print.go
```diff
@@ -30,6 +30,11 @@ func PrintError(err error) {
 	color.Red("[ERROR]: %v\n", trace.UserMessage(err))
 }
 
+// PrintWarn outputs a warning message to stdout.
+func PrintWarn(message string, format ...interface{}) {
```
a-palchikov: nit: `format` would be expceted to be the format string with placeholders - this looks like a slice of arguments, so `args`?

PR #441, opened 2019-06-11
Query pod/container before initiating ssh session.

r0mant:
I've tested this with both standalone cluster and via hub. Closes https://github.com/gravitational/gravity.e/issues/4098.


r0mant:
The ticket mentions Teleport 4 but Gravity still uses 3.2 and I'm not entirely sure what sort of protection is being referred to there, so this fix is still needed. It might be referring to the paths sanitization, which is present in 3.2, but it only applies to roundtrip client IIRC and not the web handlers.

PR #442, opened 2019-06-12
[5.5] fix restart regression

a-palchikov:
* Fix a regression with rolling update with restart not configured to happen on the target node.
 * Configure resource create/remove operations to log similar to install operation - both to `syslog` and a log file.

### DIFF
--- a/lib/install/process.go
+++ b/lib/install/process.go
@@ -35,7 +35,6 @@ func InitProcess(ctx context.Context, installerConfig Config, gravityConfig proc
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
-	setLoggingOptions()
 	err = p.InitRPCCredentials()
 	if err != nil {
 		return nil, trace.Wrap(err)
--- a/lib/process/wizard.go
+++ b/lib/process/wizard.go
@@ -83,10 +83,6 @@ func WizardTeleportConfig(clusterName, stateDir string) *telecfg.FileConfig {
 	return &telecfg.FileConfig{
 		Global: telecfg.Global{
 			DataDir: stateDir,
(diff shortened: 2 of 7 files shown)


lib/install/process.go
```diff
@@ -35,7 +35,6 @@ func InitProcess(ctx context.Context, installerConfig Config, gravityConfig proc
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
-	setLoggingOptions()
```
a-palchikov: I found it strange to be resetting logging configuration out of nowhere.
See below for changes for the similar effect.

lib/process/wizard.go
```diff
@@ -83,10 +83,6 @@ func WizardTeleportConfig(clusterName, stateDir string) *telecfg.FileConfig {
 	return &telecfg.FileConfig{
 		Global: telecfg.Global{
 			DataDir: stateDir,
-			Logger: telecfg.Log{
```
a-palchikov: This change replaces the removal of `setLoggingOptions` from above.

lib/update/internal/rollingupdate/builder.go
```diff
@@ -132,7 +132,8 @@ func (r Builder) restart(server storage.UpdateServer) update.Phase {
 	node := r.node("restart", "Restart container on node %q", server.Hostname)
 	node.Executor = libphase.RestartContainer
 	node.Data = &storage.OperationPhaseData{
-		Package: &r.App,
+		ExecServer: &server.Server,
```
a-palchikov: Inadvertently removed the link to the target server to execute the command one with the recent commit.

PR #449, opened 2019-06-14
Dmitri/4096 debug report

a-palchikov:
This PR adds automatic generation of debug reports upon failure during installation.

I opted to keep the old reporting functionality for now but removed the redundant `GetSiteOperationCrashReport` API as it can be also implemented with a single API - `GetSiteReport`.
Also, while working on it I decided to factor some of the installerclient functionality w.r.t exit strategies as not all uses of client equally want to `abort` the operation, `complete` the operation or `generate debug report`.
I moved this into a new interface named [Lifecycle](https://github.com/gravitational/gravity/blob/dmitri/4096-debug-report/lib/install/client/client.go#L266).

This also restores the functionality of the agents properly exiting with `operation aborted` if the installer process is aborted.

Requires https://github.com/gravitational/gravity.e/pull/4122.


lib/expand/join.go
```diff
@@ -98,6 +101,8 @@ type Peer struct {
 	// server is the gRPC installer server
 	server *server.Server
 
+	// abortC is a channel to signal external operation abort
+	abortC     chan chan struct{}
```
a-palchikov: I needed this channel to be able to guarantee delivery of the abort message to the client (but only if there's a client).

lib/expand/join.go
```diff
@@ -129,10 +137,21 @@ func (p *Peer) Execute(req *installpb.ExecuteRequest, stream installpb.Agent_Exe
...
+	for {
+		select {
+		case event := <-p.dispatcher.Chan():
+			err := stream.Send(event)
+			if err != nil {
+				return trace.Wrap(err)
+			}
+		case doneC := <-p.abortC:
+			err := stream.Send(installpb.AbortEvent)
+			close(doneC)
```
a-palchikov: Acknowledge that the client has been notified.

lib/install/client/automatic_lifecycle.go
```diff
@@ -0,0 +1,119 @@
...
+		// Stream done but no completion event
+		if err1 := c.shutdown(ctx); err1 != nil {
+			c.WithError(err1).Warn("Failed to shut down.")
+		}
+		return nil
+	default:
+		if err1 := r.generateDebugReport(ctx, c); err1 != nil {
+			c.WithError(err1).Warn("Failed to generate debug report.")
+		}
+		if err1 := c.shutdown(ctx); err1 != nil {
```
a-palchikov: I decided to shutdown installer (and agents) on failure so the agent clients at least exit and users have the ability to use the same terminal and look around. The agents are configured to restart and reconnect, so it's a matter of `gravity resume` on the client to reattach to the progress loop.

lib/install/client/automatic_lifecycle.go
```diff
@@ -0,0 +1,119 @@
...
+	// Aborter specifies the completion handler for when the operation is aborted
+	Aborter func(context.Context) error
+	// Completer specifies the optional completion handler for when the operation
+	// is completed successfully
+	Completer CompletionHandler
+	// DebugReportPath specifies the path to the debug report file
+	DebugReportPath string
+	// LocalDebugReporter specifies the handler for generating host-local debug
+	// report
+	LocalDebugReporter func(ctx context.Context, path string) error
```
a-palchikov: Client resorts to a local node debug report if cluster-wide report fails to be generated.

lib/install/operation.go
```diff
@@ -165,10 +157,21 @@ func (i *Installer) completeOperation(operation ops.SiteOperation, status dispat
...
 	return trace.Wrap(i.config.Process.Wait())
 }
 
+func (i *Installer) registerExitHandlersForAgents(op ops.SiteOperation) {
+	i.addAborter(signals.StopperFunc(func(ctx context.Context) error {
+		i.WithField("operation", op.ID).Info("Aborting agent service.")
+		return trace.Wrap(i.config.Process.AgentService().AbortAgents(ctx, op.Key()))
+	}))
+	i.addStopper(signals.StopperFunc(func(ctx context.Context) error {
+		i.WithField("operation", op.ID).Info("Stopping agent service.")
```
a-palchikov: Agents are restarted each time installer restarts.
Agents are aborted only if the installer has been interrupted.

lib/install/server/server.go
```diff
@@ -116,6 +134,12 @@ type Executor interface {
 	Complete(operationKey ops.SiteOperationKey) error
 }
 
+// DebugReporter allows to capture the operation state
+type DebugReporter interface {
```
a-palchikov: Made this into a separate interface so it can be optionally implemented (and it is only implemented by the installer but not agents).

lib/ops/opsservice/report.go
```diff
@@ -41,7 +41,21 @@ import (
...
-func (s *site) getSiteOperationCrashReport(op ops.SiteOperation) (io.ReadCloser, error) {
+func (s *site) getClusterReport() (io.ReadCloser, error) {
+	op, err := storage.GetLastOperationForCluster(s.backend(), s.domainName)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	switch op.Type {
+	case ops.OperationInstall:
+		return s.getClusterInstallReport((ops.SiteOperation)(*op))
```
a-palchikov: This condition lacks the test that the operation is still active - it does not make sense to go that route for a completed operation. This is TBD.

lib/report/packages.go
```diff
@@ -0,0 +1,115 @@
+/*
```
a-palchikov: Moved this to lib/report from tool/gravity/cli.

lib/install/install.go
```diff
@@ -123,6 +123,21 @@ func (i *Installer) Complete(key ops.SiteOperationKey) error {
...
+// Implements server.DebugReporter
+func (i *Installer) GenerateDebugReport(path string) error {
+	i.WithField("path", path).Info("Generate debug report.")
+	op, err := ops.GetWizardOperation(i.config.Operator)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	err = i.generateDebugReport(op.ClusterKey(), path)
+	if err != nil {
+		i.WithError(err).Warn("Failed to generate debug report.")
```
r0mant: It's a little weird that this function both logs the error and returns it below. Should the error be logged by the called instead?

lib/install/operation.go
```diff
@@ -289,6 +292,24 @@ func (i *Installer) emitAuditEvents(ctx context.Context, operation ops.SiteOpera
 	return nil
 }
 
+func (i *Installer) generateDebugReport(clusterKey ops.SiteKey, path string) error {
+	f, err := os.Create(path)
```
r0mant: Should this created file be removed in case this function returns an error?

lib/install/proto/installer.proto
```diff
@@ -36,6 +36,9 @@ service Agent {
 
     // Shutdown requests that the installer service shuts down gracefully
     rpc Shutdown(ShutdownRequest) returns (google.protobuf.Empty);
+
+    // GenerateDebugReport requests that the installer generates the debug report
```
r0mant: Nit: Reads a bit weird, how about "requests the installer to generate a debug report".

lib/ops/opsservice/report.go
r0mant: Hm, this means that if the cluster has been installed and hasn't undergone any other operations yet, `gravity report` will generate install-specific report? Will you add this check for an active operation before merging?

lib/report/collector.go
```diff
@@ -0,0 +1,179 @@
...
+// ScriptCollector is a convenience Collector to execute bash scripts
+type ScriptCollector struct {
+	name   string
+	script string
+}
+
+func tarball(pattern string) string {
+	return fmt.Sprintf(`
+#!/bin/bash
+/bin/tar cz --ignore-failed-read --ignore-command-error -f /dev/stdout -C / $(readlink -e %v) -P 2> /dev/null
```
r0mant: Hm, is this script run over ssh somewhere so it's not possible to use `archive`package instead of running bash? And where does the `pattern` come from? Just thinking about the possibility of injection attack.

lib/storage/utils.go
```diff
@@ -124,6 +124,28 @@ func GetOperations(backend Backend) ([]SiteOperation, error) {
 	return operations, nil
 }
 
+// GetLastOperation returns the last operation for the local cluster
```
r0mant: Nit: "for the **specified** cluster". Same below.

tool/gravity/cli/config.go
```diff
@@ -910,6 +911,22 @@ func InstallerCleanup() error {
...
+func InstallerGenerateLocalReport(env *localenv.LocalEnvironment) func(context.Context, string) error {
+	return func(ctx context.Context, path string) error {
+		f, err := os.Create(path)
+		if err != nil {
+			return trace.ConvertSystemError(err)
+		}
+		defer f.Close()
+		err = systemReport(env, report.AllFilters, true, f)
+		if err != nil {
+			return trace.ConvertSystemError(err)
```
r0mant: Remove created file?

tool/gravity/cli/install.go
```diff
@@ -841,3 +860,8 @@ agent by issuing 'gravity resume' command.
 See https://gravitational.com/gravity/docs/cluster/#managing-an-ongoing-operation for details.
 `))
 }
+
+// DebugReportPath returns the default path for the debug report file
+func DebugReportPath() (path string) {
+	return filepath.Join(filepath.Dir(utils.Exe.Path), "crashreport.tgz")
```
r0mant: I'd move `crashreport.tgz` to lib/defaults.

lib/expand/join.go
knisbet: Nit: It's not clear, atleast based on the comment, why this is a `chan chan struct{}` as opposed to a `chan struct{}` like a stopC that's a sort of common go pattern?

lib/install/client/automatic_lifecycle.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// HandleStatus handles the results of a completed operation.
+// It executes the handler corresponding to the outcome
+func (r *AutomaticLifecycle) HandleStatus(ctx context.Context, c *Client, status installpb.ProgressResponse_Status, err error) error {
+	switch {
+	case err == nil:
+		switch status {
+		case installpb.StatusUnknown:
+			if err1 := c.shutdown(ctx); err1 != nil {
```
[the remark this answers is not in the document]
knisbet: My 2 cents, shadowing err will correctly lead to linting errors, which can lead to bugs when things are improperly shadowed. I would consider renaming error in the function handler to something like statusErr, and then err can be used in the sort of standard pattern though.

lib/ops/opsservice/report.go
a-palchikov: Yes, I will.

lib/report/collector.go
a-palchikov: `pattern` is hardcoded and only used internally.
I will create a separate issue to refactor this using archive package.

PR #451, opened 2019-06-14
Remove port range check for 30000-32767

itay:
Kubernetes by default has the range of 30000-32767 for `NodePort` assignment. However, it is not appropriate to ensure (and fail the install) if something is bound within this range, because Kubernetes will actually only choose open ports when assigning a `NodePort`. Additionally, this range can be changed via configuration, so hardcoding verification of it being unused with no way of changing it is problematic.

It would be great to backport this to 5.5.x as well, as currently we have to maintain a fork of Gravity.

See also https://community.gravitational.com/t/port-usage-and-verification/163/3


a-palchikov:
Resubmitted internally to have the CI execute successfully. CI is not yet set up to accommodate external PRs.

PR #454, opened 2019-06-18
Dmitri/4096 dispatch single phase

a-palchikov:
This PR addresses two more issues from the umbrella issue:
  - `gravity plan execute` for the individual phase on the installer node would block
  - `gravity resume` incorrectly assumed that it should start the operation from scratch (as it failed to query the active operation) 

Updates http://github.com/gravitational/gravity.e/issues/4096.


lib/expand/join.go
```diff
@@ -390,13 +390,16 @@ func (p *Peer) executePhase(ctx context.Context, opCtx operationContext, phase i
 	if err != nil {
 		return trace.Wrap(err)
 	}
+	progressReporter := dispatcher.NewProgressReporter(ctx, disp, phaseTitle(phase))
```
a-palchikov: I'm using progress reporter universally to achieve the same progress output behavior as previously with manual phase execution (with periodic updates with `Still executing ...`).

lib/install/engine/cli/cli.go
```diff
@@ -73,7 +74,15 @@ type Config struct {
 
 // Execute executes the installer steps.
 // Implements installer.Engine
-func (r *Engine) Execute(ctx context.Context, installer install.Interface, config install.Config) (err error) {
+func (r *Engine) Execute(ctx context.Context, installer install.Interface, config install.Config) (dispatcher.Status, error) {
+	err := r.execute(ctx, installer, config)
+	if err != nil {
+		return dispatcher.StatusUnknown, trace.Wrap(err)
+	}
+	return dispatcher.StatusCompleted, nil
```
a-palchikov: I also opted to return the operation status after running the task instead of sending the completion event via a dispatcher as a way to simplify delivery of that final completion event. It is easier to reason about and to implement (bypassing the dispatcher indirection).

lib/install/install.go
```diff
@@ -103,13 +104,26 @@ func (i *Installer) Execute(req *installpb.ExecuteRequest, stream installpb.Agen
...
-		if err != nil {
-			return trace.Wrap(err)
+	for {
+		select {
+		case event := <-i.dispatcher.Chan():
+			err := stream.Send(event)
+			if err != nil {
+				return trace.Wrap(err)
+			}
+		case result := <-i.execDoneC:
```
a-palchikov: This implements the logic of handling completion of a single `operation step` (be it the complete install operation or a single phase in case the installer is resumed). The difficulty here is that this case needs to handle client which might temporarily drop and not be able to accept the notification. It also handles the sending of the final completion event directly to client (bypassing the dispatcher which is used for other but non-critical notifications).

lib/utils/progress.go
```diff
@@ -311,22 +301,32 @@ func (p *progressPrinter) Stop() {
...
 	}
 	p.currentEntry = nil
 }
 
+// printStep prints step instead of the progress bar
+func printStep(out io.Writer, current, target int, message string) {
+	if target > 0 {
+		fmt.Fprintf(out, "* [%v/%v] %v\n", current, target, message)
+	} else {
+		fmt.Fprintf(out, "%v\n", message)
```
a-palchikov: I stripped the time prefix here so progress notifications for other operations might be broken - I will fix this in the next PR.
I felt that time-prefixing should be done elsewhere (and it conflicts with the progress output in the installer).

tool/gravity/cli/install.go
```diff
@@ -782,8 +781,6 @@ func NewServiceListener() (net.Listener, error) {
 // InterruptSignals lists signals installer service considers interrupts
 var InterruptSignals = signals.WithSignals(
 	os.Interrupt,
-	syscall.SIGTERM,
```
a-palchikov: I thought it was wrong afterall that stopping the installer service ( `systemctl stop <installer-service>`) would abort it (with all the consequences).

tool/gravity/cli/journal.go
```diff
@@ -60,20 +60,20 @@ func exportRuntimeJournal(env *localenv.LocalEnvironment, outputFile string) err
 	if err := m.RoBindMount(logDir, journalDir); err != nil {
 		return trace.Wrap(err)
 	}
-
-	cleanup := func(context.Context) error {
+	defer func() {
```
a-palchikov: This was another bug I introduced when I changed the semantics of the lib/system/signals#InterruptHandler to only run stoppers on interruption (and not always as before) - so this had to migrate into a `deferred` function as it needs to run always.

tool/gravity/cli/operation.go
```diff
@@ -366,11 +366,40 @@ func getOperationFromBackend(backend storage.Backend) operationGetter {
...
+			return nil, trace.Wrap(err)
+		}
+		return (*ops.SiteOperation)(op), nil
+	})
+}
+
+func getLocalClusterFromBackend(backend storage.Backend) (cluster *storage.Site, err error) {
+	// TODO(dmitri): when cluster is created by the wizard, it is not local
+	// so resort to look up
+	clusters, err := backend.GetSites(defaults.SystemAccountID)
```
a-palchikov: As the comment explains, when the wizard is installing, the cluster record it creates is not local and multiple utility functions that expect the local cluster do not work so it resorts to iterating through the records (which is supposed to be a single anyways).

lib/install/dispatcher/dispatcher.go
```diff
@@ -89,6 +89,11 @@ type Event struct {
 	Status Status
 }
 
+// IsCompleted returns true if this status indicates completion
```
knisbet: Nit: doc doesn't seem to exactly match implementation. I assume CompletedPending has a slight semantic difference to completed (or it wouldn't exist), so I think the comment should reflect this semantic so it doesn't accidentally get miss-used?

lib/install/install.go
```diff
@@ -180,9 +191,24 @@ func (i *Installer) startExecuteLoop() {
 		for {
 			select {
 			case req := <-i.execC:
-				if err := i.execute(req); err != nil {
-					i.sendError(err)
+				status, err := i.execute(req)
+				select {
+				case <-i.execDoneC:
```
knisbet: I'm not sure if I'm missing something, but why are execResults with errors being written to this channel below, if the errors get discarded if there are more requests?

lib/install/install.go
```diff
@@ -180,9 +191,24 @@ func (i *Installer) startExecuteLoop() {
...
+					// Empty the result channel
+				default:
+				}
+				if err != nil {
 					i.WithError(err).Warn("Failed to execute.")
```
knisbet: Nit: I realize this is existing code, but this should probably be updated to include what failed to execute as fields in the error...

PR #463, opened 2019-06-25
Update install documentation to reflect recent changes

a-palchikov:


docs/5.x/installation.md
```diff
@@ -216,15 +207,43 @@ Tue Apr 10 14:01:09 UTC	Executing install phase "/" finished in 5 minutes
 
 ```
 
-The following CLI flags are useful to manage the install operation:
+To abort an installation, press Ctrl+C two times in a row in the terminal.
+If the operation is aborted, the partial cluster state will be automatically removed.
```
knisbet: Instead of partial cluster state, how about partial install?

docs/5.x/installation.md
```diff
@@ -216,15 +207,43 @@ Tue Apr 10 14:01:09 UTC	Executing install phase "/" finished in 5 minutes
...
 
+Aborting an agent (and not the installer process), will only prevent this node from joining.
```
knisbet: I think this needs to be clearer. Maybe instead of saying aborting an agent, aborting a join? Then it's linked to what the user is doing, and not the internals of the way we see the implementation?

docs/5.x/installation.md
```diff
@@ -216,15 +207,43 @@ Tue Apr 10 14:01:09 UTC	Executing install phase "/" finished in 5 minutes
...
+    has been fixed.
+
+
+Installer processes (that includes the main installer process and all agent processes that join additional nodes to the cluster)
+execute inside a systemd service so the operation will continue in background even if the terminal session has timed out.
+
+In order to reconnect to the installer, issue `./gravity resume`.
+
+!!! warning "Installer state directory":
+    It is important to run install commands from the directory with the original gravity binary.
```
knisbet: It might be a good idea to try and eliminate this limitation in the future (or atleast have a ticket open to do so). I could see users missing this concept in the docs and bumping into this.

docs/5.x/installation.md
```diff
@@ -216,15 +207,43 @@ Tue Apr 10 14:01:09 UTC	Executing install phase "/" finished in 5 minutes
...
+
+Each node runs a systemd service (and may even run two). The main installer service executes
+from a service called `gravity-installer` (agent nodes have a corresponding `gravity-agent`). The services are configured
+from the temporary state sub-directory called `.gravity` in the directory where the installer has been extracted.
+
+To manually remove installation state:
+
+```bash
+# Stop the installer service (or gravity-agent - depending on the node):
+root$ systemctl stop gravity-installer.service
```
knisbet: TODO: maybe this should be incorporated as part of `./gravity leave`?

docs/5.x/installation.md
a-palchikov: What do you see here as limiting?
For install to work, it needs to know where to find original installer data (extracted from the tarball), so running `./gravity install` makes sense.
Then, it needs to look up state just like `git` does - it defaults to a local sub-directory to the directory from where the executable is invoked by default. It does not mean it needs to always use the same binary - it can be the one already installed in `/usr/bin`, but it needs to be the same directory with the install files.

docs/5.x/installation.md
a-palchikov: It is already part of `gravity leave` (more specifically, it is part of `gravity system uninstall` which `gravity leave` implicitly executes).

PR #467, opened 2019-06-26
Do not log binary output of remote report collect call

a-palchikov:
Provide a non-logging Exec API in the agent service to use for debug report collection to silence the output of stdout.
Fixes https://github.com/gravitational/gravity.e/issues/4144.

### DIFF
--- a/lib/system/environ/validate.go
+++ b/lib/system/environ/validate.go
@@ -46,9 +46,9 @@ func ValidateInstall(env *localenv.LocalEnvironment) func() error {
 			log.WithError(err).Warn("Failed to validate state directory requirements.")
 		}
 		// FIXME: validate enough of directory contents to make an educated decision automatically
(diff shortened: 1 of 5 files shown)


lib/system/environ/validate.go
```diff
@@ -46,9 +46,9 @@ func ValidateInstall(env *localenv.LocalEnvironment) func() error {
 			log.WithError(err).Warn("Failed to validate state directory requirements.")
 		}
 		// FIXME: validate enough of directory contents to make an educated decision automatically
-		// if err := validateDirectoryEmpty(stateDir); err != nil {
```
a-palchikov: I think it's better on though this has a downside of not allowing the user to restart `gravity install` after it has created the state directory but before it could start an operation. This will require manual state directory removal with `./gravity leave --force` (which not that bad and there's a corresponding message about it).

PR #468, opened 2019-06-26
Handle server interrupt gracefully - without invoking abort mechanism

a-palchikov:
Currently, service does not handle at least `SIGTERM` to gracefully shutdown and instead is killed by the system as a result. I was mistakenly believing I could not use interrupt signals to properly shutdown just the service.

This PR handles server interrupts (`SIGTERM`/`SIGQUIT`) gracefully - without invoking abort mechanism of the client.

### DIFF
--- a/lib/install/operation.go
+++ b/lib/install/operation.go
@@ -187,7 +187,6 @@ func (i *Installer) newCompletionEvent(status dispatcher.Status) *dispatcher.Eve
 
 func (i *Installer) runStoppers(ctx context.Context, stoppers []signals.Stopper) error {
 	var errors []error
-	i.WithField("stoppers", stoppers).Info("Executing stoppers.")
 	for _, c := range stoppers {
 		if err := c.Stop(ctx); err != nil {
 			errors = append(errors, err)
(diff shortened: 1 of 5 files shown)


lib/install/operation.go
```diff
@@ -187,7 +187,6 @@ func (i *Installer) newCompletionEvent(status dispatcher.Status) *dispatcher.Eve
 
 func (i *Installer) runStoppers(ctx context.Context, stoppers []signals.Stopper) error {
 	var errors []error
-	i.WithField("stoppers", stoppers).Info("Executing stoppers.")
```
a-palchikov: This used to spew warning about `stopper.StopperFunc` being non-serializable.

tool/gravity/cli/clusterupdate.go
```diff
@@ -66,9 +66,6 @@ func updateTrigger(
 		return trace.Wrap(err)
 	}
 	defer updater.Close()
-	if !manual {
```
a-palchikov: Left-over from previous refactoring gone wrong - I think the second `if` makes more sense.

PR #479, opened 2019-07-02
Updated network section with Ports & Subnet requirements

eldios:
Updated network section with Ports & Subnet requirements


docs/5.x/requirements.md
```diff
@@ -88,27 +88,43 @@ These ports are required during initial installation and can be closed after the
...
+| 3080                    | TCP                | HTTPS                           | ext         | controllers | Teleport Web  UI                         |
+| 3008-3012               | TCP                | HTTPS / gRPC                    | all         | all         | Internal Gravity services                |
+| 32009                   | TCP                | HTTPS                           | ext         | controllers | Gravity Cluster/OpsCenter Admin UI (ext) |
+| 32009                   | TCP                | HTTPS                           | controllers | all         | Gravity internal API                     |
+| 3012                    | TCP                | HTTPS                           | all         | all         | Gravity RPC agent                        |
+
+!!! Source/Destination Legend:
+  * all - Any node which is a member of the cluster
+  * ext - Any source outside the cluster
+  * localhost - The port is only used within the host were the request started
```
knisbet: I think this can possibly be simplified, `port is only used within a host`??

docs/5.x/requirements.md
```diff
@@ -88,27 +88,43 @@ These ports are required during initial installation and can be closed after the
...
+| 8472                    | UDP                | VXLAN                           | all         | all         | Overlay network                          |
+| 7496                    | TCP/UDP            | HTTPs                           | all         | all         | Serf (Health check agents) peer to peer  |
+| 7373                    | TCP                | RPC                             | localhost   | localhost   | Serf RPC - peer to peer                  |
+| 7575                    | TCP                | gRPC                            | all         | all         | Cluster status gRPC API                  |
+| 2379, 2380, 4001, 7001  | TCP                | HTTPS                           | all         | controllers | Etcd server communications               |
+| 6443                    | TCP                | HTTPS                           | all         | controllers | Kubernetes API Server                    |
+| 30000 - 32767           | N.A                | depend on the services deployed | all         | all         | Kubernetes internal services range       |
```
r0mant: Depend**s**

Or I would rather rephrase to "Application specific". Also, I think it's "N/A", not "N.A" usually.

docs/5.x/requirements.md
```diff
@@ -88,27 +88,43 @@ These ports are required during initial installation and can be closed after the
...
-| 6443                    | HTTPS                                   | Kubernetes API Server                     |
-| 30000 - 32767           | HTTPS (depend on the services deployed) | Kubernetes internal services range        |
-| 10248 - 10250, 10255    | HTTPS                                   | Kubernetes components                     |
-| 5000                    | HTTPS                                   | Docker registry                           |
-| 3022-3025               | SSH                                     | Teleport internal SSH control panel       |
-| 3080                    | HTTPS                                   | Teleport Web  UI                          |
-| 3008-3012               | HTTPS                                   | Internal Gravity services                 |
-| 32009                   | HTTPS                                   | Gravity Cluster/OpsCenter Admin panel UI  |
-| 3012                    | HTTPS                                   | Gravity RPC  agent                        |
+| Port                    | Protocol - Layer 4 | Protocol - Layer 5              | Source      | Destination | Description                              |
```
r0mant: Not sure I follow what exactly is meant by "source" and "destination" here. Can you elaborate?

docs/5.x/requirements.md
a-palchikov: It should be also noteworthy, that we recently removed enforcement of this range as kubernetes is using in on a best-effort basis (i.e. selects unoccupied ports), so it might be taken out from the list of hard requirements.

docs/5.x/requirements.md
a-palchikov: Also `were` -> `where`

eldios:
okay, we should be good to go.. please approve as soon as you can guys. :+1:

docs/5.x/requirements.md
knisbet: The ports are still important to be listed in some way for AWS / Load balancer integrations, since the removal of the satellite check is a bit different. Even if kubernetes can bind to the port, doesn't mean the firewall allows the port to be used.

PR #487, opened 2019-07-05
Preserve original URL when making redirects to login screen

alex-kovoy:
closes https://github.com/gravitational/gravity.e/issues/4155

### DIFF
--- a/lib/webapi/webhandler.go
+++ b/lib/webapi/webhandler.go
@@ -479,7 +479,8 @@ func (h *WebHandler) needsLogin(handle webHandle) httprouter.Handle {
 			if !trace.IsAccessDenied(err) {
 				log.Error(trace.DebugReport(err))
 			}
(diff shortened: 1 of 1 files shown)


lib/webapi/webhandler.go
```diff
@@ -479,7 +479,8 @@ func (h *WebHandler) needsLogin(handle webHandle) httprouter.Handle {
 			if !trace.IsAccessDenied(err) {
 				log.Error(trace.DebugReport(err))
 			}
-			http.Redirect(w, r, "/web/login", http.StatusFound)
+
+			http.Redirect(w, r, "/web/login?redirect_uri="+r.URL.Path, http.StatusFound)
```
r0mant: Do you need to URL-encode `r.URL.Path`?

alex-kovoy: Not in this case because we do not pass or take any information from this request.

PR #493, opened 2019-07-09
Update wait phase for wait for kube-system namespace.

r0mant:
Just had an install fail on "rbac" phase with "kube-system namespace not found error".

This PR updates the "wait" phase to wait for this namespace existence as well.

### DIFF
--- a/lib/install/phases/postsystem.go
+++ b/lib/install/phases/postsystem.go
@@ -91,17 +91,32 @@ func (p *waitExecutor) waitForAPI(ctx context.Context, done chan bool) {
 	for {
 		select {
 		case <-timer.C:
-			_, err := p.Client.CoreV1().ComponentStatuses().Get("scheduler", metav1.GetOptions{})
-			if err != nil {
-				p.Info("Waiting for kubernetes API to start: ", err)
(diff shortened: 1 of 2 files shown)


lib/install/phases/postsystem.go
```diff
@@ -104,6 +109,16 @@ func (p *waitExecutor) waitForAPI(ctx context.Context, done chan bool) {
 	}
 }
 
+func (p *waitExecutor) tryQueryAPI() error {
+	_, err := p.Client.CoreV1().ComponentStatuses().Get("scheduler", metav1.GetOptions{})
+	return trace.Wrap(err)
+}
+
+func (p *waitExecutor) tryQueryNamespace() error {
+	_, err := p.Client.CoreV1().Namespaces().Get(defaults.KubeSystemNamespace, metav1.GetOptions{})
```
a-palchikov: nit: I don't really an issue with this at this point, but it would make sense to look into informers and the wait functionality as implemented in client-go/kubectl.

r0mant: Yeah, it might be useful for a more generic scenario, I feel like it'd be an overkill in this case.

PR #495, opened 2019-07-09
Fix resources subfolder deletion in initContainer

### LINKED ISSUE #452: [BUG] app that contains resources directory cannot install
**Describe the bug**

If the root directory of a gravity app contains a resources subdirectory, the built application will not install.

```
/app.yaml
/resources
/resources/something.yaml
```

**To Reproduce**

Download quickstart reference
Add resources directory: `mkdir mattermost/resources/resources`
Add file: 
```
cat mattermost/resources/resources/namespace.yaml
apiVersion: v1
kind: Namespace
metadata:
  name: test
```
Build app: `tele build -o mattermost.tar mattermost/resources/app.yaml`
Install built app.

**Expected behavior**

The install to complete successfully

**Logs**

The init container for the install hook fails with the following logs:
```
kevin-test1:/$ kctl logs install-50bbd4-5pgpx -c init
gravitational.io/mattermost:2.2.0 unpacked at /var/lib/gravity/resources
mv: can't rename '/var/lib/gravity/resources/resources/charts': Directory not empty
mv: can't rename '/var/lib/gravity/resources/resources/resources': Directory not empty
```

 - OS [e.g. Redhat 7.4]: Ubuntu 16.04
 - Gravity [e.g. 5.5.4]: 6.0.0-rc.1
 - Platform [e.g. Vmware, AWS]: GCE (local provider)

**Additional context**

The install logs should probably pull the init container logs if the installer see's the init container failing.

knisbet:
Fixes #452 

In #452 if the application directory contains a resources subfolder, hooks will fail. So change this to attempt to remove the resources folder, but ignore failures if the folder isn't empty after the mv.

### DIFF
--- a/lib/app/hooks/configure.go
+++ b/lib/app/hooks/configure.go
@@ -375,13 +375,13 @@ TMPDIR={{.StateDir}} {{.StateDir}}/gravity --state-dir={{.StateDir}} app unpack 
 	--insecure --ops-url=$ops_url \
 	{{.Package}} {{.ResourcesDir}};
 mv {{.ResourcesDir}}/resources/* {{.ResourcesDir}}
-rm -r {{.ResourcesDir}}/resources || true
+rmdir --ignore-fail-on-non-empty {{.ResourcesDir}}/resources
 `))
 
 var initInstallScriptTemplate = template.Must(template.New("sh").Parse(`
 TMPDIR={{.StateDir}} /opt/bin/gravity app unpack --service-uid={{.ServiceUser.UID}} {{.Package}} {{.ResourcesDir}}
 mv {{.ResourcesDir}}/resources/* {{.ResourcesDir}}
-rm -r {{.ResourcesDir}}/resources || true
+rmdir --ignore-fail-on-non-empty {{.ResourcesDir}}/resources
 `))
 
 type initScriptContext struct {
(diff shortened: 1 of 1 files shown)


lib/app/hooks/configure.go
```diff
@@ -375,13 +375,13 @@ TMPDIR={{.StateDir}} {{.StateDir}}/gravity --state-dir={{.StateDir}} app unpack
 	--insecure --ops-url=$ops_url \
 	{{.Package}} {{.ResourcesDir}};
 mv {{.ResourcesDir}}/resources/* {{.ResourcesDir}}
```
a-palchikov: Agree with @r0mant here.

I think we should fix the `app unpack` command to extract the contents of `<package>/resources/` and not `<package>/.` as it does now. Since this is the only user of `app unpack`, it is relatively safe to change the semantics of the command, although we can always add a new hidden flag to support this mode to make it backwards-compatible if required.
The main issue is moving sub-directory called the same as the existing directory in the target directory (`resources` in this case), so I'd simply try and eliminate the need for mv in the first place - i.e. extract the contents of `./resources` directory without the directory itself.

knisbet:
My bad, this is what I get for writing it down, having it pop back up and thinking the errors was in the wrong spot. Lets abandon this one and Dima should submit his branch as a PR then.

a-palchikov:
Ok - I'll finish the implementation and submit the PR.

PR #500, opened 2019-07-11
[5.5] unpack application resources without the 'resources' directory

### LINKED ISSUE #452: [BUG] app that contains resources directory cannot install

a-palchikov:
In #452, if the application resource directory itself contains a `resources` sub-directory, hook init containers fail.
To fix this, this PR changes the semantics of `app unpack` sub-command to unpack application resources from the tarball's `resources` directory directly into the specified directory without the `resources` directory itself. This should be a safe (and sensible!) change as the hook's init container code is the only client of `app unpack`.

Fixes #452.

### DIFF
--- a/lib/archive/archive.go
+++ b/lib/archive/archive.go
@@ -109,11 +109,39 @@ func Extract(r io.Reader, dir string) error {
 			return trace.Wrap(err)
 		}
 
-		if err := extractFile(tarball, header, dir); err != nil {
+		if err := extractFile(tarball, header, dir, header.Name); err != nil {
 			return trace.Wrap(err)
 		}
 	}
 	return nil
+}
+
+// ExtractWithPrefix extracts the contents of the specified tarball under directory dir.
+// Only the files/directories found in tarDirPrefix in the tarball matching patterns are extracted.
+// Note, files from tarDir are written directly to dir omitting tarDir.
+// The resulting files and directories are created using the current user context.
+func ExtractWithPrefix(r io.Reader, dir, tarDirPrefix string) error {
+	return trace.Wrap(TarGlobWithPrefix(tar.NewReader(r), tarDirPrefix, func(match *tar.Header, r *tar.Reader) error {
+		relpath, err := filepath.Rel(tarDirPrefix, match.Name)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		// Security: ensure tar doesn't refer to file paths outside the directory
+		err = validPath(relpath, dir, "invalid file path")
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		if match.Linkname != "" {
+			err = validPath(match.Linkname, dir, "invalid link path")
+			if err != nil {
+				return trace.Wrap(err)
+			}
+		}
+		if err := extractFile(r, match, dir, relpath); err != nil {
+			return trace.Wrap(err)
+		}
+		return nil
+	}))
 }
 
 // HasFile returns nil if the specified tarball contains specified file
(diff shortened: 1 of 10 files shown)


tool/gravity/cli/app.go
```diff
@@ -424,7 +424,7 @@ func unpackAppResources(env *localenv.LocalEnvironment, loc loc.Locator, dir, op
 		}
 		defer reader.Close()
 
-		if err := dockerarchive.Untar(reader, dir, archive.DefaultOptions()); err != nil {
+		if err := archive.ExtractGlob(reader, dir, "resources", []string{"*"}); err != nil {
```
r0mant: Nit: `defaults.ResourcesDir`?

lib/archive/archive.go
```diff
@@ -109,13 +109,31 @@ func Extract(r io.Reader, dir string) error {
...
 }
 
+// ExtractGlob extracts the contents of the specified tarball under directory dir.
+// Only the files/directories found in tarDir in the tarball matching patterns are extracted.
+// Note, files from tarDir are written directly to dir omitting tarDir.
+// The resulting files and directories are created using the current user context.
+func ExtractGlob(r io.Reader, dir, tarDir string, patterns []string) error {
+	return trace.Wrap(TarGlob(tar.NewReader(r), tarDir, patterns, func(tarRelPath string, header *tar.Header, f *tar.Reader) error {
+		// Security: ensure tar doesn't refer to file paths outside the directory
+		err := SanitizeTarPath(header, dir)
```
r0mant: Nit: I'd try to be consistent with how err is handled at least in the same function, i.e. let's use either

```go
if err := xxx(); err != nil {
```

style or

```go
err := xxx()
if err != nil {
```

in both places.

tool/gravity/cli/app.go
r0mant: Also, should the "default" directory (if dir is empty) above be updated to include "resources" directory?

```go
if dir == "" {
	unpackedDir := filepath.Join(env.StateDir, defaults.PackagesDir, defaults.UnpackedDir)
	dir = pack.PackagePath(unpackedDir, loc)
}
```

a-palchikov: I'd rather have directory argument mandatory.

lib/archive/archive.go
```diff
@@ -109,13 +109,42 @@ func Extract(r io.Reader, dir string) error {
...
+// Note, files from tarDir are written directly to dir omitting tarDir.
+// The resulting files and directories are created using the current user context.
+func ExtractWithPrefix(r io.Reader, dir, tarDirPrefix string) error {
+	err := TarGlobWithPrefix(tar.NewReader(r), tarDirPrefix, func(match *tar.Header, r *tar.Reader) error {
+		relpath, err := filepath.Rel(tarDirPrefix, match.Name)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		// Security: ensure tar doesn't refer to file paths outside the directory
+		err = validPath(relpath, dir, "invalid file path")
```
knisbet: I think my only worry is, that this replicates other sanitization code, which leaves some risk if something new is discovered missing in the tar sanitization, that the semi-duplicate code get's missed. I see why it was separated out, that's just my only thought on the risks of doing it this way.

a-palchikov: I updated to keep using `SanitizeTarPath` as is.

PR #502, opened 2019-07-11
Dmitri/install failed precondition

a-palchikov:
This PR fixes a couple of deficiencies in handling initialization failures in agent/installer services.
It implements early service failures that prevent it from restarting, adds support for detecting a failed service to the client connection loop and uses the fact that the service is `failed` to be able to restart the `gravity install` / `gravity join` commands.

Requires https://github.com/gravitational/gravity.e/pull/4177.
Fixes https://github.com/gravitational/gravity.e/issues/4164.


lib/expand/join.go
```diff
@@ -150,18 +150,23 @@ func (p *Peer) Execute(req *installpb.ExecuteRequest, stream installpb.Agent_Exe
...
-				return trace.Wrap(err)
+				return err
 			}
 		case err := <-p.execDoneC:
 			if err != nil {
-				// Phase finished with an error
-				return status.Error(codes.FailedPrecondition, err.Error())
+				// Phase finished with an error.
+				if installpb.IsRPCError(err) {
+					return trace.Unwrap(err)
```
a-palchikov: I need to relay RPC errors to client unchanged since client has specific logic around specific error types and this gets messed up if we wrap the errors at RPC API boundary with `trace.Wrap`.

lib/expand/join.go
```diff
@@ -150,18 +150,23 @@ func (p *Peer) Execute(req *installpb.ExecuteRequest, stream installpb.Agent_Exe
...
+				}
+				// Aborted code is RPC-specific here - it indicates an operation
+				// failure that can be retried
+				return status.Error(codes.Aborted, err.Error())
```
a-palchikov: I've been previously using `codes.FailedCondition` as the error type for execute failures but I reconsidered it after re-reading the comment:
```go
// FailedPrecondition indicates operation was rejected because the
// system is not in a state required for the operation's execution.
// For example, directory to be deleted may be non-empty, an rmdir
// operation is applied to a non-directory, etc.
//
// A litmus test that may help a service implementor in deciding
// between FailedPrecondition, Aborted, and Unavailable:
//  (a) Use Unavailable if the client can retry just the failing call.
//  (b) Use Aborted if the client should retry at a higher-level
//      (e.g., restarting a read-modify-write sequence).
//  (c) Use FailedPrecondition if the client should not retry until
//      the system state has been explicitly fixed. E.g., if an "rmdir"
//      fails because the directory is non-empty, FailedPrecondition
//      should be returned since the client should not retry unless
//      they have first fixed up the directory by deleting files from it.
//  (d) Use FailedPrecondition if the client performs conditional
//      REST Get/Update/Delete on a resource and the resource on the
//      server does not match the condition. E.g., conflicting
//      read-modify-write on the same resource.
```
which made me believe `codes.Aborted` is better suited here and I'm using `codes.FailedPrecondition` to mark the early stage initialization failures.

lib/install/client/automatic_lifecycle.go
```diff
@@ -52,24 +53,30 @@ func (r *AutomaticLifecycle) HandleStatus(ctx context.Context, c *Client, status
 		if err := r.generateDebugReport(ctx, c); err != nil {
 			c.WithError(err).Warn("Failed to generate debug report.")
 		}
-		if err := c.shutdown(ctx); err != nil && !isServerUnavailableError(err) {
+		var err error
+		if pb.IsFailedPreconditionError(statusErr) {
+			err = c.shutdownWithExitCode(ctx, defaults.FailedPreconditionExitCode)
```
a-palchikov: If the client was connected, it will make the service exit with the specific exit code. Otherwise, the service is supposed to exit with the same error code on its own.
Having client control the server being the preferred way allowed to simplify a lot of workflows previously, so this is sticking to it.

lib/install/proto/client.go
```diff
@@ -28,35 +28,77 @@ import (
...
 				return nil, trace.Wrap(err)
 			}
 			return conn, nil
 		}),
 	}
-	dialOptions = append(dialOptions, opts...)
+	dialOptions = append(dialOptions, config.DialOptions...)
 	conn, err := grpc.DialContext(ctx, "unix:///installer.sock", dialOptions...)
 	if err != nil {
+		if wrappedErr, ok := trace.Unwrap(err).(wrappedError); ok {
```
a-palchikov: Need this abomination to detect whether the error from Dial is `transport.ConnectionError` (so it can be formatted in a friendlier way) which is in an internal package.

lib/install/proto/client.go
```diff
@@ -28,35 +28,77 @@ import (
...
+	DialOptions []grpc.DialOption
+}
+
 // SocketPath returns the default path to the installer service socket
 func SocketPath() (path string, err error) {
 	return state.GravityInstallDir("installer.sock")
 }
+
+// Temporary returns false for this error
+func (serviceError) Temporary() bool {
```
a-palchikov: This implements the undeclared interface tested for by the gRPC client connection code to be able to tell it (by returning `false`) that the connection should not be retried after this error.

lib/install/proto/proto.go
```diff
@@ -75,26 +77,51 @@ func (r *ExecuteRequest_Phase) IsResume() bool {
...
+}
+
 // ErrAborted defines the aborted operation error
 var ErrAborted = utils.NewExitCodeErrorWithMessage(defaults.AbortedOperationExitCode, "operation aborted")
 
 // ErrCompleted defines the completed operation error.
 // This is not an error in the usual sense - rather, it indicates that the operation
 // has been completed and that the agent should not shut down and not restart
-var ErrCompleted = utils.NewExitCodeErrorWithMessage(defaults.AbortedOperationExitCode, "operation completed")
+var ErrCompleted = utils.NewExitCodeErrorWithMessage(defaults.CompletedOperationExitCode, "operation completed")
```
a-palchikov: This was awkward.

lib/utils/error.go
```diff
@@ -354,13 +355,22 @@ func ShouldReconnectPeer(err error) error {
...
+func NewFailedPreconditionError(err error) error {
+	return WrapExitCodeError(defaults.FailedPreconditionExitCode, err)
+}
+
 // ExitCodeError defines an interface for exit code errors
 type ExitCodeError interface {
 	error
 	ExitCode() int
+	// OrigError returns the original error this error wraps.
+	// Implements trace.Wrappable
```
a-palchikov: Note: there's actually no `trace.Wrappable` but should be as implementing the complete `trace.Error` interface is overkill.

lib/system/environ/validate.go
```diff
@@ -80,19 +108,29 @@ func validateNonVolatileDirectory(stateDir string, printer utils.Printer) error
...
+	manager, err := systemservice.New()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	status, err := manager.StatusService(serviceName)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	switch status {
+	case systemservice.ServiceStatusFailed, systemservice.ServiceStatusUnknown:
```
r0mant: How about "inactive"?

lib/defaults/defaults.go
```diff
@@ -1155,6 +1155,11 @@ var (
 	// successfully.
 	// The exit code is used to prevent the agent service from restarting after shut down
 	CompletedOperationExitCode = 253
```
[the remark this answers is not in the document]
knisbet: My 2 cents: I don't tend to make 0 a success code, I tend to use 0 as UNKNOWN, since if the variable is uninitialized to 0 it will stand out and not accidentally get interpreted as success when it's actually uninitialized.

lib/expand/join.go
knisbet: Nit: This answer should probably be summarized in a comment, to answer the why codes.Aborted...

lib/install/proto/installer.proto
```diff
@@ -74,6 +74,9 @@ message AbortRequest {
 message ShutdownRequest {
     // Completed indicates that the operation has been successfully completed
     bool completed = 1;
+    // ExitCode optionally specifies the exit code for the service.
+    // 0 value is considered success
+    int32 exitCode = 2;
```
knisbet: Nit on the comment, 0 is success but was is code 2?

lib/install/proto/proto.go
knisbet: lol... it's a mistake I would make :)

lib/install/server/server.go
```diff
@@ -127,7 +128,11 @@ func (r *Server) Shutdown(ctx context.Context, req *installpb.ShutdownRequest) (
 	if req.Completed {
 		r.completed(ctx)
 	} else {
-		r.done(ctx)
+		var err error
+		if req.ExitCode != 0 {
+			err = utils.NewExitCodeError(int(req.ExitCode))
```
knisbet: Nit: Why not have NewExitCodeError return a nil error on ExitCode 0

lib/system/environ/validate.go
```diff
@@ -45,9 +47,13 @@ func ValidateInstall(env *localenv.LocalEnvironment) func() error {
 			// Operational state directory requirements are advisory
 			log.WithError(err).Warn("Failed to validate state directory requirements.")
 		}
-		// FIXME: validate enough of directory contents to make an educated decision automatically
-		if err := validateDirectoryEmpty(stateDir); err != nil {
-			return trace.Wrap(err)
+		if err := validateNoActiveService(stateDir); err != nil {
```
knisbet: Is it safe to validate only the service is running, and not the contents of the stateDir?

lib/utils/error.go
knisbet: Nit: the note compared to the comment would confuse me if I encountered the comment when coming across this later...

lib/install/proto/installer.proto
a-palchikov: it's a protobuf tag value, not the value for the exit code if I understood your comment correctly.

lib/install/client/install.go
```diff
@@ -125,6 +125,17 @@ func (r *InstallerStrategy) checkAndSetDefaults() (err error) {
...
 
+// isServiceFailed returns an error if the service has failed.
+func (r *InstallerStrategy) isServiceFailed() error {
+	serviceName := service.Name(r.ServicePath)
+	failed, err := service.IsFailed(serviceName)
+	if err == nil && failed {
+		return trace.Errorf("service %q has failed. Check journal log for details.",
+			serviceName)
+	}
+	return nil
```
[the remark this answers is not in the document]
a-palchikov: Yes, it only returns an error when the service has definitely failed.

lib/defaults/defaults.go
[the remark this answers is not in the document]
a-palchikov: It is a very specific success code - one that should have a specific effect on the service. One can manually stop the service which would also exit with 0 but will not be finished (as is implied by this code).

lib/expand/join.go
a-palchikov: It is the "boundary" function.

lib/install/proto/installer.proto
knisbet: My bad.

lib/install/client/install.go
```diff
@@ -147,3 +159,30 @@ type InstallerStrategy struct {
 	// installer service connection.
 	ConnectTimeout time.Duration
 }
+
+// updateEnviron relays part of the client's environment to service
+func updateEnviron(req *systemservice.NewServiceRequest) {
```
r0mant: Nit: Instead of having this method with a side-effect here, what do you think about making this a part of `NewServiceRequest` API? I.e. it could have a field that list all environment variables (by names) that need to be propagated to the service and then systemservice implementation would basically do what this method does.

In general, I notice that our packages tend to have a lot of these one-off utility methods which seem unrelated to the main "idea" of the package they're in, while instead they could be made a part of the API (where possible) which would make them easier discoverable/reusable and lead to better structured code.

lib/install/install.go
```diff
@@ -110,7 +111,7 @@ func (i *Installer) Execute(req *installpb.ExecuteRequest, stream installpb.Agen
 		case result := <-i.execDoneC:
 			if result.err != nil {
 				// Phase finished with an error
-				return status.Error(codes.FailedPrecondition, result.err.Error())
+				return status.Error(codes.Aborted, result.err.Error())
```
r0mant: To be frank, I still have no idea why would one use one error code versus another, and it seems to be vital to know where to use what because it affects the logic of the install service AFAIU.

Can these generic codes maybe be adopted to our needs? E.g. have our own defined errors wrapping these ones, with clear names/descriptions and use them everywhere we need to return an error which affects the service behavior. We already have `Abort`, maybe also have `StopWithoutClean` (totally made up) or something.... you get the idea.

lib/install/proto/client.go
```diff
@@ -28,35 +28,92 @@ import (
...
+		}
 		return nil, trace.Wrap(err)
 	}
 	client := NewAgentClient(conn)
 	return client, nil
 }
 
+func (r *ClientConfig) checkAndSetDefaults() error {
+	if r.SocketPath == "" {
+		return trace.BadParameter("Socket path is required")
```
r0mant: Nit: "Socket" -> "socket".

lib/install/proto/client.go
r0mant: What does "undeclared interface" mean? Maybe this note should also be a part of the Godoc then?

Sorry I missed this first time around.

lib/system/environ/validate.go
```diff
@@ -45,9 +49,21 @@ func ValidateInstall(env *localenv.LocalEnvironment) func() error {
...
+			log.WithError(err).Warn("Failed to determine if service is not active.")
+			return trace.BadParameter("detected an active installer service in %v, "+
+				"please resume the agent with `gravity resume` or "+
+				"clean it up using `gravity leave --force` before proceeding "+
+				"(see https://gravitational.com/gravity/docs/cluster/#deleting-a-cluster for more details)",
+				stateDir)
+		}
+		if err := validateNoActiveOperation(stateDir); err != nil {
+			log.WithError(err).Warn("Failed to detect an active operation.")
```
r0mant: I don't want to be pedantic but this should probably only be logged in case `!trace.IsAlreadyExists(err)` :) Same for the if above.

lib/system/environ/validate.go
```diff
@@ -56,6 +72,28 @@ func ValidateInstall(env *localenv.LocalEnvironment) func() error {
 	}
 }
 
+// GetServiceName returns the name of the service configured in the specified state directory stateDir
+func GetServiceName(stateDir string) (name string, err error) {
```
r0mant: Nit: Goes to my previous comment about "random" methods in packages - is there a better place for these two (GetServiceName/GetServicePath)? They must be used elsewhere too since they're exported?

lib/system/environ/validate.go
```diff
@@ -111,6 +159,39 @@ func validateNoPackageState(packages pack.PackageService, stateDir string) error
...
+	}
+	return nil
+}
+
+func newBackendFromDir(dir string) (storage.Backend, error) {
+	backend, err := keyval.NewBolt(keyval.BoltConfig{
+		Path:     filepath.Join(dir, defaults.GravityDBFile),
+		Multi:    true,
+		Readonly: true,
+		Timeout:  -1, // No timeout
```
r0mant: Nit:

```go
Timeout: keyval.NoTimeout,
```

```go
// keyval/constants.go
NoTimeout = -1
```

lib/install/install.go
a-palchikov: Yes, but on gRPC boundary all errors are wrapped in gRPC error.

lib/system/environ/validate.go
a-palchikov: Why is it a random method? It is in the environment package which implements support for working with the environment, including detecting services based on the contents of the state directory.

lib/install/client/install.go
r0mant: My gripe is exactly with the abundance of these "utility methods" that seem out of place everywhere in our code base. As a result, we oftentimes end up with a lot of code duplication b/c a person that needs a certain feature (incl. myself) just can't know there's a random method in a random package that does that. I also think that there's nothing bad about adding some extra functionality to a service, even if it's only gonna be used in a single place yet - makes it easier to discover and reuse in future as I mentioned (and brings more structure to the code).

Anywho, I'm not picking on this particular function, just wanted to share my thoughts with you.

lib/system/environ/validate.go
r0mant: Okay, that makes sense.

lib/install/install.go
r0mant: Yeah, that makes sense. At the very least we should probably document what impact each returned code has on the service (if it's not yet).

PR #505, opened 2019-07-15
Add command to set the phase state.

r0mant:
This PR adds a new hidden subcommand that sets the specified phase state without executing it which is oftentimes useful when repairing failed upgrades or other operations. Implementation-wise this follows how execute/rollback/complete are implemented due to all the differences in how FSMs for different operations work. The new command looks like this:

```
gravity plan set --phase=/checks --state=completed
```

I've also forward-ported a change to add "short" display format for operation plan from 5.5 which I added while working on the Docker device issue.

Closes https://github.com/gravitational/gravity.e/issues/3215.

### DIFF
--- a/lib/fsm/fsm.go
+++ b/lib/fsm/fsm.go
@@ -277,6 +277,22 @@ func (f *FSM) RollbackPhase(ctx context.Context, p Params) error {
 	return nil
 }
 
+// ChangePhaseState updates the specified phase state.
+func (f *FSM) ChangePhaseState(ctx context.Context, change StateChange) error {
+	if err := change.Check(); err != nil {
+		return trace.Wrap(err)
+	}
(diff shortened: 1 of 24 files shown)


lib/fsm/fsm.go
```diff
@@ -551,6 +567,18 @@ type StateChange struct {
 	Error trace.Error
 }
 
+// StateChange verifies that state change is valid.
+func (c StateChange) Check() error {
+	if c.Phase == "" {
+		return trace.BadParameter("phase name must not be empty")
+	}
+	if !utils.StringInSlice(storage.OperationPhaseStates, c.State) {
```
a-palchikov: This needs to be an external API from lib/storage:
```go
if storage.IsValidOperationState(c.State) {
```

PR #511, opened 2019-07-16
Fix a flaky test

a-palchikov:
When the test generates the application packages for local/remote services, there's a potential race if two separate packages are generated with the same contents.
The issue is with file/directory item timestamps that can make the two packages differ. Instead, generate a single application package and replicate it in both services.

Requires https://github.com/gravitational/gravity.e/pull/4171.


lib/app/service/test/builder.go
```diff
@@ -57,6 +57,15 @@ func CreateDummyApplication2(apps app.Applications, loc loc.Locator, dependencie
 	return createApp(loc, dependencies, apps, c)
 }
 
+// CreateDummyApplicationInMultipleServices creates a test application with a valid manifest
+// in all of the specified application services
+func CreateDummyApplicationInMultipleServices(loc loc.Locator, c *C, services ...app.Applications) {
```
r0mant: Nit, I wonder why not just tweak `CreateDummyApplication` to accept variadic argument. This would require reordering its arguments though so "apps" goes last, is that a lot of refactoring?

PR #528, opened 2019-07-18
Add alertmanager to gravity status

knisbet:
For your consideration, this pulls in alerts from the prometheus alertmanager to gravity status. A couple of notes:
- On this iteration, I don't attempt to tweak the overall status to degraded if alerts are present. This can be refined if it seems appropriate. 
- The format I've started with looks like this:
```
root@kevin-test1:~# /tmp/gravity status
Cluster status:	degraded
Application:	telekube, version 6.0.0-rc.4
Join token:	3b3a6d0cc147
Last completed operation:
    * operation_expand (a1a265d9-2cfd-4838-844c-9ad1ad74946b)
      started:		Thu Jul 11 01:36 UTC (1 week ago)
      completed:	Thu Jul 11 01:38 UTC (1 week ago)
Cluster endpoints:
    * Authentication gateway:
        - 10.162.0.7:32009
        - 10.162.0.6:32009
        - 10.162.0.5:32009
    * Cluster management URL:
        - https://10.162.0.7:32009
        - https://10.162.0.6:32009
        - https://10.162.0.5:32009
Cluster nodes:	modestritchie4624
    Masters:
        * kevin-test1 / 10.162.0.7 / node
            Status:	healthy
        * kevin-test2 / 10.162.0.6 / node
            Status:	healthy
        * kevin-test3 / 10.162.0.5 / node
            Status:	healthy
    Nodes:
        * kevin-test4 / 10.162.0.4 / node
            Status:	healthy
        * kevin-test5 / 10.162.0.3 / node
            Status:	healthy
Cluster Alerts:
    * CoreDNSDown [5h19m13s]
      - CoreDNS has disappeared from Prometheus target discovery.
    * TargetDown [22m16s]
      - 20% of the nethealth targets are down.
```
- The raw alerts are also included with gravity status when printed as json
- I filter out any alerts that appear to come from satellite
- I had to bump the dns-app, since it looks like it didn't get bumped after the last time it was updated

Let me know what you think? I think it's the type of thing where we'll have to fine tune gravity status and the alerts as we go.


lib/status/status.go
```diff
@@ -33,16 +34,20 @@ import (
...
+	alertmanager "github.com/prometheus/alertmanager/api/v2/client"
+	"github.com/prometheus/alertmanager/api/v2/client/alert"
+	"github.com/prometheus/alertmanager/api/v2/models"
 	"github.com/sirupsen/logrus"
 )
 
 // FromCluster collects cluster status information.
 // The function returns the partial status if not all details can be collected
-func FromCluster(ctx context.Context, operator ops.Operator, cluster ops.Site, operationID string) (status *Status, err error) {
+func FromCluster(ctx context.Context, operator ops.Operator, cluster ops.Site, operationID, dnsAddr string) (status *Status, err error) {
```
knisbet: I'm not super happy about this, I wanted to pass the entire environment for grabbing the status, but it seems to create a circular dependency tree. So for now only passing the dnsAddress in use.

lib/defaults/defaults.go
```diff
@@ -537,6 +537,9 @@ const (
 	// PrometheusServiceAddr is the Prometheus HTTP API service address.
 	PrometheusServiceAddr = "prometheus-k8s.monitoring.svc.cluster.local:9090"
 
+	// AlertmanagerServiceAddr is the Prometheus Aletermanager HTTP API service address.
```
r0mant: Aletermanager -> Alertmanager

lib/status/status.go
r0mant: Yeah, we actually pass DNS config around quite a bit in different place around the code which is unfortunate. I would at least convert these arguments to a struct (`StatusConfig` or something) so you don't have to pass empty string `""` where it's not needed.

```go
type StatusConfig struct {
  Operator ops.Operator
  Cluster ops.Site
  OperationID String
  DNSAddress string
}

func FromCluster(ctx context.Context, conf StatusConfig) ...
```

lib/status/status.go
```diff
@@ -129,6 +134,14 @@ func FromCluster(ctx context.Context, operator ops.Operator, cluster ops.Site, o
 	}
 
 	status.State = cluster.State
+
+	// Collect information from alertmanager
+	if dnsAddr != "" {
```
r0mant: Is this a good way to determine whether we need to collect alertmanager alerts or not? Why does it depend on passing DNS address? Maybe explicit flag would be better?

Also, I wonder how does it work on older clusters? I.e. if you run `gravity status` using this binary on an older cluster - what would happen?

lib/status/status.go
```diff
@@ -557,3 +572,91 @@ const (
...
 	roleTag = "role"
 )
+
+// FromAlertManager collects alerts from the prometheus alertmanager deployed to the cluster
+func FromAlertManager(ctx context.Context, dnsAddr string) ([]*models.GettableAlert, error) {
+	client, err := httplib.GetPlanetClient(httplib.WithLocalResolver(dnsAddr))
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	transport := httptransport.NewWithClient(defaults.AlertmanagerServiceAddr, "/api/v2",
```
r0mant: What do you think about moving Alertmanager client creation to `lib/clients`? There are already a few there, this way we'll have a single place where API clients are created.

lib/status/status.go
```diff
@@ -557,3 +572,91 @@ const (
...
+					alertname: "WatchdogDown",
+					severity:  critical,
+				},
+			},
+		})
+	}
+	return filtered
+}
+
+func stringAddr(s string) *string {
```
r0mant: This method belongs in `lib/utils/units.go` (see there are `IntPtr` and other similar methods there at the bottom).

lib/status/status.go
```diff
@@ -557,3 +572,91 @@ const (
...
+	getOk, err := am.Alert.GetAlerts(alert.NewGetAlertsParams().WithContext(ctx))
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	return filterAlerts(getOk.Payload), nil
+}
+
+const (
+	alertname    = "alertname"
```
r0mant: These constants here and the method below seem to be Alertmanager-specific, how about moving them to a dedicated module (file) in this package (`alertmanager.go`)?

tool/gravity/cli/status.go
```diff
@@ -279,6 +280,11 @@ func printStatusText(cluster clusterStatus) {
 		printAgentStatus(*cluster.Agent, w)
 	}
 
+	if len(cluster.Alerts) > 0 {
+		fmt.Fprintln(w, "Cluster Alerts:")
```
r0mant: Nit: Don't want to seem too pedantic but all other sections of gravity status output capitalize only the first word so I'd rename this to `Cluster alerts:` so it looks consistent.

lib/status/status.go
knisbet: Is this a good way to determine whether we need to collect alertmanager alerts or not? 
- It's not about determining whether or not to collect the alertmanager, it's to prevent undefined behaviour if the status is called without the DNS configuration.

Why does it depend on passing DNS address?
- Alertmanager runs as a kubernetes serice/pods. So we need to discover the service address in use, which requires connecting to the dns service discovery. Alternatively it might be possibly to set alertmanager to a static IP, but we would still need dynamic config to calculate the IP based on the service range used, and I generally prefer to let kubernetes be the authority on IP assignments.

Maybe explicit flag would be better?
- Again, my thinking was it's not about turning on/off, more that if the config is incomplete that we don't run into undefined behaviour. 

Also, I wonder how does it work on older clusters?
It should fail to resolve the service discovery address, and I believe throw an error similar to if etcd is down or it's unable to reach planet. Although I'm not sure under what circumstances a gravity 6 binary will have someone run `gravity status` against a non 6+ cluster??

lib/status/status.go
knisbet: I'm not sure what value wrapping the client would add?

lib/status/status.go
r0mant: I can easily imagine someone running gravity status using new binary on an old cluster. Say, when upgrading from pre-6.0 to 6.0+ and using the new binary. I'm sure you can imagine a few too, customers are unpredictable as you know :)

Anyway, it makes sense about the dns addr, but we need to make sure that the failure is handled gracefully here and not fails the whole gravity status output.

lib/install/operation.go
```diff
@@ -217,7 +217,7 @@ func (i *Installer) getClusterStatus() (*status.Status, error) {
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
-	status, err := status.FromCluster(i.ctx, clusterOperator, *cluster, "")
+	status, err := status.FromCluster(i.ctx, clusterOperator, *cluster, "", "")
```
a-palchikov: I agree with @r0mant that we should at least move these to a configuration struct to minimize impact.

lib/status/status.go
```diff
@@ -557,3 +572,91 @@ const (
 	// roleTag is the name of the tag containing node role
 	roleTag = "role"
 )
+
+// FromAlertManager collects alerts from the prometheus alertmanager deployed to the cluster
+func FromAlertManager(ctx context.Context, dnsAddr string) ([]*models.GettableAlert, error) {
+	client, err := httplib.GetPlanetClient(httplib.WithLocalResolver(dnsAddr))
```
a-palchikov: This is an undesired implicit dependency we should avoid by passing the planet client in configuration. The client will be created by the caller where dns configuration is available (local environment has it as does the cluster metadata record).
Also, the cluster record that `FromCluster` receives already contains the cluster's DNS configuration.

lib/status/status.go
```diff
@@ -557,3 +572,91 @@ const (
...
+				}
+			}
+		}
+
+		filtered = append(filtered, alert)
+	}
+
+	// if we didn't find the watchdog alert, it indicates there is a problem with the alerting system, acting as a
+	// sort of deadman switch
+	if !hasWatchdog {
```
a-palchikov:
```go
if hasWatchdog {
	return filtered
}
// augment filtered with a failure to find watchdog alert
```

lib/status/status.go
```diff
@@ -557,3 +572,91 @@ const (
...
+		// filter satellite alerts, because they're already collected directly from satellite
+		if job, ok := alert.Labels[job]; ok && job == satellite {
+			continue
+		}
+
+		// Prometheus detects the gravity-site election process as an error since only one pod is ever ready
+		// alertname: KubeDaemonSetRolloutStuck
+		// daemonset: gravity-site
+		// message: Only 33.33333333333333% of the desired Pods of DaemonSet kube-system/gravity-site are scheduled...
+		if name, ok := alert.Labels[alertname]; ok && name == rolloutStuck {
```
a-palchikov: I wonder if there's a way to set up prometheus to ignore firing this alert in the first place?

tool/gravity/cli/status.go
```diff
@@ -154,13 +155,13 @@ func statusPeriodic(env *localenv.LocalEnvironment, printOptions printOptions, s
...
 // statusOnce collects cluster status information
-func statusOnce(ctx context.Context, operator ops.Operator, operationID string) (*statusapi.Status, error) {
+func statusOnce(ctx context.Context, operator ops.Operator, operationID string, env *localenv.LocalEnvironment) (*statusapi.Status, error) {
 	cluster, err := operator.GetLocalSite()
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
 
-	status, err := statusapi.FromCluster(ctx, operator, *cluster, operationID)
+	status, err := statusapi.FromCluster(ctx, operator, *cluster, operationID, env.DNS.Addr())
```
a-palchikov: `cluster` already has DNS configuration.

tool/gravity/cli/status.go
```diff
@@ -378,6 +384,18 @@ func printNodeStatus(node statusapi.ClusterServer, w io.Writer) {
...
 }
 
+func printPrometheusAlerts(alerts []*models.GettableAlert, w io.Writer) {
+	for _, alert := range alerts {
+		if *alert.Status.State != "active" {
+			continue
+		}
+
+		duration := time.Now().Sub(time.Time(*alert.StartsAt)).Round(time.Second)
+		fmt.Fprintf(w, "    * %v [%v]\n", alert.Labels["alertname"], duration)
```
a-palchikov: I'm wonder if this is not going to cause indentation issues in the long run when the output would need to be readjusted for some reason. Have you considered accepting indent as argument and using tabs?

lib/status/status.go
knisbet: I don't know, I look at it from the perspective that all the layers of indirection and renaming things is where I struggle the most with this code base. So I want to connect to the alertmanager, I don't just create an alertmanager client, I have to go find the gravity client for the alert manager, that doesn't really do anything but wrap the client construction to give it a new name. So that's where I struggle to see the value add. Sure, someone needs to understand what they're doing, but even with a client it just means more code to explore to figure out how it works in order to be able to call it, and inevitably will require refactoring when someone wants to add behaviours.

lib/status/status.go
knisbet: While early return works for error handling, isn't it generally considered confusing/error prone for value returns?? Presumably this function will have additional capabilities added in the future that could occur after the check for watchdog.

lib/status/status.go
knisbet: No idea.

tool/gravity/cli/status.go
knisbet: Unless I'm missing something, I attempted to write this consistent with the rest of the status formatting:
https://github.com/gravitational/gravity/blob/master/tool/gravity/cli/status.go#L336-L360
https://github.com/gravitational/gravity/blob/master/tool/gravity/cli/status.go#L362-L379
https://github.com/gravitational/gravity/blob/master/tool/gravity/cli/status.go#L313-L334

lib/status/status.go
UnderGreen: I will take a look how to silence the alert for gravity site.

PR #530, opened 2019-07-19
Add prechecks to join operation

r0mant:
This pull request adds `/checks` phase to the join operation which runs all single-node tests on the joining node as well as some multi-node tests (such as time drift) using one of masters as an anchor.

Note that much of this PR is just refactoring and moving stuff around so it was possible to reuse all the existing checks infrastructure. The actual new phase is in `lib/expand/phases/checks.go`.

Closes https://github.com/gravitational/gravity.e/issues/3767.

### DIFF
--- a/lib/checks/checks.go
+++ b/lib/checks/checks.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 Gravitational, Inc.
+Copyright 2018-2019 Gravitational, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -44,24 +44,19 @@ import (
 	"github.com/gravitational/satellite/monitoring"
 	"github.com/gravitational/trace"
 	"github.com/pborman/uuid"
-	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 )
+
+var log = logrus.WithField(trace.Component, "checks")
 
 // New creates a new checker for the specified list of servers using given
 // set of server information payloads and the specified interface for
 // running remote commands.
-func New(remote Remote, servers []Server, manifest schema.Manifest, requirements map[string]Requirements) (*checker, error) {
-	for _, server := range servers {
-		if _, exists := requirements[server.Server.Role]; !exists {
-			return nil, trace.NotFound("no requirements for node profile %q", server.Server.Role)
-		}
+func New(config Config) (*checker, error) {
+	if err := config.Check(); err != nil {
+		return nil, trace.Wrap(err)
 	}
-	return &checker{
-		remote:       remote,
-		servers:      servers,
-		manifest:     manifest,
-		requirements: requirements,
-	}, nil
+	return &checker{Config: config}, nil
 }
 
 // ValidateManifest verifies the specified manifest against the host environment.
@@ -219,27 +214,6 @@ func RunLocalChecks(ctx context.Context, req LocalChecksRequest) error {
 	return nil
 }
 
-// PortsForProfile parses ports ranges from the specified node profile
-func PortsForProfile(profile schema.NodeProfile) (tcp, udp []int, err error) {
-	for _, ports := range profile.Requirements.Network.Ports {
-		for _, portRange := range ports.Ranges {
-			parsed, err := utils.ParsePorts(portRange)
-			if err != nil {
-				return nil, nil, trace.Wrap(err)
-			}
-			switch ports.Protocol {
-			case "tcp":
-				tcp = append(tcp, parsed...)
-			case "udp":
-				udp = append(udp, parsed...)
-			default:
-				return nil, nil, trace.BadParameter("unknown protocol for port: %q", ports.Protocol)
-			}
-		}
-	}
-	return tcp, udp, nil
-}
-
 // FormatFailedChecks returns failed checks formatted as a list
 func FormatFailedChecks(failed []*agentpb.Probe) string {
 	if len(failed) == 0 {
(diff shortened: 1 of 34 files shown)


lib/checks/checks.go
```diff
@@ -310,163 +308,155 @@ type Server struct {
...
-			log.Warnf("Failed to validate remote node: %v.", trace.DebugReport(err))
-			errors = append(errors,
-				trace.BadParameter("failed to validate remote node %v", server))
-		}
-		if len(failed) != 0 {
-			errors = append(errors, trace.BadParameter("%v failed checks:\n%v",
-				server, FormatFailedChecks(failed)))
-		}
+	for _, server := range r.Servers {
+		errors = append(errors, r.RunSingleNodeChecks(ctx, server)...)
```
a-palchikov: `TestNode`?

lib/checks/checks.go
```diff
@@ -310,163 +308,155 @@ type Server struct {
...
+	}
 
-		err = checkServerProfile(server, requirements)
-		if err != nil {
-			errors = append(errors, err)
-		}
+	// run checks that take all servers into account
+	errors = append(errors, r.RunMultiNodeChecks(ctx, r.Servers)...)
```
a-palchikov: `TestNodes`?

lib/checks/checks.go
```diff
@@ -310,163 +308,155 @@ type Server struct {
...
+	requirements := r.Reqs[server.Server.Role]
+	validateCtx, cancel := context.WithTimeout(ctx, defaults.AgentValidationTimeout)
+	defer cancel()
+	failed, err := r.Remote.Validate(validateCtx, server.AdvertiseIP, ValidateConfig{
+		Manifest: r.Manifest,
+		Profile:  server.Server.Role,
+		Docker:   requirements.Docker,
+	})
+	if err != nil {
+		log.Warnf("Failed to validate remote node: %v.", trace.DebugReport(err))
```
a-palchikov: `log.WithError(err)` will not cripple the stack trace.

lib/checks/checks.go
```diff
@@ -310,163 +308,155 @@ type Server struct {
...
 
-	// run checks that take all servers into account
-	err := checkSameOS(r.servers)
+	err = checkSystemPackages(server, dockerConfig)
 	if err != nil {
 		errors = append(errors, err)
 	}
 
-	err = checkTime(time.Now().UTC(), r.servers)
+	err = r.checkTempDir(ctx, server)
```
a-palchikov: Btw, temporary directory check seems to be broken as temporary directory is not reported by agents so this basically tests as if the temporary directory == working directory of the process.
I'll create a ticket for this.

lib/checks/checks.go
```diff
@@ -310,163 +308,155 @@ type Server struct {
...
 
-			// use the maximum throughput measured over a couple of tests
-			for i := 0; i < 3; i++ {
-				speed, err := r.checkServerDisk(ctx, server.Server, target.path)
-				if err != nil {
-					return trace.Wrap(err)
-				}
-				maxBps = utils.MaxInt64(speed, maxBps)
-			}
+// checkDisks runs disk performance check on the provided server and makes
```
a-palchikov: `checkDisks verifies that disk performance satisfies the profile requirements`

lib/checks/checks.go
```diff
@@ -579,12 +569,12 @@ func (r *checker) checkBandwidth(ctx context.Context) error {
...
 	for addr, result := range resp {
 		ip, _ := utils.SplitHostPort(addr, "")
-		server, err := findServer(r.servers, ip)
+		server, err := findServer(servers, ip)
 		if err != nil {
 			return trace.Wrap(err)
 		}
 
-		requirements := r.requirements[server.Server.Role]
+		requirements := r.Reqs[server.Server.Role]
```
a-palchikov: nit: I'd simply export the old attribute: **R**equirements as abbreviation does not make it any clearer

lib/expand/phases/checks.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	var errors []error
+	errors = append(errors, checker.RunSingleNodeChecks(ctx, *node)...)
+	// Use one of master nodes as an "anchor" so when running multi-node
+	// checks the joining node will be compared against that master (e.g.
+	// for same OS check, time drift check, etc.)
+	errors = append(errors, checker.RunMultiNodeChecks(ctx, []checks.Server{*master, *node})...)
+	return trace.NewAggregate(errors...)
```
a-palchikov:
```go
return trace.NewAggregate(
	checker.RunSingleNodeChecks(ctx, *node),
	checker.RunMultiNodeChecks(ctx, []checks.Server{*master, *node}),
)
```

helgi:
@r0mant Could this `/checks` be used instead of doing a curl from `worker` to a `master`with the eqiv of `curl $GRAVITY_MASTER_HOST/telekube/gravity` to determine if the leader has completed its installation and is available (I understand I will want to download the binary potentially to get the "right" version).

r0mant:
@helgi Could you give more details on your use-case please? Feel free to hit me up in Slack. In general, the join operation will not start unless one of masters is available.

helgi:
@r0mant I was wondering if we can replace https://github.com/gravitational/terraform-gravity/blob/master/aws/cloud-init/worker.script.tpl#L15-L24 with a call to `/checks`?

In the above the `while` loop + `curl` achieving 2 things:
1. Ensure leader election is finished and the leader has completed their setup
2. Correct gravity binary is downloaded to the worker prior to running `autojoin`

In my scenario I would want to verify point 1 but I would not want to download the binary. That's why I wanted to see if `/checks` could be `curl`'ed from the workers to verify readiness of the leader.

Or if I have the proper gravity binary already, will `autojoin` now be able to better retry until the leader is ready thanks to `/checks`?

We can catch up on Slack or Keybase if this isn't clear enough.

r0mant:
@helgi Oh, in this case `/checks` refers to an operation plan phase, not the API method so you can't curl it.

But you don't necessarily need to wait for the installation to finish before running "gravity join" - it should be smart enough to detect that the installation is going and wait until it's finished before starting expand operation. I'm pretty sure "autojoin" is the same.

I guess in the code block you linked above the retry loop is more for fetching the gravity binary (cause you need gravity-site running/ready for this) than for waiting to initiate join operation.

r0mant:
@helgi So I guess to sum up what I attempted to convey above - if you manage to get a proper `gravity` binary on the joining node beforehand, you don't need the "repeat" loop and can just launch join right away - the command will wait for install to finish before proceeding.

PR #533, opened 2019-07-22
Add changelog for 5.0.34 release. Add LTS word to all lts releases

UnderGreen:


docs/5.x/changelog.md
```diff
@@ -125,12 +125,20 @@ LTS starts with `3.51.0` with minor backwards compatible changes added over time
 
 ## 5.x Releases
 
+### 5.0.34 LTS
+
+### Bugfixes
+* Fix security issue with Kapacitor service.
```
r0mant: If there was a security issue, should these notes include a warning similar to the security releases below? cc @knisbet (and maybe give more info about what the issue was?)

docs/5.x/changelog.md
```diff
@@ -125,12 +125,20 @@ LTS starts with `3.51.0` with minor backwards compatible changes added over time
...
+
+### Improvements
+* Improved validation of gravity-site upgrade.
```
r0mant: Based on the changelog b/w .33 and .34, how about:
```
Add `postUpdate` hook to `gravity-site` to make sure it's up and running after an upgrade.
```

docs/5.x/changelog.md
knisbet: I agree with Roman

UnderGreen: Can you take a look at the new sentence?

knisbet: looks better to me.

PR #538, opened 2019-07-24
Display info messages for nodes.

r0mant:
This PR revendors Satellite and makes sure informational probes (e.g. node conditions set by [Node Problem Detector](https://github.com/kubernetes/node-problem-detector)) are displayed in `gravity status`.

Requires https://github.com/gravitational/satellite/pull/112 and https://github.com/gravitational/planet/pull/442.

Closes https://github.com/gravitational/gravity.e/issues/4058.


lib/status/status.go
```diff
@@ -350,6 +350,8 @@ type ClusterServer struct {
 	Status string `json:"status"`
 	// FailedProbes lists all failed probes if the node is not healthy
 	FailedProbes []string `json:"failed_probes,omitempty"`
+	// InfoProbes lists all informational failed probes
+	InfoProbes []string `json:"info_probes,omitempty"`
```
a-palchikov: See my comment about warning vs info - I feel info is an unnecessary replacement for warnings which are already in.

r0mant: It is not really a replacement, the behavior is different: "warning" probes lead to "degraded" status, "info" probes - don't. I replied to your comment there.

PR #539, opened 2019-07-24
Dmitri/4184 expand resilience

a-palchikov:
This is the first part of changes to improve expand operation resilience and fix blunt bugs:

 - [x] Check certificate validity prior to use to avoid time sync problems between nodes upfront.
 - [x] Improve the expand operation by synchronizing remote operation creation with local state to avoid resumption on empty state. Also improve startup by making synchronizing gradually (i.e. validating and creating operation plan if none has been created, starting and waiting for operation/agents before commencing with the plan, etc.)

Updates github.com/gravitational/gravity.e/issues/4184.

### DIFF
--- a/lib/expand/bootstrap.go
+++ b/lib/expand/bootstrap.go
@@ -37,12 +37,6 @@ func (p *Peer) init(ctx operationContext) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	if !ctx.hasOperation() {
-		// This indicates the fact that the expand operation is initialized
-		// at a later moment and the operation might not be available after
-		// having successfully connected to the cluster.
-		return nil
-	}
 	return p.startAgent(ctx)
 }
 
@@ -113,7 +107,7 @@ func (p *Peer) ensureServiceUserAndBinary(ctx operationContext) error {
 
 // syncOperationPlan synchronizes operation and plan data to the local join backend
 func (p *Peer) syncOperationPlan(ctx operationContext) error {
-	err := p.syncOperation(ctx)
+	err := p.syncOperation(ctx.Operator, ctx.Cluster, ctx.Operation.Key())
 	if err != nil {
 		return trace.Wrap(err)
 	}
@@ -130,26 +124,23 @@ func (p *Peer) syncOperationPlan(ctx operationContext) error {
 }
 
 // syncOperation synchronizes operation-related data to the local join backend
-func (p *Peer) syncOperation(ctx operationContext) error {
+func (p *Peer) syncOperation(operator ops.Operator, cluster ops.Site, operationKey ops.SiteOperationKey) error {
 	// sync cluster
-	err := p.JoinBackend.DeleteSite(ctx.Cluster.Domain)
+	err := p.JoinBackend.DeleteSite(cluster.Domain)
 	if err != nil && !trace.IsNotFound(err) {
 		return trace.Wrap(err)
 	}
-	_, err = p.JoinBackend.CreateSite(ops.ConvertOpsSite(ctx.Cluster))
+	_, err = p.JoinBackend.CreateSite(ops.ConvertOpsSite(cluster))
 	if err != nil {
 		return trace.Wrap(err)
 	}
 	// sync operation
-	operation, err := ctx.Operator.GetSiteOperation(ctx.Operation.Key())
+	operation, err := operator.GetSiteOperation(operationKey)
 	if err != nil {
 		return trace.Wrap(err)
 	}
 	_, err = p.JoinBackend.CreateSiteOperation(storage.SiteOperation(*operation))
-	if err != nil {
-		return trace.Wrap(err)
-	}
-	return nil
+	return trace.Wrap(err)
 }
 
 // startAgent starts a new RPC agent using the specified operation context.
@@ -195,7 +186,7 @@ func (p *Peer) newAgent(ctx operationContext) (*rpcserver.PeerServer, error) {
 // getPeerAddrAndToken returns the peer address and token for the specified role
 func getPeerAddrAndToken(ctx operationContext, role string) (peerAddr, token string, err error) {
 	peerAddr = ctx.Peer
-	if strings.Contains(peerAddr, "http") { // peer may be an URL
+	if strings.HasPrefix(peerAddr, "http") { // peer may be an URL
 		peerURL, err := url.Parse(ctx.Peer)
 		if err != nil {
 			return "", "", trace.Wrap(err)
--- a/lib/expand/bootstrap.go
+++ b/lib/expand/bootstrap.go
@@ -19,15 +19,31 @@ package expand
 import (
 	"github.com/gravitational/gravity/lib/install"
 	"github.com/gravitational/gravity/lib/ops"
+	pb "github.com/gravitational/gravity/lib/rpc/proto"
+	rpcserver "github.com/gravitational/gravity/lib/rpc/server"
 	"github.com/gravitational/gravity/lib/state"
 	"github.com/gravitational/gravity/lib/storage"
 	"github.com/gravitational/gravity/lib/system/environ"
+	"github.com/gravitational/gravity/lib/systeminfo"
+	log "github.com/sirupsen/logrus"
 
 	"github.com/gravitational/trace"
 )
 
 // init initializes the peer after a successful connect
-func (p *Peer) init(ctx operationContext) error {
+func (p *Peer) init(ctx operationContext) (*storage.Server, error) {
+	err := p.initEnviron(ctx)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	server, err := p.startAgent(ctx)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	return server, nil
+}
+
+func (p *Peer) initEnviron(ctx operationContext) error {
 	if err := p.clearLogins(); err != nil {
 		return trace.Wrap(err)
 	}
@@ -38,9 +54,6 @@ func (p *Peer) init(ctx operationContext) error {
 		return trace.Wrap(err)
 	}
 	if err := p.ensureServiceUserAndBinary(ctx); err != nil {
-		return trace.Wrap(err)
-	}
-	if err := p.startAgent(ctx); err != nil {
 		return trace.Wrap(err)
 	}
 	return nil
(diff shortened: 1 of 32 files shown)


lib/expand/bootstrap.go
```diff
@@ -17,17 +17,33 @@ limitations under the License.
...
 )
 
 // init initializes the peer after a successful connect
 func (p *Peer) init(ctx operationContext) error {
+	err := p.initEnviron(ctx)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if !ctx.hasOperation() {
+		return nil
```
a-palchikov: this is a special case for expand when we don't have the operation after connecting to the cluster (CLI-workflow when the operation is created by the agent).

lib/expand/join.go
```diff
@@ -158,36 +162,24 @@ func (p *Peer) Execute(req *installpb.ExecuteRequest, stream installpb.Agent_Exe
...
+					return trace.Unwrap(result.Err)
+				}
+				// Flag a failed intermediate step as codes.Aborted.
+				// From gRPC documentation:
+				// Use Aborted if the client should retry at a higher-level
+				// (e.g., restarting a read-modify-write sequence).
+				// See https://github.com/grpc/grpc-go/blob/v1.22.0/codes/codes.go#L78
+				return status.Error(codes.Aborted, trace.UserMessage(result.Err))
+			}
+			if result.CompletionEvent != nil {
```
a-palchikov: I updated expand side of the service to match that of installer w.r.t handling of completion status.

lib/expand/join.go
```diff
@@ -666,19 +665,8 @@ func (p *Peer) checkAndSetServerProfile(app ops.Application) error {
 		"specified node role %q is not defined in the application manifest", p.Role))
 }
 
-// runLocalChecks makes sure node satisfies system requirements
-func (p *Peer) runLocalChecks(cluster ops.Site, installOperation ops.SiteOperation) error {
```
a-palchikov: Removing local checks as it's difficult to come up with a heuristic when it's safe to run them since expand operation state is not updated beyond `OperationStateExpandPrechecks` and they have become redundant after Roman has introduced cluster<->joining node checks phase.

lib/expand/phases/wait.go
```diff
@@ -60,7 +61,10 @@ type waitPlanetExecutor struct {
 func (p *waitPlanetExecutor) Execute(ctx context.Context) error {
 	p.Progress.NextStep("Waiting for the planet to start")
 	p.Info("Waiting for the planet to start.")
-	err := utils.Retry(defaults.RetryInterval, defaults.RetryAttempts,
+	ctx, cancel := defaults.WithTimeout(ctx)
+	defer cancel()
+	b := backoff.NewConstantBackOff(defaults.RetryInterval)
+	err := utils.RetryWithInterval(ctx, b,
```
a-palchikov: This fixes the issue with being unable to interrupt this phase otherwise (i.e. cancelling the context did not have an effect).

lib/ops/opsservice/agents.go
```diff
@@ -457,10 +457,6 @@ func (r *AgentPeerStore) checkHostname(ctx context.Context, group *agentGroup, a
 		return trace.AccessDenied("One of existing servers already has hostname %q: %q.", hostname, existingServers)
 	}
 
-	if group.hasPeer(addr, hostname) {
```
a-palchikov: I've removed this test to account for the case when the joining agent fails with an error loosing the connection to the cluster. It should be able to reconnect (rejoin) at the expense of the unlikelier case of having two agents starting from the same node. Let me know if you still think this is an omission.

lib/system/environ/uninstall.go
```diff
@@ -41,6 +41,13 @@ func UninstallSystem(printer utils.Printer, logger log.FieldLogger) (err error)
 	if err := unmountDevicemapper(printer, logger); err != nil {
 		errors = append(errors, err)
 	}
+	svm, err := systemservice.New()
```
a-palchikov: This was a bug that left package (planet/teleport) services uninstalled.

lib/expand/join.go
```diff
@@ -436,16 +452,16 @@ func (p *Peer) submit(req *installpb.ExecuteRequest) bool {
...
+func (p *Peer) execute(req *installpb.ExecuteRequest) (dispatcher.Status, error) {
 	p.WithField("req", req).Info("Execute.")
 	opCtx, err := p.operationContext()
 	if err != nil {
-		return trace.Wrap(err)
+		return dispatcher.StatusUnknown, trace.Wrap(err)
 	}
 	if req.Phase != nil {
-		return trace.Wrap(p.executePhase(p.ctx, *opCtx, *req.Phase, p.dispatcher))
+		return p.executePhase(p.ctx, *opCtx, *req.Phase, p.dispatcher)
```
r0mant: I wanted to ask you about this before: how come this logic of executing phases leaked into Peer from FSM engine? It is weird to me that now Peer also has "execute", "executePhase" and "executeConcurrently" methods.

lib/expand/join.go
```diff
@@ -512,45 +538,44 @@ func (p *Peer) printStep(format string, args ...interface{}) {
...
-// createExpandOperation creates a new expand operation
-func (p *Peer) createExpandOperation(operator ops.Operator, cluster ops.Site) (*ops.SiteOperation, error) {
-	key, err := operator.CreateSiteExpandOperation(p.ctx, ops.CreateSiteExpandOperationRequest{
-		AccountID:   cluster.AccountID,
-		SiteDomain:  cluster.Domain,
-		Provisioner: schema.ProvisionerOnPrem,
-		Servers:     map[string]int{p.Role: 1},
-	})
+// dialWizard connects to a wizard
+func (p *Peer) dialWizard(addr string) (*operationContext, error) {
```
r0mant: This methods seems to just be wrapping `connectWizard` as-is, can we get rid of it and just use `connectWizard` (or rename `connectWizard` to `dialWizard` and use the "dial" one).

lib/expand/join.go
```diff
@@ -512,45 +538,44 @@ func (p *Peer) printStep(format string, args ...interface{}) {
...
-
-// getExpandOperation returns existing expand operation created via UI
-func (p *Peer) getExpandOperation(operator ops.Operator, cluster ops.Site, operationID string) (*ops.SiteOperation, error) {
-	operation, err := operator.GetSiteOperation(ops.SiteOperationKey{
-		AccountID:   cluster.AccountID,
-		SiteDomain:  cluster.Domain,
-		OperationID: operationID,
-	})
-	if err != nil {
+	var operation *ops.SiteOperation
```
r0mant: I think that all this logic of retrieving, expand operation either by ID or from the join backend would be better off in a separate method. I know there's already `getExpandOperation`, maybe call the new one `fetchExpandOperation`.

lib/expand/join.go
r0mant: Hmm, they were actually here for a purpose. So when the user launches gravity join (in any case, expand or wizard/cli install) and there are any local problems, it exits immediately and prints failures in the console. This was a UX improvement requested by Ev.

lib/expand/join.go
```diff
@@ -744,147 +733,186 @@ func (p *Peer) stop() error {
...
+		return p.runAgent(ctx)
+	}
+	err := p.executeExpandOperation(ctx)
+	if err != nil {
+		return dispatcher.StatusUnknown, trace.Wrap(err)
+	}
+	return dispatcher.StatusCompleted, nil
+}
+
+func (p *Peer) runAgent(ctx operationContext) (dispatcher.Status, error) {
```
r0mant: Nit: I may be misunderstanding something, but looking at what this function does, it's not related to the "agent" in any way so the name is misleading.

lib/install/utils.go
```diff
@@ -360,6 +367,16 @@ type ProgressPoller struct {
...
 }
 
+// ExecResult describes the result of execution an operation (step).
+// An optional completion event can describe the completion outcome to the client
+type ExecResult struct {
+	// CompletionEvent specifies the optional completion
+	// event to send to a client
+	CompletionEvent *dispatcher.Event
+	// Err specifies the optional execution error
+	Err error
```
r0mant: Nit: Since this is a struct field, `Error` (do not shorten)?

lib/install/utils.go
```diff
@@ -436,3 +453,8 @@ func initOperationPlan(operator ops.Operator, planner engine.Planner) error {
 	}
 	return nil
 }
+
+func newInvalidCertError(err error) error {
+	return trace.BadParameter("%s. Please make sure that clocks are synchronized between the nodes.",
```
r0mant: Nit: `... by using ntp, chrony or other time-synchronization programs`.

lib/ops/opsservice/agents.go
r0mant: IIRC it wasn't for the case of starting two agents from the same node (which I agree is unlikely) but rather to catch the case when two different nodes have the same hostname (which is very likely and we've run into this in the past).

lib/rpc/tls.go
```diff
@@ -155,6 +156,11 @@ func CredentialsFromPackage(
 	if err != nil {
 		return nil, nil, trace.Wrap(err)
 	}
+	now := time.Now()
+	err = validateCerts(tlsArchive, now)
```
r0mant: Hm, I'm not entirely sure about this - the purpose of this function is just to load the certs from the specified package, valid or not, so adding validation here may be mixing concerns. Or is it hard to add validation for what this method returns (TransportCredentials)?

Also, nit: not sure why a separate variable for "now" is needed.

lib/rpc/tls.go
```diff
@@ -346,3 +358,47 @@ func upsertPackage(packages pack.PackageService, pkg loc.Locator, archive utils.
...
+	}
+	caKeyPair := archive[pb.CA]
+	if err := validateCert(caKeyPair.CertPEM, timestamp); err != nil {
+		return trace.Wrap(err, "invalid CA certificate (%v)",
+			trace.UserMessage(err))
+	}
+	return nil
+}
+
+func validateCert(pemBytes []byte, timestamp time.Time) error {
```
r0mant: Nit: `timestamp` -> `now`?

lib/rpc/tls.go
```diff
@@ -346,3 +358,47 @@ func upsertPackage(packages pack.PackageService, pkg loc.Locator, archive utils.
...
+	cert, err := tlsca.ParseCertificatePEM(pemBytes)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if timestamp.Before(cert.NotBefore) {
+		return newCertError("certificate is valid in the future")
```
r0mant: I would include `now` and `cert.NotBefore` in the error for easier troublshooting. Same for the error below.

lib/rpc/tls.go
```diff
@@ -346,3 +358,47 @@ func upsertPackage(packages pack.PackageService, pkg loc.Locator, archive utils.
...
+	}
+	return nil
+}
+
+func validateCert(pemBytes []byte, timestamp time.Time) error {
+	cert, err := tlsca.ParseCertificatePEM(pemBytes)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if timestamp.Before(cert.NotBefore) {
```
r0mant: I would add a couple of seconds of leeway to the certificate start date, both ways. IIRC we had some sort of race conditions in the past attempting to use the issued certificate right away.

lib/expand/join.go
a-palchikov: Ok, I'll try to come up with a heuristic to only run them at early operation stage.

lib/ops/opsservice/agents.go
a-palchikov: Oh - my bad. I rechecked the condition and you're right. Will revert.

lib/expand/join.go
a-palchikov: No logic has leaked but there might be some naming clashes.

lib/expand/join.go
a-palchikov: Yup - it used to be more but degenerated to `connectWizard` which I'll merge back as `dialWizard`.

lib/expand/join.go
a-palchikov: It implements the loop main loop of the joining agent during install - hence the `agent` in the name.

lib/utils/error.go
```diff
@@ -437,5 +436,6 @@ func isLicenseError(message string) bool {
 }
 
 func isHostAlreadyRegisteredError(message string) bool {
-	return strings.Contains(message, "One of existing peers already has hostname")
+	return strings.Contains(message, "One of existing peers already has hostname") ||
```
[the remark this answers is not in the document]
a-palchikov: Note sure I follow the bit about compile-time verification?

PR #545, opened 2019-07-26
Properly remove node from the cluster if expand is aborted

a-palchikov:
* Have 'plan display' and 'plan complete' use cluster plan as a fallback.
 * Invoke shrink operation after abort to properly remove the partially added node from the cluster.

Fixes http://github.com/gravitational/gravity.e/issues/4184.


lib/expand/join.go
```diff
@@ -953,14 +957,44 @@ func (p *Peer) waitForOperation(operator ops.Operator, operation ops.SiteOperati
...
 			return nil
 		case <-p.ctx.Done():
 			return trace.Wrap(p.ctx.Err())
 		}
 	}
 }
 
+func (p *Peer) leave() error {
+	p.Info("Leave cluster.")
+	ctx, cancel := context.WithTimeout(context.Background(), defaults.NodeLeaveTimeout)
```
r0mant: Pass context (instead of the background one) to `leave()`?

lib/expand/join.go
```diff
@@ -953,14 +957,44 @@ func (p *Peer) waitForOperation(operator ops.Operator, operation ops.SiteOperati
...
+		return trace.Wrap(err)
+	}
+	_, err = opCtx.Operator.CreateSiteShrinkOperation(ctx,
+		ops.CreateSiteShrinkOperationRequest{
+			AccountID:  opCtx.Cluster.AccountID,
+			SiteDomain: opCtx.Cluster.Domain,
+			Servers:    []string{operation.Servers[0].Hostname},
+			Force:      true,
+			// Have cluster avoid triggering remote node cleanup
+			NodeRemoved: true,
```
r0mant: Depending on where exactly join was interrupted, the new node may actually be already online and thus this flag will lead to new serf member staying in the cluster. I remember this b/c we fixed a similar bug with AWS autoscaling a few months ago.

Maybe omit this flag and let shrink follow the normal procedure, i.e. it should figure out itself whether to launch agent or not?

tool/gravity/cli/operation.go
```diff
@@ -163,18 +163,37 @@ func completeOperationPlan(localEnv *localenv.LocalEnvironment, environ LocalEnv
...
-		return completeEnvironPlan(localEnv, environ, *op)
+		err = completeEnvironPlan(localEnv, environ, *op)
 	case ops.OperationUpdateConfig:
-		return completeConfigPlan(localEnv, environ, *op)
+		err = completeConfigPlan(localEnv, environ, *op)
 	default:
 		return trace.BadParameter("operation type %q does not support plan completion", op.Type)
 	}
+	if trace.IsNotFound(err) {
+		return completeClusterOperationPlan(localEnv, *op)
```
r0mant: Does this address part 2 of https://github.com/gravitational/gravity.e/issues/3215 by chance? I.e. can we "complete" arbitrary operation using a CLI command now?

lib/expand/join.go
a-palchikov: The purpose of leave is to encapsulate this - there's no context to pass from outside.

lib/expand/join.go
a-palchikov: I cannot do this without introducing another flag to avoid launching the remote agent (which I understand this flag is already about).
As for your concern - it will not leave the node (i.e. no serf member) as aborting cleans up the services on the node and all of the state (hence the reason why the agent part of the shrink is redundant as it basically does the same).

tool/gravity/cli/operation.go
a-palchikov: Yes.

lib/expand/join.go
r0mant: Okay, that makes sense.

PR #562, opened 2019-08-02
Implement fio-based disk performance check for etcd

r0mant:
This PR implements an additional preflight check that uses `fio` to make sure that the disk used for etcd data meets etcd performance requirements - the source of many issues in customers' clusters. The check verifies sequential write IOPS and fsync latency and the test is constructed specifically for etcd use-case.

Fio binary is bundled with the cluster images now. I have introduced a new `/init` phase for install/expand that pulls it on the nodes before preflight checks run.

Closes https://github.com/gravitational/gravity.e/issues/3927.


lib/checks/disks/result.go
```diff
@@ -0,0 +1,72 @@
...
+	return j.Sync.Latency.Percentile[bucket99] / 1000000
+}
+
+// FioReadResult contains reads-related metrics.
+type FioReadResult struct {
+	// IOPS is the number of read iops.
+	IOPS float64 `json:"iops"`
+}
+
+// FioWriteResult contains writes-related metrics.
```
a-palchikov: write~~s~~-related
Also, for me it would be FioIOResult - irrelevant whether for read/write operation. Then you'd only need a single struct.

lib/network/validation/proto/validation.proto
```diff
@@ -12,11 +12,26 @@ service Validation {
...
+// CheckDisksRequest describes a disks performance test request.
+message CheckDisksRequest {
+    // Spec is a fio job spec describing the test to perform.
+    bytes spec = 1;
+}
+
+// CheckDisksResponse is the result of the disk performance test.
+message CheckDisksResponse {
+    // Result is the fio job result output in json format.
+    bytes result = 1;
```
a-palchikov: Why not define it in protobuf format?

a-palchikov:
Based on robotest failures you should consider extending upgrade's `/init` with the same set of steps to create/pull `fio` binary for checks.

lib/network/validation/proto/validation.proto
r0mant: Yeah, I was actually considering that but just wasn't sure how the output changes based on the fio configuration and whether it's possible to describe it with a single struct. But I think it would be better, yes, so I'll do that for our use-case.

build.assets/Makefile
```diff
@@ -520,6 +522,64 @@ $(BANDWAGON_OUT):
...
+
+.PHONY: pull-fio-sub
+pull-fio-sub:
+	-$(AWS) s3 cp $(S3_OPTS) \
+	    $(BUILD_BUCKET_URL)/fio/$(FIO_VER)/fio \
+	    $(FIO_OUT)
+
+.PHONY: build-fio
+build-fio:
+	if [ ! -e "$(FIO_OUT)" ]; then \
```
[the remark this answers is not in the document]
r0mant: It is needed to make sure "build" does not trigger on the first run after download. I tried to achieve the same using make's dependencies but this way turned out to be the cleanest.

build.assets/Makefile
```diff
@@ -520,6 +522,64 @@ $(BANDWAGON_OUT):
...
+
+$(FIO_OUT):
+	mkdir -p $(FIO_BUILDDIR) && \
+	    $(MAKE) pull-fio && \
+	    $(MAKE) build-fio
+
+.PHONY: pull-fio
+pull-fio:
+	if [ ! -e "$(FIO_OUT)" ]; then \
+	    $(MAKE) pull-fio-sub; \
```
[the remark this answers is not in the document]
r0mant: It's not used anywhere else atm, I just wanted to keep it similar to "build" targets below, and the same pattern is used somewhere above too.

lib/checks/disks.go
```diff
@@ -0,0 +1,119 @@
...
+			iops, EtcdMinWriteIOPS))
+	}
+	if latency > EtcdMaxFsyncLatencyMs {
+		errors = append(errors, fmt.Sprintf("  * High fsync latency of %vms, required maximum is %vms.",
+			latency, EtcdMaxFsyncLatencyMs))
+	}
+	return fmt.Sprintf(
+		`It looks like on node %v etcd data resides on a disk (%v) that does not satisfy recommended performance requirements:
+%v
+For optimal performance, please make sure that the directory with etcd data resides on the device that meets the aforementioned
```
knisbet: So the way this error is worded, to me sounds like a warning. However, looking at the code, to me this looks like it will return an error.

I would consider simplifying the message / making it consistent with other checks, such as https://github.com/gravitational/gravity/blob/23989ef97ed7fcc65d0aa560b4c008e957c70b87/lib/checks/checks.go#L455-L458

For example:
`server %q disk IOPS on %q/s is lower than required %v", server.Hostname, filepath.Dir(testPath), strings.Join(errors, "\n")` And then if it formats ok return an aggregate error with the multiple failures for IOPS and latency.

lib/install/phases/init.go
```diff
@@ -0,0 +1,108 @@
...
+	locator, err := p.Application.Manifest.Dependencies.ByName(constants.FioPackage)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	_, reader, err := p.Packages.ReadPackage(*locator)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	defer reader.Close()
+	path := filepath.Join(os.TempDir(), constants.FioBin)
```
knisbet: Does the packaging system when writing files maintain the build dates or the time they're written to disk? Extracting to the configured TempDir can be a bit dangerous, because if a file is written from a tar that maintains the data it was packaged, the tmpfs cleaned can run and delete the file. As I understand it the window for this should be pretty small, but I have seen it happen where customers extract gravity to /tmp and then files are missing by the time the install is running. 

To get this far in the install, isn't /var/lib/gravity initialized by this point, and fio can be placed in a predictable path within /var/lib/gravity?

a-palchikov: I think it's a good point.

r0mant:
@a-palchikov @knisbet I have addressed your comments. LGTY?

lib/network/validation/proto/proto.go
```diff
@@ -55,14 +58,71 @@ func (r CheckBandwidthRequest) Check() error {
...
+}
+
+// GetWriteIOPS returns number of write iops.
+func (r FioJobResult) GetWriteIOPS() float64 {
+	return r.Write.Iops
+}
+
+// GetFsyncLatency returns 99th percentile of fsync latency in milliseconds.
+func (r FioJobResult) GetFsyncLatency() int64 {
+	return r.Sync.Latency.Percentile[bucket99] / 1000000
```
a-palchikov: nit: do you need to validate the percentile map before using the `bucket99` slot?

lib/network/validation/disk.go
```diff
@@ -0,0 +1,78 @@
...
+	"encoding/json"
+	"os"
+	"path/filepath"
+
+	"github.com/gravitational/gravity/lib/defaults"
+	"github.com/gravitational/gravity/lib/network/validation/proto"
+	"github.com/gravitational/gravity/lib/utils"
+
+	"github.com/gravitational/trace"
+	"golang.org/x/net/context"
```
a-palchikov: this should not be needed anymore?..
But there's more than a handful places where this is used - maybe a task on its own.

r0mant: Yeah, I saw it is also used in a few more places and decided not to go down that road for now.

lib/network/validation/proto/proto.go
r0mant: Fio always populates this bucket unless you override the percentiles list via a parameter. Probably a better way would be to pass bucket to this method but it should be fine for now.

Issue #536, opened 2019-07-23
[BUG] Prevent upgrades that skip versions


knisbet:
**Describe the bug**
Gravity doesn't seem to currently prevent upgrades that skip too many versions. This should be enforced in the upgrade scripts and/or tele build to prevent the cluster from breaking when using unsupported upgrade paths.

a-palchikov:
As we already have [a basic way](https://github.com/gravitational/gravity/blob/6.1.1/tool/gravity/cli/clusterupdate.go#L355) of enforcing the installed version minimum we could follow the same version skew rules as kubernetes (using the version of k8s bundled by the runtime application).
But after the work on allowing multiple intermediate kubernetes upgrades, it will probably matter less so we could stay with a single version minimum known to support them.

knisbet:
I suspect the really tricky part though, is the intermediate upgrade stages requires tele build flags, so installation restrictions should react to that. IE an upgrade for 5.0 -> 5.5 that doesn't specify 5.2 as an intermediate stage should prevent the upgrade, where as one that does have the intermediate stage to do 5.0 -> 5.2 -> 5.5 should complete the upgrade.

Also, the way the basic check is setup now leads to the variable not keeping up to date. So process wise or test wise, we should likely ensure that the minimum version is set to a version that we actually test upgrades from in robotest.

a-palchikov:
I agree. One way we could expose the minimum version if in the `gravity version` command. There could be some extended view with additional metadata.

wadells:
This work is being tracked in #1064 now.

PR #570, opened 2019-08-06
Dmitri/5.5.x/4146 tele build

a-palchikov:
This PR implements support for `tele build --upgrade-via=<runtime-version>` which enables to embed another runtime version inside the resulting installer.
The package dependencies of the specified runtime package are marked with a `intermediate-upgrade:<runtime-version>` label to be able to filter on the runtime version when building an operation plan for upgrade.

Since the changes are relatively stand-alone, I'd like to commit them before working on the upgrade part.

Requires https://github.com/gravitational/gravity.e/pull/4195.

### DIFF
--- a/build.assets/Makefile
+++ b/build.assets/Makefile
@@ -155,7 +155,7 @@ endef
 .PHONY: build
 build:
 	@if [ "$(OS)" = "darwin" ]; then $(MAKE) build-on-host; else $(MAKE) build-in-container; fi
-	ln -sf $(GRAVITY_BUILDDIR) $(GRAVITY_CURRENT_BUILDDIR)
+	ln --symbolic --force --no-target-directory $(GRAVITY_BUILDDIR) $(GRAVITY_CURRENT_BUILDDIR)
 
 #
 # generate gRPC code
--- a/lib/app/dependency.go
+++ b/lib/app/dependency.go
@@ -78,14 +78,16 @@ func (r Dependencies) AsPackages() (result []loc.Locator) {
 	return result
 }
 
-// UniqPackages returns packages without duplicates
+// UniqPackages returns packages without duplicates.
+// packages is sorted in-place as a result of this operation
 func UniqPackages(packages []pack.PackageEnvelope) []pack.PackageEnvelope {
 	sort.Sort(packagesByLocator(packages))
 	n := set.Uniq(packagesByLocator(packages))
 	return packages[:n]
 }
 
-// UniqApps returns apps without duplicates
+// UniqApps returns apps without duplicates.
+// apps is sorted in-place as a result of this operation
 func UniqApps(apps []Application) []Application {
 	sort.Sort(appsByLocator(apps))
 	n := set.Uniq(appsByLocator(apps))
--- a/lib/app/dependency.go
+++ b/lib/app/dependency.go
@@ -25,82 +25,99 @@ import (
 )
 
 // GetDependencies transitively collects dependencies for the specified application package
-func GetDependencies(app *Application, apps Applications) (result *Dependencies, err error) {
-	state := &state{
-		visitedPackages: map[string]struct{}{},
-		visitedApps:     map[string]struct{}{},
-	}
-	if err = getDependencies(app, apps, state); err != nil {
+func GetDependencies(app Application, apps Applications) (result *Dependencies, err error) {
+	dependencies, err := GetFullDependencies(GetDependenciesRequest{
+		App:  app,
+		Apps: apps,
+	})
+	if err != nil {
 		return nil, trace.Wrap(err)
 	}
-	result = &Dependencies{}
-	result.Packages = append(result.Packages, state.packages...)
-	if state.runtimePackage != nil {
-		result.Packages = append(result.Packages, *state.runtimePackage)
-	}
-	for _, locator := range state.apps {
-		if !locator.IsEqualTo(app.Package) {
-			result.Apps = append(result.Apps, locator)
-		}
-	}
-	result.Packages = loc.Deduplicate(result.Packages)
-	result.Apps = loc.Deduplicate(result.Apps)
-	return result, nil
+	return dependencies, nil
 }
 
 // VerifyDependencies verifies that all dependencies for the specified application are available
 // in the provided package service.
-func VerifyDependencies(app *Application, apps Applications, packages pack.PackageService) error {
-	dependencies, err := GetDependencies(app, apps)
-	if err != nil {
-		return trace.Wrap(err)
+func VerifyDependencies(app Application, apps Applications, packages pack.PackageService) error {
+	_, err := GetFullDependencies(GetDependenciesRequest{
+		App:  app,
+		Apps: apps,
+		Pack: packages,
+	})
+	return trace.Wrap(err)
+}
+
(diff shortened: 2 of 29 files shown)


build.assets/Makefile
```diff
@@ -155,7 +155,7 @@ endef
 .PHONY: build
 build:
 	@if [ "$(OS)" = "darwin" ]; then $(MAKE) build-on-host; else $(MAKE) build-in-container; fi
-	ln -sf $(GRAVITY_BUILDDIR) $(GRAVITY_CURRENT_BUILDDIR)
+	ln --symbolic --force --no-target-directory $(GRAVITY_BUILDDIR) $(GRAVITY_CURRENT_BUILDDIR)
```
a-palchikov: This fixes an issue with infinite recursion in the symlink.

lib/app/docker/distribution.go
```diff
@@ -120,7 +120,10 @@ func alive(path string, handler http.Handler) http.Handler {
 			return
 		}
 
-		handler.ServeHTTP(w, r)
+		ctx, cancel := defaultContext()
```
a-palchikov: This fixes an issue with distribution always logging at the default (`Debug`) level in HTTP handlers.

lib/builder/builder.go
```diff
@@ -128,6 +130,7 @@ func (c *Config) CheckAndSetDefaults() error {
 		c.FieldLogger = logrus.WithField(trace.Component, "builder")
 	}
 	if c.Progress == nil {
+		// FIXME: 6
```
a-palchikov: Currently, progress reporting output is broken w.r.t to number of build steps which I'm no longer sure how many atm (and I should not care really - this needs fixing).

lib/install/install.go
```diff
@@ -573,6 +573,7 @@ func generateInstallToken(service ops.Operator, accountID, installToken string)
 func GetApp(service appservice.Applications) (*appservice.Application, error) {
 	apps, err := service.ListApps(appservice.ListAppsRequest{
 		Repository: defaults.SystemAccountOrg,
+		Type:       storage.AppUser,
```
a-palchikov: This API was always about user applications.

lib/app/dependency.go
```diff
@@ -126,13 +139,35 @@ func getDependencies(app *Application, apps Applications, state *state) error {
 	if runtimePackage, _ := app.Manifest.DefaultRuntimePackage(); runtimePackage != nil {
 		state.runtimePackage = runtimePackage
 	}
-
 	return nil
 }
 
+func (r *state) isPackage(pkg loc.Locator) bool {
```
r0mant: Nit: I think `hasPackage` would be a bit more appropriate name for this method. Same for `hasApp` below.

lib/app/service/pull.go
```diff
@@ -407,9 +435,12 @@ type pullState struct {
...
 }
 
-// IsMetadataPackage determines if the specified package is a metadata package.
-// A metadata package describes a remote package and deserves
-// special handling in certain cases.
-func IsMetadataPackage(envelope pack.PackageEnvelope) bool {
-	return envelope.RuntimeLabels[pack.PurposeLabel] == pack.PurposeMetadata
+// copyRuntimeLabels copies labels from runtimeLabels to labels for each
+// label non-existent in labels
+func copyRuntimeLabels(labels map[string]string, runtimeLabels map[string]string) {
```
r0mant: Nit: `(labels, runtimeLabels map[string]string)`. Also, maybe `srcLabels` and `dstLabels` would be a bit more suitable names?

lib/builder/builder.go
```diff
@@ -460,8 +497,36 @@ There are a few ways to resolve the issue:
 	return nil
 }
 
+func (b *Builder) collectUpgradeDependencies() (result *libapp.Dependencies, err error) {
```
r0mant: Could you add a comment about the purpose of this function?

lib/builder/builder.go
```diff
@@ -512,3 +559,79 @@ type GetRepositoryFunc func(*Builder) (string, error)
...
+}
+
+func filterUpgradePackageDependencies(packages []pack.PackageEnvelope) (result []pack.PackageEnvelope) {
+	result = packages[:0]
+	for _, pkg := range packages {
+		if pkg.Locator.Repository != defaults.SystemAccountOrg {
+			continue
+		}
+		switch {
+		case pkg.Locator.Name == constants.TeleportPackage,
```
r0mant: Why not do a switch directly on `pkg.Locator.Name`?

lib/builder/builder.go
```diff
@@ -460,8 +497,36 @@ There are a few ways to resolve the issue:
...
+			},
+			Apps: apps,
+			Pack: b.Env.Packages,
+		}
+		dependencies, err := libapp.GetDependencies(req)
+		if err != nil {
+			return nil, trace.Wrap(err)
+		}
+		addUpgradeVersionLabel(dependencies, version.String())
+		result.Packages = append(result.Packages, filterUpgradePackageDependencies(dependencies.Packages)...)
```
r0mant: Should this do some sort of deduplication, or it'll be taken care of later?

lib/defaults/defaults.go
```diff
@@ -1177,7 +1177,11 @@ var BaseTaintsVersion = semver.Must(semver.NewVersion("4.36.0"))
 
 // BaseUpdateVersion sets the minimum version that this binary
 // can update
-var BaseUpdateVersion = semver.Must(semver.NewVersion("3.51.0"))
+var BaseUpdateVersion = semver.Must(semver.NewVersion("5.0.0"))
```
knisbet: Question: I opened this issue a couple weeks ago (https://github.com/gravitational/gravity/issues/536) and it looks like it's already covered by this variable. The question is, how procedurally do we ensure this variable is kept up to date? 

Just thinking out loud, can we expose it in a way that the robotest fails if it's test matrix doesn't test the BaseUpdateVersion? Or some other way to ensure that it's flagged as needing an update?

lib/app/dependency.go
```diff
@@ -17,106 +17,136 @@ limitations under the License.
...
+	for _, app := range r.Apps {
+		result = append(result, app.Package)
 	}
-	return nil
+	return result
+}
+
+// UniqPackages returns packages without duplicates
+func UniqPackages(packages []pack.PackageEnvelope) []pack.PackageEnvelope {
+	sort.Sort(packagesByLocator(packages))
```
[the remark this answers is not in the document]
a-palchikov: My uses are such that mutating the slices in-place is fine but I'll leave a note in the comment.
And if a future user is unhappy with such behavior, the API can receive a copy instead (otherwise it would be making a decision for the caller it might not like).

lib/defaults/defaults.go
a-palchikov: Currently this variable is a baseline of sorts - it does not identify a specific version so even if exposed it would not be tested against. We could instead use a specific version so this became testable as you described.
And there're no provisions to keep it automatically up-to-date as afaic there's no good condition to know when to switch to the new base version other than maybe the version of kubernetes that the bundled runtime application sports.
We should continue discussing this in the issue you [created](https://github.com/gravitational/gravity/issues/536) though.

PR #574, opened 2019-08-09
[5.5.x] 'upload update' changes for direct upgrades from 5.0

a-palchikov:
This PR updated `upload update` to take intermediate upgrade package dependencies into account and upload/sync the proper set of packages - this another step towards enabling direct upgrades from 5.0.

I took the liberty of refactoring app pull/sync magic to make it simpler as it was daunting when I first tried extending it to suit my needs. I also moved those functionalities directly under `lib/app` where I believe they're more appropriate (although I kept pull tests under `lib/app/service` to keep being able to set up test environment.

Requires https://github.com/gravitational/gravity.e/pull/4199.


lib/localenv/tarballenv.go
```diff
@@ -0,0 +1,80 @@
+package localenv
```
r0mant: Missing license header.

lib/localenv/tarballenv.go
```diff
@@ -0,0 +1,80 @@
...
+	libapp "github.com/gravitational/gravity/lib/app"
+	"github.com/gravitational/gravity/lib/pack"
+	"github.com/gravitational/gravity/lib/pack/encryptedpack"
+	"github.com/gravitational/gravity/lib/utils"
+
+	"github.com/gravitational/license"
+	"github.com/gravitational/trace"
+)
+
+// NewTarballEnvironment creates new environment with specified configuration
```
r0mant: I would say something along the lines of "creates a new local environment for the specified unpacked cluster image tarball".

lib/app/service/installer.go
```diff
@@ -455,5 +455,5 @@ var uploadScriptTemplate = template.Must(template.New("uploadScript").Parse(`#!/
 #
 #    http://www.apache.org/licenses/LICENSE-2.0
 #
-./gravity --insecure update upload --state-dir=.
+./gravity --insecure update upload
```
r0mant: Will this work if the upgrade/upload script invoked from some directory other than unpacked tarball? E.g.
```
$ cd ~
$ sudo ~/installer/upload
```

lib/pack/utils.go
```diff
@@ -722,6 +722,13 @@ func IsPlanetConfigPackage(loc loc.Locator, labels map[string]string) bool {
 		loc.Repository != defaults.SystemAccountOrg
 }
 
+// IsMetadataPackage determines if the specified package is a metadata package.
+// A metadata package describes a remote package and deserves
+// special handling in certain cases.
```
r0mant: Nit: TBH "special handling in certain cases" doesn't tell the reader much :smile: I'd either try to be more specific or remove this phrase altogether.

tool/gravity/cli/ops.go
```diff
@@ -87,110 +88,155 @@ func appPackage(env *localenv.LocalEnvironment) error {
 }
 
 func uploadUpdate(env *localenv.LocalEnvironment, opsURL string) error {
-	// create local environment with gravity state dir because the environment
-	// provided above has upgrade tarball as a state dir
-	localStateDir, err := localenv.LocalGravityDir()
+	clusterOperator, err := env.SiteOperator()
```
r0mant: If `env` has the tarball as its state dir (as the comment that you removed above said), then this will attempt to look for "ops" credentials in the tarball state (and won't find them), no?

tool/gravity/cli/ops.go
```diff
@@ -87,110 +88,155 @@ func appPackage(env *localenv.LocalEnvironment) error {
...
 
-	defaultEnv, err := localenv.New(localStateDir)
+	cluster, err := clusterOperator.GetLocalSite()
 	if err != nil {
 		return trace.Wrap(err)
 	}
 
-	clusterOperator, err := defaultEnv.SiteOperator()
+	tarballEnv, err := localenv.NewTarballEnvironment(localenv.TarballEnvironmentArgs{
+		License: cluster.License.Raw,
```
r0mant: Why is "state-dir" that comes from the command's CLI flag not passed in here in the config? That flag is supposed to indicate the state dir of the unpack tarball so it would make sense to propagate it here, no?

tool/gravity/cli/ops.go
```diff
@@ -87,110 +88,155 @@ func appPackage(env *localenv.LocalEnvironment) error {
...
-	tarballApps, err := env.AppServiceLocal(localenv.AppConfig{
-		Packages: tarballPackages,
-	})
+	deps, err := getUploadDependencies(tarballEnv, *appPackage, *installedRuntimeVersion)
 	if err != nil {
 		return trace.Wrap(err)
 	}
 
-	appPackage, err := install.GetAppPackage(tarballApps)
+	env.PrintStep("Importing application %v v%v", appPackage.Name, appPackage.Version)
```
r0mant: "Importing cluster image..." Looks like we've missed this one when renaming applications to images.

tool/gravity/cli/ops.go
```diff
@@ -87,110 +88,155 @@ func appPackage(env *localenv.LocalEnvironment) error {
...
 	}
 
-	env.PrintStep("Importing application %v v%v", appPackage.Name, appPackage.Version)
-	_, err = appservice.PullApp(appservice.AppPullRequest{
-		SrcPack: tarballPackages,
-		SrcApp:  tarballApps,
-		DstPack: clusterPackages,
-		DstApp:  clusterApps,
-		Package: *appPackage,
+	env.PrintStep("Application has been uploaded")
```
r0mant: "Cluster image has been uploaded"

tool/gravity/cli/ops.go
```diff
@@ -201,21 +247,25 @@ func uploadUpdate(env *localenv.LocalEnvironment, opsURL string) error {
...
 	}
-
-	env.PrintStep("Application has been uploaded")
 	return nil
 }
 
+func canUpload(cluster ops.Site, app loc.Locator) error {
+	if cluster.State == ops.SiteStateDegraded {
+		return trace.BadParameter("The cluster is in degraded state so " +
+			"uploading new applications is prohibited. Please check " +
```
r0mant: "new cluster images"

tool/gravity/cli/system.go
```diff
@@ -367,14 +368,13 @@ func applyUpdates(env *localenv.LocalEnvironment, updates []storage.PackageUpdat
...
-	pullReq := appservice.PackagePullRequest{
+func pullUpdate(ctx context.Context, localPackages, remotePackages pack.PackageService, reporter pack.ProgressReporter, update storage.PackageUpdate) error {
+	puller := libapp.Puller{
 		SrcPack:  remotePackages,
 		DstPack:  localPackages,
-		Package:  update.To,
 		Progress: reporter,
 	}
-	_, err := appservice.PullPackage(pullReq)
+	err := libapp.PullPackage(context.TODO(), update.To, puller)
```
r0mant: Use `ctx` from above?

lib/app/pull.go
```diff
@@ -0,0 +1,268 @@
...
+	deps, err := GetDependencies(GetDependenciesRequest{
+		App:  *app,
+		Apps: puller.SrcApp,
+		Pack: puller.SrcPack,
+	})
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	deps.Apps = append(deps.Apps, *app)
+	puller.Dependencies = *deps
```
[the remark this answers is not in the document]
a-palchikov: It is not internal state - puller specified as parameter in this case is used as a template.

lib/app/service/installer.go
a-palchikov: This won't and it did not work before as well. Is this a requirement? The script will error on being unable to find the `gravity` binary (if there's none in the $HOME directory) or might error (subject to version of the binary from home directory) on being unable to find installer state if there's a binary directly in $HOME.

tool/gravity/cli/ops.go
a-palchikov: `env` is not (and should not be) the tarball directory - it is whatever state directory has been specified with `--state-dir` (and would be the local gravity state directory if unspecified).
I'm adding an explicit (but optional) `--data-dir` to the command to specify the tarball directory and avoid confusion with the state directory.

PR #575, opened 2019-08-09
Additional hook variables.

r0mant:
This pull request makes sure that all hook containers receive environment variables with certain basic cluster information which people keep asking about. The variables which are currently available in hooks are cluster name, cluster provider and initial flavor name.

The way it works is a new config map (I called it `cluster-info`) is created during initial cluster installation (or during upgrade for older clusters) which then gets mounted into hook containers as environment.

This approach is more flexible and extensible than just injecting env vars directly and also helps with separation of concerns because "app service" (the one that creates hooks) doesn't know anything about the cluster.

I will update the documentation in another PR. Closes https://github.com/gravitational/gravity.e/issues/3599.

### DIFF
--- a/lib/constants/constants.go
+++ b/lib/constants/constants.go
@@ -551,6 +551,15 @@ const (
 	// ClusterConfigurationMap is the name of the ConfigMap that hosts cluster configuration resource
 	ClusterConfigurationMap = "cluster-configuration"
 
+	// ClusterInfoMap is the name of the ConfigMap that contains cluster information.
+	ClusterInfoMap = "cluster-info"
+	// ClusterNameEnv is the environment variable that contains cluster domain name.
+	ClusterNameEnv = "GRAVITY_CLUSTER_NAME"
+	// ClusterProviderEnv is the environment variable that contains cluster provider.
+	ClusterProviderEnv = "GRAVITY_CLUSTER_PROVIDER"
+	// ClusterFlavorEnv is the environment variable that contains initial cluster flavor.
+	ClusterFlavorEnv = "GRAVITY_CLUSTER_FLAVOR"
+
 	// SMTPSecret specifies the name of the Secret with cluster SMTP configuration
 	SMTPSecret = "smtp-configuration-update"
 
--- a/lib/constants/constants.go
+++ b/lib/constants/constants.go
@@ -555,7 +555,7 @@ const (
 	ClusterInfoMap = "cluster-info"
 	// ClusterNameEnv is the environment variable that contains cluster domain name.
 	ClusterNameEnv = "GRAVITY_CLUSTER_NAME"
-	// ClusterProviderEnv is the environment variable that contains cluster provider.
+	// ClusterProviderEnv is the environment variable that contains cluster cloud provider.
 	ClusterProviderEnv = "GRAVITY_CLUSTER_PROVIDER"
 	// ClusterFlavorEnv is the environment variable that contains initial cluster flavor.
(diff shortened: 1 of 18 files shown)


lib/constants/constants.go
```diff
@@ -551,6 +551,15 @@ const (
 	// ClusterConfigurationMap is the name of the ConfigMap that hosts cluster configuration resource
 	ClusterConfigurationMap = "cluster-configuration"
 
+	// ClusterInfoMap is the name of the ConfigMap that contains cluster information.
+	ClusterInfoMap = "cluster-info"
+	// ClusterNameEnv is the environment variable that contains cluster domain name.
+	ClusterNameEnv = "GRAVITY_CLUSTER_NAME"
+	// ClusterProviderEnv is the environment variable that contains cluster provider.
```
a-palchikov: Provider as in cloud provider?

lib/install/phases/resources.go
```diff
@@ -26,17 +26,106 @@ import (
...
+func (r *systemResources) Execute(ctx context.Context) error {
+	r.Progress.NextStep("Configuring system Kubernetes resources")
+	r.Info("Configuring system Kubernetes resources.")
+	if err := r.createClusterInfoMap(); err != nil {
+		return trace.Wrap(err)
+	}
+	return nil
+}
+
+// createClusterInfoMap creates a config map with basic cluster information.
```
a-palchikov:
```go
// createClusterInfoMap creates a config map with cluster information to be made available to every cluster hook.
```

lib/install/planbuilder.go
```diff
@@ -349,14 +351,29 @@ func (b *PlanBuilder) AddRBACPhase(plan *storage.OperationPlan) {
...
-func (b *PlanBuilder) AddResourcesPhase(plan *storage.OperationPlan) {
+// AddSystemResourcesPhase appends phase that creates system Kubernetes
+// resources to the provided plan.
+func (b *PlanBuilder) AddSystemResourcesPhase(plan *storage.OperationPlan) {
+	plan.Phases = append(plan.Phases, storage.OperationPhase{
+		ID:          phases.SystemResourcesPhase,
+		Description: "Create system Kubernetes resources",
+		Data: &storage.OperationPhaseData{
+			Server:  &b.Master,
+			Cluster: &b.Cluster,
```
a-palchikov: Why not store cluster record directly in the operation plan? I'm sure it's useful in other places and will even minimize redundancy.

lib/update/cluster/phases/init.go
```diff
@@ -270,6 +282,30 @@ func (p *updatePhaseInit) updateClusterDNSConfig() error {
...
+	if err == nil {
+		p.Info("Config map %v already exists.", constants.ClusterInfoMap)
+		return nil
+	}
+	err = rigging.ConvertError(err)
+	if !trace.IsNotFound(err) {
+		return trace.Wrap(err)
+	}
+	// Cluster info config map doesn't exist yet, create it.
+	configMap := phases.ClusterInfoMap(ops.ConvertOpsSite(p.Cluster))
```
a-palchikov: Not to be too nit-picky, but it's puzzling that lib/update is using lib/install for this - I'd think that this configuration should be stored in another package which both lib/install and lib/update could use. Could be even a totally new package like `lib/kubernetes` or `lib/config` or lib/cluster for Gravity-specific cluster configuration.

lib/install/planbuilder.go
r0mant: I'm actually not sure why I did that at all. ClusterID is available in the plan so I can just query it.

PR #576, opened 2019-08-09
Add Zoho Pagesense

benarent:
Adds support for a page survey to gather feedback from Gravity docs.

### DIFF
--- a/docs/theme/src/third-party.js
+++ b/docs/theme/src/third-party.js
@@ -15,7 +15,9 @@ if (location.origin === "https://gravitational.com"){
   ga('create', "UA-65867304-1", 'auto');
   ga('send', 'pageview');
 
-
+  //ZOHO Page Sense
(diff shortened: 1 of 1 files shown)


benarent:
@alexwolfe just a friendly ping to remind to review this.

knisbet:
@benarent Just FYI, loading external resources does create some risks around security. I know from talking to managers at other security oriented vendors, they've been working in the direction of removing any third party content from their sites entirely. So please be mindful that any third party resources add significant value, or try to mirror or protect the loading of the resources from changes using embedded checksums.

r0mant:
@benarent @alexwolfe @knisbet This has been sitting for a while, let's do something about it? I.e. close if we can live without it or merge if it's essential (something tells me it's not since it's been here for a month). If we're to merge this, I agree with Kevin's suggestions, some pointers to look at: https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity.

PR #582, opened 2019-08-13
Fix hook vars.

r0mant:
Instead of mounting config map with cluster info into hooks (which won't work if the hook is launched from any namespace other than kube-system), read vars from it and populate environment in the request.

### DIFF
--- a/lib/app/service/app.go
+++ b/lib/app/service/app.go
@@ -43,8 +43,10 @@ import (
 	"github.com/gravitational/gravity/lib/utils"
 
 	dockerarchive "github.com/docker/docker/pkg/archive"
+	"github.com/gravitational/rigging"
 	"github.com/gravitational/trace"
 	log "github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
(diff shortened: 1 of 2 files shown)


lib/app/service/app.go
```diff
@@ -329,6 +333,24 @@ func (r *applications) StartAppHook(ctx context.Context, req appservice.HookRunR
 	}, nil
 }
 
+// injectEnvVars updates the provided hook run request with additional
+// environment variables such as cluster information.
+func (r *applications) injectEnvVars(req *appservice.HookRunRequest, client *kubernetes.Clientset) error {
+	configMap, err := client.Core().ConfigMaps(metav1.NamespaceSystem).Get(
+		constants.ClusterInfoMap, metav1.GetOptions{})
```
a-palchikov: But do we need the configmap now - maybe use the cluster information bypassing the configmap? What if it gets removed or updated in an incompatible way?

r0mant: I thought about this. The thing is, even though the app service has a backend, there may be no cluster in it (e.g. in local case) so it can't rely on querying cluster from the backend. We could add cluster to the hook request but I'm not sure that's a good approach... it would also move the burden of retrieving/passing it on the callers...

a-palchikov: Which means we have now more things to worry about w/o for reconciliation.

a-palchikov: Besides, what you describe as an unforunate lack of data is fixable by providing enough information to the application service so it can provide hooks with that data - we're not limited to using what we've developed thus far - we can tweak and fix things to our advantage.

PR #587, opened 2019-08-14
[5.2.x] Generate configuration package for teleport/planet offline

a-palchikov:
This PR adds two additional (hidden) system commands to enable offline configuration package generation for upgrades executed by a later version (5.5.x upwards):

```
$ gravity update rotate-planet-config --help
usage: gravity update rotate-planet-config --package=PACKAGE --runtime-package=RUNTIME-PACKAGE --addr=ADDR --id=ID

Generate new configuration package for planet

Flags:
      --log-file="/var/log/telekube-install.log"  
      --package=PACKAGE      Name of the package to generate
      --runtime-package=RUNTIME-PACKAGE  
                             Name of the runtime package to generate configuration for
      --addr=ADDR            Address of this node as used by the update operation
      --id=ID                Update oepration ID
```
```
$ gravity update rotate-teleport-config --help
usage: gravity update rotate-teleport-config --package=PACKAGE --addr=ADDR --id=ID

Generate new configuration package for teleport

Flags:
      --log-file="/var/log/telekube-install.log"  
      --package=PACKAGE      Name of the package to generate
      --addr=ADDR            Address of this node as used by the update operation
      --id=ID                Update oepration ID

```
Updates https://github.com/gravitational/gravity.e/issues/4146.


lib/ops/ops.go
```diff
@@ -759,35 +762,49 @@ func (r RotateSecretsRequest) SiteKey() SiteKey {
...
+type RotateTeleportConfigRequest struct {
+	// Key identifies the cluster operation
+	Key SiteOperationKey `json:"key"`
 	// Server is the server to rotate configuration for
 	Server storage.Server `json:"server"`
 	// Servers is all cluster servers
-	Servers []storage.Server `json:"servers"`
+	Servers storage.Servers `json:"servers"`
+	// Node specifies the configuration package for teleport nodes
+	Node *loc.Locator `json:"node,omitempty"`
```
a-palchikov: I left out the changes for the master teleport configuration since it's out of scope for offline generation.

tool/gravity/cli/register.go
```diff
@@ -166,17 +166,24 @@ func RegisterCommands(app *kingpin.Application) *Application {
 	g.UpdateUploadCmd.CmdClause = g.UpdateCmd.Command("upload", "Upload update package to locally running site").Hidden()
 	g.UpdateUploadCmd.OpsCenterURL = g.UpdateUploadCmd.Flag("ops-url", "Optional OpsCenter URL to upload new packages to (defaults to local gravity site)").Default(defaults.GravityServiceURL).String()
 
-	// manual update flow commands
-	g.UpdateCompleteCmd.CmdClause = g.UpdateCmd.Command("complete", "Mark update operation as completed").Hidden()
```
a-palchikov: This command had no implementation hence I removed it.

lib/pack/constants.go
```diff
@@ -44,8 +44,10 @@ const (
 	PurposePlanetConfig = "planet-config"
 	// PurposeRuntime marks a package as a runtime container package
 	PurposeRuntime = "runtime"
-	// PurposeTeleportConfig marks packages with teleport config
+	// PurposeTeleportConfig marks package with teleport master config
 	PurposeTeleportConfig = "teleport-config"
+	// PurposeTeleportNodeConfig marks package with teleport node config
```
a-palchikov: I only added the node teleport configuration marker - master configuration is still generated using the old label.

lib/ops/opsservice/update.go
```diff
@@ -70,13 +69,59 @@ func (o *Operator) RotateSecrets(req ops.RotateSecretsRequest) (*ops.RotatePacka
...
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	masterIPs := (storage.Servers)(req.Servers).MasterIPs()
+	if len(masterIPs) == 0 {
+		return nil, trace.NotFound("no master server found")
+	}
+
+	nodeConfig, err = cluster.getTeleportNodeConfig(ctx, masterIPs[0], *nodeConfigPackage, node)
```
a-palchikov: @r0mant not sure if I needed to pull the change to use all master addresses - do you think it's necessary for 5.2.x?

lib/ops/ops.go
```diff
@@ -698,6 +698,9 @@ type Updates interface {
 	// RotatePlanetConfig rotates planet configuration package for the server specified in the request
 	RotatePlanetConfig(RotatePlanetConfigRequest) (*RotatePackageResponse, error)
 
+	// RotateTeleportConfig rotates teleport configuration package for the server specified in the request
+	RotateTeleportConfig(RotateTeleportConfigRequest) (nodeConfig *RotatePackageResponse, err error)
```
knisbet: nit: nodeConfig and err variables in interface description I assume are not needed? Or nodeConfig seems like a sort of weird name for RotatePackageResponse.

lib/ops/ops.go
```diff
@@ -759,35 +762,49 @@ func (r RotateSecretsRequest) SiteKey() SiteKey {
...
-// RotatePlanetConfigRequest is a request to rotate server's planet configuration package
-type RotatePlanetConfigRequest struct {
-	// AccountID is the account id of the local cluster
-	AccountID string `json:"account_id"`
-	// ClusterName is the local cluster name
-	ClusterName string `json:"cluster_name"`
-	// OperationID is the id of the operation
-	OperationID string `json:"operation_id"`
+// Check validates this request
+func (r RotateTeleportConfigRequest) Check() error {
```
knisbet: nit: I believe we use CheckAndSetDefaults as a standard name in other locations in the codebase, even if Defaults are not being set.

lib/ops/opsservice/configure.go
```diff
@@ -811,82 +801,73 @@ type masterConfig struct {
...
+	}
+	reader, err := pack.GetConfigPackage(s.packages(), config.planetPackage, config.configPackage, args)
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
+	labels := map[string]string{
+		pack.PurposeLabel:     pack.PurposePlanetConfig,
+		pack.ConfigLabel:      config.planetPackage.ZeroVersion().String(),
+		pack.AdvertiseIPLabel: config.server.AdvertiseIP,
+		pack.OperationIDLabel: config.installExpand.ID,
```
knisbet: I'm not sure I understand why AdvertiseIP and OperationID are being added as labels / what purpose they serve? Usually when I think of a labelling system, I think of it as a way to attach selectors to retrieve a set or individual object. To me, I'm not sure what embedding server config in the label is meant to do, and the operation to me seems more likely it should maybe be metadata, which we may not have support for?

a-palchikov: This is a good question. I think `OperationID` was added as a means of being able to pull the package during init/bootstrap step of the corresponding operation (@r0mant correct me if I'm wrong), I would think advertise IP serves similar purpose but I'm not completely sure. I will see if I can do away with that (note though that this PR is a backport of existing changes - hence I'd avoid making too many deviations).

knisbet: As a backport I wouldn't worry about it then.

lib/ops/ops.go
r0mant: Nit: Bit weird that the field `Node` has type `Locator`, I got confused initially. Maybe `NodePackage` or `NodeConfig`?

lib/ops/opsservice/configure.go
r0mant: These labels are used by install FSM when pulling packages for specific operation/server, yes.

lib/ops/opsservice/update.go
r0mant: I think it depends on what will happen with the nodes if this `masterIPs[0]` master goes down / is removed. Presumably they won't be able to contact auth server anymore? We can address this in another PR though as it seems tangential.

PR #588, opened 2019-08-14
Fix flaky rpc/proxy test

a-palchikov:
This PR attempts to fix an issue with flaky rpc/proxy tests observed in robotest.
It adds explicit `io.Closer` implementation for the proxy local link and moves to use `sync.WaitGroup` instead of channels to observe lifespan of internal connection handlers.

### DIFF
--- a/lib/rpc/proxy/proxy.go
+++ b/lib/rpc/proxy/proxy.go
@@ -134,7 +134,10 @@ func (r *Proxy) serve(listener net.Listener) {
 
 		c2, err := r.link.Dial()
 		if err != nil {
-			r.WithError(err).Warn("Failed to dial: %v.")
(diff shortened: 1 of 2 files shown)


lib/rpc/proxy/proxy.go
```diff
@@ -110,35 +120,47 @@ func (r NetLink) String() string {
...
 	for {
 		c1, err := listener.Accept()
 		if err != nil {
-			r.Errorf("Failed to accept: %v.", err)
+			select {
+			case <-r.doneCh:
+				return
+			default:
+			}
+			r.WithError(err).Warnf("Failed to accept.")
```
r0mant: Nit: `Warn`.

PR #590, opened 2019-08-14
[6.1] Do not set auth gateway public addrs.

r0mant:
https://github.com/gravitational/gravity/pull/589 port to 6.1.


itay:
@r0mant is this going to be in a published 6.1.x release (e.g. 6.1.2)?

PR #593, opened 2019-08-15
6.0.4 changelog.

r0mant:


docs/6.x/changelog.md
```diff
@@ -32,6 +32,13 @@ LTS starts with `3.51.0` with minor backwards compatible changes added over time
 
 ## 6.x Releases
 
+### 6.0.4 (August 15th, 2019)
+
+#### Bugfixes
+
+* Fix an issue with ssh'ing into some nodes via Cluster Control Panel.
+* Fix an issue with some audit events not being returned when using custom date picker.
```
a-palchikov: not being returned -> filtered out?

docs/6.x/changelog.md
```diff
@@ -32,6 +32,13 @@ LTS starts with `3.51.0` with minor backwards compatible changes added over time
 
 ## 6.x Releases
 
+### 6.0.4 (August 15th, 2019)
+
+#### Bugfixes
+
+* Fix an issue with ssh'ing into some nodes via Cluster Control Panel.
```
a-palchikov: with ssh'ing -> with ssh connections to nodes
Would be helpful to clarify `some` in this context.

docs/6.x/changelog.md
r0mant: I think these two mean different things and "not being returned" fits better in this case.

PR #599, opened 2019-08-19
logo for docs

alexwolfe:
@russjones added logo in docs nav


russjones:
lgtm, but probably @r0mant or @alex-kovoy should review since this is a Gravity change?

Issue #619, opened 2019-09-03
Add ability to enable privileged containers


r0mant:
### Problem

By default Gravity clusters do not allow users to run privileged containers by starting kube-apiserver with `--allow-privileged=false` flag.

Privileged containers are required though in order to be able to run things like custom CNI plugins and dynamic volume provisioners.

In the ideal use-case, a user would be able to write a proper PSP spec to run these services but it is oftentimes quite challenging (if possible at all) to figure out the required set of capabilities, and plus privileged containers enable some additional features on top of that such as ability to see and manage all host devices which is required for volume plugins.

### Solution

Keep privileged containers off by default but add an ability to enable them by adding the flag to the cluster manifest, extending the existing "systemOptions" section:

```yaml
systemOptions:
  allowPrivileged: true
```

This way the cluster will have an ability to run privileged containers and users will still be able to control this ability in a fine-grained way by using PSPs.

kontsevoy:
Some thoughts:

1. You've used different verbs for the manifest and the CLI: "enable" vs "allow". I would pick "allow". 
2. But do we need a verb at all? What if we trade it for more descriptive name: `--privileged-containers=false` or `privilegedContainers: false`?

@benarent may add some more.

r0mant:
@kontsevoy @benarent The `--allow-privileged` flag is not ours - it's a flag of `kube-apiserver` which is a Kubernetes binary - we're not adding a CLI flag to our binaries, only the manifest field.

We could use `allowPrivileged` instead, yeah, to keep it consistent with the flag name.

PR #620, opened 2019-09-04
Add ability to allow privileged containers via a manifest flag.

r0mant:
This PR adds an additional boolean manifest field `systemOptions.allowPrivileged` that enables privileged containers in the cluster. Requires https://github.com/gravitational/planet/pull/476. Closes https://github.com/gravitational/gravity/issues/619.

### DIFF
--- a/docs/6.x/cluster.md
+++ b/docs/6.x/cluster.md
@@ -4,12 +4,12 @@ Every Cluster created from a Gravity Cluster Image is a fully featured
 Kubernetes environment. It contains the following components:
 
 1. All user applications that have been packaged into the Cluster Image.
-2. All Kubernetes daemons like `kube-scheduler`, `kube-apiserver`, and others. 
+2. All Kubernetes daemons like `kube-scheduler`, `kube-apiserver`, and others.
 3. The Kubernetes CLI tool: `kubectl`.
 4. The Gravity Cluster "hypervisor", called `gravity`, for managing the Cluster.
 5. A web application to monitor and manage the Cluster state ("Control
 Panel")
-6. The Gravity Authentication Gateway for integrating Kubernetes authentication 
+6. The Gravity Authentication Gateway for integrating Kubernetes authentication
    and SSH access to Cluster nodes with corporate identity providers via SSO.
 
 You can also use familiar Kubernetes tools such as `kubectl` to perform regular
@@ -29,8 +29,8 @@ However, Gravity pre-configures Kubernetes to be as reliable as possible,
 greatly reducing the need for ongoing active management. To make this possible,
 the `gravity` hypervisor launches and manages all Kubernetes Linux daemons.
 
-When the `gravity` daemon starts, it creates its own "Master Container". The 
-Master Container itself is a containerized `systemd` instance. It launches the required Kubernetes daemons and contains their dependencies as well as monitoring and support tools. 
+When the `gravity` daemon starts, it creates its own "Master Container". The
+Master Container itself is a containerized `systemd` instance. It launches the required Kubernetes daemons and contains their dependencies as well as monitoring and support tools.
 
 Running Kubernetes services from inside the Master Container brings several
 advantages as opposed to installing Kubernetes components in a more traditional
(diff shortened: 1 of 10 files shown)


r0mant:
@knisbet Yeah, I was gonna do them in a separate PR but I guess here would be fine too.

r0mant:
@knisbet I've added the docs too.

docs/6.x/cluster.md
```diff
@@ -4,12 +4,12 @@ Every Cluster created from a Gravity Cluster Image is a fully featured
 Kubernetes environment. It contains the following components:
 
 1. All user applications that have been packaged into the Cluster Image.
-2. All Kubernetes daemons like `kube-scheduler`, `kube-apiserver`, and others.
```
r0mant: That's just my text editor auto-truncating trailing whitespaces. There's a bunch of these here.

lib/schema/schema.go
```diff
@@ -644,6 +644,7 @@ const manifestSchema = `
       "additionalProperties": false,
       "properties": {
         "baseImage": {"type": "string"},
+        "allowPrivileged": {"type": "boolean"},
```
a-palchikov: Did you try updating with this flag on? I remember `additionalProperties: false` being the culprit for update failures as the existing cluster will probably fail to validate the new application's manifest.

docs/6.x/cluster.md
benarent: 😡

docs/6.x/pack.md
```diff
@@ -465,13 +465,16 @@ systemOptions:
     args: ["--system-reserved=memory=500Mi"]
     hairpinMode: "promiscuous-bridge"
 
+  # When set to true, allows running privileged containers in the deployed
+  # clusters, defaults to false
+  allowPrivileged: true
```
benarent: It's hard to see the context here, but do we want to set this true? If it's a default example?

docs/6.x/cluster.md
r0mant: @benarent FWIW I think everyone should configure their editors to truncate trailing whitespaces :)

docs/6.x/pack.md
r0mant: Well, the purpose of this particular sample manifest is to show how to enable everything you can possibly enable. So this is not technically "default" manifest you'd want to use. I'm okay with setting this to false here though.

lib/schema/schema.go
r0mant: @a-palchikov Good call, yeah - it didn't work obviously due to this schema issue.

I took a stab at fixing it, check out my latest commit: https://github.com/gravitational/gravity/pull/620/commits/561f1b3928dfd204f149e3f6d1f6be9032cfcf83. With these changes the upgrade works (I tried 6.0 to this new version with privileged flag enabled).

lib/localenv/localenv.go
```diff
@@ -579,6 +585,9 @@ type AppConfig struct {
 	// Packages allow to override default env.Packages when creating
 	// an app service
 	Packages pack.PackageService
+	// Backend allows to override default env.Backend when creating
```
a-palchikov: `... allows to override default backend when creating the service`

tool/gravity/cli/clusterupdate.go
```diff
@@ -322,11 +321,26 @@ func checkForUpdate(
...
+		return nil, trace.Wrap(err)
+	}
+
+	// Create the *local* app service that uses the cluster's backend and
+	// packages.
+	//
+	// The local service is needed to handle cases such as newly introduced
+	// manifest field which gravity-site which may be running the old code
+	// does not recognize.
+	apps, err := env.AppServiceLocal(localenv.AppConfig{
```
a-palchikov: I'm now wondering why this is not part of the cluster environment - why not have `ClusterApps` (similar to `ClusterPackages` that uses the cluster application service)?

r0mant: Well... this app service is kind of a weird beast cause it uses cluster's packages/backend, but by itself is local... I guess I could put it in there anyway.

docs/6.x/pack.md
knisbet: I think if you look at the way we do other samples, such as extensions.logs.disabled we show the default value (false). So it would seem to match to have allowPrivileged show the default value as well.

r0mant: Yeah, I'll change it to false.

Issue #606, opened 2019-08-23
[BUG] CoreDNS configmap missing upstream resolvers


knisbet:
**Describe the bug**

The auto-generated coredns configuration is missing upstream resolvers. This prevents pods from being able to resolve names from outside the cluster.

**To Reproduce**

```
sudo GRAVITY_CHECKS_OFF=1 ./gravity install --cloud-provider=generic --pod-network-cidr="10.244.0.0/16" --service-cidr="10.100.0.0/16" --vxlan-port=8472 --cluster=cluster.local
```

**Expected behavior**

The coredns configmap template should be generated with upstream nameservers.

**Logs**

```
kubectl -n kube-system get configmap/coredns -o yaml | sed 's/\\n/\n/g'
apiVersion: v1
data:
  Corefile: "
.:53 {
  reload
  errors
  health
  prometheus :9153
  cache 30

    \ loop
  reload
  loadbalance
  hosts {
    fallthrough
  }
  kubernetes
    cluster.local in-addr.arpa ip6.arpa {
    pods verified
    fallthrough in-addr.arpa
    ip6.arpa
  }

}
"
kind: ConfigMap
metadata:
  creationTimestamp: "2019-08-23T14:37:30Z"
  managedFields:
  - apiVersion: v1
    fields:
      f:data:
        .: null
        f:Corefile: null
    manager: gravity
    operation: Update
    time: "2019-08-23T14:37:30Z"
  name: coredns
  namespace: kube-system
  resourceVersion: "208"
  selfLink: /api/v1/namespaces/kube-system/configmaps/coredns
  uid: 891f52f0-c5b3-11e9-b1ba-42010aa20fe5
```

**Environment (please complete the following information):**
 - OS [e.g. Redhat 7.4]: Ubuntu 18.04
 - Gravity [e.g. 5.5.4]: 6.0.2
 - Platform [e.g. Vmware, AWS]: GCE

**Additional context**

knisbet:
Doing a bit of digging, it appears this is due to ubuntu 18.04 using systemd-resolved as a local caching/stub resolver and configures /etc/resolv.conf to use the localhost resolver and not the upstream DNS servers. This creates a problem for gravity, because the /etc/resolv.conf is how gravity detects the upstream DNS servers, so coredns can be configured to communicate with them directly.

```
knisbet@knisbet-kubeadm1:~/6.1.2$ cat /etc/resolv.conf
# This file is managed by man:systemd-resolved(8). Do not edit.
#
[...]

PR #627, opened 2019-09-08
support loading upstream DNS servers when systemd-resolved stub resol…

knisbet:
…ver is enabled

Updates #606 
Requires Backport

The systemd-resolved stub resolver when configured points /etc/resolv.conf at 127.0.0.53 with a systemd-resolved listener. The configuration for upstream DNS appears to be embedded somewhere in DBUS, so it would require more research to see if there is a correct programatic way to pull the upstream DNS configuration into gravity. For now, it seems we can read a compatibility more resolv.conf from /run/systemd/resolve/resolv.conf which has the populated list of upstream DNS servers.


lib/install/phases/coredns.go
```diff
@@ -93,12 +93,24 @@ func (r *corednsExecutor) Execute(ctx context.Context) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
+	servers := resolvConf.Servers
+
+	// Optionally try and load upstream nameservers from systemd-resolved by reading the compatibility resolv.conf
+	// More Info: https://github.com/gravitational/gravity/issues/606#issuecomment-529171440
+	// TODO(knisbet) is there a better way to pull upstream resolvers directly from systemd?
+	systemdResolvConf, err := systeminfo.ResolvFromFile("/run/systemd/resolve/resolv.conf")
```
a-palchikov: Even though I also have `systemd-resolved` active on my ubuntu `18.04`, I don't have any upstream nameservers listed in `/run/systemd/resolve/resolv.conf` - instead it just links back to `systemd-resolved` with `127.0.0.53`.
I'm not sure which of the [compatibility modes](https://www.freedesktop.org/software/systemd/man/systemd-resolved.service.html#/etc/resolv.conf) is on but the file does not list any link-specific DNS servers.

One way I saw this information being queried from `systemd-resolved` is by fetching `DNS` servers property for the specific link from [DBus](https://www.freedesktop.org/wiki/Software/systemd/resolved/). Here's an [example](https://github.com/FiloSottile/captive-browser/blob/master/cmd/systemd-networkd-dns/main.go) of the query.
Also, on my system the `resolve1` DBus service was not registered and I subsequently failed to query it with DBus although `systemd-resolve --status` is reporting the DNS servers properly.

knisbet: Well that sucks, I was hoping regardless of compatibility mode it was writing that file. thanks for checking. Guess we'll have to go the dbus route, but the problem I have with that is I found out how to pull the per interface DNS over dbug (although that go code is a good reference), I didn't see a clear indication on how to pull the global resolver when set...

```
Global
         DNS Servers: 8.8.8.8
          DNSSEC NTA: 10.in-addr.arpa
                      16.172.in-addr.arpa
                      168.192.in-addr.arpa
```

`systemd-resolved --status` obviously pulls it from somewhere, just need to waste time figuring out where that is.

a-palchikov:
Ok, so after restarting the `NetworkManager` service on my system, the `systemd-resolved` did update the `/run/systemd/resolve/resolv.conf` file with the list of external nameservers and seems to keep them in sync if they're updated in the settings.

lib/install/phases/coredns.go
a-palchikov: Ok, so after restarting the `NetworkManager` service on my system, the `systemd-resolved` did update the `/run/systemd/resolve/resolv.conf` file with the list of external nameservers and seems to keep them in sync if they're updated in the settings.

lib/install/phases/coredns.go
```diff
@@ -93,12 +93,24 @@ func (r *corednsExecutor) Execute(ctx context.Context) error {
...
+	if err != nil && !trace.IsNotFound(err) {
+		return trace.Wrap(err)
+	}
+	if systemdResolvConf != nil {
+		servers = append(servers, systemdResolvConf.Servers...)
```
r0mant: Do we need to do some kind of deduplication on this list now?

knisbet:
I request re-review, since I added deuplication logic. Also I think this change may need to be replicated into planet and it's resolv.conf as well.

PR #633, opened 2019-09-10
[5.5.x] package dependency/pull updates

a-palchikov:
* Enable parallel pull for application packages.
 * Update package conflict handling during pulls to avoid introducing inflexible control attributes (`SkipIfExists`). This also reverts the application pull logic behavior.

### DIFF
--- a/lib/app/pull.go
+++ b/lib/app/pull.go
@@ -185,7 +185,7 @@ func (r Puller) pullPackage(loc loc.Locator) error {
 			return trace.Wrap(err)
 		}
 	}
-	logger.Debug("Pull package.")
+	logger.Info("Pull package.")
 	reader := ioutil.NopCloser(utils.NopReader())
 	if r.MetadataOnly {
 		env, err = r.SrcPack.ReadPackageEnvelope(loc)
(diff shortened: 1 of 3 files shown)


lib/app/pull.go
```diff
@@ -232,13 +252,15 @@ func (r Puller) pullApp(loc loc.Locator) error {
...
 			return nil
 		}
-		logger.Info("Application already exists.")
-		return trace.AlreadyExists("application %v already exists", loc)
+		if err != nil {
+			return trace.Wrap(err)
+		}
 	}
-	logger.Info("Pull application.")
+	logger.Debug("Pull application.")
```
r0mant: Nit: IMO it was quite helpful to see this log entry without having to enable debug mode. Same for "pull package" above. Unless this is a duplicate and it's already logged elsewhere?

lib/app/pull.go
```diff
@@ -266,3 +288,43 @@ func (r Puller) pullApp(loc loc.Locator) error {
...
 }
+
+func onConflictDependencies(upsert bool) conflictHandler {
+	if upsert {
+		return onConflictContinue
+	}
+	return onConflictSkip
+}
+
+func onConflict(upsert bool) conflictHandler {
```
r0mant: Nit: I would add comments on this and the above functions explaining our reasoning behind handling "already exists" cases for the app and its dependencies.

PR #638, opened 2019-09-11
Add mTLS support.

r0mant:
This PR add support for client certificate authentication to Gravity web handlers and teaches "tele" to use Teleport-issued user certificates when talking to those APIs.

The high-level user-facing impact of this is that now "tele" can be used right after "tsh login" - "tele login" is no longer required.

Some details:

* Web handlers for pack/app/ops services can now authenticate incoming requests by x509 certificates provided by the clients. The way it works is:
  * All handlers are wrapped in Teleport's AuthMiddleware that looks for optional x509 cert in the HTTP request, extracts username from it and adds it to the request context.
  * Handlers then look if username is present in the request context and use it to get user info. Otherwise they fallback to regular auth methods such as bearer token or basic auth.

* All our handlers had exactly the same "needsAuth" logic so I factored it out into a dedicated "Authenticator" which all of them now use. No more code duplication!

* The "tele" client now knows how to search the standard Teleport's key store (~/.tsh) for the current cluster and user-certificates, and uses them when creating pack/app/ops clients which then triggers the authentication flow described above on the server side. The "old" token-based authentication still takes precedence though, for the time being for backwards compatibility.

* I felt like the "LocalEnvironment" was becoming too messy so I also factored out all credentials-related logic to a "CredentialsService" which is responsible to retrieving user credentials from different sources - Gravity config (which will become obsolete), Teleport key store or bolt key store.

There are still quite a few things outstanding so fully replacing "tele login" with "tsh login" is far from done, but this chunk should be fine to go in on its own.

References https://github.com/gravitational/gravity.e/issues/4198.


lib/ops/resources/gravity/gravity_test.go
```diff
@@ -51,9 +53,14 @@ var _ = check.Suite(&GravityResourcesSuite{
 func (s *GravityResourcesSuite) SetUpSuite(c *check.C) {
 	s.s.SetUp(c)
 	// start up ops server using configured ops handler
-	s.server = httptest.NewServer(s.s.Handler)
+	s.server = httptest.NewTLSServer(s.s.Handler)
 	// create the ops client that uses admin agent creds
-	client, err := opsclient.NewBearerClient(s.server.URL, s.s.Creds.Password)
+	client, err := opsclient.NewBearerClient(s.server.URL, s.s.Creds.Password,
+		opsclient.HTTPClient(&http.Client{
```
a-palchikov: I've this at least 3 times, `newInsecureHTTPClient`?

lib/builder/builder.go
```diff
@@ -124,6 +126,14 @@ func (c *Config) CheckAndSetDefaults() error {
 	if c.GetRepository == nil {
 		c.GetRepository = getRepository
 	}
+	if c.Credentials == nil {
+		c.Credentials, err = localenv.NewCredentials(localenv.CredentialsServiceConfig{
```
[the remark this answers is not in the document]
r0mant: This might be better, yes.

lib/localenv/credentials.go
```diff
@@ -0,0 +1,352 @@
...
+	if currentCluster != "" {
+		return currentCluster, nil
+	}
+	currentProfile, err := s.currentProfile()
+	if err == nil {
+		return fmt.Sprintf("https://%v", currentProfile.WebProxyAddr), nil
+	}
+	if s.dbKeyStore != nil {
+		entries, err := s.dbKeyStore.GetLoginEntries()
+		if err == nil && len(entries) == 1 {
```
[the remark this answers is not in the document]
r0mant: This is actually how it's worked before too, this logic was just ported from localenv as-is.

lib/localenv/localenv.go
```diff
@@ -438,9 +311,9 @@ func (env *LocalEnvironment) OperatorService(opsCenterURL string, options ...htt
 
 // SiteOperator returns Operator for the local gravity site
 func (env *LocalEnvironment) SiteOperator() (*opsclient.Client, error) {
-	operator, err := env.OperatorService(
-		defaults.GravityServiceURL, httplib.WithLocalResolver(env.DNS.Addr()), httplib.WithInsecure())
-	return operator, trace.Wrap(err)
+	return env.OperatorService(defaults.GravityServiceURL,
+		httplib.WithLocalResolver(env.DNS.Addr()),
+		httplib.WithInsecure())
```
knisbet: I realize not in scope for this PR, but I really need to work on getting rid of these hardcoded insecure calls everywhere.

lib/ops/opshandler/opshandler.go
```diff
@@ -2307,62 +2335,25 @@ func (s *WebHandler) needsAuth(fn ServiceHandle) httprouter.Handle {
...
 // GetHandlerContext authenticates the user that made the request and returns
 // the appropriate handler context
-func GetHandlerContext(w http.ResponseWriter, r *http.Request, backend storage.Backend, operator ops.Operator, auth httplib.Authenticator, usersService users.Identity) (*HandlerContext, error) {
-	authCreds, err := httplib.ParseAuthHeaders(r)
+func GetHandlerContext(w http.ResponseWriter, r *http.Request, backend storage.Backend, operator ops.Operator, authenticator users.Authenticator, usersService users.Identity) (*HandlerContext, error) {
+	authResult, err := authenticator.Authenticate(w, r)
 	if err != nil {
-		return nil, trace.Wrap(err, "credentials error")
+		log.WithError(err).Debug("Authentication error.")
+		return nil, trace.AccessDenied("bad username or password") // Hide the actual error.
```
knisbet: When suppressing an error, I think logging of that error should not be debug, IE something high enough that it would actually be present in the logs. Possibly even with a correlation ID, where the correlation ID can be used to find the full error.

lib/ops/opshandler/opshandler_test.go
```diff
@@ -99,12 +101,21 @@ func (s *OpsHandlerSuite) SetUpTest(c *C) {
...
 
 	// for regular test, let's be admins, so tests
 	// won't be affected by auth issues
 	s.client, err = opsclient.NewAuthenticatedClient(
-		s.webServer.URL, s.adminUser, "admin-password")
+		s.webServer.URL, s.adminUser, "admin-password",
+		opsclient.HTTPClient(&http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: true,
```
knisbet: Nit: I usually try to generate an x509 cert inside the test suite, so that the tests don't run with x09 verification disabled. 

https://github.com/gravitational/satellite/blob/c383450edfb3dc4354be14d141dfd004b923e4c9/agent/server_test.go#L724-L778

lib/users/ap.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// GetDomainName returns the CA cluster name.
+func (a *accessPoint) GetDomainName() (string, error) {
+	clusterName, err := a.GetClusterName()
+	if err != nil && !trace.IsNotFound(err) {
+		return "", trace.Wrap(err)
+	}
+	if trace.IsNotFound(err) {
+		return "", nil
```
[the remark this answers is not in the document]
r0mant: That's how I had done it originally, but this method is called by the internal Teleport's middleware and it does not tolerate not handle not found errors atm. I'll add a comment here.

lib/users/auth.go
```diff
@@ -0,0 +1,190 @@
...
+	cookie, err := r.Cookie(constants.SessionCookie)
+	return err == nil && cookie != nil && cookie.Value != ""
+}
+
+// authenticateContext attempts to authenticate the provided request by looking
+// at whether it has a user name already set in its context.
+//
+// The user is set in the request context by an authentication middleware that
+// extracts it from the verified client-supplied x509 certificate.
+func (a *authenticator) authenticateContext(r *http.Request) (storage.User, services.AccessChecker, error) {
```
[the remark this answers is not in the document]
r0mant: Good call.

lib/ops/opshandler/opshandler_test.go
r0mant: Yeah, it's actually even easier as the test TLS server provides the underlying client which is already configured with proper certs/pool.

r0mant:
@a-palchikov @knisbet I've addressed your comments: https://github.com/gravitational/gravity/pull/638/commits/6ff41afec38ca8ceceb10e3d6054758c708c82a2.

PR #646, opened 2019-09-13
Backported TimeDrift check

eldios:
Up-ped Planet version to backport TimeDrift check to 5.5.x


a-palchikov:
Upgrades are failing with:
```
{"checker":"ping-checker","status":"running"},{"checker":"time-drift","detail":"failed to check time drift","status":"failed","error":"rpc error: code = Unimplemented desc = unknown method Time"}
```

eldios:
it seems that upgrades are failing when upgrading from 5.2.x cause those checks are non-existent in there yet... I'll investigate why.

r0mant:
@eldios I remember when working on this check, I accounted for the rolling update use-case and made sure that it tolerates a "not implemented" error (which is what gRPC API returns for unknown methods): https://github.com/gravitational/satellite/blob/7.0.1/monitoring/timedrift.go#L205. Maybe in 5.2 the error is somehow not wrapped correctly (I think trace.NotImplemented was introduced not too long ago)?

eldios:
closing the PR for now to fix it properly.

PR #649, opened 2019-09-16
[5.2.x] remove teleport configuration generation

a-palchikov:
The generation relied on the fact that the existing cluster has all teleport node configuration packages in cluster package service which appears to only be true if the cluster was installed as a single node and expanded on a node-by-node basis. If the cluster was initially instead installed using a different flavor the node configuration packages are only selectively available in cluster package service.

Instead, do not generate the package name in the first place and let the system upgrade work out the upgrade itself - i.e. keep the existing configuration package if the rotation API does not return anything.


tool/gravity/cli/update.go
```diff
@@ -291,27 +291,8 @@ func rotatePlanetConfig(env *localenv.LocalEnvironment, pkg *loc.Locator, runtim
...
-	// Find and report the installed teleport configuration package name
-	configPackage, err := pack.FindAnyTeleportConfigPackage(env.Packages, cluster.Domain)
-	if err != nil {
-		return trace.Wrap(err)
-	}
-	configPackageForServer := teleportConfigPackageForServer(*configPackage, serverAddr, cluster.Domain)
-	env.Println(configPackageForServer.String())
+	// Do not return a package - the currently active teleport configuration
+	// package will be used
 	return nil
```
r0mant: What's the point of this function now that all it does is just return nil without doing anything?

a-palchikov: Guess it's already late - will remove completely.

r0mant: Okay, lgtm then :)

PR #657, opened 2019-09-17
[5.x] update documentation about intermediate Kubernetes upgrades

a-palchikov:


docs/6.x/pack.md
```diff
@@ -912,3 +912,32 @@ When building a Cluster Image, `tele build` will discover `custom-planet:1.0.0`
...
 installation all nodes with the role `worker` will use the custom "planet" Docker
 image instead of the default one.
+
+## Support For Intermediate Upgrades
+
+Kubernetes has a strict version skew [support policy](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-version-skew) which makes
+cluster upgrades with versions of Kubernetes being more than a couple of minor versions apart complicated if not impossible. This can become a big issue
+if clusters are not kept constantly up-to-date.
+
+Before, the way to upgrade such a cluster had been upgrading by using cluster images with compatible Kubernetes versions in lock step
```
r0mant: Nit: IMO this paragraph reads like an excuse for this feature to exist and belongs more to the design document's "rationale" section than to the user guide. I would either try to rephrase it so it explains how upgrade works normally (i.e. you can go only from 5.0->5.2 or 5.2->5.5) or just remove it.

docs/6.x/pack.md
```diff
@@ -912,3 +912,32 @@ When building a Cluster Image, `tele build` will discover `custom-planet:1.0.0`
...
+until the target image has been applied. This is a time-consuming and not automation-friendly process.
+
+Starting with version `5.5.20`, installers built with `tele` can embed all intermediate versions of Kubernetes necessary to automatically upgrade older
```
r0mant: installers -> cluster images

docs/6.x/pack.md
```diff
@@ -912,3 +912,32 @@ When building a Cluster Image, `tele build` will discover `custom-planet:1.0.0`
...
+clusters.
+
+In order to enable intermediate upgrade support, `tele build` should be supplied with a list of Gravity runtime versions to use as intermediate hops.
```
r0mant: runtime versions -> base image versions

docs/6.x/pack.md
```diff
@@ -912,3 +912,32 @@ When building a Cluster Image, `tele build` will discover `custom-planet:1.0.0`
...
+For example, to upgrade a cluster based on Gravity `5.0.35` (which is based on Kubernetes `1.9.13`) to a cluster based on Gravity `5.5.20` (which, in turn,
+is based on Kubernetes `1.13.10`), the following flag adds an intermediate step based on Gravity `5.2.15` (which is based on Kubernetes `1.11.9`):
+
+```bash
+$ tele build ... --upgrade-via=5.2.15
+```
+
+!!! note "Intermediate Upgrade Support":
+    Only LTS versions of Gravity can be used as intermediate upgrade hops.
```
r0mant: ... "Check [Releases](link) page to see which LTS versions are available".

Also, can the "target" hop (i.e. final upgrade version) be non-LTS, or does it have to be LTS as well? Worth mentioning here I think.

docs/6.x/pack.md
```diff
@@ -912,3 +912,32 @@ When building a Cluster Image, `tele build` will discover `custom-planet:1.0.0`
 Docker image and vendor it along with other dependencies. During the Cluster
 installation all nodes with the role `worker` will use the custom "planet" Docker
 image instead of the default one.
+
+## Support For Intermediate Upgrades
```
r0mant: This is more of a nitpick but I think that "intermediate upgrades" term we coined here is kinda confusing :) Maybe say something like "Direct Upgrades From Older LTS Versions"? I.e. so a user reading the title (or searching the docs) would immediately understand what this is about. Maybe it's just me though.

docs/6.x/pack.md
```diff
@@ -912,3 +912,32 @@ When building a Cluster Image, `tele build` will discover `custom-planet:1.0.0`
...
+```
+
+!!! note "Intermediate Upgrade Support":
+    Only LTS versions of Gravity can be used as intermediate upgrade hops.
+
+
+!!! note:
+    Upgrades with support for intermediate Kubernetes runtimes is available since Gravity version `5.5.20`.
+    Gravity version `5.2.15` and later can be used as an intermediate upgrade hop.
+    The Gravity version `6.x` will receive support for intermediate upgrades in near future.
```
r0mant: I would just say "Later/newer Gravity versions will receive support..." w/o specifying concrete version here.

docs/6.x/pack.md
```diff
@@ -912,3 +912,36 @@ When building a Cluster Image, `tele build` will discover `custom-planet:1.0.0`
...
+
+Kubernetes has a strict version skew [support policy](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-version-skew) which makes
+cluster upgrades to versions of Kubernetes being more than a couple of minor versions apart complicated if not impossible, hence
+upgrading such a cluster to the version of Kubernetes that would break the version skew policy is not recommended. This can become a big issue if
+clusters are not kept up-to-date.
+
+To correctly upgrade such a cluster, it first needs to be upgraded to a cluster image with a compatible version of Kubernetes (according to the version
+skew policy). This step should be repeated until the cluster has been upgraded to the target version.
+
+For example, to upgrade a cluster based on Gravity `5.0.35` (which is based on Kubernetes `1.9.13`) to a cluster based on Gravity `5.5.20` (which
```
benarent: This is a handy paragraph, I wonder if we should also put Gravity Version & K8 version in a table for easier scanning for customers.

docs/6.x/pack.md
```diff
@@ -912,3 +912,38 @@ When building a Cluster Image, `tele build` will discover `custom-planet:1.0.0`
...
+clusters are not kept up-to-date.
+
+To correctly upgrade such a cluster, it first needs to be upgraded to a cluster image with a compatible version of Kubernetes (according to the version
+skew policy). This step should be repeated until the cluster has been upgraded to the target version.
+
+For example, to upgrade a cluster based on Gravity `5.0.35` (which is based on Kubernetes `1.9.13`) to a cluster based on Gravity `5.5.21` (which
+is based on Kubernetes `1.13.11`), the cluster needs to be upgraded to Gravity version `5.2.15` (with Kubernetes `1.11.9`) first before upgrading
+to `5.5.21`.
+
+Since version `5.5.21` `tele` is capable of producing cluster image tarballs that can upgrade such clusters directly by embedding all the required
```
knisbet: This text seems to be embedded twice, once as a doc item, and also as a note. I think I would just leave it as a note that can be removed at some point when it's no longer relevant.

docs/6.x/pack.md
```diff
@@ -912,3 +912,38 @@ When building a Cluster Image, `tele build` will discover `custom-planet:1.0.0`
...
+intermediate versions of Kubernetes.
+
+For example, the following command embeds Kubernetes runtime from Gravity version `5.2.15` which will allow direct upgrades of clusters
+based on Gravity version `5.0.x`:
+
+```bash
+$ tele build ... --upgrade-via=5.2.15
+```
+
+The flag can be specified multiple times to add as many intermediate hops as required.
```
knisbet: Nit, I think an example would be good.

docs/6.x/pack.md
```diff
@@ -912,3 +912,38 @@ When building a Cluster Image, `tele build` will discover `custom-planet:1.0.0`
 Docker image and vendor it along with other dependencies. During the Cluster
 installation all nodes with the role `worker` will use the custom "planet" Docker
 image instead of the default one.
+
+## Direct Upgrades From Older LTS Versions
+
+Kubernetes has a strict version skew [support policy](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-version-skew) which makes
+cluster upgrades to versions of Kubernetes being more than a couple of minor versions apart complicated if not impossible, hence
```
knisbet: Sorry, I think this conceptually could be introduced a bit cleaner. I'm not a huge fan of the part about kubernetes upgrades being complicated / impossible, I think that portion doesn't really add value to the docs.

I would consider starting by introducing the gravity requirement on upgrades between adjacent LTS releases (or link to where we document this, but I don't think we do), and then that can indicate its partially driven by the kubernetes version skew policy, but also that's the upgrade paths we officially test. Then I wouldn't bother with the content about what a "correct" upgrade process is, as this is a feature that can be used to bypass those requirements.



Also, questions I can envision being asked I don't think are covered:
- Can this be used with any cluster, or only with "old" clusters that need the intermediate step.
- Must the --upgrade-via be an LTS version?
- Will the upgrade take longer?
- Will the image be larger?

a-palchikov: Thanks @knisbet - I'll update!

docs/6.x/changelog.md
```diff
@@ -410,6 +410,13 @@ LTS starts with `3.51.0` with minor backwards compatible changes added over time
     This release fixes a security vulnerability in kubernetes. Please see
     [Kubernetes Announcement](https://groups.google.com/forum/#!topic/kubernetes-security-discuss/Vf31dXp0EJc) for more information.
 
+### 5.2.15 LTS (September 19th, 2019)
+
+#### Improvements
+
+* Update github.com/gravitational/monitoring-app to 5.2.5
```
r0mant: Nit: Missing dot.

docs/6.x/changelog.md
```diff
@@ -299,7 +304,19 @@ LTS starts with `3.51.0` with minor backwards compatible changes added over time
     This release fixes a security vulnerability in teleport. Please see
     [Teleport Announcement](https://github.com/gravitational/teleport/releases/tag/v4.0.5) for more information.
 
-### 5.5.20 (September 18th, 2019)
+### 5.5.21 LTS (September 26th, 2019)
+
+#### Improvements
+
```
UnderGreen: * Update github.com/gravitational/monitoring-app to 5.5.3 ([#644](https://github.com/gravitational/gravity/pull/644)).
  * Disable NodePort access to Kapacitor service.
  * Randomly generate passwords for superuser, telegraf, heapster and grafana users in InfluxDB database during installation and upgrades.

PR #683, opened 2019-09-17
Return cluster package info when shallow flag is present

### LINKED ISSUE #682: [BUG] HUB may show the error screen right after the new cluster joins
HUB Cluster screen may show the error right after the new cluster joins the HUB.

image

The issue is related to API request `shallow` flag which omits the logos and other information from cluster JSON.

This section should be populated when `shallow` is present

```
 app: {
      package: {
        repository: '',
        name: '',
        version: '',
      },
```

alex-kovoy:
fixes #682

### DIFF
--- a/lib/webapi/webapi.go
+++ b/lib/webapi/webapi.go
@@ -1490,7 +1490,9 @@ func newWebCluster(cluster ops.Site, releases []webRelease, shallow bool) webClu
 	// This significantly reduces amount of traffic b/w frontend and server and
 	// improves the web application performance.
 	if shallow {
-		webCluster.App = ops.Application{}
+		webCluster.App = ops.Application{
+			Package: webCluster.App.Package,
+		}
 		for i := range webCluster.Releases {
 			webCluster.Releases[i].ChartIcon = ""
 		}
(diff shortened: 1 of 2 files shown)


lib/webapi/webapi.go
```diff
@@ -1490,7 +1490,9 @@ func newWebCluster(cluster ops.Site, releases []webRelease, shallow bool) webClu
 	// This significantly reduces amount of traffic b/w frontend and server and
 	// improves the web application performance.
 	if shallow {
-		webCluster.App = ops.Application{}
+		webCluster.App = ops.Application{
+			Package: webCluster.App.Package,
```
r0mant: Hm, not sure what this does? `webCluster.App.Package` is empty cause you're just assigning to `webCluster.App` for the first time here?

alex-kovoy: It's not empty and we are overriding an existing `App` value with the new object. This object has all empty fields except the `Package` which we preserve.

r0mant: Ah, the `App` field comes from the embedded `Site`. Makes sense then, the variable declaration above threw me off.

PR #684, opened 2019-09-18
[5.5.x] do not drop CAP_DAC_OVERRIDE to be able to access journal files

a-palchikov:
This fixes an issue with planet journal capture in `gravity system report` resulting in:
```
No journal files were opened due to insufficient permissions.
```

### DIFF
--- a/lib/system/caps_linux.go
+++ b/lib/system/caps_linux.go
@@ -26,10 +26,14 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// DropCapabilitiesForChroot drops capabilities except those required to do chroot
-func DropCapabilitiesForChroot() error {
+// DropCapabilitiesForJournalExport drops capabilities except those required
+// to export a systemd journal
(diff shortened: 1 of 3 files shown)


lib/system/caps_linux.go
```diff
@@ -91,5 +95,12 @@ func maxCap() (cap int, err error) {
 	return strconv.Atoi(strings.TrimSpace(string(buf[:n])))
 }
 
-// capSysChroot defines the number of the linux capability to execute chroot
-const capSysChroot = 18
+const (
+	// Capability to bypass file read, write, and execute permission checks.
+	// (DAC is an abbreviation of "discretionary access control".)
+	capDACOverride = 1
```
r0mant: Are there no constants already available somewhere for these caps we could use?

a-palchikov: Not that I know of - we have an integration with an external package in planet but not in gravity.

PR #685, opened 2019-09-18
Use Debian Buster in containers

UnderGreen:


assets/telekube/resources/app.yaml
```diff
@@ -82,7 +82,7 @@ nodeProfiles:
         - name: ubuntu-core
           versions: ["16"]
         - name: debian
-          versions: ["8", "9"]
+          versions: ["8", "9", "10"]
```
r0mant: Did you actually test it on Debian 10? :) If it works, we should probably also add it to our supported distros table in the docs.

UnderGreen: Good catch. It is an artifact from another branch.

Issue #580, opened 2019-08-13
Replace "tele login" with "tsh login"


benarent:
### Objectives

- [x] `tele login` command is no longer needed to login into Hub.
- [x] `tsh login` is used to login into Hub and regular clusters.
- [x] `tele` commands (such as `build`, `ls`, `pull`, `push`) work after `tsh login`.
- [ ] `tsh login` can login users into Docker/Helm as well.

### Implementation notes

* `tele` should be taught to read user's x509 certificate from `~/.tsh` and use it for client auth when talking to Gravity API.
* Gravity web servers (pack/app/ops) should be extended to support client-cert auth and extract user information from the client certificate. Maybe use Teleport's `lib/auth/middleware.go:AuthMiddleware` for that.

PR #688, opened 2019-09-18
Provide tctl inside cluster.

r0mant:
This PR makes sure that `tctl` can be used inside gravity clusters and is pre-configured so it can be used without any configuration from user's side. Refs https://github.com/gravitational/gravity.e/issues/4198, https://github.com/gravitational/gravity/issues/580.

Also a couple of other things:

* Update some parts of "cli/system" to use "update/system" instead (specifically the "system reinstall" command) which removed a lot of duplicated code and allowed to reuse the same logic during upgrade and install.

I'm actually thinking we should just nuke the whole "cli/system" module as 90% of its commands are not used anymore and moreover they're not kept up-to-date with the logic in "update/system" so it's probably even harmful to use them. Refs https://github.com/gravitational/gravity.e/issues/4175.

* Fix the issue with helm script symlink not being updated during upgrade (which was also a result of discrepancy between cli/system and update/system). Closes https://github.com/gravitational/gravity.e/issues/4173.

* Update system uninstall to cleanup kubectl, helm and tctl links/scripts as well.


lib/update/system/system.go
```diff
@@ -356,6 +381,61 @@ func (r *System) updateTeleportPackage(update storage.PackageUpdate) (labelUpdat
...
+			r.WithField("path", path).Info("Updated tctl script.")
+			return nil
+		}
+		errors = append(errors, err)
+	}
+	return trace.NewAggregate(errors...)
+}
+
+// renderTctlScript renders the contents of the script that invokes tctl binary
+// with apporpriate configuration.
```
a-palchikov: apporpriate -> appropriate

lib/update/system/system.go
```diff
@@ -499,12 +579,23 @@ func updateKubectl(planetPath string, logger logrus.FieldLogger) (err error) {
...
 		out, err = exec.Command("ln", "-sfT", kubectlPath, path).CombinedOutput()
 		if err == nil {
+			logger.Infof("Updated kubectl symlink: %v -> %v.", path, kubectlPath)
 			break
 		}
-		logger.WithFields(logrus.Fields{
-			logrus.ErrorKey: err,
-			"output":        string(out),
-		}).Warn("Failed to update kubectl symlink.")
+		logger.WithError(err).WithField("output", string(out)).Warn(
```
a-palchikov: What if I want to add another field - should I chain it with another `WithFeild`? Each `WithField` is also potentially allocating memory.

lib/update/system/system.go
```diff
@@ -713,3 +804,8 @@ func unpack(packages update.LocalPackageService, loc loc.Locator) error {
 	}
 	return trace.Wrap(pack.Unpack(packages, loc, path, nil))
 }
+
+// tctlScript is the template of the script that invokes tctl binary with
+// configuration supplied via an environment variable.
+var tctlScript = template.Must(template.New("tctl").Parse(`#!/bin/bash
+TELEPORT_CONFIG={{.conf}} {{.path}} "$@"`))
```
a-palchikov: Will `path` be properly escaped if it contains special characters? Ditto for teleport configuration.

tool/gravity/cli/system.go
```diff
@@ -220,17 +215,17 @@ func systemHistory(env *localenv.LocalEnvironment) error {
...
 			From:   newPackage,
 			To:     newPackage,
 			Labels: labels,
 		}
-		return trace.Wrap(systemBlockingReinstall(env, update))
+		return trace.Wrap(systemBlockingReinstall(env, update, clusterRole))
 	}
 
-	args := []string{"system", "reinstall", newPackage.String()}
+	args := []string{"system", "reinstall", newPackage.String(), "--cluster-role", clusterRole}
```
a-palchikov: According to flag description it's optional.

lib/update/system/system.go
r0mant: Not necessarily, you can still use `WithFields()`, I just think `WithField()` looks cleaner when you have only 1 field.

lib/update/system/system.go
r0mant: I don't think path can contain special characters. It points to the unpacked teleport package directory inside our state dir, and if somebody overrides state dir to have special characters in it, I think they'll have bigger problems than non-working tctl script.

And the configuration is base64-encoded so it doesn't need to be escaped.

PR #716, opened 2019-09-23
[5.5.x] teleport configuration naming changes

a-palchikov:
This PR updates the naming scheme for teleport configuration packages to match planet config/secrets package naming in that the package version equals to that of the bundled package (`teleport` or `planet`) and all config updates are performed by issuing increments in semver metadata.

Updates https://github.com/gravitational/gravity.e/issues/4219.

### DIFF
--- a/lib/constants/constants.go
+++ b/lib/constants/constants.go
@@ -416,10 +416,6 @@ const (
 	// EnvGravityTeleportConfig is environment variable setting debugging mode
 	EnvGravityTeleportConfig = "GRAVITY_TELEPORT_CONFIG"
 
-	// EnvGravityTeleportVersion is the environment variable that specifies the version of
-	// the bundled teleport
-	EnvGravityTeleportVersion = "GRAVITY_TELEPORT_VERSION"
-
 	// EnvNetworkingType specifies the name of environment variable for defining networking type for kubernetes
 	EnvNetworkingType = "KUBERNETES_NETWORKING"
 
--- a/lib/constants/constants.go
+++ b/lib/constants/constants.go
@@ -416,6 +416,10 @@ const (
 	// EnvGravityTeleportConfig is environment variable setting debugging mode
 	EnvGravityTeleportConfig = "GRAVITY_TELEPORT_CONFIG"
 
+	// EnvGravityTeleportVersion is the environment variable that specifies the version of
+	// the bundled teleport
+	EnvGravityTeleportVersion = "GRAVITY_TELEPORT_VERSION"
+
 	// EnvNetworkingType specifies the name of environment variable for defining networking type for kubernetes
 	EnvNetworkingType = "KUBERNETES_NETWORKING"
 
--- a/lib/ops/opsservice/configure.go
+++ b/lib/ops/opsservice/configure.go
@@ -1462,31 +1462,6 @@ func (s *site) planetConfigPackage(node remoteServer, version string) loc.Locato
 	}
 }
 
-// FIXME: remove me
-// // serverPackages returns a list of package locators specific to the provided server
-// func (s *site) serverPackages(server *ProvisionedServer) ([]loc.Locator, error) {
-// 	masterConfigPackage := s.teleportMasterConfigPackage(server)
-// 	nodeConfigPackage := s.teleportNodeConfigPackage(server)
-// 	if err != nil {
-// 		return nil, trace.Wrap(err)
-// 	}
-// 	planetPackage, err := s.app.Manifest.RuntimePackageForProfile(server.Role)
-// 	if err != nil {
-// 		return nil, trace.Wrap(err)
-// 	}
-// 	planetSecretsPackage, err := s.planetSecretsPackage(server, planetPackage.Version)
-// 	if err != nil {
-// 		return nil, trace.Wrap(err)
-// 	}
-// 	planetConfigPackage := s.planetConfigPackage(server, planetPackage.Version)
-// 	return []loc.Locator{
-// 		masterConfigPackage,
-// 		nodeConfigPackage,
-// 		planetSecretsPackage,
-// 		planetConfigPackage,
-// 	}, nil
-// }
-
 func (s *site) addCloudConfig(config clusterconfig.Interface) (args []string) {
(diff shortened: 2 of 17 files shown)


lib/ops/opsservice/configure.go
```diff
@@ -1309,11 +1304,20 @@ func (s *site) configureTeleportNode(ctx *operationContext, masterIPs []string,
...
+}
+
+func (s *site) teleportNodeConfigPackage(node remoteServer, teleportVersion string) loc.Locator {
+	return loc.Locator{
+		Repository: s.siteRepoName(),
+		Name: fmt.Sprintf("%v-%v",
+			constants.TeleportNodeConfigPackage,
+			PackageSuffix(node, s.domainName),
+		),
+		Version: teleportVersion,
```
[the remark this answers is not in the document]
a-palchikov: Yes. The metadata is supposed to be an ever increasing counter.

lib/ops/opsservice/configure.go
```diff
@@ -1198,11 +1187,20 @@ func toObject(in interface{}) (map[string]interface{}, error) {
...
-	configPackage, err := loc.ParseLocator(
-		fmt.Sprintf("%v/%v:0.0.%v-%v", s.siteRepoName(), constants.TeleportMasterConfigPackage,
-			time.Now().UTC().Unix(), PackageSuffix(master, s.domainName)))
-	return configPackage, trace.Wrap(err)
+func (s *site) teleportNextMasterConfigPackage(master remoteServer, teleportVersion string) loc.Locator {
+	teleportVersion = fmt.Sprintf("%v+%v", teleportVersion, time.Now().UTC().Unix())
+	return s.teleportMasterConfigPackage(master, teleportVersion)
+}
+
+func (s *site) teleportMasterConfigPackage(master remoteServer, teleportVersion string) loc.Locator {
```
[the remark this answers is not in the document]
a-palchikov: I'm going to move these to a new file for starters.

lib/constants/constants.go
```diff
@@ -416,6 +416,10 @@ const (
 	// EnvGravityTeleportConfig is environment variable setting debugging mode
 	EnvGravityTeleportConfig = "GRAVITY_TELEPORT_CONFIG"
 
+	// EnvGravityTeleportVersion is the environment variable that specifies the version of
+	// the bundled teleport
+	EnvGravityTeleportVersion = "GRAVITY_TELEPORT_VERSION"
```
r0mant: I don't think this is used anymore?

lib/ops/opsservice/packages.go
```diff
@@ -0,0 +1,143 @@
+package opsservice
```
r0mant: License header missing.

lib/ops/opsservice/packages.go
```diff
@@ -0,0 +1,143 @@
...
+			PackageSuffix(node, clusterName),
+		),
+		Version: teleportVersion,
+	}
+}
+
+// suffixer replaces characters unacceptable as a package suffix
+var suffixer = strings.NewReplacer(".", "", ":", "")
+
+func PackageSuffix(node remoteServer, domain string) string {
```
r0mant: Nit: Exported function w/o godoc.

lib/ops/opsservice/packages.go
```diff
@@ -0,0 +1,156 @@
...
+		Name: fmt.Sprintf("%v-%v",
+			constants.TeleportNodeConfigPackage,
+			PackageSuffix(node, clusterName),
+		),
+		Version: teleportVersion,
+	}
+}
+
+// suffixer replaces characters unacceptable as a package suffix
+var suffixer = strings.NewReplacer(".", "", ":", "")
```
knisbet: Nit: is it more common to simply remove the character, or replace with an `_` or other substitute character?

lib/ops/opsservice/shrink.go
```diff
@@ -635,3 +658,25 @@ func (s *site) serfNodeLeave(runner *serverRunner) error {
...
+	return configPackage.Name == loc.Name && configPackage.Repository == loc.Repository
+}
+
+func (s *site) isTeleportNodeConfigPackageFor(server *ProvisionedServer, loc loc.Locator) bool {
+	configPackage := s.teleportNodeConfigPackage(server)
+	return configPackage.Name == loc.Name && configPackage.Repository == loc.Repository
+}
+
+func (s *site) isPlanetConfigPackageFor(server *ProvisionedServer, loc loc.Locator) bool {
+	// Version omitted on purpose since only repository/name are used for comparison
```
knisbet: nit: comment doesn't indicate why the version is ommitted from comparisons.

second nit: The comment is included on isPlanet... but not isTeleport... even though they appear to be the same?

lib/ops/opsservice/packages.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+// planetSecretsNextPackage generates a new planet secrets package name for the specified
+// node and planet package version.
+//
+// The package is named as '<cluster-name>/planet-<node-addr>-secrets:<planet-version>(+<increment>)?'
+// where increment is an ever-increasing counter and node-addr is a combination of the node address
+// and cluster name
+func (s *site) planetSecretsNextPackage(node *ProvisionedServer, planetVersion string) loc.Locator {
+	planetVersion = fmt.Sprintf("%v+%v", planetVersion, time.Now().UTC().Unix())
```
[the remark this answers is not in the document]
a-palchikov: Can you elaborate on clock moving backwards?

lib/ops/opsservice/packages.go
a-palchikov: I don't know which is more common (in what context), but I'd say that replacing with an underscore is less destructive but both approaches will fall short if reversibility is desired.
At least in package names, I see no benefit in replacing with underscore vs removing.

lib/ops/opsservice/shrink.go
a-palchikov: I think it does indicate why the version is not used - since only repository name/package name are relevant.
And teleport packages do not take the version parameter - hence no comment.

PR #718, opened 2019-09-24
Add ability for tele to accept credentials on CLI.

r0mant:
Supporting changes for https://github.com/gravitational/gravity.e/pull/4220.

This PR extends CredentialsService to support "static" credentials - i.e. a preconfigured credentials entry, for example set on the CLI. This is needed for ability to provide credentials to tele commands in the enterprise version, see PR linked above.

Refs https://github.com/gravitational/gravity.e/issues/4198.

### DIFF
--- a/lib/builder/builder.go
+++ b/lib/builder/builder.go
@@ -72,8 +72,6 @@ type Config struct {
 	Overwrite bool
 	// Repository represents the source package repository
 	Repository string
-	// StaticCredentials is the credentials set on the CLI
-	StaticCredentials *credentials.Credentials
 	// SkipVersionCheck allows to skip tele/runtime compatibility check
 	SkipVersionCheck bool
 	// VendorReq combines vendoring options
@@ -84,8 +82,10 @@ type Config struct {
 	NewSyncer NewSyncerFunc
 	// GetRepository is a function that returns package source repository
 	GetRepository GetRepositoryFunc
-	// Credentials provides access to user credentials
-	Credentials credentials.Service
+	// CredentialsService provides access to user credentials
+	CredentialsService credentials.Service
+	// Credentials is the credentials set on the CLI
+	Credentials *credentials.Credentials
 	// FieldLogger is used for logging
 	logrus.FieldLogger
 	// Progress allows builder to report build progress
@@ -129,10 +129,10 @@ func (c *Config) CheckAndSetDefaults() error {
 	if c.GetRepository == nil {
 		c.GetRepository = getRepository
 	}
-	if c.Credentials == nil {
-		c.Credentials, err = credentials.New(credentials.Config{
+	if c.CredentialsService == nil {
+		c.CredentialsService, err = credentials.New(credentials.Config{
 			LocalKeyStoreDir: c.StateDir,
-			Static:           c.StaticCredentials,
+			Credentials:      c.Credentials,
 		})
 		if err != nil {
 			return trace.Wrap(err)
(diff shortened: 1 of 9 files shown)


lib/localenv/credentials/credentials.go
```diff
@@ -75,6 +75,18 @@ func credentialsFromEntry(entry users.LoginEntry) *Credentials {
 	}
 }
 
+// FromTokenAndHub creates new credentials from the provided token and address.
+func FromTokenAndHub(token, hub string) *Credentials {
```
a-palchikov: Will this work if only one of the parameters set?

lib/builder/builder.go
```diff
@@ -72,6 +72,8 @@ type Config struct {
 	Overwrite bool
 	// Repository represents the source package repository
 	Repository string
+	// StaticCredentials is the credentials set on the CLI
+	StaticCredentials *credentials.Credentials
```
a-palchikov: nit: Why `static` and not just `credenials`?

lib/localenv/credentials/credentials.go
```diff
@@ -134,6 +148,13 @@ func (s *credentialsService) For(clusterURL string) (*Credentials, error) {
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
+	// If the preconfigured credentials are set, try to use them.
+	if s.Static != nil && utils.StringInSlice([]string{url.normalized, url.original}, s.Static.URL) {
```
a-palchikov: `Static` does hint at credentials at all tbh.

lib/localenv/credentials/credentials.go
r0mant: Yeah, you can for example set only `--hub` and it will search for its credentials in regular key stores. You can't only set `--token` though - the CLI will ask to set `--hub` as well.

lib/localenv/credentials/credentials.go
```diff
@@ -165,7 +187,7 @@ func (s *credentialsService) For(clusterURL string) (*Credentials, error) {
 		s.Debugf("Returning default credentials for %v.", clusterURL)
 		return defaultCredentials, nil
 	}
-	return nil, trace.NotFound("no credentials for %v", clusterURL)
+	return nil, trace.AccessDenied("no credentials for %v", clusterURL)
```
knisbet: Nit: I sort of feel like the original NotFound is a more appropriate error here. It's not a I tried to access something so I was denied access, I tried to locate a set of credentials in various locations and didn't find any??

r0mant: Yeah, the problem with returning NotFound errors everywhere is that it becomes impossible to distinguish between different cases which the caller has to act differently upon (e.g. "no currently active cluster" vs "no credentials for specified cluster") so I felt like access denied would fit here and also allow to separate the two scenarios.

knisbet: So my hesitation with something like this, is it creates a silent contract between the caller/callee, on what a NotFound / AccessDenied errors don't really mean AccessDenied/NotFound, but are returned in a specific way to convey some other context to the caller.

Where as I believe the current go idomatic pattern (not the go2 stuff), would be to have Is functions that allowed the caller to ask what type of error it got (like IsTimeout, IsPermission, IsExist in the os package) and then the caller can use that behavior accordingly, as an explicit contract between packages.

Alternatively, I added the ability to embed metadata in the errors, as custom fields, and an envisioned use case around this would be for embedding additional metadata that a caller could use, although we should likely settle on some conventions around that so we're consistent. But it's now possible to add kvp fields to trace errors, and read those from a caller to have the caller make choices.

As a very rough example, but ideally we would converge on a common pattern:
```
const ErrNoCurrentActiveCluster = "no_current_active_cluster"

func E() error {
     return trace.NotFound("no credentials found").AddFields({
        "cluster": clusterURL,
        ErrNoCurrentActiveCluster: true,
     })
}

func IsErrNoCurrentActiveCluster(err trace.TraceError) bool {
    _, ok := trace.GetFields(err)[ErrNoCurrentActiveCluster];
    return ok
}

// some other package
Func D() {
    err := package.E()
    if err != nil && package.IsErrNoCurrentActiveCluster(err) {
        // Do something
    }
}
```

Or we could go the other way, and make a convention for an error field of "sub_type", and embed the error as a string.

PR #739, opened 2019-09-30
support loading upstream DNS servers when systemd-resolved stub resol…

knisbet:
…… (#627)

* support loading upstream DNS servers when systemd-resolved stub resolver is enabled

* use better variable name for upstream servers

* deduplicate upstream resolvers

* handle nil ResolvConf

### DIFF
--- a/lib/install/phases/coredns.go
+++ b/lib/install/phases/coredns.go
@@ -94,16 +94,15 @@ func (r *corednsExecutor) Execute(ctx context.Context) error {
 		return trace.Wrap(err)
 	}
 
-	// Filter out local nameservers to avoid CoreDNS forwarding requests
-	// to itself and triggering loop detection, see for more details:
-	// https://github.com/coredns/coredns/tree/master/plugin/loop#troubleshooting
-	var upstreams []string
-	for _, nameserver := range resolvConf.Servers {
-		if !utils.IsLocalhost(nameserver) {
-			upstreams = append(upstreams, nameserver)
-		}
+	// Optionally try and load upstream nameservers from systemd-resolved by reading the compatibility resolv.conf
+	// More Info: https://github.com/gravitational/gravity/issues/606#issuecomment-529171440
+	// TODO(knisbet) is there a better way to pull upstream resolvers directly from systemd?
+	systemdResolvConf, err := systeminfo.ResolvFromFile("/run/systemd/resolve/resolv.conf")
+	if err != nil && !trace.IsNotFound(err) {
+		return trace.Wrap(err)
 	}
 
+	upstreams := mergeUpstreamResolvers(resolvConf, systemdResolvConf)
 	conf, err := GenerateCorefile(CorednsConfig{
 		UpstreamNameservers: upstreams,
 		Rotate:              resolvConf.Rotate,
@@ -128,6 +127,28 @@ func (r *corednsExecutor) Execute(ctx context.Context) error {
 	}
 
 	return nil
+}
+
+func mergeUpstreamResolvers(configs ...*storage.ResolvConf) []string {
+	var upstreams []string
+	dedup := make(map[string]bool)
+	for _, config := range configs {
+		if config != nil {
+			for _, nameserver := range config.Servers {
+				if _, ok := dedup[nameserver]; !ok {
+					// Filter out local nameservers to avoid CoreDNS forwarding requests
+					// to itself and triggering loop detection, see for more details:
+					// https://github.com/coredns/coredns/tree/master/plugin/loop#troubleshooting
+					if !utils.IsLocalhost(nameserver) {
+						dedup[nameserver] = true
+						upstreams = append(upstreams, nameserver)
+					}
+				}
+			}
+		}
+	}
+
+	return upstreams
 }
 
 // Rollback deletes the coredns configmap that was created in the execute step
--- a/lib/install/phases/coredns_test.go
+++ b/lib/install/phases/coredns_test.go
@@ -19,6 +19,7 @@ package phases
 import (
 	"testing"
 
+	"github.com/gravitational/gravity/lib/storage"
 	"gopkg.in/check.v1"
 )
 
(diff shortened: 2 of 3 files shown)


lib/install/phases/coredns.go
```diff
@@ -94,16 +94,15 @@ func (r *corednsExecutor) Execute(ctx context.Context) error {
...
-	// to itself and triggering loop detection, see for more details:
-	// https://github.com/coredns/coredns/tree/master/plugin/loop#troubleshooting
-	var upstreams []string
-	for _, nameserver := range resolvConf.Servers {
-		if !utils.IsLocalhost(nameserver) {
-			upstreams = append(upstreams, nameserver)
-		}
+	// Optionally try and load upstream nameservers from systemd-resolved by reading the compatibility resolv.conf
+	// More Info: https://github.com/gravitational/gravity/issues/606#issuecomment-529171440
+	// TODO(knisbet) is there a better way to pull upstream resolvers directly from systemd?
```
a-palchikov: I still have doubts about the validity of reading from this source directly.
Stating the documentation:
```
Note that /run/systemd/resolve/resolv.conf should not be used directly by applications, but only through a symlink from /etc/resolv.conf.
```
Could it be that `/run/systemd/resolve/resolv.conf` contains stale information not directly related to the used DNS nameservers in case of `/etc/resolv.conf` not being provided by systemd - i.e. not being a symlink from `/run/systemd/resolve/resolv.conf` and hence potentially managed by another package?
So, if we read from it - maybe make sure that's it has been made available by systemd (by checking whether it's a symlink)?

lib/install/phases/coredns.go
```diff
@@ -130,6 +129,28 @@ func (r *corednsExecutor) Execute(ctx context.Context) error {
 	return nil
 }
 
+func mergeUpstreamResolvers(configs ...*storage.ResolvConf) []string {
+	var upstreams []string
+	dedup := make(map[string]bool)
+	for _, config := range configs {
+		if config != nil {
```
a-palchikov: if you handle the `nil` case first, the code would not have to be so far to the right:
```go
if config == nil {
	continue
}
```

lib/install/phases/coredns.go
knisbet: Well, I'm not sure how to address these doubts, as I don't see a particularily cleaner path forward.

Let's remind ourselves what the issue really is. 
- Hosts using the systemd-resolved service, has a local stub resolver within systemd available on dbus, for doing local DNS routing and caching.
- For comparability, this resolver is bound to 127.0.0.53 for DNS protocol clients to access the systemd resolver using the DNS protocol. Localhost addresses are not reachable from network namespaces.
- Systems in this configuration, the /etc/resolv.conf file is a symlink to the systemd-resolved configuration depending on the mode, but in our problematic case pointing to the 127.0.0.53 stub resolver and not the actual upstream servers.

According to https://www.freedesktop.org/software/systemd/man/systemd-resolved.service.html systemd-resolved maintains a compatibility resolv.conf (`/run/systemd/resolve/resolv.conf`) which lists all upstream DNS servers known by systemd-resolved, but due to the format, misses the contextual information for per-interface bindings.


> Note that /run/systemd/resolve/resolv.conf should not be used directly by applications, but only through a symlink from /etc/resolv.conf.

My guess at the meaning of this warning, is that `/etc/resolv.conf` can be pointed at a number of configurations, and as such `/run/systemd/resolve/resolv.conf` may mismatch the behavior of normal resolution by say libc. I'm not sure it's meant to convey, if you read the DNS configuration from multiple sources, or want to read the DNS servers that systemd-resolved knows about, it's unsafe to do so by reading this file location. 


> Could it be that /run/systemd/resolve/resolv.conf contains stale information not directly related to the used DNS nameservers in case of /etc/resolv.conf not being provided by systemd - i.e. not being a symlink from /run/systemd/resolve/resolv.conf

I have no idea. I suppose it's possible that a system was originally configured with systemd-resolved, and then in an attempt to disable it, it's still running, but differs from the resolv.conf, which allows most functionality to work normally that's based on a libc resolver, and anything using the systemd resolver is broken. I guess I don't know at what point we draw the line on being unable to read/interpret the systems configuration.

As a reminder, the reason we're reading `/run/systemd/resolve/resolv.conf` is specifically because the system configuration doesn't point to it. The system configuration points to `/run/systemd/resolve/stub-resolv.conf` which we can't use, because it hardcodes 127.0.0.53 as the resolver to force the use of the stub resolver. This address isn't available to network namespaces.

> and hence potentially managed by another package?

I'm not sure I follow, is the suggestion that some other process on the machine may write to `/run/systemd/resolve/resolv.conf` that isn't a resolv.conf file or contains specifically incorrect upstream DNS servers?


> So, if we read from it - maybe make sure that's it has been made available by systemd (by checking whether it's a symlink)?

Again, /run/systemd/resolve/resolv.conf isn't the symlink, and isn't the destination of the symlink. It's a way to read from the systemd resolved what it knows about upstream DNS servers, since some short searches didn't uncover a way to do this. The only alternative method I saw would be to parse the cli output, which risks differences between versions, vs a format like resolv.conf which is atleast somewhat standardized. There is likely a dbus method somewhere for this that the cli uses, but I didn't see a relatively clean way to do this on my original search.


So I don't know where this leaves us. 
- We can't use the stub resolver, it binds to a loopback address not reachable from kubernetes pods
- We need the ability to read the DNS servers configured on the system. 

So what method do we have that's better?

a-palchikov: I did not mean ` /run/systemd/resolve/resolv.conf` to be the destination of the symlink, but if I understand systemd [behavior](https://github.com/systemd/systemd/blob/v239/src/resolve/resolved-resolv-conf.c) correctly, when systemd "manages" `/etc/resolv.conf` - it is a symlink to one of (`/run/systemd/resolve/resolv.conf`, `/run/systemd/resolve/stub-resolv.conf` or `/resolv.conf`) - hence was my question whether we can avoid interpreting the uplink file if `/etc/resolv.conf` is not a symlink.

knisbet: I don't know. 

My understanding of that scenario, is it means some services would still be using systemd-resolved (anything using dbus as a resolver), and libc like resolvers would be using the /etc/resolv.conf for resolution. So the system would be in a fairly particular miss-configuration, where the DNS servers read from `/run/systemd/resolve/resolv.conf` are still partially being used by the system, but possibly in a way that isn't noticeable? Since I think systemd-resolved is relatively new, it may not be super common vs the libc resolver at present, but I have no idea if that would continue to be the case.

PR #744, opened 2019-10-01
[6.1.x] configurable HTTP timeout for monitoring checks.

a-palchikov:
Updates https://github.com/gravitational/gravity.e/issues/4222.


Gopkg.toml
```diff
@@ -65,15 +65,19 @@ ignored = [
 
 [[constraint]]
   name = "github.com/gravitational/satellite"
-  version = "=5.5.4"
+  version = "=6.1.1"
+
+[[constraint]]
+  name = "github.com/miekg/dns"
```
a-palchikov: That's an unfortunate consequence of satellite not pinning the version of `miekg/dns` - will need to fix that in another update.

lib/testutils/auth.go
```diff
@@ -405,3 +405,6 @@ func (c *AuthClient) UploadSessionRecording(r events.SessionRecording) error {
 func (c *AuthClient) GetClusterCACert() (*auth.LocalCAResponse, error) {
 	return nil, nil
 }
+func (c *AuthClient) GenerateUserCerts(key []byte, user string, ttl time.Duration, compatibility string) (ssh []byte, tls []byte, err error) {
```
a-palchikov: For some reason, the [revision](https://github.com/gravitational/teleport/commit/0fd17) this branch was referring to had been deleted (probably with the sub-branch or something similar) hence I had to update the teleport's revision as well which introduced this change.

PR #746, opened 2019-10-02
[6.1.x] Provide all error context upon operation failure

a-palchikov:
* Fix a couple of typos.
 * Return precondition failed errors for initialization failures in service for better visibility in client.
 * Return all error context upon operation failure to the client.

Updates https://github.com/gravitational/gravity.e/issues/4205.
Requires https://github.com/gravitational/gravity.e/pull/4226.

### DIFF
--- a/lib/install/utils.go
+++ b/lib/install/utils.go
@@ -17,6 +17,7 @@ package install
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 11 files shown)


lib/install/utils.go
```diff
@@ -62,6 +63,7 @@ func GetAppPackage(service app.Applications) (*loc.Locator, error) {
 func GetApp(service app.Applications) (*app.Application, error) {
 	apps, err := service.ListApps(app.ListAppsRequest{
 		Repository: defaults.SystemAccountOrg,
+		Type:       storage.AppUser,
```
a-palchikov: This ensures that we don't use an unrelated application package if the user application is actually missing from the tarball.

lib/rpc/server/peers.go
```diff
@@ -109,7 +109,7 @@ func (r *peers) tryPeer(ctx context.Context, peer *peer) error {
 }
 
 func (r *peers) monitorPeers() {
-	log := r.WithField("health.checker", r)
+	log := r.WithField("health.checker", r.String())
```
a-palchikov: This fixes `Failed to fire hook: json: unsupported type: func() string` in service logs.

PR #751, opened 2019-10-03
[5.5.x] fix node mismatch when using encrypted installer

a-palchikov:
* Do not count a peer re-sending the join request as a new node.
 * Use specific formatting implementation instead of generic dump based on spew for readability.

Updates https://github.com/gravitational/gravity.e/issues/4227.

### DIFF
--- a/lib/install/flow.go
+++ b/lib/install/flow.go
@@ -440,7 +440,7 @@ func (i *Installer) waitForAgents() error {
 			}
 			report, err := i.Operator.GetSiteInstallOperationAgentReport(i.OperationKey)
 			if err != nil {
-				log.Warningf("Failed to get agent report: %v.", err)
+				log.Warnf("Failed to get agent report: %v.", trace.UserMessage(err))
 				continue
 			}
 			if !i.canContinue(report) {
--- a/lib/ops/opsservice/agents.go
+++ b/lib/ops/opsservice/agents.go
@@ -308,7 +308,8 @@ func NewAgentPeerStore(backend storage.Backend, users users.Users,
 
 // NewPeer adds a new peer
 func (r *AgentPeerStore) NewPeer(ctx netcontext.Context, req pb.PeerJoinRequest, peer rpcserver.Peer) error {
-	r.Infof("NewPeer(%v).", peer.Addr())
+	logger := r.WithField("peer", peer.Addr())
+	logger.Info("NewPeer.")
 
 	token, user, err := r.authenticatePeer(req.Config.Token)
 	if err != nil {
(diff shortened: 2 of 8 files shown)


lib/install/flow.go
```diff
@@ -440,7 +440,7 @@ func (i *Installer) waitForAgents() error {
 			}
 			report, err := i.Operator.GetSiteInstallOperationAgentReport(i.OperationKey)
 			if err != nil {
-				log.Warningf("Failed to get agent report: %v.", err)
+				log.Warnf("Failed to get agent report: %v.", trace.UserMessage(err))
```
a-palchikov: Otherwise (due to unexpected behavior of trace for error traces that traveled over the wire), the output is less than readable.

lib/update/engine.go
```diff
@@ -126,13 +126,6 @@ func (r *Engine) Complete(fsmErr error) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	err = r.operator.ActivateSite(ops.ActivateSiteRequest{
```
a-palchikov: This and below  should not be necessary as CompleteOperation/FailOperation transition the cluster to specified state.

lib/ops/opsservice/agents.go
```diff
@@ -394,6 +396,10 @@ func (r *AgentPeerStore) authenticatePeer(token string) (*storage.ProvisioningTo
 
 func (r *AgentPeerStore) validatePeer(ctx context.Context, group *agentGroup, info storage.System,
 	req pb.PeerJoinRequest, clusterName string) error {
+	if group.hasPeer(req.Addr, info.GetHostname()) {
```
r0mant: Hmm, I'm not sure about this. Let's say I have two VMs with the same hostname (say, `node-1`) and the same internal IP, which is not uncommon (e.g. all my local VMs have IP `10.0.2.15`). With this change they will both be allowed to join the installer?

a-palchikov: In this case we need another differentiator apart from the tuple (address, hostname) as this is not going to work regardless of this change. What this is fixing is the race between the usual agent reconnect and the test connection that the agent sends at the very beginning (which is the difference between 5.5.x and 5.0.x) to verify the possibility of the reconnect.
Besides, an agent can drop off any time (without triggering `PeerLeave` - i.e. due to temporary networking issue) and then reconnect (effectively resending the `PeerJoin`) so it is supposed to be idempotent.

a-palchikov: Another question regarding your VM setup - how do set up routing between the VMs or VM and the host on that IP if it's the same for all of them?

r0mant: My VMs also have another host-only adapter for VM-to-VM communication but that was just an example - I was mostly wondering how we could detect similar situation and prevent a user from making such mistake? This is definitely an edge case though.

Just as an idea, if (ip, hostname) is not enough, could we add a mac address too? I think it can be spoofed as well but it has to be intentional and would protect against mistakes. This might be out of scope of this particular issue though.

PR #753, opened 2019-10-03
Mark Role as Enterprise Only

knisbet:
It looks like role is an enterprise only resource, but was missing in the docs as Enterprise only.

https://community.gravitational.com/t/unable-to-create-roles-with-6-x/378

### DIFF
--- a/docs/6.x/config.md
+++ b/docs/6.x/config.md
@@ -10,11 +10,11 @@ Resource Name             | Resource Description
 `github`                  | GitHub connector
 `oidc`                    | OIDC connector (Enterprise only)
 `saml`                    | SAML connector (Enterprise only)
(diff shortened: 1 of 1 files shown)


docs/6.x/config.md
```diff
@@ -10,7 +10,7 @@ Resource Name             | Resource Description
 `github`                  | GitHub connector
 `oidc`                    | OIDC connector (Enterprise only)
 `saml`                    | SAML connector (Enterprise only)
-`role`                    | Cluster role
+`role`                    | Cluster role (Enterprise only)
```
r0mant: Nit: "User role" or "Cluster user role" would be a bit more accurate? Also, since you're here anyway, could you please fix inconsistent capitalization of first words in the descriptions in this table? :)

Issue #647, opened 2019-09-13
Add ability to define default cloud-provider in app.yaml


adamlamar:
In `app.yaml`, the `providers` section allows setting options for the various cloud provider implementations. By default, the provider is autodetected.

Although I'm deploying on AWS for testing, I don't want AWS APIs to be used when building the cluster because my assumption is the image will be deployed on bare metal. To achieve this today, I can provide the `--cloud-provider=generic` flag to `gravity install`, but if this flag is forgotten when running on an ec2 instance without any roles defined, the installation fails.

Ideally I would like to be able to set a key in `app.yaml` somewhere under `providers` that would override autodetection and force `--cloud-provider=generic` for my cluster image.

adamlamar:
FYI I also ran into this issue when running `gravity join`. When running the provided join command with the join token, the join will fail mysteriously due to a timeout unless `--cloud-provider=generic` is provided.

PR #755, opened 2019-10-03
Add ability to specify default provider.

r0mant:
* Add a new manifest field `.providers.default` which specifies the provider to use if none is supplied on the command-line. If the default provider is not specified, the old auto-detect logic kicks in.
* Deprecate `--cloud-provider` flag for the join command. It will now query the provider from the cluster/installer. The flag is still there for backwards compatibility but is ignored.

Closes https://github.com/gravitational/gravity/issues/647.

### DIFF
--- a/tool/gravity/cli/config.go
+++ b/tool/gravity/cli/config.go
@@ -564,7 +564,7 @@ func (i *InstallConfig) validateDNSConfig() error {
 }
 
 func (i *InstallConfig) validateCloudConfig() (err error) {
-	i.CloudProvider, err = validateOrDetectCloudProvider(i.CloudProvider)
+	i.CloudProvider, err = i.validateOrDetectCloudProvider(i.CloudProvider)
 	if err != nil {
 		return trace.Wrap(err)
 	}
@@ -671,7 +671,6 @@ func NewJoinConfig(g *Application) JoinConfig {
 		SystemDevice:  *g.JoinCmd.SystemDevice,
 		DockerDevice:  *g.JoinCmd.DockerDevice,
 		Mounts:        *g.JoinCmd.Mounts,
-		CloudProvider: *g.JoinCmd.CloudProvider,
 		OperationID:   *g.JoinCmd.OperationID,
 		FromService:   *g.JoinCmd.FromService,
 	}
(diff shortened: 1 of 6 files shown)


tool/gravity/cli/config.go
```diff
@@ -851,9 +836,8 @@ func (r *agentConfig) checkAndSetDefaults() (err error) {
 	if r.token == "" {
 		return trace.BadParameter("token is required")
 	}
-	r.cloudProvider, err = validateOrDetectCloudProvider(r.cloudProvider)
-	if err != nil {
-		return trace.Wrap(err)
+	if r.cloudProvider == "" {
```
r0mant: As far as I looked, the `gravity ops agent` command that uses this config is only called by shrink operation which always sets this flag to the cluster's provider. So it should be fine to require the flag here - didn't want to over-complicate this logic for retrieving app manifest (to check default provider) if it's used only in shrink.

tool/gravity/cli/config.go
```diff
@@ -564,7 +564,7 @@ func (i *InstallConfig) validateDNSConfig() error {
 }
 
 func (i *InstallConfig) validateCloudConfig() (err error) {
```
a-palchikov: I think this call can go to `InstallConfig.NewInstallerConfig` so we have a consistent place for validate-and-set (the role is validated and set in `NewInstallerConfig`) or both need to be in `CheckAndSetDefaults`.

tool/gravity/cli/config.go
```diff
@@ -736,19 +726,14 @@ func (j *JoinConfig) GetPeers() ([]string, error) {
...
-	config := &proto.RuntimeConfig{
+func (j *JoinConfig) GetRuntimeConfig() proto.RuntimeConfig {
+	return proto.RuntimeConfig{
 		Token:        j.Token,
 		Role:         j.Role,
 		SystemDevice: j.SystemDevice,
 		DockerDevice: j.DockerDevice,
 		Mounts:       convertMounts(j.Mounts),
 	}
-	err := install.FetchCloudMetadata(j.CloudProvider, config)
```
a-palchikov: Why is this being removed?

r0mant: I found that this is already being done by the `install.NewAgent()` function which join peer calls so it was extra work. Moreover, cloud provider is actually unknown at this particular point b/c join peer retrieves it from the cluster now (which is done after connecting to wizard or cluster inside the peer).

tool/gravity/cli/config.go
r0mant: Yeah, it's a mess. I'll see if I can unify this somehow.

PR #762, opened 2019-10-07
[5.5.x] update buildbox to fetch pinned versions of GRPC-dependencies

a-palchikov:
Updates https://github.com/gravitational/gravity.e/issues/4216.


Gopkg.toml
```diff
@@ -72,7 +72,7 @@ ignored = [
 
 [[constraint]]
   name = "github.com/gogo/protobuf"
-  version = "=v1.1.0"
+  version = "=v1.3.0"
```
a-palchikov: To go in hand with updated versions of GRPC-dependencies.

lib/network/validation/proto/Makefile
```diff
@@ -8,4 +8,4 @@ all: $(IDL)
 	protoc -I=. -I=$$PROTO_INCLUDE \
 		-I=/gopath/src/github.com/gravitational/gravity/vendor/github.com/gravitational/satellite/agent/proto \
 		$^ \
-		--gofast_out=plugins=grpc,$(deps):. 
+		--gogo_out=plugins=grpc,$(deps):.
```
a-palchikov: `gofast` seems to generate a reference to golang protobuf package while we're [using](https://github.com/gravitational/gravity/blob/dmitri/5.5.x/4216-grpc-reproducible-build/Gopkg.toml#L74) the one from gogo repository (and so does the vendored k8s.io).

PR #775, opened 2019-10-08
[5.5.x] mark intermediate packages correctly

a-palchikov:
Use the actual runtime application for collecting dependencies instead of wrapping it into a host application manifest to avoid marking unrelated dependencies as related to a specific intermediate step.

Fixes https://github.com/gravitational/gravity.e/issues/4231.


lib/builder/builder.go
```diff
@@ -260,10 +260,7 @@ func (b *Builder) SyncPackageCache(ctx context.Context, runtimeVersion semver.Ve
 	for _, runtimeVersion := range append([]semver.Version{runtimeVersion}, intermediateVersions...) {
 		b.NextStep("Syncing packages for %v", runtimeVersion)
 		if err := b.syncPackageCache(ctx, runtimeVersion, syncer, apps); err != nil {
-			if trace.IsNotFound(err) {
-				return trace.NotFound("runtime version %v not found", runtimeVersion)
-			}
-			return trace.Wrap(err)
+			return trace.Wrap(err, "failed to sync packages for runtime version %v", runtimeVersion)
```
a-palchikov: This also fixes an issue with shadowing package errors from the user using `tele build`.

lib/builder/builder.go
```diff
@@ -513,20 +508,21 @@ func (b *Builder) collectUpgradeDependencies() (result *libapp.Dependencies, err
 		return nil, trace.Wrap(err)
 	}
 	result = &libapp.Dependencies{}
-	for _, version := range b.UpgradeVia {
+	for _, runtimeVersion := range b.UpgradeVia {
+		app, err := apps.GetApp(loc.Runtime.WithVersion(runtimeVersion))
```
a-palchikov: Cannot use the host application to avoid pulling in and marking unrelated packages are part of this (`runtimeVersion`) intermediate step.

PR #777, opened 2019-10-09
[5.5.x] do not treat context expiration as terminal when validating agent connection

a-palchikov:
* Make agent connect timeout configurable via environment (`GRAVITY_AGENT_CONNECT_TIMEOUT`).
 * Ignore deadline exceeded errors when validating connection. As the original motivation for validating the connection to the peer (server) has been early detecting of terminal conditions, connection timeout should not be considered a terminal condition and the agent should be allowed to continue even if connection takes longer than the originally allotted time.

Requires https://github.com/gravitational/gravity.e/pull/4235.
Updates https://github.com/gravitational/gravity.e/issues/4229.

### DIFF
--- a/lib/defaults/defaults.go
+++ b/lib/defaults/defaults.go
@@ -213,6 +213,10 @@ const (
 	// PeerConnectTimeout is the timeout of an RPC agent connecting to its peer
 	PeerConnectTimeout = 10 * time.Second
 
+	// EnvPeerConnectTimeout is the environment variable that overrides the value of the
+	// agent timeout for the validating connect
+	EnvPeerConnectTimeout = "GRAVITY_AGENT_CONNECT_TIMEOUT"
+
 	// GravityPackagePrefix defines base prefix of gravity package
 	GravityPackagePrefix = "gravitational.io/gravity"
 
--- a/lib/defaults/defaults.go
+++ b/lib/defaults/defaults.go
@@ -215,7 +215,7 @@ const (
 
 	// EnvPeerConnectTimeout is the environment variable that overrides the value of the
 	// agent timeout for the validating connect
-	EnvPeerConnectTimeout = "GRAVITY_AGENT_CONNECT_TIMEOUT"
+	EnvPeerConnectTimeout = "GRAVITY_PEER_CONNECT_TIMEOUT"
 
 	// GravityPackagePrefix defines base prefix of gravity package
 	GravityPackagePrefix = "gravitational.io/gravity"
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -495,11 +495,9 @@ func (p *Peer) getAgent(opCtx operationContext) (*rpcserver.PeerServer, error) {
 		return nil, trace.Wrap(err)
 	}
 	// make sure that connection to the RPC server can be established
-	ctx, cancel := context.WithTimeout(p.Context, defaults.PeerConnectTimeout)
-	defer cancel()
-	err = agent.ValidateConnection(ctx)
+	err = validateConnection(p.Context, agent)
 	if err != nil {
-		return agent, trace.Wrap(err)
+		return nil, trace.Wrap(err)
 	}
 	return agent, nil
 }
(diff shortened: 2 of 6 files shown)


lib/defaults/defaults.go
```diff
@@ -213,6 +213,10 @@ const (
 	// PeerConnectTimeout is the timeout of an RPC agent connecting to its peer
 	PeerConnectTimeout = 10 * time.Second
 
+	// EnvPeerConnectTimeout is the environment variable that overrides the value of the
+	// agent timeout for the validating connect
+	EnvPeerConnectTimeout = "GRAVITY_AGENT_CONNECT_TIMEOUT"
```
r0mant: Nit: Use either "peer" or "agent" consistently?

lib/rpc/server/peer.go
```diff
@@ -426,3 +434,15 @@ type peer struct {
 	// doneCh is the channel that is closed when this peer shuts down
 	doneCh chan struct{}
 }
+
+func isDeadlineExceededError(err error) bool {
```
r0mant: Nit: Would utils/errors.go be a better place for this?

lib/expand/join.go
```diff
@@ -860,3 +858,18 @@ func watchReconnects(ctx context.Context, cancel context.CancelFunc, watchCh <-c
...
+func validateConnection(ctx context.Context, agent *rpcserver.PeerServer) error {
+	connectTimeoutS := utils.GetenvWithDefault(defaults.EnvPeerConnectTimeout, defaults.PeerConnectTimeout.String())
+	connectTimeout, err := time.ParseDuration(connectTimeoutS)
+	if err != nil {
+		return trace.Wrap(err, "invalid agent connect timeout %q", connectTimeoutS)
+	}
+	ctx, cancel := context.WithTimeout(ctx, connectTimeout)
+	defer cancel()
+	err = agent.ValidateConnection(ctx)
+	if err != nil && !trace.IsLimitExceeded(err) {
```
r0mant: Not sure I understand this: so if the "validate" fails with timeout, we return "nil" as if it succeeded anyway?

a-palchikov: Yes, because it could still be connecting for all we know. There should be a separate (HTTP) timeout to set on all HTTP requests in general - out of scope here.

lib/rpc/server/peer.go
a-palchikov: It is pretty specific in the implementation - not really something I'd use elsewhere.

lib/expand/join.go
knisbet: Do we have any idea why the connection is timing out? Seems odd to me that validating connectivity to the server can really timeout, or as in the original report, that this is somehow related to the number of nodes joining the cluster.

knisbet:
I don't know, it seems odd to me that that this is a process for ding some early validation, and if it fails to do so that we should just ignore it and continue. I'm also really curious why it's timing out at all, shouldn't the other side be answering within the timeout?

a-palchikov:
As I [wrote](https://github.com/gravitational/gravity.e/issues/4229#issuecomment-540547468) in the issue - I've been able to reproduce this with an increased network latency so it is quite a possible scenario. Although the validation connect times out (in my test it did so right after successfully creating a client but before it was able to send `PeerJoin`), the operation then succeeds with the change (albeit slower than usual).

lib/utils/env.go
```diff
@@ -84,6 +84,15 @@ func CheckInPlanet() bool {
 	return runningInsideContainer
 }
 
+// GetenvWithDefault returns the value the environment variables given
+// with name or defaultValue is the variable does not exist
```
r0mant: **if** the variable

lib/utils/env.go
```diff
@@ -84,6 +84,15 @@ func CheckInPlanet() bool {
 	return runningInsideContainer
 }
 
+// GetenvWithDefault returns the value the environment variables given
```
r0mant: value **of** the environment variable

PR #779, opened 2019-10-09
[5.5.x] do not consider non-existent directories for file system usage stats

a-palchikov:
Skip entries if target directory does not exist when computing filesystem usage.

Requires forward port to 6.x, master.

Updates https://github.com/gravitational/gravity.e/issues/4232.

### DIFF
--- a/lib/systeminfo/systeminfo.go
+++ b/lib/systeminfo/systeminfo.go
@@ -238,6 +238,10 @@ func collectFilesystemUsage(fs []storage.Filesystem) (result storage.FilesystemS
 	for _, mount := range fs {
 		usage := sigar.FileSystemUsage{}
 		if err := usage.Get(mount.DirName); err != nil {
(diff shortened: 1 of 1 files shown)


lib/systeminfo/systeminfo.go
```diff
@@ -238,6 +238,10 @@ func collectFilesystemUsage(fs []storage.Filesystem) (result storage.FilesystemS
 	for _, mount := range fs {
 		usage := sigar.FileSystemUsage{}
 		if err := usage.Get(mount.DirName); err != nil {
+			if os.IsNotExist(err) {
```
r0mant: Just thinking, maybe it would be better to treat any error as non-critical here and just log it. In case it fails with like a permission error or something else.

a-palchikov: Yeah, maybe you're right.

Issue #780, opened 2019-10-09
[BUG] Autojoin fails with "bad username or password" on 6.0+


knisbet:
```
Original Error: *trace.AccessDeniedError bad username or password
Stack Trace:
    /home/knisbet/go/src/github.com/gravitational/gravity/lib/ops/opsclient/opsclient.go:138 github.com/gravitational/gravity/lib/ops/opsclient.(*Client).GetAccounts
    /home/knisbet/go/src/github.com/gravitational/gravity/lib/expand/join.go:1124 github.com/gravitational/gravity/lib/expand.(*Peer).validateWizardState
    /home/knisbet/go/src/github.com/gravitational/gravity/lib/expand/join.go:585 github.com/gravitational/gravity/lib/expand.(*Peer).dialWizard
    /home/knisbet/go/src/github.com/gravitational/gravity/lib/expand/join.go:819 github.com/gravitational/gravity/lib/expand.(*Peer).tryConnect
    /home/knisbet/go/src/github.com/gravitational/gravity/lib/expand/join.go:782 github.com/gravitational/gravity/lib/expand.(*Peer).connectLoop
```

https://github.com/gravitational/gravity/blob/832f080453655eef896881dfdec31af4affc8017/lib/expand/join.go#L800-L835

The Join process appeals to fail, because it first attempts to join the wizard, and only if that fails with a connection failure / non permanent failure, continues on to try the cluster service. In the autojoin case, the peer for the wizard is the cluster, so when it attempts to contact the wizard, the cluster sends back the bad username or password error, which the join logic correctly see's as a permanent failure.

PR #781, opened 2019-10-09
indicate to the join process whether it should use the wizard or not

knisbet:
Add an indicator that AutoJoin can use, to indicate to the expand process that it's connecting to an existing cluster and shouldn't attempt to use the wizard. This hint is needed because right now the logic is based around attempting to connect to the wizard first, and if that produces a connectivity failure, then to move onto the cluster join.

Updates https://github.com/gravitational/gravity/issues/780


knisbet:
> The code's fine but I'm curious, why was this working before then? The logic of first connecting to wizard and then to the cluster was here from the very start.

@r0mant  In DIma's PR it changed the behaviour slightly, where IIRC before it would continue sort of regardless of error or with more error conditions, and post the changes it tries to select unrecoverable errors and exit. The bad username or password though should be a hard failure, otherwise a user with a bad token may not notice.

PR #802, opened 2019-10-15
[6.1.x] followup to #768

a-palchikov:
* Fix teleport config fetching code (lib/process/import.go) to support fetching of legacy teleport config packages if no packages are using new versioning scheme.
 * Add tests for the added functionality.
 * Bonus: fix flaky lib/rpc/proxy test by backporting changes from master.

Updates https://github.com/gravitational/gravity/pull/768 to fix an issue with upgrades over a version that does not use the new configuration package versioning scheme.

### DIFF
--- a/lib/process/import.go
+++ b/lib/process/import.go
@@ -112,12 +112,12 @@ func (i *importer) Close() error {
 
 // getMasterTeleportConfig extracts configuration from teleport package
 func (i *importer) getMasterTeleportConfig(clusterName string) (*telecfg.FileConfig, error) {
-	configPackage, err := pack.FindLatestPackageCustom(pack.FindLatestPackageRequest{
-		Packages:   i.packages,
-		Repository: clusterName,
-		Match:      matchTeleportConfigPackage(*defaults.TeleportVersion),
-	})
+	configPackage, err := i.findLatestTeleportConfigPackage(clusterName, *defaults.TeleportVersion)
 	if err != nil {
+		if trace.IsNotFound(err) {
+			return nil, trace.Wrap(err,
+				"failed to find latest teleport configuration for %v", clusterName)
+		}
 		return nil, trace.Wrap(err)
 	}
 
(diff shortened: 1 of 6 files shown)


lib/process/import.go
```diff
@@ -250,3 +273,9 @@ func matchTeleportConfigPackage(teleportVersion semver.Version) pack.MatchFunc {
 		return verBase.Compare(teleportVersion) == 0
 	}
 }
+
+func matchLegacyTeleportConfigPackage() pack.MatchFunc {
+	return func(env pack.PackageEnvelope) bool {
+		return env.HasLabel(pack.PurposeLabel, pack.PurposeTeleportMasterConfig)
```
r0mant: For the legacy packages - do we need to compare them somehow, to make sure an appropriate one is selected rather than just the first one with the label? Or were they "uncomparable" before? I don't remember honestly.

lib/process/import.go
```diff
@@ -112,12 +112,12 @@ func (i *importer) Close() error {
 
 // getMasterTeleportConfig extracts configuration from teleport package
 func (i *importer) getMasterTeleportConfig(clusterName string) (*telecfg.FileConfig, error) {
-	configPackage, err := pack.FindLatestPackageCustom(pack.FindLatestPackageRequest{
-		Packages:   i.packages,
-		Repository: clusterName,
-		Match:      matchTeleportConfigPackage(*defaults.TeleportVersion),
-	})
+	configPackage, err := i.findLatestTeleportConfigPackage(clusterName, *defaults.TeleportVersion)
```
r0mant: To ease debugging, do you know if we eventually log the package that was selected somewhere? If not, I'd log it here.

a-palchikov: Yes, it is logged at the caller's site.

lib/process/import.go
a-palchikov: This is just a matcher implementation that selects the package which will then be evaluted in the caller. The caller searches the package with the maximum available version among the selected ones - so yes, it selects the latest which is how it should have worked before.

PR #803, opened 2019-10-15
autojoin stale retry loop

knisbet:
This change codifies a workaround previously placed in the cloud-init config when creating a cluster via terraform on AWS.

The AWS autojoin integrations use AWS SSM to store the URL of the cluster and token used for joining the cluster by name. When a cluster is uninstalled and being re-installed, new parameters may not have replaced the old ones yet, causing nodes to attempt to join the stale cluster which will fail. The workaround in the bash script operated by testing that the URL in SSM is a valid cluster, and then only launching autojoin once this succeeded. 

I don't see a better way to do this, so I've essentially copied the same logic here, when retrieving the parameters from SSM, test them for validity by trying to connect to the URL and seeing if a successfully http result comes back. Loop indefinitely until that condition is met, which is appropriate behaviour for autojoin.

### DIFF
--- a/tool/gravity/cli/config.go
+++ b/tool/gravity/cli/config.go
@@ -885,15 +885,19 @@ func retryUpdateJoinConfigFromCloudMetadata(ctx context.Context, config *autojoi
 		// Test that the serviceURL is reachable
 		// When re-installing a cluster into AWS, the serviceURL can point to an old cluster
 		// until the new cluster overwrites the SSM values. So test the ServiceURL is reachable before using it.
-		req, _ := http.NewRequest("GET", config.serviceURL, nil)
+		req, err := http.NewRequest("GET", config.serviceURL, nil)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		// TODO(Knisbet) replace with NewRequestWithContext when on golang 1.13
 		req = req.WithContext(ctx)
 		_, err = client.Do(req)
 		if err != nil {
-			return trace.WrapWithMessage(err, "Waiting for service URL to become available.").
+			return trace.Wrap(err, "Waiting for service URL to become available.").
 				AddField("service_url", config.serviceURL)
 		}
 
-		return err
+		return trace.Wrap(err)
 	}
 	return trace.Wrap(utils.RetryWithInterval(ctx, b, f))
 }
--- a/tool/gravity/cli/config.go
+++ b/tool/gravity/cli/config.go
@@ -37,6 +37,7 @@ import (
 	"github.com/gravitational/gravity/lib/constants"
 	"github.com/gravitational/gravity/lib/defaults"
 	"github.com/gravitational/gravity/lib/expand"
+	"github.com/gravitational/gravity/lib/httplib"
 	"github.com/gravitational/gravity/lib/install"
 	installerclient "github.com/gravitational/gravity/lib/install/client"
 	installpb "github.com/gravitational/gravity/lib/install/proto"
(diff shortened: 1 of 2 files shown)


tool/gravity/cli/config.go
```diff
@@ -865,12 +868,42 @@ type agentConfig struct {
 	cloudProvider string
 }
 
+func retryUpdateJoinConfigFromCloudMetadata(ctx context.Context, config *autojoinConfig) error {
+	// Security: Insecure is OK here because we're only testing reachability to the cluster
+	client := httplib.GetClient(true,
+		httplib.WithTimeout(5*time.Second),
+		httplib.WithInsecure())
```
a-palchikov: nit: this will add `WithInsecure` twice (boolean argument also means `insecure`). Would be nice to get rid of the boolean argument in the first place (although this is probably out of this scope) as redundant and better make do with functional options.

tool/gravity/cli/config.go
```diff
@@ -865,12 +868,42 @@ type agentConfig struct {
...
+		err := updateJoinConfigFromCloudMetadata(ctx, config)
+		if err != nil && !trace.IsRetryError(err) {
+			return &backoff.PermanentError{Err: err}
+		}
+
+		// Test that the serviceURL is reachable
+		// When re-installing a cluster into AWS, the serviceURL can point to an old cluster
+		// until the new cluster overwrites the SSM values. So test the ServiceURL is reachable before using it.
+		req, _ := http.NewRequest("GET", config.serviceURL, nil)
+		req = req.WithContext(ctx)
```
a-palchikov: There's [NewRequestWithContext](https://golang.org/pkg/net/http/#NewRequestWithContext) - or did you want to avoid it for some reason?
Also I'm not sure about ignoring an error here - is it guaranteed to succeed?

tool/gravity/cli/config.go
```diff
@@ -865,12 +868,42 @@ type agentConfig struct {
...
+	f := func() error {
+		err := updateJoinConfigFromCloudMetadata(ctx, config)
+		if err != nil && !trace.IsRetryError(err) {
+			return &backoff.PermanentError{Err: err}
+		}
+
+		// Test that the serviceURL is reachable
+		// When re-installing a cluster into AWS, the serviceURL can point to an old cluster
+		// until the new cluster overwrites the SSM values. So test the ServiceURL is reachable before using it.
+		req, _ := http.NewRequest("GET", config.serviceURL, nil)
```
r0mant: Can we not ignore the error please?

tool/gravity/cli/config.go
```diff
@@ -865,12 +868,42 @@ type agentConfig struct {
...
+		req = req.WithContext(ctx)
+		_, err = client.Do(req)
+		if err != nil {
+			return trace.WrapWithMessage(err, "Waiting for service URL to become available.").
```
r0mant: Why not just Wrap()? It accepts the message too.

tool/gravity/cli/config.go
```diff
@@ -865,12 +868,42 @@ type agentConfig struct {
...
+				AddField("service_url", config.serviceURL)
+		}
+
+		return err
```
r0mant: trace.Wrap?

tool/gravity/cli/config.go
knisbet: NewRequestWithContext was added in go 1.13, I think we're still using go 1.12.

knisbet: NewRequestWithContext was added in go1.13, I think we're still on golang 1.12 on this repo.

PR #814, opened 2019-10-17
Fix remote support issues.

r0mant:
This PR is to fix issues with inability to access remote clusters via `tsh` or web UI terminal via a hub.

The issue was twofold:

* TLS certificates generated by gravity didn't have all required information encoded into it e.g. user traits. Now gravity uses teleport's auth API to generate certs instead of its own implementation.
* Local users created by gravity did not have "traits" field on them which is now required by teleport auth to work properly. Now gravity always assigns traits when creating local users. There is also a migration that will add traits to the users that do not have them upon startup.


lib/users/usersservice/migrate.go
```diff
@@ -68,18 +73,22 @@ func (u *UsersService) Migrate() error {
 	return nil
 }
 
-func (u *UsersService) updateUserWithRoles(user storage.User, roles ...teleservices.Role) error {
-	for i := range roles {
-		if err := u.backend.UpsertRole(roles[i], storage.Forever); err != nil {
-			return trace.Wrap(err)
-		}
-		user.AddRole(roles[i].GetName())
+// updateUserWithTraits sets traits for the provided user that doesn't have them.
```
a-palchikov: nit: looks like it **always** sets traits - not when a user does not have them.

Issue #750, opened 2019-10-02
Improve gravity status command after network partition


bernardjkim:
While testing gravity behavior after a network partition, I noticed that the `status` command has a delay of ~15 minutes. 

From my understanding, the issue seems to be that once a network partition is created, it takes a while for kubernetes to realize that a node is unreachable. During that time, if status is requested, the gravity `statusOperator` will attempt to send an HTTP request to the node that kubernetes has not yet realized is unreachable. The `statusOperator` has no timeout value for an HTTP request and ends up waiting until kubernetes has discovered the network partition.

Here is a relevant article better explaining the scenario: https://medium.com/expedia-group-tech/improving-kubernetes-responsiveness-under-network-partitioning-df00403a6d97

It is not ideal to have users waiting ~15 minutes for a response from a gravity command even if the kubernetes cluster itself is not yet in a functional state.

We can improve the user experience by adding a timeout when initializing the `statusOperator` so that users aren't left hanging and wondering what is wrong. With this change, after timing out, `gravity status` should respond with a status that lets the user know that the cluster is currently 'offline'?

a-palchikov:
That's a very good point - I see no reason why we'd create an HTTP client to a service without a timeout by default.

PR #817, opened 2019-10-18
Add timeout when collecting cluster status

bernardjkim:
This PR addresses timeout portion of #750 .
As it is now, after a status command times out, the returned status is incorrect.

Here is an example of the returned status after timing out:
```
Cluster status: degraded
Cluster image:  telekube, version 7.0.0-alpha.1.122
Join token:     longtoken
Last completed operation:
    * operation_install (4d662bd6-2e4f-4bc4-b349-d16ba5b70369)
      started:          Tue Oct  8 19:09 UTC (3 hours ago)
      completed:        Tue Oct  8 19:09 UTC (3 hours ago)
Cluster endpoints:
    * Authentication gateway:
        - 172.28.128.3:32009
        - 172.28.128.4:32009
        - 172.28.128.5:32009
    * Cluster management URL:
        - https://172.28.128.3:32009
        - https://172.28.128.4:32009
        - https://172.28.128.5:32009
Cluster nodes:  dev.test
    Masters:
        * gravity1 / 172.28.128.4 / node
            Status:     degraded
        * gravity2 / 172.28.128.5 / node
            Status:     offline
    Nodes:
        * gravity0 / 172.28.128.3
            Status:     offline
...
```
In this scenario, we started off with three master nodes and `gravity2` was the initial leader. `gravity2` was then isolated from the cluster by a network partition. This status was returned when `gravity status` was requested from `gravity0`. This status response doesn't seem to correctly reflect the current state of the cluster and may be more confusing than helpful.

Will work on improving status response in separate PR.


lib/localenv/localenv.go
```diff
@@ -319,6 +319,15 @@ func (env *LocalEnvironment) OperatorService(opsCenterURL string, options ...htt
 	return client, nil
 }
 
+// SiteOperatorWithTimeout returns Operator for the local gravity site with a
+// timeout duration.
+func (env *LocalEnvironment) SiteOperatorWithTimeout(duration time.Duration) (*opsclient.Client, error) {
```
a-palchikov: nit: we've been moving away from `Site` in favor of `Cluster` for some time now - would be cool to avoid `[Ss]ite` in new method names.
Another question - why not hardcode a timeout in the service clients universally - is there a reason to have a client with no timeout in the first place?

lib/defaults/defaults.go
```diff
@@ -252,6 +252,15 @@ const (
 	// is accessible
 	ClusterCheckTimeout = 5 * time.Second
 
+	// ClusterOperatorTimeout specifies the timeout for a ClusterOperator.
+	ClusterOperatorTimeout = 10 * time.Second
+
+	// StatusCollectionTimeout specifies the timeout for collecting gravity status.
+	StatusCollectionTimeout = 20 * time.Second
```
a-palchikov: I think it's a bit on the side of optimism. This value does not take any context into consideration (network latency etc.) and should be more accommodating in my opinion.

lib/httplib/client.go
```diff
@@ -182,6 +182,9 @@ func GetClient(insecure bool, options ...ClientOption) *http.Client {
 	for _, o := range options {
 		o(client)
 	}
+	if client.Timeout == 0 {
+		client.Timeout = defaults.ClientTimeout
```
r0mant: Hmm, I'm not sure there should be a default timeout on the client. Esp. the one as short as 30 seconds. This is overall timeout, not just the connect timeout, right? There are some potentially long-running gravity commands, like downloading packages, which may be timing out because of this?

lib/localenv/localenv.go
bernardjkim: Good point, I've modified the GetClient func to use the default ClientTimeout if no timeout was given. 58b7e3c

lib/defaults/defaults.go
bernardjkim: Wasn't sure what values to use for the default timeouts. I've increased the status and audit collection timeout to one minute. What do you think would be a safe value for timeouts?

lib/httplib/client.go
bernardjkim: Oh... yea I guess this wouldn't work then.

a-palchikov: Running without timeout is also not ideal. In case of long-running commands, they would either disable or set a longer timeout explicitly.
@bernardjkim I think setting the default should be done **before** applying the options to let the caller override the defaults.
@r0mant let's discuss that in more detail.

tool/gravity/cli/status.go
```diff
@@ -75,7 +76,9 @@ func status(env *localenv.LocalEnvironment, printOptions printOptions) error {
 		}
 	}
 	if status.Agent == nil {
-		status.Agent, err = statusapi.FromPlanetAgent(context.TODO(), nil)
+		ctx, cancel := context.WithTimeout(context.TODO(), defaults.StatusCollectionTimeout)
```
knisbet: By Re-creating the context, is the intention that when falling back to give an additional 1-minute for collecting status to complete?

bernardjkim: Yeah, that's what I was thinking.

lib/httplib/client.go
r0mant: @a-palchikov @bernardjkim My proposal is to have a default *connect* timeout and no default overall timeout which is very request-specific.

a-palchikov: @bernardjkim Ok, so let's set the dialer's timeout to some reasonable default and leave the client's timeout a caller's responsibility for the time being so this gets merged.

lib/defaults/defaults.go
```diff
@@ -252,6 +252,12 @@ const (
 	// is accessible
 	ClusterCheckTimeout = 5 * time.Second
 
+	// StatusCollectionTimeout specifies the timeout for collecting gravity status.
+	StatusCollectionTimeout = 1 * time.Minute
+
+	// AuditCollectionTimeout specifies the timeout for collecting audit logs.
+	AuditCollectionTimeout = 1 * time.Minute
```
r0mant: Nit: Looking at where it is being used, "audit collection" is not very accurate name as it's used for the audit log client which is used for submitting audit events. Something like AuditLogClientTimeout would probably be more appropriate name. Also, I think 1 minute for this is too much, I'd make it like 5 seconds or so.

PR #824, opened 2019-10-22
pod access to etcd

knisbet:
add kubernetes service SANs to etcd certs so that etcd is reachable from pods inside the cluster

Requires: forward-port

This is a result of the change in etcd 3 to using the etcd gateway. The etcd gateway acts as a L4 proxy to a master, which means the master cert needs SANs for the gateway nodes, which we don't currently support. As an alternative, have clients within pods access etcd via the kubernetes svc, and contact the master etcd servers directly.

I would discourage this mode of operation for most services, which should instead should deploy an etcd cluster within kubernetes, but in the stolon case it historically uses the cluster etcd.


knisbet:
Yea, I considered that, but I don't think we really have the infra in place for an external service to track node uptime and membership changes and sync with the kubernetes API. I think in an ideal world you're right, but unless we want to add first class support for the service, we risk regressions and complexity :(.

Issue #840, opened 2019-10-24
gravity check should verify upgrade related checks


klizhentas:
## Description

Whenever users would like to make sure that upgrade will succeed, they would like to run `gravity check upgrade` from the install tarball.

This command should run the upgrade to the point of running preflight checks without performing any actions.

This will allow users to very that upgrade in `dry run` mode could succeed.

r0mant:
This is available starting from [5.5.27](https://gravitational.com/gravity/docs/changelog/#5527-november-1st-2019). Will be forward-ported to 6.1/6.2 and master as well.

PR #843, opened 2019-10-25
(5.5) Join prechecks.

r0mant:
Backport https://github.com/gravitational/gravity/pull/530 to 5.5.

I've started looking into https://github.com/gravitational/gravity/issues/840 and remembered that when adding join prechecks a few months ago I did some refactoring in the checks-related code which would be good to have before working on making `gravity check` command works for upgrades. Hence, this backport. Will need to port to 6.1/6.2 as well.

### DIFF
--- a/lib/expand/builder.go
+++ b/lib/expand/builder.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 Gravitational, Inc.
+Copyright 2018-2019 Gravitational, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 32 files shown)


lib/expand/builder.go
```diff
@@ -167,7 +181,6 @@ func (b *planBuilder) AddStartAgentPhase(plan *storage.OperationPlan) {
 				OpsCenterURL: fmt.Sprintf("https://%v", b.Peer),
 			},
 		},
-		Requires: []string{SystemPhase},
```
a-palchikov: Does this depend on another phase or does not have any dependency?

lib/install/fsm.go
```diff
@@ -248,12 +247,10 @@ func (f *fsmEngine) GetExecutor(p fsm.ExecutorParams, remote fsm.Remote) (fsm.Ph
...
-func (f *fsmEngine) RunCommand(ctx context.Context, runner fsm.RemoteRunner, server storage.Server, p fsm.Params) error {
-	args := []string{"install", "--phase", p.PhaseID, fmt.Sprintf("--force=%v", p.Force)}
-	if f.RemoteOpsURL != "" && f.RemoteOpsToken != "" {
-		args = append(args,
-			fmt.Sprintf("--ops-url=%v", f.RemoteOpsURL),
-			fmt.Sprintf("--ops-token=%v", f.RemoteOpsToken))
+func (f *fsmEngine) RunCommand(ctx context.Context, runner rpc.RemoteRunner, server storage.Server, p fsm.Params) error {
+	args := []string{"plan", "execute", "--phase", p.PhaseID}
+	if p.Force {
+		args = append(args, "--force")
```
a-palchikov: If I remember correctly, this change alone broke etcd upgrades so the etcd plan had to be tweaked to fix it. Should be taken into account if not yet.

lib/expand/builder.go
r0mant: It does not have any dependency anymore. Because we start the agent as a first phase now that we run pre-flight checks.

lib/install/fsm.go
r0mant: Discussed offline that the necessary changes are already present in 5.5 so we should be ok.

PR #845, opened 2019-10-28
(5.5) Make gravity check command work for upgrades

r0mant:
This PR updates the existing `gravity check` command to make sure that it can be used for running upgrade preflight checks too. The way it works is:

* When `gravity check` is launched, it detects whether there is an installed cluster:
  * If there is no cluster, it runs install prechecks like before.
  * If there is a cluster, it runs upgrade prechecks.

Other notes:

- I've added a new script, `check`, to the cluster images so it's more convenient to run prechecks.
- Had to refactor the guts of the Checker a bit to make sure that we can report the failed checks properly and not like just a bunch of concatenated errors.

Closes https://github.com/gravitational/gravity/issues/840. Needs port to 6.1, 6.2 and master.


lib/checks/checks.go
```diff
@@ -319,119 +329,145 @@ func (r *checker) Run(ctx context.Context) error {
...
+			FormatFailedChecks(failed))
+	}
 
-	return trace.NewAggregate(errors...)
+	return nil
 }
 
 // CheckNode executes checks for the provided individual server.
-func (r *checker) CheckNode(ctx context.Context, server Server) error {
+func (r *checker) CheckNode(ctx context.Context, server Server) (failed []*agentpb.Probe) {
```
r0mant: Changed this (and `CheckNodes` below) from returning a generic error to a list of failed probes instead so various callers have a chance to format the errors properly.

lib/checks/checks.go
```diff
@@ -319,119 +329,145 @@ func (r *checker) Run(ctx context.Context) error {
...
 // CheckNodes executes checks that take all provided servers into account.
-func (r *checker) CheckNodes(ctx context.Context, servers []Server) error {
+func (r *checker) CheckNodes(ctx context.Context, servers []Server) (failed []*agentpb.Probe) {
 	if ifTestsDisabled() {
 		log.Infof("Skipping multi-node checks due to %q set.",
 			constants.PreflightChecksOffEnvVar)
 		return nil
 	}
 
-	if len(servers) < 2 {
```
r0mant: Turns out this check up here was preventing the `checkPorts` for a single node from executing so bound ports weren't detected.

tool/gravity/cli/rpcagent.go
```diff
@@ -130,30 +130,40 @@ var agentFunctions map[string]agentFunc = map[string]agentFunc{
...
-	clusterEnv, err := localEnv.NewClusterEnvironment()
+func rpcAgentDeploy(localEnv *localenv.LocalEnvironment, leaderParams, nodeParams string) error {
+	_, err := rpcAgentDeployHelper(context.Background(), localEnv, leaderParams, nodeParams)
 	if err != nil {
 		return trace.Wrap(err)
 	}
+	return nil
+}
+
+func rpcAgentDeployHelper(ctx context.Context, localEnv *localenv.LocalEnvironment, leaderParams, nodeParams string) (credentials.TransportCredentials, error) {
```
r0mant: Couldn't really come up with a better name for this. There is already `deployAgents` method below :)

lib/pack/utils.go
```diff
@@ -256,6 +256,13 @@ func ExecutePackageCommand(p PackageService, cmd string, loc loc.Locator, confLo
 	return out, nil
 }
 
+// FindPackageByName returns the first package with the specified name.
+func FindPackageByName(packages PackageService, name string) (*PackageEnvelope, error) {
```
a-palchikov: `FindFirstPackageByName`

lib/update/cluster/phases/checks.go
```diff
@@ -88,26 +79,22 @@ func (p *updatePhaseChecks) PostCheck(context.Context) error {
 
 // Execute runs preflight checks
 func (p *updatePhaseChecks) Execute(ctx context.Context) error {
-	installedApp, err := p.apps.GetApp(p.installedPackage)
+	p.Infof("Executing preflight checks on %v.", storage.Servers(p.servers))
+	checker, err := ops.NewUpgradeChecker(ctx, ops.UpgradeCheckerConfig{
```
a-palchikov: or `NewChecker` had it been implemented in `lib/update/cluster`

tool/gravity/cli/checks.go
```diff
@@ -17,28 +17,46 @@ limitations under the License.
...
 
-func checkManifest(env *localenv.LocalEnvironment, manifestPath, profileName string, autoFix bool) error {
+func checkManifest(env *localenv.LocalEnvironment, manifestPath, imagePath, profileName string, autoFix bool) error {
+	_, err := GetLocalClusterInfo(env)
+	if err != nil {
+		env.PrintStep("Running install preflight checks")
+		return checkInstall(env, manifestPath, profileName, autoFix)
+	}
+	env.PrintStep("Detected deployed cluster, running upgrade preflight checks")
+	return checkUpgrade(context.Background(), env, manifestPath, imagePath)
```
a-palchikov: Limit the runtime with a non-dummy context?

tool/gravity/cli/checks.go
```diff
@@ -17,28 +17,46 @@ limitations under the License.
...
+}
+
+func checkInstall(env *localenv.LocalEnvironment, manifestPath, profileName string, autoFix bool) error {
+	if profileName == "" {
```
a-palchikov: Use the first profile as a fallback if none specified?

tool/gravity/cli/rpcagent.go
```diff
@@ -130,30 +130,40 @@ var agentFunctions map[string]agentFunc = map[string]agentFunc{
 	constants.RPCAgentSyncPlanFunction: executeSyncOperationPlan,
 }
 
-func rpcAgentDeploy(localEnv, updateEnv *localenv.LocalEnvironment, leaderParams, nodeParams string) error {
-	clusterEnv, err := localEnv.NewClusterEnvironment()
+func rpcAgentDeploy(localEnv *localenv.LocalEnvironment, leaderParams, nodeParams string) error {
+	_, err := rpcAgentDeployHelper(context.Background(), localEnv, leaderParams, nodeParams)
```
a-palchikov: This should also avoid using the unlimited context.

tool/gravity/cli/checks.go
```diff
@@ -57,15 +78,91 @@ func checkManifest(env *localenv.LocalEnvironment, manifestPath, profileName str
...
+		return trace.Wrap(err)
+	}
+	packages, err := env.ClusterPackages()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	// Need to upload gravity package from the upgrade image, otherwise
+	// RPC agents may fail to deploy because they will be looking for
+	// this gravity package in the cluster's package service.
+	err = uploadGravity(ctx, env, tarballEnv.Packages, packages)
```
a-palchikov: Why not use the manifest and deal with the specific gravity package version? You won't need `FindPackageByName` in this case.

tool/gravity/cli/checks.go
```diff
@@ -17,28 +17,46 @@ limitations under the License.
...
 
+	"github.com/fatih/color"
 	pb "github.com/gravitational/satellite/agent/proto/agentpb"
 	"github.com/gravitational/trace"
 )
 
-func checkManifest(env *localenv.LocalEnvironment, manifestPath, profileName string, autoFix bool) error {
+func checkManifest(env *localenv.LocalEnvironment, manifestPath, imagePath, profileName string, autoFix bool) error {
+	_, err := GetLocalClusterInfo(env)
+	if err != nil {
```
a-palchikov: Ignoring the error will make it difficult to understand why the command executes installer checks on an installed cluster if it has issues. Do we want to suggest running status in case of specific error conditions? (I don't really feel comfortable treating **any** error as a cluster not being there condition).

r0mant: Yeah, that's a bit tough. I feel like we don't have a bulletproof way to determine whether we have a cluster locally or now. For now, I've opted for the same heuristic as "gravity status" uses which is trying to query the planet agent status.

lib/checks/checks.go
```diff
@@ -255,6 +255,16 @@ func DockerConfigFromSchemaValue(dockerSchema schema.Docker) (config storage.Doc
 	}
 }
 
+// Checker defines a preflight checker interface.
+type Checker interface {
+	// Run runs a full set of checks on the nodes configured in the checker.
+	Run(ctx context.Context) error
+	// CheckNode executes checks for the provided individual server.
+	CheckNode(ctx context.Context, server Server) []*agentpb.Probe
```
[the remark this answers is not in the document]
r0mant: I've updated the comments to be more clear.

tool/gravity/cli/commands.go
```diff
@@ -670,6 +670,8 @@ type CheckCmd struct {
 	AutoFix *bool
 	// ImagePath is path to unpacked cluster image
 	ImagePath *string
+	// Timeotu is checks timeout
```
a-palchikov: Timeout is the time allotted to run preflight checks.

tool/gravity/cli/register.go
```diff
@@ -208,11 +208,12 @@ func RegisterCommands(app *kingpin.Application) *Application {
...
 
-	g.CheckCmd.CmdClause = g.Command("check", "check host environment to match manifest")
-	g.CheckCmd.ManifestFile = g.CheckCmd.Arg("manifest", "application manifest in YAML format").Default(defaults.ManifestFileName).String()
-	g.CheckCmd.Profile = g.CheckCmd.Flag("profile", "profile to check").Short('p').String()
-	g.CheckCmd.AutoFix = g.CheckCmd.Flag("autofix", "attempt to fix some of the problems").Bool()
-	g.CheckCmd.ImagePath = g.CheckCmd.Flag("image-path", "Path to unpacked cluster image.").String()
+	g.CheckCmd.CmdClause = g.Command("check", "Execute preflight checks")
+	g.CheckCmd.ManifestFile = g.CheckCmd.Arg("manifest", "Cluster image manifest file").Default(defaults.ManifestFileName).String()
+	g.CheckCmd.Profile = g.CheckCmd.Flag("profile", "Name of the node profile to check against").Short('p').String()
+	g.CheckCmd.AutoFix = g.CheckCmd.Flag("autofix", "Attempt to fix some of the discovered problems").Bool()
```
a-palchikov: `Attempt to fix discovered problems.` It does this on best effort - otherwise, I'd mention that not all problems are automatically fixable or something along those lines.

tool/gravity/cli/checks.go
```diff
@@ -17,37 +17,89 @@ limitations under the License.
...
+}
+
+// detectsCluster attempts to detect whether the current node has a cluster deployed.
+//
+// Returns true if the cluster is detected and false otherwise.
+func detectCluster(ctx context.Context) bool {
+	_, err := statusapi.FromPlanetAgent(ctx, nil)
 	if err != nil {
-		return trace.Wrap(err)
+		log.WithError(err).Info("No cluster detected: failed to query planet agent.")
```
knisbet: Hmm, I think logging this could cause some confusion. It's tough, because this particular log is completely normal on a new install, and abnormal on an upgrade.

tool/gravity/cli/checks.go
```diff
@@ -17,37 +17,89 @@ limitations under the License.
...
+
+	env.PrintStep("Detected deployed cluster, running upgrade preflight checks")
+	return checkUpgrade(ctx, env, config)
+}
+
+// detectsCluster attempts to detect whether the current node has a cluster deployed.
+//
+// Returns true if the cluster is detected and false otherwise.
+func detectCluster(ctx context.Context) bool {
+	_, err := statusapi.FromPlanetAgent(ctx, nil)
```
knisbet: Hmm, I'm not sure this logic should be based on reaching a service, doesn't it make sense to detect the cluster based on whether the cluster has been installed to disk, and not based on the status/health of a service?

r0mant: Yeah, we were discussing it a bit with Dima above as well, and as I mentioned I don't think we have the "silver bullet" way to determine whether the cluster is installed right now.

For example, the way you mentioned also won't work for scenarios where, say, a cluster was partially installed/uninstalled and still has some leftovers on disk.

At least by checking the planet agent status we know that if it succeeds, the cluster **for sure** is deployed. If it doesn't succeed, yeah, it may mean that the cluster is either not installed or in trouble, but I think that's a fine trade-off for now. Gravity status also does it this way currently.

tool/gravity/cli/checks.go
r0mant: Yeah, that's why I logged it as info. The error is there to ease possible troubleshooting in cases we discussed below (and above).

lib/app/service/installer.go
```diff
@@ -403,6 +435,22 @@ You also need a direct network connection to the servers
...
+Executing preflight checks
+==========================
+
+Before launching install or upgrade operation, you can execute preflight
+checks to make sure the infrastructure satisfies all requirements.
+
+For example, to see if the node satisfies requirements for a particular
+node profile before initial installation, run:
+
+./checks --profile=worker
```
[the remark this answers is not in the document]
r0mant: I've updated the description to be more clear.

tool/gravity/cli/checks.go
knisbet: > For example, the way you mentioned also won't work for scenarios where, say, a cluster was partially installed/uninstalled and still has some leftovers on disk.

I'm not sure what we should consider "working" in this particular scenario. Should pre-install checks be run and give a clean bill of health, for the installer to detect the install state and barf later?

I think based on this discussion, there are 4 scenarios, and the way it works now:
1. Clean system - pre-install checks
2. Partial install / Non working system - pre-install checks
3. Partial install - planet-agent online - upgrade checks (could report system as healthy or non healthy depending on where the install failed, unless we set something like install state in etcd, I've never checked that)
4. completed install - upgrade checks

Using some level of installation state on disk
1. Clean system - pre-install checks
2. Partial install / Non working system - report cluster as non-functioning
3. Partial install - planet-agent online - could report healthy/non healthy
4. completed install - upgrade checks

The installer errors are here:
```
[ERROR]: detected previous installation state in /var/lib/gravity/local, please clean it up using `gravity leave --force` before proceeding (see https://gravitational.com/gravity/docs/cluster/#deleting-a-cluster for more details)
```
https://github.com/gravitational/gravity/blob/master/lib/system/environ/validate.go#L67-L71

https://github.com/gravitational/gravity/blob/master/lib/system/environ/validate.go#L136-L141


Perhaps the issue is that we're continuing to overload logic that operates in non-intuitive ways. The pre-install and pre-upgrade checks are similar but not identical, but we're trying to create a single UX pattern for both which will confuse those who encounter it. Because it has to guess what the user intends to do, and there is always potential for that guess to be wrong. The `gravity status` behavior to report pre-install checks when it can't connect to the cluster is already extremely confusing to those who encounter it. 


I dunno, to me it makes more sense to report the system as non functioning on a partial install, than it does to give it a clean bill of health only to fail when someone tries to run the install a moment later.

lib/app/service/installer.go
```diff
@@ -381,14 +390,37 @@ main "$@"
...
 scriptdir=$(dirname $(realpath $0))
-app=$($scriptdir/gravity app-package --state-dir=$scriptdir)
-$scriptdir/upload && $scriptdir/gravity --insecure update trigger $app
+app=$("$scriptdir/gravity" app-package --state-dir="$scriptdir")
+"$scriptdir/upload" && "$scriptdir/gravity" --insecure update trigger $app
+`
+
+	checkScript = `#!/bin/bash
+#
+# Script for executing preflight checks.
```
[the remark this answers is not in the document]
knisbet: > First, it's very easy to fire up the checks just by invoking the script, without remembering the command / trying to figure out what flags you might need. And the command is there for you for more involved scenarios.
Well this could be embedded in the logic of the binary itself. If it were me, for the pre-install checks, I would even add the ability to ask the user which profile they want to use. If only a single profile is defined, just use it, if someone wants to automate it accept the flag, and if someone doesn't know, give a numbered list that they can select from. Although I would do this on the installer as well.

> Also, it serves a sort of "educational" purpose for the user, i.e. like you said we don't really advertise anywhere in the installer that you can run preflight checks before launching operation while seeing the script will make them realize they can do that. Maybe I should call it something more prominent though like run_preflight_checks rather than generic check.
I guess my line of thinking is why do we want to educate someone on this. If I see a script called `checks` or a script called `run_preflight_checks` as someones first interaction with a new product, these things mean nothing to me. 

Where as someone asking for the ability to run pre-flight checks manually, so it creates less work for them when the checks would fail, seems to point to a UX problem in the upgrade itself. That this is a common pain point in the upgrade, that checks will fail, and then someone is left with a partial upgrade to clean up, especially if rooted in a bug in the checks themselves as recently occurred. 

To me the only reason to introduce a new UX pattern specifically around checks is:
- If I want to run upgrade checks using the new version of the gravity binary, but don't want to actually do an upgrade, or make the decision on continuing with an upgrade separately. Is this is what's being solved with this pattern?
- I want the ability to trigger upgrade checks outside of doing an actual upgrade (IE to check if things are healthy). Our UX pattern for this is already to use `gravity status`.
- I want the ability as a developer to test a new version of checks without as much hassle (Something like this probably shouldn't be at the forefront).

tool/gravity/cli/checks.go
r0mant: @knisbet Yeah, I think it would make sense to detect partial installation and give the instructions  on how to clean up before running the checks. I've already merged this but I'll make a ticket to improve this behavior.

lib/app/service/installer.go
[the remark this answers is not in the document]
r0mant: > If I want to run upgrade checks using the new version of the gravity binary, but don't want to actually do an upgrade, or make the decision on continuing with an upgrade separately. Is this is what's being solved with this pattern?

Yep. It was requested by a customer.

PR #860, opened 2019-11-02
Retry agent start.

r0mant:
The issue: if "join" is launched before installer has exited, there is a (high) chance that it will attempt to start an agent on the first node (which now happens as a first join phase) while wizard is still running and occupying port 3012. The bind will fail and the gravity-agent service will not restart b/c it's a one-shot service. This PR fixes (or rather, mitigates) the issue by retrying agent start.


r0mant:
Closing for now.

Issue #863, opened 2019-11-04
[BUG] kubelet fails to start with unsupported labels


knisbet:
labels set via kubelet are currently having some issues, such as a miss-match between kubelet and the kubernetes ecosystem on what consistutes a valid label. Because of this, labels set within the app.yaml that used to work, can now cause kubelet to fail to start and installs to fail.

```
Oct 16 03:31:28 ip-10-30-4-39.ec2.internal kubelet[3792]: F1016 03:31:28.074425    3792 server.go:185] unknown 'kubernetes.io' or 'k8s.io' labels specified with --node-labels: [kubernetes.io/role]
Oct 16 03:31:28 ip-10-30-4-39.ec2.internal kubelet[3792]: --node-labels in the 'kubernetes.io' namespace must begin with an allowed prefix (kubelet.kubernetes.io, node.kubernetes.io) or be in the specifically allowed set (beta.kubernetes.io/arch, beta.kubernetes.io/instance-type, beta.kubernetes.io/os, failure-domain.beta.kubernetes.io/region, failure-domain.beta.kubernetes.io/zone, failure-domain.kubernetes.io/region, failure-domain.kubernetes.io/zone, kubernetes.io/arch, kubernetes.io/hostname, kubernetes.io/instance-type, kubernetes.io/os)
Oct 16 03:31:28 ip-10-30-4-39.ec2.internal systemd[1]: kube-kubelet.service: Main process exited, code=exited, status=255/EXCEPTION
Oct 16 03:31:28 ip-10-30-4-39.ec2.internal systemd[1]: kube-kubelet.service: Failed with result 'exit-code'.
Oct 16 03:31:33 ip-10-30-4-39.ec2.internal systemd[1]: kube-kubelet.service: Service RestartSec=5s expired, scheduling restart.
Oct 16 03:31:33 ip-10-30-4-39.ec2.internal systemd[1]: kube-kubelet.service: Scheduled restart job, restart counter is at 54.
Oct 16 03:31:33 ip-10-30-4-39.ec2.internal systemd[1]: Stopped Kubernetes Kubelet.
```

The overall direction appears to be that the NodeAuthorizer is slowly limiting the labels that kubelet is allowed to control itself, due to security concerns if a node is breached a node could be relabelled to take additional workloads. So we should go back in the direction of having the node object created by the install process.

There is a risk to this though, as we previously ran into an issue that we failed to isolate, where a node had lost some it's node object, which is why we hardened the config in the first place.

PR #864, opened 2019-11-04
Kevin/master/experimental remove node registration2

knisbet:
Updates https://github.com/gravitational/gravity/issues/863
Requires Backport
Note: Robotest will fail until PR updated with
- https://github.com/gravitational/gravity.e/pull/4245
- https://github.com/gravitational/planet/pull/526

This PR moves node object creation into the install and join processes. I also did some overall label cleanup so we're a bit more consistent / using the latest standard labels. For example, by default `kubectl get nodes` will show the name of the node profile in the role field.


lib/defaults/defaults.go
```diff
@@ -666,15 +666,33 @@ const (
...
+	// KubernetesNodeRoleLabelFormat is the kubernetes node label shown in `kubectl get nodes`
+	KubernetesNodeRoleLabelFormat = "node-role.kubernetes.io/%v"
+
+	//
+	// Kubernetes Well Known Labels
+	// https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/
+	//
+
+	// KubernetesArchLabel is the kubernetes well-known label for arch
+	KubernetesArchLabel = "kubernetes.io/arch"
```
a-palchikov: nit: I would reference the label consts in api:
```go
import "k8s.io/api/core/v1"
...

KubernetesArchLabel = v1.LabelArchStable
// ditto for other labels
```

lib/defaults/defaults.go
```diff
@@ -666,15 +666,33 @@ const (
...
+
+	// KubernetesOSLabel is the kubernetes well known label for os
+	KubernetesOSLabel = "kubernetes.io/os"
+
+	// KubernetesHostnameLabel is the name of kubernetes well-known label what contains host's IP
+	KubernetesHostnameLabel = "kubernetes.io/hostname"
+
+	//
```
a-palchikov: Remove comment placeholder?

lib/install/phases/node.go
```diff
@@ -0,0 +1,115 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:   r.ExecutorParams.Phase.Data.Server.KubeNodeID(),
+				Labels: r.Labels,
+			},
+			Spec: v1.NodeSpec{
+				Taints: r.Taints,
+			},
+		})
+		if err != nil {
+			return trace.Wrap(err)
```
a-palchikov: nit: `rigging.ConvertError(err)`

lib/install/phases/node.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+		return nil
+	})
+
+	return trace.Wrap(err)
+}
+
+// Rollback deletes the coredns configmap that was created in the execute step
```
a-palchikov:
```go
// Rollback is a no-op for this executor
```

lib/install/planbuilder.go
```diff
@@ -324,6 +324,48 @@ func (b *PlanBuilder) AddWaitPhase(plan *storage.OperationPlan) {
 	})
 }
 
+// AddRegisterNodesPhase appends phases to register each node with the cluster
+func (b *PlanBuilder) AddRegisterNodesPhase(plan *storage.OperationPlan) error {
+	var newPhases []storage.OperationPhase
+
+	for i, node := range b.Masters {
```
a-palchikov: Could be:
```go
for i, node := range append(b.Masters, b.Nodes...) {
```

lib/storage/storage.go
```diff
@@ -1607,6 +1607,28 @@ func (s *Server) IsMaster() bool {
...
 
+// GetNodeLabels returns a consistent set of labels that should be applied to the node
+func (s *Server) GetNodeLabels(profileLabels map[string]string) map[string]string {
+	labels := map[string]string{
+		defaults.KubernetesAdvertiseIPLabel:                         s.AdvertiseIP,
+		defaults.KubernetesRoleLabel:                                s.ClusterRole,
+		defaults.KubernetesHostnameLabel:                            s.KubeNodeID(),
+		defaults.KubernetesArchLabel:                                "amd64", //Only amd64 is currently supported
+		defaults.KubernetesOSLabel:                                  "linux", // Only linux is currently supported
+		fmt.Sprintf(defaults.KubernetesNodeRoleLabelFormat, s.Role): s.Role,
```
a-palchikov: This could instead be:
```go
defaults.FormatKubernetesNodeRoleLabel(s.Role)
```

lib/defaults/defaults.go
knisbet: Hmm, in this case, I think I would go the direction of just using the kubernetes constants directly, not sure we buy much by aliasing them into defaults.

lib/install/planbuilder.go
knisbet: So this doesn't work in this case, atleast without some other tweaks. I think it has to do with the Server object in OperatePhaseData being a pointer, this seems to cause the pointer to point to the temporary node location, and then get changed by subsequent loop iterations. By keeping this as separate loops it's just pointing to the prebuilt structure of masters and servers.

I'm not sure if I'm missing something, but I had tried this change and ran into that glitch.

lib/install/phases/node.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+		return nil
+	})
+
+	return trace.Wrap(err)
+}
+
+// Rollback is a noop for this executor
+func (r *registerNodeExecutor) Rollback(context.Context) error {
+	return nil
```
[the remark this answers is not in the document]
knisbet: I was under the impression none of the install phases implemented rollback?

lib/ops/opsservice/configure.go
```diff
@@ -115,6 +116,16 @@ func (o *Operator) ConfigurePackages(req ops.ConfigurePackagesRequest) error {
 		}
 		return trace.Wrap(err)
 	}
+
+	if operation.Type == ops.OperationExpand {
+		// TODO (knisbet) temporary timeout since configure packages overall isn't time limited
+		c, cancel := context.WithTimeout(context.TODO(), 5*time.Minute)
+		defer cancel()
+		err = o.registerKubernetesNode(c, *operation)
```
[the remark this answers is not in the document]
knisbet: Honestly, because invoking a phase on a master node (which has admin certs) is completely broken, and I gave up on figuring out why because nothing else within the join seems to use it. 

To me it makes sense to run the registration when creating the node within gravity, that it also creates it within kubernetes. This can be done during join, since kubernetes is already available. On install though, the node object needs to specifically be created after the kubernetes api starts, but before any pods are created.

knisbet:
@r0mant 

> I remember we had some issues with the nodes "disappearing" before and then re-registering w/o labels and taints. Do you know if this was a Kubernetes bug that was fixed, or can we still encounter this behavior?

We never identified or isolated why this occurred. This is largely why I dropped the node auto-registration, so if we lose a node object it doesn't autoregister with missing labels, and we see this clearly happen so we can hopefully isolate and fix it. Because the direction kubernetes is going restricts the ability to auto-register a node, I don't see much choice.


> I notice we're changing some labels - do the actual values change, or we just use Kubernetes-provided constants? In the former case, should these labels be updated when upgrading from older versions?

Sorry, I'm not exactly sure what exactly you're referring to. Some of the kubernetes labels have stabilized, such as beta.kubernetes.io/os=linux -> kubernetes.io/os=linux. So we add the stabalized labels as part of the upgrade in the migration phase (https://github.com/gravitational/gravity/pull/864/files#diff-6fac0dc897a3a085495cf21b90654a38R180-R182). I think it's too early to drop the beta labels, as there might still be software floating around that uses them. Other than that, the label values are the same, we're just being a bit more consistent in making sure they're set.

lib/ops/opsservice/configure.go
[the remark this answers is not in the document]
r0mant: Hmm, can you define "completely broken"? :) I understand it's easier to just put it wherever but I'd really like us to use this phase-based approach, i.e. to me registering the node doesn't seem to belong to the "configure" phase.

I don't think it's actually even broken - join actually runs stuff on a master node via an agent when the 2nd master joins. Maybe in your case the agent wasn't just launched on the master? It is currently only launched when expanding 1->2 so chances are you just need to reorder the "startAgent" phase so it's executed before "registerNode". Unless it's some other error?

lib/install/phases/node.go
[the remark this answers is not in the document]
r0mant: If it's just for install, then yeah. I thought this executor is also reused by join. Join reuses some install executors.

lib/ops/opsservice/configure.go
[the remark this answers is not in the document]
knisbet: This does use the phase based approach, it's just during the phase that creates the node within the site object, the node also gets created within kubernetes. It's happening in code that already has to run to join a node to the cluster, and is related to node creation within the cluster since this is where the node gets added to the site object.

The remote execution on join is indeed broken, I wasted a bunch of time checking this. When I looked through the join code, I didn't see any phases that actually are set to run on the master node. Instead, in the 1->2 scenario you describe, the phase is set to run on the joining node, but does an arbitrary command execution on the master node. Not the same thing as running a phase on the master. I didn't save the logs, but as I remember it, the phase execution is broken on remote nodes, which appears to be due to slight differences in the install/update agent behaviour. The join process is launching the agent as if it's an install, and as such has a problem actually starting the phase within the agent process (I didn't keep the logs).  Because nothing else actually uses this, I'm not sure it's worth the investment and risk of breaking something else.

The other thing to consider, is as I've mentioned before and flagged in the most recent security audit, we shouldn't be enabling arbitrary command executions between workers to masters at all. The only reason this got rated low, is because the agent is off most of the time and only started when in use. The remote phase invocation and the current behaviour you described both rely on arbitrary command executions, and probably should be re-considered.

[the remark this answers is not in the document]
r0mant: I guess if we treat this phase as "node creation", it sort of does make sense then. The only thing is should we then update the rollback for this phase to remove the Kubernetes node object too? Don't think I saw it.

knisbet:
robotest passed, just needed to fix a unit test

PR #865, opened 2019-11-04
[5.5] Support nodes_count variable.

UnderGreen:
Moves clound config into template file. Use one big disk instead of
adding another one for /var/lib/gravity.

### DIFF
--- a/terraform-dev/cluster.tf
+++ b/terraform-dev/cluster.tf
@@ -1,11 +1,11 @@
 variable "image_name" {
   type = "string"
-  default = "ubuntu-16.04-server-cloudimg-amd64-disk1.img"
+  default = "ubuntu-18.04-server-cloudimg-amd64.img"
 }
 
 variable "disk_size" {
   type = "string"
-  default = "40000000000"
+  default = "15000000000"
 }
 
 variable "memory_size" {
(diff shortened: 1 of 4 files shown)


terraform-dev/cluster.tf
```diff
@@ -18,109 +18,68 @@ variable "cpu_count" {
...
-  name = "gravity-disk-${count.index}.qcow2"
-  pool = "default"
-  size = "${var.gravity_dir_size}"
-  count = 3
-}
-
-# 12 GB volume for tmp
-resource "libvirt_volume" "tmp" {
-  name = "gravity-tmp-disk-${count.index}.qcow2"
+# 40 GB volume for root disk
```
a-palchikov: Since this is configurable, I'd not mention it or say that this is a default.

terraform-dev/cluster.tf
```diff
@@ -18,109 +18,68 @@ variable "cpu_count" {
...
-    - 'parted -a opt /dev/vdc mkpart primary ext4 0% 100%'
-    - 'mkfs.ext4 -L TMP /dev/vdc1'
-    - 'echo "/dev/vdb1 /var/lib/gravity ext4 discard,noatime,nodiratime 0 0" >> /etc/fstab'
-    - 'echo "/dev/vdc1 /tmp ext4 discard,noatime,nodiratime 0 0" >> /etc/fstab'
-    - 'mkdir -p /var/lib/gravity'
-    - 'mount -a'
-    - 'chmod 777 /tmp'
-    EOF
+  name           = "commoninit-${count.index}.iso"
+  user_data = "${templatefile("cloudinit.cfg", {
```
a-palchikov: Do you have to quote it and put inside of ${ .. }$?

Issue #668, opened 2019-09-17
Dynamic volume provisioner in Gravity cluster


benarent:
Need a full guide on how to deploy a dynamic volume provisioner in Gravity cluster, e.g. Local Volume Provisioner. Maybe add Rook/OpenEBS too later on.

benarent:
After discussing at product meeting this issue has been noted as one of the more important issues. 

After discussion having a guide with https://openebs.io/ will be a good starting point.

PR #883, opened 2019-11-16
Add OpenEBS integration.

r0mant:
This PR adds a new system application dependency to gravity clusters - storage-app that installs and configures OpenEBS.

The gist of what's new:

* storage-app is packaged during tele build. We should consider excluding it if it's not enabled (cause it's almost 1GB) but that's for later.
* New manifest section `storage.openebs` that allows to enable it.
* New resource type `PersistentStorage` that allows to tweak OpenEBS configuration.
* Logic necessary to ensure OpenEBS can be installed during upgrade.
* The docs.

Closes https://github.com/gravitational/gravity/issues/668. Depends on https://github.com/gravitational/storage-app/pull/1 and https://github.com/gravitational/planet/pull/529.


lib/schema/manifest.go
```diff
@@ -1208,6 +1249,9 @@ func ShouldSkipApp(manifest Manifest, app loc.Locator) bool {
 		if ext != nil && ext.Catalog != nil && ext.Catalog.Disabled {
 			return true
 		}
+	case defaults.StorageAppName:
+		// do not install storage-app if no storage providers are enabled
+		return !manifest.OpenEBSEnabled()
```
a-palchikov: Is this complicated to also filter out the application package as a dependency if not enabled?

lib/storage/persistentstorage.go
```diff
@@ -0,0 +1,471 @@
...
+	return &NDMConfig{
+		ProbeConfigs: []*NDMProbe{
+			{Name: "udev probe", Key: "udev-probe", State: true},
+			{Name: "searchest probe", Key: "searchest-probe", State: false},
+			{Name: "smart probe", Key: "smart-probe", State: true},
+		},
+		FilterConfigs: []*NDMFilter{
+			{
+				Name:    "os disk exclude filter",
+				Key:     "os-disk-exclude-filter",
```
a-palchikov: nit: move these to local const?

lib/storage/persistentstorage.go
```diff
@@ -0,0 +1,471 @@
...
+				Exclude: strings.Join(defaultExcludeDevices, ","),
+			},
+		},
+	}
+}
+
+// NDMConfigFromConfigMap creates NDM config from the provided config map.
+func NDMConfigFromConfigMap(cm *v1.ConfigMap) (*NDMConfig, error) {
+	data, ok := cm.Data["node-disk-manager.config"]
+	if !ok || len(data) == 0 {
```
a-palchikov: nit `ok` is pretty redundant here.

lib/storage/persistentstorage.go
```diff
@@ -0,0 +1,471 @@
...
+type NDMFilter struct {
+	// Name is the filter name.
+	Name string `yaml:"name"`
+	// Key is the filter id.
+	Key string `yaml:"key"`
+	// State is the filter state (enabled/disabled).
+	State bool `yaml:"state"`
+	// Include is a list of includes for this filter.
+	Include string `yaml:"include,omitempty"`
+	// Exclude is a list of excludes for htis filter.
```
a-palchikov: **this**

docs/6.x/storage.md
```diff
@@ -0,0 +1,346 @@
...
+`storage.yaml` file:
+
+```yaml
+kind: PersistentStorage
+version: v1
+spec:
+  openebs:
+    filters:
+      devices:
+        include: ["/dev/sdd", "/dev/sde"]
```
knisbet: I'm not sure I follow this. The above says the filters get merged with the default filters. 

So if unset, it will include all devices but the defaults to be denied?
Then if set, it will whitelist only those explicitly listed, with an implicit deny all?

docs/6.x/storage.md
```diff
@@ -0,0 +1,346 @@
...
+      mountPoints:
+        exclude: ["/", "/etc/hosts", "/boot"]
+      # Include/exclude devices from specific vendors.
+      vendors:
+        include: []
+        exclude: ["CLOUDBYT", "OpenEBS"]
+```
+
+!!! note "Default filters"
+    Default OpenEBS filters will be merged with the filters configured via
```
knisbet: What are the defaults?

lib/app/apps.go
```diff
@@ -134,24 +134,22 @@ func DeleteAppHookJob(ctx context.Context, client *kubernetes.Clientset, req Del
...
+// GetUpdatedDependencies compares dependencies of the "installed" and "update"
+// apps and returns locators of updated (or new) dependencies.
 //
 // Only direct dependencies are compared, without base app resolution.
-func GetUpdatedDependencies(installed, update Application) ([]loc.Locator, error) {
-	if installed.Package.IsEqualTo(update.Package) {
-		return nil, trace.NotFound("no update for %v", update)
-	}
-
+func GetUpdatedDependencies(installed, update Application, installedManifest, updateManifest schema.Manifest) ([]loc.Locator, error) {
```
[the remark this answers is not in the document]
r0mant: It may not be the same manifest though. For example, the way upgrade plan builder uses it is it passes "runtime" ("base") applications here but the "main" app manifests.

lib/install/plan.go
```diff
@@ -82,6 +82,14 @@ func (r *Planner) GetOperationPlan(operator ops.Operator, cluster ops.Site, oper
 	builder.AddRBACPhase(plan)
 	builder.AddCorednsPhase(plan)
 
+	// create OpenEBS configuration if it's enabled, it has to be done
+	// before OpenEBS is installed during the runtime phase
+	if cluster.App.Manifest.OpenEBSEnabled() {
```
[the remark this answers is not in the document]
r0mant: Yeah, forgot to PR those.

docs/6.x/storage.md
```diff
@@ -0,0 +1,346 @@
...
+and tweak its parameters, if needed.
+
+## cStor Engine
+
+cStor is the most sophisticated storage engine supported by OpenEBS that provides
+features such as data replication, thin provisioning, snapshotting and so on.
+cStor serves block storage to containers over iSCSI interface.
+
+!!! note "iSCSI tools":
+    iSCSI tools must be installed on the cluster nodes in order for cStor volumes
```
knisbet: Just curious, why doesn't this work by just having the iscsi tools within planet??

lib/defaults/defaults.go
```diff
@@ -610,6 +610,8 @@ const (
 	KubeSystemNamespace = "kube-system"
 	// MonitoringNamespace is the name of k8s namespace for the monitoring-related resources
 	MonitoringNamespace = "monitoring"
+	// OpenEBSNamespace is the name of k8s namespace where OpenEBS is deployed.
+	OpenEBSNamespace = "openebs"
```
knisbet: Should we tie openebs to an "openebs" namespace, or simply tie it to a "storage" namespace, that's used by the "storage" app. Not a huge deal or anything, but might be slightly more consistent to match the two. Either openebs app -> openebs ns, or storage -> storage.

Thoughts?

lib/install/phases/openebs.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// createNamespace creates OpenEBS namespace if it doesn't exist.
+func (r *openebs) createNamespace() error {
+	_, err := r.Client.CoreV1().Namespaces().Create(&v1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{Name: defaults.OpenEBSNamespace},
+	})
+	if err != nil {
+		if !trace.IsAlreadyExists(rigging.ConvertError(err)) {
+			return trace.Wrap(err)
```
knisbet: I don't remember if the kubernetes error will include the namespace name, just in case should maybe `.AddField("namespace", defaults.OpenEBSNamespace)` for additional context

lib/schema/manifest.go
r0mant: You mean so it's not packaged during tele build? Yeah, this is something that's on my list of improvements, I haven't looked at what it'd take to implement this yet.

docs/6.x/storage.md
r0mant: Yes, if `include` is present, then only matching devices will be considered. If it's empty, then all devices will be considered, except those matching `exclude` filter. I think I'll add this explanation here and also link back to the OpenEBS docs.

docs/6.x/storage.md
```diff
@@ -0,0 +1,346 @@
...
+    Default OpenEBS filters will be merged with the filters configured via
+    a `PersistentStorage` resource so you don't need to add them explicitly.
+
+Suppose that we want to include only two specific devices, `/dev/sdd` and
+`/dev/sde`, in the OpenEBS pool. Let's define the following resource in a
+`storage.yaml` file:
+
+```yaml
+kind: PersistentStorage
+version: v1
```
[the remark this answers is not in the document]
r0mant: It's not a CR. It's just like most of our other resources - a "virtual" resource that gets translated directly into config map with openebs configuration.

docs/6.x/storage.md
r0mant: I've expanded the docs on this a bit, let me know if that's more clear now.

docs/6.x/storage.md
r0mant: It does :) At least in my brief testing I think it worked. The reason I didn't include it in this "first round" of integration, is I think just adding iscsi tools to planet is not enough, it should be a bit smarter, for example, the planet should start iscsid conditionally, only if it's not already present on host, otherwise it will fail and this will lead to degraded status. I just wanted to focus on local pv initially, we can work on smoother cstor integration after that.

lib/defaults/defaults.go
r0mant: That's an interesting thought, I'm not sure we should necessarily tie those together though. The way I think about it:

* We could change namespace to a generic `storage` but that would complicate upgrading a bit as we'd have to diligently change all namespaces in their operator.yaml. Probably wouldn't be such a big deal if/when we migrate to Helm though.
* If we ever add another storage provider, it'd make sense to put it in a separate namespace. For the same reason, I think renaming the app to specifically "openebs" is also not ideal.

On the other hand, we do put all prometheus/alertmanager stuff into "monitoring" namespace... A nice side effect with using `openebs` namespace is that this is their default namespace so all command in their documentation just work.

lib/ops/opsservice/update.go
```diff
@@ -526,13 +526,66 @@ func (s *site) checkUpdateParameters(update *pack.PackageEnvelope, provisioner s
...
+	// specific checks are required.
+	if !installedEnabled && updateEnabled {
+		// TODO(r0mant): Check if OpenEBS was installed "out of band" and bail out?
+		return nil
+	}
+
+	// OpenEBS was enabled before, but disabled in the new version, we don't
+	// support uninstalling it at the moment.
+	if installedEnabled && !updateEnabled {
+		return trace.BadParameter(`The cluster has OpenEBS integration enabled but it's disabled in the version you're trying to upgrade to.
```
a-palchikov: This means rebuilding the cluster image. Can we make it a warning and continue as if it was not disabled?

r0mant: We sure can, but I'm not sure we can safely make the assumption that the updated application will work properly if it says that OpenEBS should be disabled but it's, in fact, enabled. We'll also end up with discrepancy b/w the declared state (manifest) and the actual state.

a-palchikov:
The robotests are consistently failing on resize:
```
[ERROR]: failed to execute phase "/configure" {"message":"etcdserver: request timed out"}
```
```
[ERROR]: failed to execute phase "/configure" {"message":"rpc error: code = Unavailable desc = etcdserver: leader changed"}
```

r0mant:
Hmm, robotest actually passed on the second to the last commit, but one of unit tests was broken. Might be unrelated to the changes. I’ll launch again.

PR #885, opened 2019-11-18
[6.1.x] filter out the unrelated peer on join

a-palchikov:
Updates https://github.com/gravitational/gravity.e/issues/4247.
Optionally depends on https://github.com/gravitational/gravity.e/pull/4248.

### DIFF
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -1138,14 +1138,17 @@ func (p *Peer) validateWizardState(operator ops.Operator) (*ops.Site, *ops.SiteO
 
 	switch operation.State {
 	case ops.OperationStateInstallInitiated,
-		ops.OperationStateInstallProvisioning:
-		// Consider these states for resuming the installation
-	case ops.OperationStateInstallPrechecks,
+		ops.OperationStateInstallProvisioning,
+		ops.OperationStateInstallPrechecks,
+		ops.OperationStateInstallDeploying,
 		ops.OperationStateFailed:
+		if len(operation.Servers) == 0 {
+			return nil, nil, trace.NotFound("no servers joined yet")
+		}
 		// Consider these states for resuming the installation
 		// (including failed that puts the operation into manual mode).
 		// Be careful about including unrelated peers though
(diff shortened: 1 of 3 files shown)


lib/expand/join.go
```diff
@@ -1138,11 +1138,17 @@ func (p *Peer) validateWizardState(operator ops.Operator) (*ops.Site, *ops.SiteO
...
-		ops.OperationStateInstallPrechecks,
+		ops.OperationStateInstallProvisioning:
+		// Consider these states for resuming the installation
+	case ops.OperationStateInstallPrechecks,
 		ops.OperationStateFailed:
 		// Consider these states for resuming the installation
-		// (including failed that puts the operation into manual mode)
+		// (including failed that puts the operation into manual mode).
+		// Be careful about including unrelated peers though
+		if thisPeerPartOfInstallState(p.AdvertiseAddr, operation.Servers) {
```
r0mant: Why is this check only performed if the operation is in the `prechecks` and `failed` states? Trying to understand whether this covers all scenarios. E.g. what about the two states above (`initiated` and `provisioning`) and `deploying`?

lib/expand/join.go
```diff
@@ -1286,6 +1292,10 @@ func isExpandOperationReady(state string) (bool, error) {
 	}
 }
 
+func thisPeerPartOfInstallState(addr string, servers []storage.Server) bool {
```
r0mant: Nit: I think you could just accept `storage.Servers` and Go would do the type conversion automatically and you won't have to do it below.

tool/tele/main.go
```diff
@@ -34,7 +34,7 @@ func main() {
 	stdlog.SetOutput(log.StandardLogger().Writer())
 	app := kingpin.New("tele", "Gravity tool for building and publishing cluster and application images.")
 	if err := run(app); err != nil {
-		log.Error(trace.DebugReport(err))
+		log.WithError(err).Warn("Command failed.")
```
r0mant: Same q as on the "gravity.e" PR - did you mean to change this to warning?

Issue #884, opened 2019-11-16
[BUG] `tele build` cannot be used in GitLab CI with Kubernetes


zimmski:
**Describe the bug**
We are trying to use `tele build` inside of our CI (GitLab with the CI jobs running inside Kubernetes with Docker). However, the following error occurs in step 3 of the build `failed to push images to local registry`.

**To Reproduce**
Use a GitlabCI with Docker and Kubernetes

**Expected behavior**
This should just work. According to https://community.gravitational.com/t/cannot-build-app-tarballs-in-a-docker-container/68/3 the "host" network is needed. In the context of doing the build in Kubernetes, this does not make sense to me.

**Environment (please complete the following information):**
- Gravity 6.2.2

militarpancho:
Hi!, just in case it would be helpful I think that flag is mandatory to tele to build correctly...

militarpancho:
Another problem that I think I've encountered setting gravity with CI is that tele build does not support Docker TCP mode to use dind....I created this pr: https://github.com/gravitational/gravity/pull/886

zimmski:
Thanks for your answer. I will definitely have a look at your PR.

What I am wondering is why the "host" network is needed at all and how Gravity can be changed to simply not require this option anymore.

militarpancho:
I think the reason is performance...a lot of ports are used to vendor docker images detected in the yaml resources at build time. If your CI is running in Kubernetes, using dind instead of the unix socket in the same pod that the container that runs "tele" could avoid "host" flag to be needed.... (I suppose this because containers in the same pod shares network port space)

PR #886, opened 2019-11-18
Accept DOCKER_HOST env var in tele build

militarpancho:
Hi, concerning #884 I think it would be nice tele build read DOCKER_HOST variable to allow tcp mode in order to use dind in a CI environment and do not mess up with a lot of images the ci agents/runners. I don't write code in go usually but just in case would be helpful I start this pull request....

P.S: Sorry if this is not the way you guys usually use to contribute or something...

Thanks in advance

### DIFF
--- a/lib/builder/build.go
+++ b/lib/builder/build.go
@@ -121,7 +121,12 @@ func checkBuildEnv() error {
 		return trace.BadParameter("tele build is not supported on %v, only "+
 			"Linux is supported", runtime.GOOS)
 	}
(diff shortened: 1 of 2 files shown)


a-palchikov:
Thanks for your contribution. Unfortunately the CI is not set up to build external PRs so I'll have to make another with this change.

a-palchikov:
Superseded by https://github.com/gravitational/gravity/pull/905

Issue #676, opened 2019-09-17
Support Customer add-on image in bundle. 


benarent:
Support custom addon images to the bundle (via @knisbet : This comes up alot with lots of different customers. Hacking it by creating empty pods doesn’t seem like a realistic solution. I would consider either adding a CRD type object that is specifically for adding additional images that tele can understand, or adding additional images to the app.yaml (or both)

Issue #680, opened 2019-09-17
Use tele build with private docker registries


benarent:
How to use tele build with private docker registries.

PR #889, opened 2019-11-19
Add ability to provide additional images to vendor.

r0mant:
@a-palchikov @knisbet Take a look guys. I've added ability to provide additional images to vendor via manifest so you don't have to have fake pod spec anymore. I've extended the existing "dependencies" field as it seemed most fitting. This is how it looks like:

```yaml
dependencies:
  additionalImages:
  - nginx:1.17.5
  - quay.io/bitnami/redis:5.0
```

Also, added some docs on vendoring and this feature.

This can also be backported to 6.x.

Closes https://github.com/gravitational/gravity/issues/676, closes #https://github.com/gravitational/gravity/issues/680.

### DIFF
--- a/docs/6.x/pack.md
+++ b/docs/6.x/pack.md
@@ -131,6 +131,54 @@ docker run -e OPS_URL=<opscenter url> \
     by setting `--state-dir`. You can use unique temporary directory
     to avoid sharing state between builds, or use parallel builds instead.
 
+## Vendoring
+
+When you execute the `tele build` command, `tele` discovers all Docker images
+referenced by your cluster or application image resources (Helm charts or plain
+Kubernetes spec files) and packages them inside the resulting cluster or
+application image tarball. This process is called "vendoring" and it gives the
+Gravity images their "self-sufficiency" property: all vendored images are
+synchronized with the in-cluster private Docker registry when the cluster is
+installed and Kubernetes pulls them from this local registry when spinning up
+pods.
+
+#### Private Docker Registries
+
+To be able to vendor a Docker image, it needs to be available via the local
+Docker client. To see a list of images for the local Docker client, you can
+typically use the `docker images` command.
+
+If an image being vendored is not available via the local Docker client, `tele`
+will attempt to pull it from a remote registry specified by the image, or the
+default [Docker Hub](https://hub.docker.com) if the registry is not specified.
+
+In case the image belongs to a private Docker registry, your local Docker client
+must be configured with proper credentials for it (e.g. via `docker login` or
+[TLS certificates](https://docs.docker.com/engine/security/certificates/)) in
+order for `tele` to be able to pull it.
+
+#### Image References Discovery
+
+The `tele` tool can extract image references from all core Kubernetes objects
+such as pods, deployments, replica and daemon sets and so on.
+
+!!! note "Helm chart images":
+    If the cluster image resources include Helm charts, `tele` will render the
+    Helm templates before extracting the image references. Providing a custom
+    values file for template rendering is not supported at the moment.
+
+For some objects, `tele` can't extract images because their schema is not known
+beforehand. An example would be Kubernetes custom resources. To handle this
+scenario, you can declare additional images that should be vendored in the
+manifest file:
+
+```yaml
+dependencies:
+  additionalImages:
+  - nginx:1.17.5
+  - quay.io/bitnami/redis:5.0
+```
+
 ## Image Manifest
 
 The Image Manifest is a YAML file that is passed as an input to `tele build`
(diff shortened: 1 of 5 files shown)


docs/6.x/pack.md
```diff
@@ -131,6 +131,54 @@ docker run -e OPS_URL=<opscenter url> \
...
+
+To be able to vendor a Docker image, it needs to be available via the local
+Docker client. To see a list of images for the local Docker client, you can
+typically use the `docker images` command.
+
+If an image being vendored is not available via the local Docker client, `tele`
+will attempt to pull it from a remote registry specified by the image, or the
+default [Docker Hub](https://hub.docker.com) if the registry is not specified.
+
+In case the image belongs to a private Docker registry, your local Docker client
```
a-palchikov: `belongs to` -> resides in

docs/6.x/pack.md
```diff
@@ -131,6 +131,54 @@ docker run -e OPS_URL=<opscenter url> \
...
+
+The `tele` tool can extract image references from all core Kubernetes objects
+such as pods, deployments, replica and daemon sets and so on.
+
+!!! note "Helm chart images":
+    If the cluster image resources include Helm charts, `tele` will render the
+    Helm templates before extracting the image references. Providing a custom
+    values file for template rendering is not supported at the moment.
+
+For some objects, `tele` can't extract images because their schema is not known
```
a-palchikov: objects -> resources (or resource files)

docs/6.x/pack.md
```diff
@@ -131,6 +131,54 @@ docker run -e OPS_URL=<opscenter url> \
...
+beforehand. An example would be Kubernetes custom resources. To handle this
```
a-palchikov: nit: custom Kubernetes resources

knisbet:
I thought the original discussion around this was using something more like a CRD or custom object, with the potential for exploring a mutating admission controller that could attempt to transparently rewrite some of these image paths at run time (I think Sasha mentioned he had a request along those lines)??

IE, the development flow for something like a helm chart with an operator, would be to drop the CRD with required images as part of the helm chart itself, and tele when it scans the chart can then identify the images used by that operator/chart. The additional images then just become an object file, that can be included by individual app developers, works with "gravity applications" and not just images, etc.

r0mant:
@knisbet Yeah, I remember that. However, what I'm thinking is that instead of introducing another CRD specifically for images, I think we should rather make our manifest a proper CRD, eventually. We already ask users to write a single "gravity-specific" resource file - manifest - and we were talking before about adding manifest support to our "application images" (which you build out of Helm charts) as well so they can take same advantages as our "cluster images" in regards to requirements checks, dependencies and so on. And then you'll be able to drop your manifest file among your resources, just like you said, and there may be a some kind of controller in the cluster that monitors it. With this approach I feel like having additional image deps in manifest makes sense.

knisbet:
Well, the reason I think it makes some sense to separate out, is it makes things less tightly coupled. If this operates as it's own object, it's now just an object anyone can include as part of their resources where relevant. If a customer wants three objects, they can just create three objects. The controller itself wouldn't even need to be part of gravity, it's just a behavior.

 When we're looking at customers doing hybrid applications, where a team might just be responsible for developing a helm chart, should they be editing the application manifest to include additional images. Also, if including a mutating controller is part of the future, we may need to be aware of additional fields or structures that are needed. Such as vendor/download the image from internal.registry/gravitational, but rewrite request that are really for quay.io/gravitational/debian

r0mant:
@knisbet Yeah-yeah, it does make sense... as does having it in the manifest I think... IIRC the idea was kinda the same that users would just drop it among their resources and we'll register it as a CRD and the controller like you said doesn't even have to be a part of gravity cluster. The sort of downsides of separating it out I see is that _a)_ this is yet another resource type for us to maintain/support and for users to learn and _b)_ to add other features (like additional deps, requirements, etc.) we'll then also end up introducing CRDs and end up with a zoo of those as opposed to having a per-cluster or per-app manifest that contains everything.

I think this is ultimately the product decision. Since @benarent is not here to make this call, I wonder if @klizhentas or @kontsevoy have an opinion on this.

r0mant:
@knisbet I spent more time thinking about it and came to the conclusion that using a custom resource is indeed a much better approach in this case. It will also play really nice with using admission controller where tele build would just auto-create an instance of this custom resource when scanning charts/specs and we won't have to do any rewriting at all.

So I'm gonna close this one and work on the CR way instead. Let me know what you think. I have created a design doc to collect this and other tele build improvements: https://docs.google.com/document/d/1ms8zW37DWOvH4nWPbQZxYBVxlkqQ1NOZ1f1KfQdxynY/edit?usp=sharing.

Issue #698, opened 2019-09-18
Default UID for planet and conflicts with system users


benarent:
Default UID for planet and conflicts with system users.

PR #901, opened 2019-11-20
use a uid/gid less likely to conflict with a system user

knisbet:
Fixes https://github.com/gravitational/gravity/issues/698

Change the default UID/GID to be one far less likely to conflict with one already configured on the system. I'm not sure if there is a good reason to keep 1000, but when I did basic install/upgrade sanity testing I didn't see anything stand out.

980665 is from the standard gravitational constant with decimal removed (https://en.wikipedia.org/wiki/Standard_gravity), which seemed fitting.


helgi:
I have a few places where I used `1000` for operations to make things work with gravity as I never made a user that matched that ID on the host. Implication here on any automation, user-data or charts that pin to 1000 (as k8s can only do numerical IDs, not named, in securityContext)

knisbet:
> I have a few places where I used `1000` for operations to make things work with gravity as I never made a user that matched that ID on the host. Implication here on any automation, user-data or charts that pin to 1000 (as k8s can only do numerical IDs, not named, in securityContext)

Well, this can be maintained by using the installer flags to set the uid/gid at install time. Or creating a planet user on the host with uid 1000, which will then not invoke these defaults.

> My only "concern" is that it's over 65535 which I don't think should cause any problems on any modern distros that support 32-bit uids but I wonder if we'd be on a bit safer side picking a number in the 16-bit range just in case. Like I said, maybe not a concern though (since it can always be overridden by a flag).

I believe all the distros we support have 32 bit UIDs. According to https://en.wikipedia.org/wiki/User_identifier the 32 bit UIDs were introduced with linux 2.4 in 2001. Although some distros set the  /etc/login.defs for local users to create uid's less than 65k, I believe that only affects the useradd commands for generating a uid.


> Also: planet needs this as a default for the container image.

Maybe I'm missing something, but why? If this is the case wouldn't the installer flags to set the UID to a custom value not work?

PR #905, opened 2019-11-22
Accept DOCKER_HOST env var in tele build

a-palchikov:
Double of https://github.com/gravitational/gravity/pull/886 to run robotests.

Requires https://github.com/gravitational/gravity.e/pull/4252.


lib/builder/build.go
```diff
@@ -121,7 +121,8 @@ func checkBuildEnv() error {
 		return trace.BadParameter("tele build is not supported on %v, only "+
 			"Linux is supported", runtime.GOOS)
 	}
-	client, err := docker.NewClient(constants.DockerEngineURL)
+	endpoint := utils.GetenvWithDefault("DOCKER_HOST", constants.DockerEngineURL)
+	client, err := docker.NewClient(endpoint)
```
r0mant: I think that Vendorer (inside Builder) also uses the hardcoded default Docker URL (https://github.com/gravitational/gravity/blob/master/lib/builder/builder.go#L322). I would consider passing Docker URL to the Builder via `builder.Config`.

tool/gravity/cli/app.go
```diff
@@ -181,11 +182,22 @@ func importApp(env *localenv.LocalEnvironment, registryURL, dockerURL, source st
 	}
 
 	if req.Vendor {
+		dockerClient, err := docker.NewDefaultClient()
```
r0mant: Should this consider `dockerURL` passed to this method? It's used above, when creating AppService.

Issue #852, opened 2019-10-30
tele build bad progress reporter UX


r0mant:
In `tele build` we currently hardcode the number of steps the progress reporter expects and if it happens that the actual number of steps does not match, it says "aborted" at the end which gives the impression that the build failed:

```
* [9/6] Saving the snapshot as installer.tar
Still saving the snapshot as installer.tar (10 seconds elapsed)
Still saving the snapshot as installer.tar (20 seconds elapsed)
Still saving the snapshot as installer.tar (30 seconds elapsed)
...
Still saving the snapshot as installer.tar (1 minute elapsed)
Still saving the snapshot as installer.tar (1 minute elapsed)
• [9/6] Build aborted after 16 minutes 
```

I think we should get rid of the number of steps altogether and just print timestamps (`env.PrintStep()`) like we do for all operations.

Issue #679, opened 2019-09-17
Tele build error is confusing.


benarent:
Tele build error is confusing: 

- [ ] Will skip unrecognised object in 
```
Helm chart charts/provisioner: apiVersion=/, kind=
```
- [ ]  tele build non-inutive error:
```
[ERROR]: error converting YAML to JSON: yaml: line 32: did not find expected key (On an apparently malformed DaemonSet)
```

PR #907, opened 2019-11-22
tele build UX updates.

r0mant:
A few UX improvements to tele build:

* Fix the confusing `unrecognized object: apiVersion=/, kind=` error which was due to empty resource files which our decoder didn't ignore and returned empty resources instead.
* Do not print "unrecognized resource" warning for each resource in tele build output, they can accumulate, pollute the output and confuse the user. Instead, log them and print a single warning to tell them to look in the logs if needed.
* Get rid of the steps in the tele build output which always got out of sync. Instead, use our standard progress output with timestamps.
* Add more context to resource parsing errors. Previously, it would just print a generic "json parse error" with no way of telling which resource file it's for. Now it prints the name of the file with the error, and the actual error.
* Improve how errors with multiple user messages are displayed. See https://github.com/gravitational/trace/pull/53 for details (which this PR requires).

Closes https://github.com/gravitational/gravity/issues/679, closes https://github.com/gravitational/gravity/issues/852.

### DIFF
--- a/lib/app/service/vendor.go
+++ b/lib/app/service/vendor.go
@@ -340,7 +340,6 @@ func analyzeResources(resourceFiles, chartFiles resources.ResourceFiles, req Ven
 //    has been detected
 func printResourceStatus(resourceFile resources.ResourceFile, req VendorRequest) error {
 	relPath := utils.TrimPathPrefix(resourceFile.Path(), filepath.Dir(req.ManifestPath))
-	//req.ProgressReporter.PrintSubDebug("Detected %v %v", resourceFile.Kind(), relPath)
 	extractedImages, err := resourceFile.Images()
 	if err != nil {
 		return trace.Wrap(err)
--- a/lib/app/service/vendor.go
+++ b/lib/app/service/vendor.go
@@ -17,7 +17,6 @@ limitations under the License.
 package service
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -28,7 +27,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/ghodss/yaml"
+	"github.com/gravitational/gravity/lib/app/docker"
 	"github.com/gravitational/gravity/lib/app/hooks"
 	"github.com/gravitational/gravity/lib/app/resources"
 	"github.com/gravitational/gravity/lib/archive"
@@ -43,6 +42,7 @@ import (
 	"github.com/gravitational/gravity/lib/utils"
 
 	dockerarchive "github.com/docker/docker/pkg/archive"
+	"github.com/ghodss/yaml"
 	teleutils "github.com/gravitational/teleport/lib/utils"
 	"github.com/gravitational/trace"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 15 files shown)


lib/app/service/vendor.go
```diff
@@ -342,10 +342,25 @@ func printResourceStatus(resourceFile resources.ResourceFile, manifestPath strin
...
 	for _, o := range extractedImages.UnrecognizedObjects {
 		gvk := o.GetObjectKind().GroupVersionKind()
-		progressReporter.PrintSubStep("Will skip unrecognized object in %v %v: apiVersion=%v/%v, kind=%v",
-			resourceType, relPath, gvk.Group, gvk.Version, gvk.Kind)
+		unk, err := resources.ToUnknown(o)
+		if err != nil {
+			log.WithError(err).WithFields(log.Fields{
+				"apiVersion": fmt.Sprintf("%v/%v", gvk.Group, gvk.Version),
+				"kind":       gvk.Kind,
+			}).Error("Failed to convert object to unknown resource.")
```
a-palchikov: For me, logging an error should be the final step of the failing process. If it's not terminal, it is not an error anymore.

lib/app/service/vendor.go
```diff
@@ -701,13 +716,22 @@ func isChartDirectory(path string) (bool, error) {
...
 		return nil, trace.Wrap(err)
 	}
-	log.WithField("path", path).Debug(string(out))
-	return resources.Decode(bytes.NewReader(out))
+	var result resources.ResourceFiles
+	for k, v := range rendered {
+		resource, err := resources.Decode(strings.NewReader(v))
+		if err != nil {
+			return nil, trace.Wrap(err).
+				AddUserMessage("%v", err).
```
a-palchikov: hmm, this is a weird use of the user message. Isn't this what `Wrap` is supposed to preserve in the first place?

lib/app/service/vendor.go
```diff
@@ -747,11 +771,12 @@ func resourcesFromPath(root string, includePatterns []string, ignorePatterns []s
 				return nil
 			}
 			log.Infof("Extracting images from Helm chart directory %v.", path)
-			resource, err := resourceFromChart(path)
+			resource, err := resourcesFromChart(path)
 			if err != nil {
-				return trace.Wrap(err)
+				return trace.Wrap(err).AddUserMessage("failed to parse Helm chart resources %q",
```
a-palchikov: here, saving a field would've been more appropriate:
```go
return trace.Wrap(err, "failed to parse Helm chart resources").AddField("path", path)
```

lib/app/service/vendor.go
```diff
@@ -769,8 +794,8 @@ func resourcesFromPath(root string, includePatterns []string, ignorePatterns []s
 		}
 		resourceFile, err := resources.NewResourceFile(path)
 		if err != nil {
-			return trace.Wrap(err, "failed to parse resource file %q: %v",
-				utils.TrimPathPrefix(path, root, defaults.ResourcesDir), err)
+			return trace.Wrap(err).AddUserMessage("%v", err).
```
a-palchikov: Are you worried that this loses the error context so it is additionally logged as a string?

lib/app/service/vendor.go
r0mant: No, actually - the way it works now is Wrap does not add a user message unless it was provided explicitly like `trace.Wrap(err, "user message", ...)`. We could change that to extract a user message from the error although I think it might mess up all our current error reporting. Or I could just do `trace.Wrap(err, "%v", err)` here which would have the same effect.

lib/app/service/vendor.go
r0mant: Fields are not printed when the error is printed though (I don't think). Also, this is the error that users will see so I think it should be formatted properly and fields are more suitable for logs than for user-facing messages.

lib/app/service/vendor.go
r0mant: No, the error's call stack is logged in full, but the users get a nice message (like, the example on the trace PR).

lib/app/service/vendor.go
a-palchikov: Fields are only dumped with `DebugReport` yes.

lib/app/service/vendor.go
```diff
@@ -329,10 +329,25 @@ func printResourceStatus(resourceFile resources.ResourceFile, manifestPath strin
 	if len(extractedImages.Images) == 0 {
 		progressReporter.PrintSubStep("Found no images to vendor in %v %v", resourceType, relPath)
 	}
+	if len(extractedImages.UnrecognizedObjects) > 0 {
+		progressReporter.PrintSubWarn("Some resources in %v %v weren't recognized, check --debug output",
```
knisbet: So I'm not a huge fan of suppressing the resource warnings behind the debug flag. I interpret the debug UX pattern as I want to troubleshoot tele's behaviour, and get a bunch of debug info about tele itself to troubleshoot, not that this is a pattern that should be used for digging through and debugging an application.

My 2 cents, is I think it's better to print the output like we were doing before, but now that the unrecognized object stuff is fixed, just make sure it's clean, including the filename, line number if possible, etc. If you really want to hide it, then a UX pattern more like `tele build --verbose` for better application debug information may be warranted.

r0mant: Yeah, I wanna hide it by default b/c otherwise people who have a lot of custom resources will be constantly seeing this and even with cleaned up output, these can amount to a lot of noise.

I like the idea of having `--verbose` flag that will print more detailed info including these warning, can do that too.

lib/app/service/vendor.go
```diff
@@ -347,6 +365,8 @@ func printResourceStatus(resourceFile resources.ResourceFile, manifestPath strin
 				"kind":       gvk.Kind,
 				"name":       unk.Metadata.Name,
 			}).Info("Skip unrecognized object.")
+			req.ProgressReporter.PrintSubDebug(color.BlueString("Unrecognized object: apiVersion=%v; kind=%v; name=%v",
```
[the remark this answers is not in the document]
r0mant: lol, it's the same but in blue 😄I just wanted to have these lines stand out from the rest of debug output so it's easier to spot them.

Issue #903, opened 2019-11-21
Add ability to use custom values files in tele build for helm charts


UnderGreen:
For now, tele build does not respect any custom values files for Helm charts, only the default `values.yaml` inside chart. It can cause problems with vendoring images if one uses different images or tags for images in the custom files. Only default images from values.yaml will be vendored in a tarball.

PR #909, opened 2019-11-26
Add ability to provide helm values to tele build.

r0mant:
Add ability to override Helm values during tele build via `--set` and `--values` flags - flag names are the same as in helm commands and gravity app commands. Closes https://github.com/gravitational/gravity/issues/903.

Should be pretty easy to backport to 6.x as well.

### DIFF
--- a/tool/tele/cli/register.go
+++ b/tool/tele/cli/register.go
@@ -54,6 +54,8 @@ func RegisterCommands(app *kingpin.Application) Application {
 	tele.BuildCmd.Parallel = tele.BuildCmd.Flag("parallel", "Specifies the number of concurrent tasks. If < 0, the number of tasks is not restricted, if unspecified, then tasks are capped at the number of logical CPU cores.").Int()
 	tele.BuildCmd.Quiet = tele.BuildCmd.Flag("quiet", "Suppress any output to stdout.").Short('q').Bool()
 	tele.BuildCmd.Verbose = tele.BuildCmd.Flag("verbose", "Produce more detailed build output, can be useful for troubleshooting.").Short('v').Bool()
+	tele.BuildCmd.Set = tele.BuildCmd.Flag("set", "Set Helm chart values on the command line. Can be specified multiple times and/or as comma-separated values: key1=val1,key2=val2.").Strings()
+	tele.BuildCmd.Values = tele.BuildCmd.Flag("values", "Set Helm chart values from the provided YAML file. Can be specified multiple times.").Strings()
 
 	tele.ListCmd.CmdClause = app.Command("ls", "List cluster and application images published to Gravity Hub.")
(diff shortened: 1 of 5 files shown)


tool/tele/cli/register.go
```diff
@@ -53,6 +53,8 @@ func RegisterCommands(app *kingpin.Application) Application {
 	tele.BuildCmd.SkipVersionCheck = tele.BuildCmd.Flag("skip-version-check", "Skip version compatibility check.").Hidden().Bool()
 	tele.BuildCmd.Parallel = tele.BuildCmd.Flag("parallel", "Specifies the number of concurrent tasks. If < 0, the number of tasks is not restricted, if unspecified, then tasks are capped at the number of logical CPU cores.").Int()
 	tele.BuildCmd.Quiet = tele.BuildCmd.Flag("quiet", "Suppress any output to stdout.").Short('q').Bool()
+	tele.BuildCmd.Set = tele.BuildCmd.Flag("set", "Set Helm chart values on the command line. Can specify multiple or comma-separated: key1=val1,key2=val2.").Strings()
```
a-palchikov: `Can be specified multiple times or as a comma-separated value.`

tool/tele/cli/register.go
```diff
@@ -53,6 +53,8 @@ func RegisterCommands(app *kingpin.Application) Application {
...
+	tele.BuildCmd.Values = tele.BuildCmd.Flag("values", "Set Helm chart values from the provided YAML file(-s).").Strings()
```
a-palchikov: `... from the provided YAML file. Can be specified multiple times.`

McKeownDesign:
Will this provide the same functionality that executing a "helm install" with value overwriting from multiple value files ex.
helm install . -n nodde_name --namespace dev -f ./values.yaml -f ./dev.yaml --debug --dry-run

With this command dev.yaml overwrites values in values.yaml, if this the functionality that was built into this change? 

Would I do something like:
tele build -o project.tar mattermost/resources/app.yaml --values values.yaml, dev.yaml

Sorry for the question but I know it just got merged and I don't see it in the documentation.

r0mant:
@McKeownDesign Yes, you should be able to do that and internally these arguments are just passed to the same API used by helm template command so the behavior should be identical.

Just to clarify though, these values are only used to render helm templates in order to extract image references from them during tele build. During cluster installation, these values should still be passed to the helm install command.

McKeownDesign:
@r0mant Where do you pass these to to helm install command? Is that using the --config flag on the gravity install? I don't see in the documentation where I can pass in similar helm install commands.

r0mant:
@McKeownDesign Would you mind asking this question on https://community.gravitational.com for easier future discoverability please? And I'll reply there. Thanks!

PR #910, opened 2019-11-26
lib/blob/cluster: fix dropped test errors in RCPSuite.SetUpTest()

alrs:
This picks up two dropped test errors in `lib/blob/cluster`.


r0mant:
@alrs Good catch 👍 Thank you for your contribution!

PR #917, opened 2019-11-27
Expose kubeproxy port.

r0mant:
https://github.com/gravitational/gravity/pull/913 backport.

### DIFF
--- a/lib/constants/constants.go
+++ b/lib/constants/constants.go
@@ -19,6 +19,7 @@ limitations under the License.
 package constants
 
 import (
+	"fmt"
 	"time"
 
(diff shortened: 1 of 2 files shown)


lib/constants/constants.go
```diff
@@ -774,6 +778,12 @@ var (
 	HubImageName = "hub"
 )
 
+// ExternalDNS formats the provided hostname as a wildcard A record for use
+// with external DNS provisioners.
+func ExternalDNS(hostname string) string {
```
a-palchikov: nit: I thought this would somehow use `ExternalDNSHostnameAnnotation` - otherwise it belongs in `lib/utils/net` or something like that.

PR #939, opened 2019-12-06
Dmitri/test nightly

a-palchikov:
This PR wants to merge the branch that has been running as the nightly.
It fixes the following issues:
  * `gravity status` to exit with a non-zero code if the cluster is degraded
  * `gravity status` to output errors to `stderr` so they don't get mingled with the output which is especially critical for alternative output formats like `JSON`
  * `gravity report` to grab the host journal logs for everything (and not a select few `SYSLOG_IDENTIFIER`s which caused more issues than it was worth - mostly failing to capture the information that was actually required to troubleshoot)

It also temporarily switches to a robotest branch which I'll update once the robotest branch has been merged.


lib/report/system.go
```diff
@@ -97,27 +102,21 @@ func planetServices() Collectors {
...
-		return fmt.Sprintf("SYSLOG_IDENTIFIER=%v", id)
-	}
-	matches := []string{
-		syslogID("./gravity"),
-		syslogID("gravity"),
-		syslogID(defaults.GravityBin),
-	}
-
-	return Script("gravity-system.log.gz", fmt.Sprintf(script, strings.Join(matches, " ")))
+/bin/journalctl --no-pager --output=export --since=yesterday | /bin/gzip -f`
```
r0mant: Why not everything but since yesterday?

lib/utils/exec.go
```diff
@@ -128,6 +131,27 @@ func RunStream(ctx context.Context, w io.Writer, args ...string) error {
...
 }
 
+// ExecUnprivileged executes the specified command as unprivileged user
+func ExecUnprivileged(ctx context.Context, command string, args []string, opts ...CommandOptionSetter) error {
+	nobody, err := user.Lookup("nobody")
+	if err != nil {
+		return trace.ConvertSystemError(err)
+	}
+	cmd := exec.CommandContext(ctx, command, args...)
+	uid, _ := strconv.Atoi(nobody.Uid)
```
r0mant: Do we have to ignore the error here and below?

lib/report/system.go
```diff
@@ -69,11 +70,13 @@ func basicSystemInfo() Collectors {
...
 		Cmd("lscpu", "lscpu"),
 		Cmd("lsmod", "lsmod"),
-		Cmd("running-processes", "/bin/ps", "aux", "--forest"),
-		Cmd("systemctl-host", "/bin/systemctl", "status"),
-		Cmd("dmesg", "cat", "/var/log/dmesg"),
+		Cmd("running-processes", "/bin/ps", "auxZ", "--forest"),
+		Cmd("host-system-status", "/bin/systemctl", "status", "--full"),
+		Cmd("host-system-failed", "/bin/systemctl", "--failed", "--full"),
+		Cmd("host-system-jobs", "/bin/systemctl", "list-jobs", "--full"),
+		Cmd("dmesg", "dmesg", "--raw"),
```
r0mant: `/bin/dmesg`?

PR #950, opened 2019-12-10
Extract images from ImageSet.

r0mant:
Consider custom `ImageSet` resources when extracting image references for vendoring.

Requires https://github.com/gravitational/lens/pull/1, closes https://github.com/gravitational/gravity/issues/676, closes https://github.com/gravitational/gravity/issues/680.

@benarent Please review the docs.


Gopkg.toml
```diff
@@ -222,6 +222,10 @@ ignored = [
   name = "github.com/gravitational/tail"
   version = "1.0.1"
 
+[[constraint]]
+  name = "github.com/gravitational/lens"
+  branch = "roman/crd"
```
r0mant: Will update when the lens PR gets merged.

docs/6.x/pack.md
```diff
@@ -131,6 +131,59 @@ docker run -e OPS_URL=<opscenter url> \
...
+Kubernetes spec files) and packages them inside the resulting cluster or
+application image tarball. This process is called "vendoring" and it gives the
+Gravity images their "self-sufficiency" property: all vendored images are
+synchronized with the in-cluster private Docker registry when the cluster is
+installed and Kubernetes pulls them from this local registry when spinning up
+pods.
+
+#### Private Docker Registries
+
+To be able to vendor a Docker image, it needs to be available via the local
```
a-palchikov: `To vendor a docker image, the image needs to be available locally`.

docs/6.x/pack.md
```diff
@@ -131,6 +131,59 @@ docker run -e OPS_URL=<opscenter url> \
...
+Docker client. To see a list of images for the local Docker client, you can
+typically use the `docker images` command.
+
+If an image being vendored is not available via the local Docker client, `tele`
+will attempt to pull it from a remote registry specified by the image, or the
```
a-palchikov: `the remote registry`

docs/6.x/pack.md
```diff
@@ -131,6 +131,59 @@ docker run -e OPS_URL=<opscenter url> \
...
+
+The `tele` tool can extract image references from all core Kubernetes objects
+such as pods, deployments, replica and daemon sets and so on.
+
+!!! note "Helm chart images":
+    When building cluster or application image out of a Helm chart, `tele` will
+    render the Helm templates before extracting the image references. You can
+    use `--set` and `--values` flags to provide custom Helm values to `tele build`.
+
+For some objects, `tele` can't extract images because their schema is not known
```
a-palchikov: `tele` can only extract images from resource types it knows about. The `ImageSet` resource allows to list additional
image references which `tele` would otherwise not be able to extract (i.e. from custom resource types):

docs/6.x/pack.md
```diff
@@ -131,6 +131,59 @@ docker run -e OPS_URL=<opscenter url> \
...
+resources:
+
+```yaml
+apiVersion: lens.gravitational.io/v1beta1
+kind: ImageSet
+metadata:
+  name: extra-images
+spec:
+  images:
+  - image: nginx:1.11.0
```
a-palchikov: Does this support digest specification?
I know we have (had?) issues with pulling image references with digests.

a-palchikov:
Need more time for lens PR though.

docs/6.x/pack.md
```diff
@@ -131,6 +131,59 @@ docker run -e OPS_URL=<opscenter url> \
     by setting `--state-dir`. You can use unique temporary directory
     to avoid sharing state between builds, or use parallel builds instead.
 
+## Vendoring
```
knisbet: Are you planning to backport to 6.x?

docs/6.x/pack.md
```diff
@@ -131,6 +131,59 @@ docker run -e OPS_URL=<opscenter url> \
...
+
+When you execute the `tele build` command, `tele` discovers all Docker images
+referenced by your cluster or application image resources (Helm charts or plain
+Kubernetes spec files) and packages them inside the resulting cluster or
+application image tarball. This process is called "vendoring" and it gives the
+Gravity images their "self-sufficiency" property: all vendored images are
+synchronized with the in-cluster private Docker registry when the cluster is
```
knisbet: I don't know if it's necessary to describe this as synchronized, vs simply stating that this is how we ensure all images referenced at build time are available after installation.

docs/6.x/pack.md
```diff
@@ -131,6 +131,59 @@ docker run -e OPS_URL=<opscenter url> \
...
+installed and Kubernetes pulls them from this local registry when spinning up
+pods.
+
+#### Private Docker Registries
+
+To be able to vendor a Docker image, it needs to be available via the local
+Docker client. To see a list of images for the local Docker client, you can
+typically use the `docker images` command.
```
knisbet: I'm not sure we should be pointing someone to the `docker images` command, as I know this has lead to confusion in the past. Docker images only shows what is locally available, not necessarily what is available for docker to pull from a registry.

lib/app/resources/resourcefiles.go
```diff
@@ -343,6 +344,10 @@ func extractImages(objects []runtime.Object) (*ExtractedImages, error) {
 					containers = append(containers, job.Spec.Template.Spec.InitContainers...)
 				}
 			}
+		case *lensv1beta1.ImageSet:
+			for _, spec := range resource.Spec.Images {
```
knisbet: I'm not sure if I'm missing something, it's unclear to me how the registry field comes into play?

docs/6.x/pack.md
knisbet: no docs on the registry field?

lib/app/resources/resourcefiles.go
r0mant: So I was thinking the registry field will be useful when we add the actual admission server which will be looking at it and pulling images from the respective registry if it's specified. For users who just want to vendor additional images for now it's not really needed.

Or are you suggesting that if the registry field is specified, these images should be rewritten to point to this registry currently? We could do that but I don't think there is a real use-case for it at the moment, and the rewriting code will be going away soon anyway.

docs/6.x/pack.md
r0mant: No, forgot to add a note that this new resource is only supported in 7.0.

docs/6.x/pack.md
r0mant: I haven't touched pulling logic so whatever issues we have/had are likely still there. Might be a good  candidate to fix in the scope of these vendoring improvements though (but not in this PR).

lib/app/resources/resourcefiles.go
knisbet: Sorry, wasn't suggesting anything, I couldn't see how it was intended to be used, so I wasn't sure if this was for what we discussed in the design, whether we want the ability to pull from an alternate location, and whether the field registry was for that intended purpose.

docs/6.x/pack.md
benarent: If it's only in 7.x I would fork the docs and only put these updates there.

docs/6.x/pack.md
```diff
@@ -131,6 +131,57 @@ docker run -e OPS_URL=<opscenter url> \
...
+    When building cluster or application image out of a Helm chart, `tele` will
+    render the Helm templates before extracting the image references. You can
+    use `--set` and `--values` flags to provide custom Helm values to `tele build`.
+
+A special `ImageSet` custom resource allows to list additional images to vendor,
+which `tele` would otherwise not be able to extract (for example, from custom
+resource types):
+
+```yaml
+apiVersion: lens.gravitational.io/v1beta1
```
benarent: Maybe we should a reference to what 'lens.gravitational.io` is, I would add a code comment above `kind: ImageSet` and  put the warning higher.

r0mant:
@a-palchikov @knisbet FYI moved everything from the [lens PR](https://github.com/gravitational/lens/pull/1) here. I think it's actually quite nice as we now have a framework in Gravity to add more custom resources should we need to.

I'm going to merge this once CI passes.

Issue #944, opened 2019-12-10
[BUG] gravity status response time increases with additional namespaces in 6.2.x clusters


tosih:
**Describe the bug**
The response time for `gravity status` requests increases linearly with the amount of namespaces installed in a cluster. As a result, gravity status can timeout on clusters with a full set of helm releases. This causes issues for our infrastructure, as we depend on information from gravity for processes such as scaling and rolling our clusters.

**To Reproduce**
On a fresh 6.2.5 gravitational cluster.
1. Create 30+ namespace objects.
```
$ tele login --hub <opscenter_url> --auth Keycloak <cluster_name>
$ echo {1..30} | xargs -n1 kubectl create namespace
namespace/1 created
namespace/2 created
...
namespace/29 created
namespace/30 created
```
2. Run `gravity status` on a master node and note the time it takes to get a response.

**Expected behavior**
The response from `gravity status` should be within a second.

**Logs**
```[root@ip-10-0-2-173 ~]# gravity status --debug
2019-12-06T14:45:05-08:00 DEBU             got search paths: [/var/lib/gravity assets/local] processconfig/config.go:57
2019-12-06T14:45:05-08:00 DEBU             look up configs in /var/lib/gravity processconfig/config.go:59
2019-12-06T14:45:05-08:00 DEBU             /var/lib/gravity/gravity.yaml not found in search path processconfig/config.go:67
2019-12-06T14:45:05-08:00 DEBU             look up configs in assets/local processconfig/config.go:59
2019-12-06T14:45:05-08:00 DEBU             assets/local/gravity.yaml not found in search path processconfig/config.go:67
2019-12-06T14:45:05-08:00 DEBU [LOCAL]     Creating local environment. args:{ /var/lib/gravity/local false false true 0s 0s 0x2586bf0 {[] 0} false} localenv/localenv.go:152
2019-12-06T14:45:05-08:00 DEBU             got search paths: [/var/lib/gravity assets/local] processconfig/config.go:57
2019-12-06T14:45:05-08:00 DEBU             look up configs in /var/lib/gravity processconfig/config.go:59
2019-12-06T14:45:05-08:00 DEBU             /var/lib/gravity/gravity.yaml not found in search path processconfig/config.go:67
2019-12-06T14:45:05-08:00 DEBU             look up configs in assets/local processconfig/config.go:59
2019-12-06T14:45:05-08:00 DEBU             assets/local/gravity.yaml not found in search path processconfig/config.go:67
[...]

PR #952, opened 2019-12-10
(6.1) More efficient app endpoints retrieval.

r0mant:
When retrieving endpoints, make a single API call to get services across all namespaces rather than iterating over namespaces.

Closes https://github.com/gravitational/gravity/issues/944. Needs port to 6.2 and master.

### DIFF
--- a/lib/ops/opsservice/endpoints.go
+++ b/lib/ops/opsservice/endpoints.go
@@ -75,10 +75,6 @@ func (o *Operator) GetApplicationEndpoints(key ops.SiteKey) ([]ops.Endpoint, err
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
-	namespaceList, err := client.CoreV1().Namespaces().List(metav1.ListOptions{})
-	if err != nil {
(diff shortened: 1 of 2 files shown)


lib/ops/opsservice/endpoints.go
```diff
@@ -75,30 +75,18 @@ func (o *Operator) GetApplicationEndpoints(key ops.SiteKey) ([]ops.Endpoint, err
...
-			})
-			if err != nil {
-				return nil, trace.Wrap(err)
-			}
-			if serviceList == nil {
-				serviceList = services
-			} else {
-				serviceList.Items = append(serviceList.Items, services.Items...)
-			}
+		serviceList, err := client.CoreV1().Services(constants.AllNamespaces).List(metav1.ListOptions{
```
a-palchikov: I'm not so sure using the empty string will query across all namespaces - correct me if I'm wrong here.

r0mant: Yep, it will: https://github.com/kubernetes/client-go/blob/master/rest/request.go#L451-L453.

PR #954, opened 2019-12-10
Kevin/6.1.x/backport 864

knisbet:


a-palchikov:
I think the tests are failing due to the same reason as here: https://github.com/gravitational/gravity.e/pull/4251/files

PR #961, opened 2019-12-12
Version 5.5.x - add Load Average

eldios:
Version 5.5.x - add Load Average


eldios:
Fixed and updated version.

Issue #660, opened 2019-09-17
Pass configuration arguments to the installer / application.


benarent:
We should provide an easy way to pass configuration arguments to the installer / application. @klizhentas says we support this under the hood via Helm but it’s not exposed, not documented.

klizhentas:
@benarent this might be an opportunity to build a simple UI/dialog generator based on helm values.yaml.

klizhentas:
so people can have easy ways to supply some configuration parameters when installing apps

r0mant:
As a first step I'm going to add the same `--values` and `--set` flags to `gravity install` and `gravity upgrade` commands (and respective scripts), now that `tele build` supports them too.

This should give users a nice and familiar way to pass their Helm values like they're used to. For example, `gravity app install` command (that installs an app image) already supports them so it will be consistent with everything.

And we can build the UI for it anytime later when we have spare cycles.

Issue #696, opened 2019-09-18
Script within planet to list the registry server contents.


benarent:
We should have a script or tool within planet to list the registry server contents. Also, users confuse docker images (local images) with the images that are present in the registry.

PR #971, opened 2019-12-18
Add support for Helm values to install/upgrade

r0mant:
Add `--values` and `--set` flags to the install/upgrade commands so users can pass Helm values at install/upgrade time. The provided values are merged into a single values file and mounted into hooks under `/var/lib/gravity/helm/values.yaml` (chose directory other than "resources" to avoid collisions with people's resources) where hooks can use it from. See included docs for more info.

Closes https://github.com/gravitational/gravity/issues/660 (the CLI part, anyway).

### DIFF
--- a/docs/6.x/pack.md
+++ b/docs/6.x/pack.md
@@ -733,9 +733,8 @@ To see more examples of specific hooks, please refer to the following documentat
 
 ## Helm Integration
 
-It is possible to use [Helm](https://docs.helm.sh/) charts as a way to package
-and install applications as every Gravity Cluster comes with a preconfigured
-Tiller server and its client, Helm.
+Gravity has a first-class [Helm](https://docs.helm.sh/) support and lets you use Helm
+charts as a way to package and install applications.
 
 Suppose you have the application resources directory with the following layout:
 
@@ -752,16 +751,25 @@ example/
 
 When building the Cluster Image, the `tele build` command will find
 directories with Helm charts (determined by the presence of `Chart.yaml` file)
-and vendor all Docker images they reference into the resulting installer
-tarball.
+and vendor all Docker images they reference into the resulting image tarball.
 
-!!! note:
-    The machine running `tele build` must have Helm binary [installed](https://docs.helm.sh/using_helm/#installing-helm)
-    and available in PATH as well as its [template plugin](https://docs.helm.sh/using_helm/#installing-a-plugin).
+The `tele build` command also allows to override Helm chart values at build time
+via `--values` and `--set` flags. These values will be taken into account when
+rendering Helm templates which is useful if you need to vendor a specific version
+of a certain Docker image or pull it from a specific registry.
+
+The flags have the same meaning and syntax as the Helm flags of the same names:
+`--values` specifies a YAML file with custom values and `--set` sets values directly
+on the command-line. Both can be provided multiple times:
+
+```bash
+$ tele build example/app.yaml --values=custom-values.yaml --set=nginx.image=1.9.1 --set=postgres.registry=internal.registry.io
+```
 
 During the installation, the vendored images will be pushed to the Cluster's local
-Docker registry which is available inside the Cluster at `leader.telekube.local:5000`.
+Docker registry which is available inside the Cluster at `registry.local:5000`.
 Helm templating engine can be used to tag images with an appropriate registry.
+
 For example, `example.yaml` may contain the following image reference:
 
 ```yaml
@@ -771,11 +779,9 @@ image: {{.Values.registry}}postgres:9.4.4
 And `values.yaml` may define the `registry` templating variable that can be set
 during application installation:
 
-
 ```
 registry: ""
 ```
-
 
 An install hook can then use the `helm` binary (which gets mounted into every hook
 container under `/usr/local/bin`) to install these resources:
@@ -804,6 +810,33 @@ correct image references.
 !!! tip:
     There is a sample application available on [GitHub](https://github.com/gravitational/quickstart/tree/master/mattermost)
     that demonstrates this workflow.
+
+### Customizing Helm values
+
+It is possible to customize values of your Helm charts when installing or
+upgrading the application. To provide custom Helm values at install time,
+pass them via `--values` and `--set` flags to `gravity install` command:
+
+```bash
+unpacked-image$ ./gravity install --values=custom-values.yaml --set=nginx.image=1.9.1 --set=postgres.registry=internal.registry.io
+```
+
+The provided values are merged into a single values file that is mounted
+into install and post-install hooks under `/var/lib/gravity/helm/values.yaml`
+so the install hook can use it in the `helm install` command:
+
+```yaml
+...
+command: ["/usr/local/bin/helm", "install", "/var/lib/gravity/resources/charts/example", "--values", "/var/lib/gravity/helm/values.yaml"]
+```
+
+The same is true for upgrades: both the `./upgrade` script included with the cluster
+image and `gravity upgrade` commands support providing custom Helm values which
+will get mounted at the same location in the upgrade/post-upgrade hooks:
+
+```bash
+unpacked-image$ ./upgrade --values=custom-values.yaml --set=nginx.image=1.11.0
+```
 
 ## Custom Installation Screen
 
--- a/docs/6.x/pack.md
+++ b/docs/6.x/pack.md
@@ -736,6 +736,9 @@ To see more examples of specific hooks, please refer to the following documentat
 Gravity has a first-class [Helm](https://docs.helm.sh/) support and lets you use Helm
 charts as a way to package and install applications.
 
+!!! note "Helm version":
+    Gravity 6 works with Helm 2. We are currently working on Helm 3 integration.
+
 Suppose you have the application resources directory with the following layout:
 
 ```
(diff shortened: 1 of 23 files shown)


docs/6.x/pack.md
```diff
@@ -805,6 +811,33 @@ correct image references.
...
+It is possible to customize values of your Helm charts when installing or
+upgrading the application. To provide custom Helm values at install time,
+pass them via `--values` and `--set` flags to `gravity install` command:
+
+```bash
+unpacked-image$ ./gravity install --values=custom-values.yaml --set=nginx.image=1.9.1 --set=postgres.registry=internal.registry.io
+```
+
+The provided values are merged into a single values file that is mounted
+into install and post-install hooks under `/var/lib/gravity/helm/values.yaml`
```
knisbet: Nit: Something to think about, is since this runs within hooks, whether it really makes sense long term to keep `/var/lib/gravity/...` here or whether a flatter structure might make sense at some point.

docs/6.x/pack.md
```diff
@@ -805,6 +811,33 @@ correct image references.
...
     that demonstrates this workflow.
 
+### Customizing Helm values
+
+It is possible to customize values of your Helm charts when installing or
+upgrading the application. To provide custom Helm values at install time,
+pass them via `--values` and `--set` flags to `gravity install` command:
+
+```bash
+unpacked-image$ ./gravity install --values=custom-values.yaml --set=nginx.image=1.9.1 --set=postgres.registry=internal.registry.io
```
knisbet: Hmmm, doesn't this imply that the --set will get applied to every helm chart contained within the cluster image? What happens if multiple charts have overlapping values?

docs/6.x/pack.md
```diff
@@ -805,6 +811,33 @@ correct image references.
...
+...
+command: ["/usr/local/bin/helm", "install", "/var/lib/gravity/resources/charts/example", "--values", "/var/lib/gravity/helm/values.yaml"]
+```
+
+The same is true for upgrades: both the `./upgrade` script included with the cluster
+image and `gravity upgrade` commands support providing custom Helm values which
+will get mounted at the same location in the upgrade/post-upgrade hooks:
+
+```bash
+unpacked-image$ ./upgrade --values=custom-values.yaml --set=nginx.image=1.11.0
```
knisbet: Does this imply that any customization / values needs to be saved and maintained for an upgrade and provided by the customer?

docs/6.x/pack.md
r0mant: What do you mean by flatter structure?

docs/6.x/pack.md
r0mant: Not really, all these values (provided by values/set flags) will be merged into a single values.yaml file and how they are applied really depends on the command that runs in the hook - we don't apply those values automatically.

So for multiple charts with possibly overlapping values, the user would need to create appropriate nesting for chart specific values, which is I think a regular practice in the Helm world anyway, for example:

```yaml
postgres: // values for chart 1
  image: 12.0.0
mariadb: // values for chart 2
  image: 11.2.3
```

and then when running helm install/upgrade, the appropriate values would get applied to the appropriate charts.

docs/6.x/pack.md
r0mant: If the user needs to customize some values during upgrade, they will need to provide those to the upgrade command, yes. The "regular" (or "default") values file can still be packaged together with their chart.

docs/6.x/pack.md
```diff
@@ -733,9 +733,8 @@ To see more examples of specific hooks, please refer to the following documentat
 
 ## Helm Integration
 
-It is possible to use [Helm](https://docs.helm.sh/) charts as a way to package
-and install applications as every Gravity Cluster comes with a preconfigured
-Tiller server and its client, Helm.
```
benarent: Have we removed Tiller because we are on Helm 3.0 . If so, I would call that our specifically.

docs/6.x/pack.md
```diff
@@ -752,16 +751,25 @@ example/
...
+`--values` specifies a YAML file with custom values and `--set` sets values directly
+on the command-line. Both can be provided multiple times:
+
+```bash
+$ tele build example/app.yaml --values=custom-values.yaml --set=nginx.image=1.9.1 --set=postgres.registry=internal.registry.io
+```
 
 During the installation, the vendored images will be pushed to the Cluster's local
-Docker registry which is available inside the Cluster at `leader.telekube.local:5000`.
+Docker registry which is available inside the Cluster at `registry.local:5000`.
```
benarent: Is there an easy to discover what docker images are available in the registry.local?

docs/6.x/pack.md
r0mant: No, not yet. We haven't upgraded to Helm 3 yet.

r0mant: But I figured this is implementation detail and not really necessary here and also this way we won't have to change it when we actually do upgrade.

docs/6.x/pack.md
r0mant: If by "easy" you mean calling the appropriate registry API with proper authentication then yes :) But really, no - we have a [ticket](https://github.com/gravitational/gravity/issues/696) in this milestone to provide a convenience script for that.

docs/6.x/pack.md
benarent: I would suggest we add a  

```
!!! note
 Gravity 6.0 works with Helm 2.0. We are currently working on Helm 3.0 Integration.

File: docs/6.x/pack.md (hunk 1 above)
Comment (a-palchikov) [reply docs/6.x/pack.md]: I guess this means that instead of mounting the resources as `/var/lib/gravity/resources/..` they're instead available as `resources/...` - ditto for `helm`.

File: docs/6.x/pack.md (hunk 4 above)
Comment (r0mant) [reply docs/6.x/pack.md]: Yeah, that would be good to have. I'll add this note.

File: tool/gravity/cli/config.go [hunk 6]
```diff
@@ -607,8 +616,12 @@ func (i *InstallConfig) validateCloudConfig(manifest schema.Manifest) (err error
 }
 
 // NewWizardConfig returns new configuration for the interactive installer
-func NewWizardConfig(env *localenv.LocalEnvironment, g *Application) InstallConfig {
-	return InstallConfig{
+func NewWizardConfig(env *localenv.LocalEnvironment, g *Application) (*InstallConfig, error) {
+	values, err := helm.Vals(*g.InstallCmd.Values, *g.InstallCmd.Set, nil, nil, "", "", "")
```
a-palchikov: Yes, but why from `InstallCmd`? Is `wizard` an alias?

tool/gravity/cli/register.go
```diff
@@ -96,6 +96,8 @@ func RegisterCommands(app *kingpin.Application) *Application {
 	g.InstallCmd.DNSZones = g.InstallCmd.Flag("dns-zone", "Specify an upstream server for the given zone within the cluster. Accepts <zone>/<nameserver> format where <nameserver> can be either <ip> or <ip>:<port>. Can be specified multiple times.").Strings()
 	g.InstallCmd.Remote = g.InstallCmd.Flag("remote", "Do not use this node in the cluster.").Bool()
 	g.InstallCmd.FromService = g.InstallCmd.Flag("from-service", "Run in service mode.").Hidden().Bool()
+	g.InstallCmd.Set = g.InstallCmd.Flag("set", "Set Helm chart values on the command line. Can be specified multiple times and/or as comma-separated values: key1=val1,key2=val2.").Strings()
+	g.InstallCmd.Values = g.InstallCmd.Flag("values", "Set Helm chart values from the provided YAML file. Can be specified multiple times.").Strings()
```
a-palchikov: nit: aren't `values` and `set` a bit generic for the gravity commands? I'm sure they're fine in helm context, but here the context is gone.

docs/6.x/pack.md
r0mant: You mean like instead of mounting everything under `/var/lib/gravity` just have `/resources` and `/helm` under root? We could do that at some point, shouldn't be too hard via symlinking to preserve backwards compatibility.

docs/6.x/pack.md
knisbet: Well there is that aspect, but the angle I was thinking of, is I customize a bunch of values when I install the chart. What happens later when I upgrade, does it keep the values file from install time, or do I as a user need to keep a copy and remember to use it again during an upgrade?

docs/6.x/pack.md
knisbet: yea, some sort of structure that makes sense with some thought. `/config/helm` /config/resources`, etc.

tool/gravity/cli/config.go
r0mant: Indeed it is. Good catch :+1:

tool/gravity/cli/register.go
r0mant: I don't know, I think having same-named flags is helpful for Helm users. Also the context is really not that different IMO: helm install/upgrade vs gravity install/upgrade, helm render vs tele build, etc.

docs/6.x/pack.md
r0mant: @knisbet Good question - not necessarily, we don't need to save values customized during the install ourselves. These values are saved by Helm and can be retrieved for any release using `helm get values` command for any revision of the release. So the upgrade hook can just take these into account if needed.

What this makes me think about though, is that we should probably save Helm values provided during `tele build`, because these are not available anywhere else. So if let's say you override an image during tele build but forget to override it during install, we may end up attempting to pull it from the internet cause different version was vendored.

I'll address this separately after this one (I'm thinking to either keep a separate values file with values set during tele build and mount it inside hooks as well, or just merge it into the same values.yaml that this PR adds).

PR #979, opened 2019-12-19
generate separate keys for each certificate

knisbet:
Fixes a security issue where:
- master nodes re-use the same secret for a majority of certificates. Now only private keys that are required to be the same are maintained.
- Each certificate assigned to a particular node reuses the same private key. Now a private key is generated for each issued certificate.
- Rotation did not rotate private keys. Note: rotation of the root key or apiserver key is not yet supported.

### DIFF
--- a/lib/ops/opsservice/configure.go
+++ b/lib/ops/opsservice/configure.go
@@ -524,6 +524,8 @@ func (s *site) configurePlanetCertAuthority(ctx *operationContext) error {
 	// we have to share the same private key for various apiservers
 	// due to this issue:
 	// https://github.com/kubernetes/kubernetes/issues/11000#issuecomment-232469678
+	// TODO(security) separate this out to a separate secret for serviceaccounttokens
+	// For reference: https://github.com/kelseyhightower/kubernetes-the-hard-way/issues/248
 	apiServer, err := authority.GenerateCertificate(csr.CertificateRequest{
 		CN:    constants.APIServerKeyPair,
 		Hosts: []string{"127.0.0.1"},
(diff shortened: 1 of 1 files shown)


lib/ops/opsservice/configure.go
```diff
@@ -673,6 +679,10 @@ func (s *site) getPlanetMasterSecretsPackage(ctx *operationContext, p planetMast
 						s.domainName, host}, "."))
 				}
 			}
+
+			// Don't rotate the APIServer key, the secret is currently used for validating serviceaccounttokens
+			// TODO(security) enable rotation of secret for apiserver/serviceaccounttokens
+			privateKeyPEM = apiserverKeyPair.KeyPEM
```
a-palchikov: I know this way the changes are kept to a minimum, but I'd want the apiserver case handled visually separately:
```go
if name == constants.APIServerKeyPair {
	keyPair, err := authority.GenerateCertificate(req, caKeyPair, apiserverKeyPair.KeyPEM, defaults.CertificateExpiry)
	if err != nil {
		return nil, trace.Wrap(err)
	}
} else {
	keyPair, err := authority.GenerateCertificate(req, caKeyPair, nil, defaults.CertificateExpiry)
	if err != nil {
		return nil, trace.Wrap(err)
	}
}
```

lib/ops/opsservice/configure.go
```diff
@@ -599,7 +601,9 @@ func (s *site) getPlanetMasterSecretsPackage(ctx *operationContext, p planetMast
 		return nil, trace.Wrap(err)
 	}
 
-	baseKeyPair, err := archive.GetKeyPair(constants.APIServerKeyPair)
+	// Don't rotate apiserver secrets, as this secret is currently used to authenticate service account tokens
+	// TODO(securty) support rotation of apiserver / serviceaccount secrets
```
helgi: In 1.17 they rotate off disk if you put it down that way, ever hour iirc if this is for k8s apiserver

PR #983, opened 2019-12-20
Add gravity registry list command.

r0mant:
Add a convenience `gravity registry list` command that displays images published in the cluster registry. Closes https://github.com/gravitational/gravity/issues/696.

### DIFF
--- a/docs/6.x/cluster.md
+++ b/docs/6.x/cluster.md
@@ -46,7 +46,8 @@ way:
 
 * `gravity` runs its own local Docker registry which is used as a Cluster-level
   cache for container images. This makes application upgrades and restarts
-  faster and more reliable.
+  faster and more reliable. See [Interacting with Cluster Registry](#interacting-with-cluster-registry)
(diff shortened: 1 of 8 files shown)


docs/6.x/cluster.md
```diff
@@ -1437,6 +1438,32 @@ The CoreDNS configuration can be edited after installation, by updating the `kub
...
+```bash
+$ gravity registry list
+```
+
+By default the command will contact the registry server running at `registry.local:5000` address
+which resolves to the registry running on the currently active master node. To list images in a
+specific registry, provide a `--registry` flag to the command:
+
+```bash
+$ gravity registry list --registry=192.168.1.1:5000
```
a-palchikov: Should we make it possible to use just the hostname/IP address (w/o the port) as this part is fixed?

r0mant: Yeah, good point. Could default to 5000 if not specified.

Issue #993, opened 2019-12-24
Gravity missing required fields when registering a node


r0mant:
There was recently a change that disabled automatic node registration in kubelet b/c starting with 1.16(?) kubelet is not allowed to set certain labels on the nodes by itself, due to security reasons. The change, instead, moved the responsibility of registering the nodes onto Gravity so it creates the node objects itself during one of the install phases.

This approach started costing us a bunch of regressions b/c kubelet does all sorts of logic when registering a node, which Gravity does not. One issue encountered in the past is that kubelet sets certain annotations on the nodes. The most recent issue is `spec.providerID` field that kubelet sets when cloud-integration is enabled - w/o it the load balancers don't work properly.

We have 2 ways forward:

1. Have Gravity set `spec.providerID` duplicating kubelet's logic. My concern is that these sorts of issues will keep happening, esp. with upgrades to new versions when kubelet logic changes.
2. Restore the automatic node registration and instead of having Gravity register the entire node object, just set the appropriate labels/taints that kubelet cannot set itself.

I'm leaning towards option number 2. I'm pretty sure this will save us a lot of headache in the future.

@a-palchikov @knisbet What do you guys think?

a-palchikov:
It definitely looks like kubelet has a fair share of logic for [setting up](https://github.com/kubernetes/kubernetes/blob/v1.15.5/pkg/kubelet/kubelet_node_status.go#L217) a node object. While this could be replicated, I agree with @r0mant that it then should either be done with each release or we let kubelet set the node up with the list of labels it is allowed to set while setting the rest externally my updating the node object.

knisbet:
Hmmm, so part of the original motivation for going this direction was due to the actual problem, where kubelet implements it's own label validation logic that differs from kubernetes itself. However, a second component to be considered, is it sounded to me like kubelet / kubeneretes may be heading in this direction anyways. The security risk with the automatic registration process, is that a compromised nodes might be able to affect it's labels to get workloads assigned that would not normally be assigned to the particular node. Although I'm not sure where that currently stands, and it did look like a much more complicated mechanism was going to be in place on which labels / taints can be controlled by kubelet, and filter to specific fields that cannot.

So I'm not opposed in any way with option 2, we just need to be weary of if the node object is inadvertently deleted it will only partially be re-created, and that long term we may be driven back towards the current registration process. Doesn't mean we shouldn't avoid some short term pain points though.

helgi:
[...]

PR #1001, opened 2020-01-06
(6.3) Remove node registration and add labels reconciliation loop.

r0mant:
This PR removes the node registration phase from Gravity and restores the automatic registration by kubelet (requires https://github.com/gravitational/planet/pull/539).

The way it works now is:

* Kubelet is started with a set of labels it is allowed to set. Gravity uses a method directly from Kubernetes source code to filter the labels.
* Inside gravity-site there is a reconciliation routine running that periodically checks if any of the node labels need to be updated. This routine will set labels that kubelet itself is not allowed to set and will also take care of scenarios if a node object gets deleted and recreated so the labels won't be lost.

Closes https://github.com/gravitational/gravity/issues/993. Will need to be ported everywhere the automatic node registration was disabled.


lib/process/process.go
```diff
@@ -621,6 +621,67 @@ func (p *Process) runRegistrySynchronizer(ctx context.Context) {
...
+		return trace.Wrap(err)
+	}
+	nodes, err := utils.GetNodes(client.CoreV1().Nodes())
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	for ip, node := range nodes {
+		server, err := cluster.ClusterState.FindServerByIP(ip)
+		if err != nil {
+			return trace.Wrap(err)
```
a-palchikov: I wonder what this error actually means? It can mean borked cluster state or unexpectedly added kubernetes node?

lib/process/process.go
```diff
@@ -621,6 +621,67 @@ func (p *Process) runRegistrySynchronizer(ctx context.Context) {
...
+		}
+		profile, err := cluster.App.Manifest.NodeProfiles.ByName(server.Role)
```
a-palchikov: L638-L648 could be better encapsulated as:
```go
missingLabels, err := getMissingLabels(cluster, server)
```

lib/process/process.go
```diff
@@ -621,6 +621,67 @@ func (p *Process) runRegistrySynchronizer(ctx context.Context) {
...
+				missingLabels[key] = val
+			}
+		}
+		if len(missingLabels) != 0 {
+			p.Infof("Adding missing labels to node %v/%v: %v.", node.Name, ip, missingLabels)
+			for key, val := range missingLabels {
+				node.Labels[key] = val
+				_, err := client.CoreV1().Nodes().Update(&node)
+				if err != nil {
+					return trace.Wrap(err)
```
a-palchikov: nit: `rigging.ConvertError`

r0mant:
@a-palchikov I thought about the same, but ultimately decided it's not worth it to add another phase for it at this point. The only use case I can see is if user uses `*.kubernetes.io` labels to schedule their stuff on the nodes but the reconciler runs for the first time as soon as gravity-site starts up which happens before any of the user pods are scheduler. And for our stuff we use `gravitational.io/k8s-role` label which is set by kubelet (and which also actually contains cluster role rather than app role).

lib/process/process.go
r0mant: Yeah, not sure how that can actually happen but the practice shows that everything that can go wrong, will. I've modified this loop a bit so the failure to reconcile a node is just logged and does not stop the entire loop.

PR #1007, opened 2020-01-09
lib/httplib: fix dropped error

alrs:
This picks up a dropped error in `lib/httplib`.


a-palchikov:
This is part of changes from [here](https://github.com/gravitational/gravity/pull/1048)

PR #1023, opened 2020-01-15
Update based on product feeback.

benarent:


docs/6.x/config.md
```diff
@@ -289,7 +289,8 @@ $ gravity resource create admin.yaml
 The new user can now log into the Cluster via the Web UI with the user
 credentials created above.
 
-!!! tip "Password Restrictions":
+!!! tip "Username and Password Restrictions":
+    Usernames should be composed of characters, hyphens, the at symbol and dots.
```
a-palchikov: The examples use emails as user names (although this is not enforced in code) - this should be hinted at that since saying names should have an at symbol is vague.

benarent: Thanks @a-palchikov It's been a while since i updated this, I'm sure it was something specific. I'll do better to explain the changes next time.

PR #1033, opened 2020-01-21
[6.1.x] a couple of fixes

a-palchikov:
Improves the following:
  - configures symlink for helm on host as it as regressed after I moved most of the system package code to lib/update/system and the symlink code was only added in the old location (in tool/gravity/cli/system.go)
  - bumps satellite to include [improvements](https://github.com/gravitational/satellite/pull/129) to handling boot configuration parameter validation

Potentially requires forward-porting.


knisbet:
@a-palchikov IIRC the node-role.kubernetes.io/role change was to fix that the node roles don't show up in `kubectl get nodes`. So I think that change was the correct label structure for the way kubernetes considers node roles, atleast in `kubectl get nodes` case. Node role isn't documented though in the kubernetes well known labels (https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/) docs.

I'm not sure what gravity behaviour you're referring to, but possibly it needs to consider the kubernetes labels, or we should be using gravity specific labels.

PR #1040, opened 2020-01-23
[5.5.x] update pod CIDR

a-palchikov:
This PR fixes the cluster configuration upgrade path for pod CIDR.
In addition to existing configuration package upgrade, it ensures proper pod subnet reset by removing the `cni0` bridge on each node prior to planet container restart - this ensures that flannel is able to reset its configuration accordingly.

Will need forward porting to 6.x LTS and master.
Requires https://github.com/gravitational/planet/pull/543.


helgi:
Can this be ported to 6.3 as well?

lib/update/clusterconfig/phases/restart.go
```diff
@@ -0,0 +1,121 @@
...
+	return nil
+}
+
+// PostCheck is a no-op
+func (*restart) PostCheck(context.Context) error {
+	return nil
+}
+
+func (r *restart) removeCNIBridge() error {
+	link, err := netlink.LinkByName("cni0")
```
r0mant: Just a random thought, it seems like this "netlink" package brought +18K additional dependencies, while deleting the interface could've be just done with os.Exec("ip", "link", "del", ...). Just a thought, I don't mind keeping it if you think it's a better approach.

a-palchikov: I initially thought I would have issues with interpreting the output if I go that route (we've had issues with english-only queries and localization before) but I think I can make it work with just commands.

PR #1048, opened 2020-01-27
Fix lint warnings

a-palchikov:
Requires https://github.com/gravitational/gravity.e/pull/4270.

Next step would be integrating lint as part of the build.


lib/storage/keyval/constants.go
```diff
@@ -37,46 +37,43 @@ const (
...
-	localClusterP               = "localcluster"
-	planP                       = "plan"
-	trustedClustersP            = "trustedclusters"
-	tunnelConnectionsP          = "tunnelconnections"
-	remoteClustersP             = "remoteclusters"
-	systemP                     = "system"
-	dnsP                        = "dns"
-	chartsP                     = "charts"
-	indexP                      = "index"
+	//activeOperationsP           = "activeops"
```
r0mant: Remove if this is not used?

lib/storage/role.go
```diff
@@ -287,7 +287,9 @@ func (*roleMarshaler) UnmarshalRole(data []byte) (teleservices.Role, error) {
 			return nil, trace.BadParameter(err.Error())
 		}
 		// we are ignoring error from this function on purpose here
-		role.CheckAndSetDefaults()
+		if err := role.CheckAndSetDefaults(); err != nil {
```
r0mant: How about the comment right above? :)

> we are ignoring error from this function on purpose here

I would keep all `CheckAndSetDefaults()` calls as it is everywhere. Otherwise we'll start getting errors when creating/retrieving resources from the database.

lib/storage/user.go
```diff
@@ -530,6 +531,7 @@ func (u *UserV2) Check() error {
 			u.GetType(), SupportedUserTypes)
 	}
 	if u.GetType() != AgentUser && u.GetPassword() == "" {
+		fmt.Printf("Failed to validate user %#v: %s\n.", u, string(debug.Stack()))
```
r0mant: Debug leftover?

tool/gravity/cli/install.go
```diff
@@ -428,7 +432,9 @@ func autojoin(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, d
 	})
 	if utils.IsContextCancelledError(err) {
 		// We only end up here if the initialization has not been successful - clean up the state
-		InstallerCleanup()
+		if err2 := InstallerCleanup(); err2 != nil {
```
r0mant: You can just use `err` here (and below).

lib/storage/role.go
a-palchikov: Do we want to have validation logic that is wrong?

r0mant: Thing is, these functions are called when retrieving resources from the database which means if there's a resource that doesn't pass this check for whatever reason, the user will start getting errors in unexpected places -> more support load. This has actually happened several times in the past, that's why I remember, and I'm sure we still have clusters out with such resources. I think it was an unfortunate decision to do CheckAndSetDefaults on the unmarshal in general - the resource should be rejected in the first place if it's invalid, which I think we do now.

Issue #943, opened 2019-12-09
Lingering serf members cause degraded status


r0mant:
If a node has been removed from the cluster as a part of the shrink operation but its planet is still running for whatever reason (which most usually happens when using ASG) the serf agent may remain a part of the serf cluster and will enter "failed" state when the planet finally shuts down.

This means that the removed node is not among Kubernetes or Gravity nodes anymore but planet still thinks it's got a failed member and thus the status will be degraded until the member is manually force-leave'd.

The "gravity status" won't show this "rogue" node b/c as far as Gravity's concerned, it's not a part of the cluster and the status will just say "degraded" without a clue of why. The failed member should be visible in "planet status" but it's not a user-facing command.

We should extend "gravity status" to give better visibility into why a cluster is degraded in such cases.

As an idea, instead of using "cluster state" as a single source of truth for the cluster nodes, we could compare a list of gravity/k8s/etcd/serf nodes and alert to any found discrepancies.

helgi:
This happened again on a 6.2 cluster; @tosih will provide more info.

helgi:
I've seen this happen various times now on 6.2, primarily when nodes are being rolled using AWS ASGs

tariq1890:
Any updates on this issue?

r0mant:
Going to prioritize this higher, customers keep running into this.

helgi:
Did this get backported to 5.5? I saw the serf bump but not the code change

PR #1059, opened 2020-01-30
Run gravity remove before instance is terminated in AWS

helgi:
AWS can take a while to terminate an instance, doing the remove first will trigger the shrink operation and let the cluster generally settle itself while AWS does its business. I have a feeling it will help resolve https://github.com/gravitational/gravity/issues/943 to some small extent

In internal asg roller I have doing `gravity remove` before AWS terminated (codified) has helped a lot with stability. When I had the roller only depend on the AWS lifecycle hook (this functionality) I would have `failed` nodes in `serf` frequently or other funky behaviours.

### DIFF
--- a/lib/autoscale/aws/events.go
+++ b/lib/autoscale/aws/events.go
@@ -107,10 +107,10 @@ func (a *Autoscaler) processEvent(ctx context.Context, operator Operator, event 
 			return trace.Wrap(err)
 		}
 	case InstanceTerminating:
-		if err := a.ensureInstanceTerminated(ctx, event); err != nil {
+		if err := a.removeInstance(ctx, operator, event); err != nil && !trace.IsNotFound(err) {
 			return trace.Wrap(err)
 		}
-		if err := a.removeInstance(ctx, operator, event); err != nil && !trace.IsNotFound(err) {
+		if err := a.ensureInstanceTerminated(ctx, event); err != nil {
 			return trace.Wrap(err)
 		}
 		if err := a.DeleteEvent(ctx, event); err != nil {


helgi:
Understood that https://github.com/gravitational/gravity/pull/315 specifically made the code wait for instance termination before the shrink operation; I'm not sure of the contest but given my previous comment I think `remove` needs to happen first but then again I can't see the `e` ticket referenced :)

knisbet:
Hmm, so I'm not sure if this will help with #943, and there are also some unknowns on other possible issues. My concern is, the removeInstance instance code starts a shrink operation, which will if the node is online start an uninstall process on the particular node using teleport. If the node goes offline mid-uninstall, it's hard for me to predict what side effects may occur. 

So to me the current ordering does make sense, where when we see the node is fully terminated, we launch the remove process (for removing a dead node) after the instance has terminated. Although it does beg the question, if we're waiting for the node to be terminated, I'm not sure our theory that the node still existing causing the node to rejoin the serf cluster is accurate. 

The other route to go is like we discussed a few weeks ago, is to launch removeInstance, but to properly heartbeat the AWS ASG termination event to keep the node alive while running the uninstall process on the node. Although I'm of the opinion that when terminating an AWS node we shouldn't need to rely on a clean uninstallation happening.

helgi:
In a different code base doing the above what I do is I have a "wait until node is removed from gravity state" and then I run AWS terminate. In the gravity code as it is right now it basically does the inverse.

This might not be the serf issue but it would explain why I have seen gravity consider a node to be available but offline beyond it's AWS existance where a **worker** that was expanding would use **master** that was available but marked offline; I only saw that when I use the AWS lifecycle hook  to deal with the `gravity remove` aspect, as soon as I did the "gravity remove, wait until node is gone, do aws terminate" I stopped seeing **workers** attempt to connect to a non-existent **master**.

--- Above is just context against why this change made sense (knowing it's not fully baked) ---

The option we talked about a few weeks back would be more than sufficient I think, and I agree, a clean uninstall isn't a requirement as long as `gravity remove` (or similar) can cleanup all the internals properly. Is there any anything essential `uninstall` if the node is being fully destroyed?

I could imagine a new `phase` that would put a node (especially when it's a **master**) into a `pre-shrink ` of sorts mode, which in turn takes it out of any normal interaction with the cluster, i.e. new **workers** would not be told to use a node (**master**) in `pre-shrink` state and could even be instructed on the fly to restart it's `expand` against a new master if its current master is being put down.

helgi:
https://github.com/gravitational/gravity/pull/1251 will resolve this much better

Issue #1016, opened 2020-01-14
[Enhancement] Allow certain fields as optional for `gravity resource create`


aelkugia:
**What Happened**

When the SMTP server replies with an empty AUTH option 250-AUTH Kapacitor it just sends the QUIT command without attempting delivery. The server is sending this empty as authentication is not required, but the gravity command line tool does not allow to leave username and password values empty:

`$ sudo gravity resource create -f alert-smtp.yaml`
`[ERROR]: failed to validate: username: username is required,password: password is required`

**Expected Behavior**

Configuration tool should accept these values as optional instead of enforcing them. Manually touching Kubernetes config maps addressed this issue but is not scalable.

**Environment**

Gravity (5.5.x)

r0mant:
This was done.

PR #1060, opened 2020-01-30
Make username/password for SMTP resource optional

UnderGreen:
Fixes https://github.com/gravitational/gravity/issues/1016


lib/storage/smtp.go
```diff
@@ -81,12 +81,8 @@ func (r *SMTPConfigV2) CheckAndSetDefaults() error {
 		return trace.BadParameter("missing parameter Host")
 	}
 
-	if r.Spec.Username == "" {
-		return trace.BadParameter("Username cannot be empty")
-	}
-
-	if r.Spec.Password == "" {
-		return trace.BadParameter("Password cannot be empty")
+	if r.Spec.Port < 0 {
```
r0mant: If you check for port validity, make sure it's less than 65535 as well then?

r0mant:
@UnderGreen Also don't forget to forward-port to 6.1, 6.3 and master please.

PR #1061, opened 2020-01-31
[5.5.x] extend ClusterConfiguration resource validation

a-palchikov:
This PR updates the `ClusterConfiguration` validation to catch more errors upfront prior to creating the operation.
The relevant validation code was moved to `lib/validate`.

Also fixes a bug in that upon update, depending on the updating `ClusterConfiguration`, (parts of) previous configuration would be overwritten with empty values. Now, the resources are merged to keep previous configuration.


lib/ops/opsservice/site.go
```diff
@@ -644,6 +647,57 @@ func (s site) gid() string {
...
+		config, err = clusterconfig.Unmarshal([]byte(spec))
+		if err != nil {
+			return nil, trace.Wrap(err)
+		}
+		s.Infof("Existing cluster configuration: %#v (%s).", config, spec)
+	} else {
+		config = clusterconfig.NewEmpty()
+		s.Infof("Existing (empty) cluster configuration: %#v (%s).", config, spec)
+	}
+	if err := s.setClusterConfigDefaults(config); err != nil {
```
a-palchikov: This is to workaround issues with previous state after the upgrade.
Ideally, the installer would keep all now disparate attributes in the `ClusterConfiguration` from the get-go - which is something to improve in future (and I remember starting this work already for a related issue).

lib/ops/resources/gravity/collection.go
```diff
@@ -626,28 +626,27 @@ func (r configCollection) WriteText(w io.Writer) error {
...
-		}
-		if len(config.ServiceNodePortRange) != 0 {
-			fmt.Fprintf(t, "Service Node Port Range:\t%v\n", config.ServiceNodePortRange)
-		}
-		if len(config.ProxyPortRange) != 0 {
-			fmt.Fprintf(t, "Proxy Port Range:\t%v\n", config.ProxyPortRange)
-		}
-		if len(config.FeatureGates) != 0 {
-			fmt.Fprintf(t, "FeatureGates:\t%v\n", formatFeatureGates(config.FeatureGates))
+	config := r.GetGlobalConfig()
```
a-palchikov: I made `ClusterConfiguration.Spec.GlobalConfig` a non-pointer since it was getting in a way with no real benefit.

lib/validate/clusterconfig.go
```diff
@@ -0,0 +1,83 @@
...
+
+package validate
+
+import (
+	"github.com/gravitational/gravity/lib/storage/clusterconfig"
+
+	"github.com/gravitational/trace"
+)
+
+// ClusterConfiguration validates that the update can update existing
```
r0mant: Incomplete sentence?

lib/ops/resources/gravity/collection.go
```diff
@@ -626,28 +626,27 @@ func (r configCollection) WriteText(w io.Writer) error {
...
+		formatCloudConfig(t, config.CloudConfig)
+	}
+	if len(config.ServiceNodePortRange) != 0 {
+		fmt.Fprintf(t, "Service Node Port Range:\t%v\n", config.ServiceNodePortRange)
+	}
+	if len(config.ProxyPortRange) != 0 {
+		fmt.Fprintf(t, "Proxy Port Range:\t%v\n", config.ProxyPortRange)
+	}
+	if len(config.FeatureGates) != 0 {
+		fmt.Fprintf(t, "FeatureGates:\t%v\n", formatFeatureGates(config.FeatureGates))
```
r0mant: Nit: `Feature Gates` (with a space).

lib/ops/opsservice/site.go
```diff
@@ -644,6 +647,53 @@ func (s site) gid() string {
...
+func (s *site) setClusterConfigDefaults(config *clusterconfig.Resource) error {
+	if config.Spec.Global.CloudProvider == "" {
+		config.Spec.Global.CloudProvider = s.provider
+	}
+	installOp, _, err := ops.GetInstallOperation(s.key, s.service)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if config.Spec.Global.PodCIDR == "" {
+		config.Spec.Global.PodCIDR = installOp.InstallExpand.Vars.OnPrem.PodCIDR
```
r0mant: Nit: I'd check for `InstallExpand` being not nil, even though this is an install operation.

lib/validate/clusterconfig.go
a-palchikov: Even with `update` and `existing` being the variable names?

PR #1066, opened 2020-02-03
Status history command

bernardjkim:
Integrate satellite timeline feature gravitational/satellite#151


lib/status/timeline.go
```diff
@@ -0,0 +1,58 @@
...
+// Timeline queries the currently stored cluster timeline.
+func Timeline(ctx context.Context) (*pb.TimelineResponse, error) {
+	addr := fmt.Sprintf("%v:%v", constants.Localhost, defaults.SatelliteRPCAgentPort)
+
+	stateDir, err := state.GetStateDir()
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	caFile := state.Secret(stateDir, defaults.RootCertFilename)
```
r0mant: There is `lib/httplib.GetPlanetClient()` that does all this IIRC. Let's reuse that.

r0mant:
@bernardjkim You also need to bump `PLANET_TAG`in the Makefile to point to the version of your planet.

tool/gravity/cli/register.go
```diff
@@ -212,6 +212,9 @@ func RegisterCommands(app *kingpin.Application) *Application {
 	g.StatusCmd.Seconds = g.StatusCmd.Flag("seconds", "Continuously display status every N seconds.").Short('s').Int()
 	g.StatusCmd.Output = common.Format(g.StatusCmd.Flag("output", "Output format: json or text.").Default(string(constants.EncodingText)))
 
+	// display cluster status history
+	g.StatusHistoryCmd.CmdClause = g.Command("status-history", "Display cluster status history.")
```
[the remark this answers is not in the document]
bernardjkim: Will work on making this a sub-command then. The `history` cmd is already taken.

lib/status/timeline.go
bernardjkim: Looks like `GetPlanetClient` returns an http client. We want to use the satellite rpc client here.

tool/gravity/cli/status.go
```diff
@@ -170,6 +170,65 @@ func statusOnce(ctx context.Context, operator ops.Operator, operationID string,
...
+
+	w := new(tabwriter.Writer)
+	w.Init(out, 0, 8, 1, '\t', 0)
+
+	timestamp := event.GetTimestamp().ToTime()
+
+	switch event.GetData().(type) {
+	case *pb.TimelineEvent_ClusterDegraded:
+		eventString := fmt.Sprintf("[%s] Cluster Degraded\n", timestamp)
+		fmt.Fprintf(w, color.RedString(eventString))
```
a-palchikov: `fmt.Fprintf` will be flagged by lint as superfluous (should use `fmt.Fprint` instead).
Also, `color.XYZ` support formatting so explicit `fmt.Sprintf` is also redundant:
```go
fmt.Fprint(w, color.RedString("[%s] Cluster Degraded\n", timestamp))
```
here and below.

tool/gravity/cli/status.go
```diff
@@ -170,6 +170,65 @@ func statusOnce(ctx context.Context, operator ops.Operator, operationID string,
...
+	case *pb.TimelineEvent_ProbeFailed:
+		e := event.GetProbeFailed()
+		eventString := fmt.Sprintf("[%s] Probe Failed\t[%s]\t[%s]\n", timestamp, e.GetNode(), e.GetProbe())
+		fmt.Fprintf(w, color.RedString(eventString))
+	case *pb.TimelineEvent_ProbeSucceeded:
+		e := event.GetProbeSucceeded()
+		eventString := fmt.Sprintf("[%s] Probe Succeeded\t[%s]\t[%s]\n", timestamp, e.GetNode(), e.GetProbe())
+		fmt.Fprintf(w, color.GreenString(eventString))
+	default:
+		fmt.Fprintf(w, color.YellowString("[%s] UnknownEvent", timestamp))
```
a-palchikov: `UnknownEvent` hints at the possibility there being such a type. I would use plain english like `unknown event` and log the object.

Gopkg.toml
```diff
@@ -77,7 +77,7 @@ ignored = [
 
 [[constraint]]
   name = "github.com/gravitational/satellite"
-  version = "=7.0.3"
+  version = "7.0.5"
```
r0mant: Let's keep the leading `=` there. Otherwise dep might update the dependencies to a newer patch version unintentionally.

lib/status/timeline.go
r0mant: Then let's create a `GetGRPCPlanetClient()` there so it can be reused and use it here.

docs/6.x/cluster.md
```diff
@@ -208,6 +208,41 @@ $ curl -sk https://localhost:7575 | python -m json.tool
 
 In this case the response HTTP status code will be `503 Service Unavailable`.
 
+### Cluster Status History
+
+Running `gravity status history` will display a history of changes to the 
+cluster status. 
+
```
benarent: I would add in a command similar to Cluster Status

```
$ tsh --cluster=production ssh admin@master gravity status history
```

I would recommend adding a sentence as to When and Why to run this command. 

Lastly, I would add a quick FAQ on using this command to https://gravitational.com/gravity/docs/faq/#common-issues

docs/6.x/cluster.md
```diff
@@ -208,6 +208,41 @@ $ curl -sk https://localhost:7575 | python -m json.tool
...
+Feb  6 21:37:11 UTC [Node Recovered]    node=node-3
+```
+
+There are just a few event types that are currently being tracked.
+- `Node Degraded` / `Node Recovered` specifies a change in the node status. The node
+key specifies the name of the node (node-1, node-2, node-3).
+- `Probe Succeeded` / `Probe Failed` specifies a change in a probe result. The checker
+key specifies the name of the health check (time-drift, kube-apiserver).
+- A few other events will be tracked and are WIP. We plan to track cluster 
+resizing, gravity upgrades, network partitions, etc...
```
benarent: I would suggest we add a link to create a Github Issue, if someone has a feature request.   e.g. https://github.com/gravitational/gravity/issues/new/choose

benarent: Is it possible to forward these `gravity status history` events to centralised system or is it just meant as a real-time debugging tool?

bernardjkim: Right now its just a debugging tool. I don't think it would take too much effort to forward these events to a centralized logging system. I'll take a look in a later time.

benarent: In that case, I would add a note that this has been designed as debugging tool.

bernardjkim: > The `gravity status history` command is an additional tool to help debug issues
with a Cluster.

 I think this portion states that the main purpose of this tool is for debugging. Should we reword this to specify this is command is only for debugging and nothing more?

docs/6.x/cluster.md
```diff
@@ -208,6 +208,54 @@ $ curl -sk https://localhost:7575 | python -m json.tool
...
+Feb  6 21:36:59 UTC [Probe Succeeded]   node=node-3 checker=node-status
+Feb  6 21:37:07 UTC [Probe Succeeded]   node=node-1 checker=kube-apiserver
+Feb  6 21:37:07 UTC [Node Recovered]    node=node-1
+Feb  6 21:37:08 UTC [Probe Succeeded]   node=node-2 checker=kube-apiserver
+Feb  6 21:37:08 UTC [Node Recovered]    node=node-2
+Feb  6 21:37:11 UTC [Probe Succeeded]   node=node-3 checker=kube-apiserver
+Feb  6 21:37:11 UTC [Node Recovered]    node=node-3
+```
+
+Here's an example of how to view the history remotely via `tsh`:
```
knisbet: Nit: I would leave documenting how to use tsh to the tsh sections of the docs.

docs/6.x/cluster.md
```diff
@@ -208,6 +208,54 @@ $ curl -sk https://localhost:7575 | python -m json.tool
...
+lets you observe when and where problems have occurred within the Cluster.
+
+There are just a few event types that are currently being tracked.
+- `Node Degraded` / `Node Recovered` specifies a change in the node status. The node
+key specifies the name of the node (node-1, node-2, node-3).
+- `Probe Succeeded` / `Probe Failed` specifies a change in a probe result. The checker
+key specifies the name of the health check (time-drift, kube-apiserver).
+- A few other events will be tracked and are WIP. We plan to track cluster 
+resizing, gravity upgrades, network partitions, etc...
+- If you have suggestions for other types of events to be tracked, let us know [here](https://github.com/gravitational/gravity/issues/new/choose).
```
knisbet: I don't think I would document statements like this, or that this tool is WIP above. I don't think we should be publishing docs that say what the product might do, just what it's capable of.

tool/gravity/cli/status.go
```diff
@@ -170,6 +170,70 @@ func statusOnce(ctx context.Context, operator ops.Operator, operationID string,
...
+		return
+	}
+	for _, event := range events {
+		printEvent(os.Stdout, event)
+	}
+}
+
+func printEvent(w io.Writer, event *pb.TimelineEvent) {
+	// stamp defines default timestamp format.
+	const stamp = "Jan _2 15:04:05 UTC"
```
knisbet: I don't believe we do this consistently throughout gravity, but my preferred date format would be iso 8601.

docs/6.x/cluster.md
bernardjkim: @benarent good to go?

PR #1072, opened 2020-02-04
Better support for external registries auth in app catalog.

r0mant:
This PR implements a few improvements for the "gravity app" set of commands so they play better with external registries. Before, our registry connection code was pretty much tailored to working with registries inside Gravity clusters.

* Add support for basic auth with registries. I used registry provided with OpenShift online to test this.
* Add support for registry prefixes/namespaces, which is also required by OpenShift for example so the images have to be pushed like `<namespace>/debian-tall`.
* More straightforward connection logic - always try to connect to the registry in a secure way, unless `--insecure` flag (or respective configuration parameter) was specified (which is what Docker does as well). I don't think the "insecure" path was actually before so hopefully nothing will break.
* Add ability to override tiller namespace (used to be hardcoded to system ns).

### DIFF
--- a/lib/docker/imageservice.go
+++ b/lib/docker/imageservice.go
@@ -26,6 +26,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -41,6 +42,9 @@ import (
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/registry/api/errcode"
 	registryclient "github.com/docker/distribution/registry/client"
+	registryauth "github.com/docker/distribution/registry/client/auth"
+	registryauthchallenge "github.com/docker/distribution/registry/client/auth/challenge"
+	registrytransport "github.com/docker/distribution/registry/client/transport"
 	registrystorage "github.com/docker/distribution/registry/storage"
 	"github.com/docker/distribution/registry/storage/cache/memory"
 	"github.com/docker/distribution/registry/storage/driver/filesystem"
@@ -62,6 +66,19 @@ type RegistryConnectionRequest struct {
 	ClientCertPath string
 	// ClientKeyPath is the full path to the client private key
 	ClientKeyPath string
+	// Username specifies optional registry username for basic auth
+	Username string
+	// Password specifies optional registry password for basic auth
+	Password string
+	// Domain specifies optional registry prefix when pushing images
+	Domain string
+	// Insecure indicates a plain http registry
+	Insecure bool
+}
+
+// HasBasicAuth returns true if the request contains basic auth credentials.
+func (r *RegistryConnectionRequest) HasBasicAuth() bool {
+	return r.Username != "" && r.Password != ""
 }
 
 // CheckAndSetDefaults makes sure the request is valid and sets some defaults
@@ -192,7 +209,19 @@ func (r *imageService) Sync(ctx context.Context, dir string, progress utils.Prin
 		if err != nil {
 			return nil, trace.Wrap(err)
 		}
-		remoteRepo, err := r.remoteStore.Repository(ctx, localRepoName)
+		remoteRepoName := localRepoName
+		// If the registry prefix was specified, rewrite all images to point
+		// to that domain - some registries (such as OpenShift) require a
+		// specific namespace where images should be pushed so an image like
+		// gravitational/debian-tall would become <namespace>/debian-tall.
+		if r.Domain != "" {
+			named, err := parseNamed(localRepoName)
+			if err != nil {
+				return nil, trace.Wrap(err)
+			}
+			remoteRepoName = fmt.Sprintf("%v/%v", r.Domain, Path(named))
+		}
+		remoteRepo, err := r.remoteStore.Repository(ctx, remoteRepoName)
 		if err != nil {
 			return nil, trace.Wrap(err)
 		}
@@ -240,7 +269,7 @@ func (r *imageService) Sync(ctx context.Context, dir string, progress utils.Prin
 			if remoteManifest == nil || !compareManifests(localManifest, remoteManifest) {
 				progress.PrintStep("Pushing image %s", tagSpec)
 				if err = r.remoteStore.updateRepo(ctx, remoteRepo, localRepo, localManifest, tag); err != nil {
-					return nil, trace.Wrap(err, "failed to update remote for tag %q", tagSpec)
+					return nil, trace.Wrap(err, "failed to update remote for tag %q: %v", tagSpec, err)
 				}
 			} else {
 				progress.PrintStep("Image %s is up-to-date", tagSpec)
@@ -282,7 +311,7 @@ func (r *imageService) connect(ctx context.Context) (err error) {
 // remoteStore defines a remote distribution registry
 type remoteStore struct {
 	log.FieldLogger
-	transport *http.Transport
+	transport http.RoundTripper
 	registry  registryclient.Registry
 	addr      string
 }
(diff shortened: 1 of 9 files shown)


tool/gravity/cli/helm.go
```diff
@@ -187,9 +203,10 @@ func releaseInstall(env *localenv.LocalEnvironment, conf releaseInstallConfig) e
 	return nil
 }
 
-func releaseList(env *localenv.LocalEnvironment, all bool) error {
+func releaseList(env *localenv.LocalEnvironment, all bool, tillerNamespace string) error {
 	helmClient, err := helm.NewClient(helm.ClientConfig{
-		DNSAddress: env.DNS.Addr(),
+		DNSAddress:      env.DNS.Addr(),
```
a-palchikov: Sort of off-topic but this really needs to be hidden behind some sort of address resolver interface with `env` implementing it using the specific DNS configuration - w/o to need to sprinkle the address everywhere.

lib/docker/imageservice.go
```diff
@@ -336,29 +365,75 @@ func initTransport(req RegistryConnectionRequest) (*http.Transport, string) {
 		DisableKeepAlives:   true,
 	}
 
+	// Figure out the registry address scheme (http or https). If the scheme
+	// was specified explicitly, keep it as-is. Otherwise, default to https
+	// unless the insecure flag was provided.
+	registryAddress, err := url.Parse(req.RegistryAddress)
```
a-palchikov: `url.Parse` will fail on addresses like `<ip>:5000` without a scheme.

r0mant: Good catch, this is actually why CI is failing.

lib/docker/imageservice.go
```diff
@@ -192,7 +226,19 @@ func (r *imageService) Sync(ctx context.Context, dir string, progress utils.Prin
 		if err != nil {
 			return nil, trace.Wrap(err)
 		}
-		remoteRepo, err := r.remoteStore.Repository(ctx, localRepoName)
+		remoteRepoName := localRepoName
+		// If the registry prefix was specified, rewrite all images to point
+		// to that domain - some registries (such as OpenShift) require a
+		// specific namespace where images should be pushed so an image like
+		// gravitational/debian-tall would become <namespace>/debian-tall.
```
knisbet: I wonder what the potential for conflicts here would be. Ie gravitational/debian-tall and quay.io/debian-tell overwrite eachother into <namespace>/debian-tall?? I don't see it in the online registries, but I think distribution supports multiple levels, so would something like gravitational/debian-tall to <namespace>/gravitational/debian-tall work?

lib/docker/imageservice.go
```diff
@@ -192,7 +226,19 @@ func (r *imageService) Sync(ctx context.Context, dir string, progress utils.Prin
 		if err != nil {
 			return nil, trace.Wrap(err)
 		}
-		remoteRepo, err := r.remoteStore.Repository(ctx, localRepoName)
+		remoteRepoName := localRepoName
+		// If the registry prefix was specified, rewrite all images to point
```
knisbet: Nit: field inconsistency, this seems to call it a registry prefix, but the variable is called domain?

lib/docker/imageservice.go
```diff
@@ -336,29 +382,59 @@ func initTransport(req RegistryConnectionRequest) (*http.Transport, string) {
...
 
-	cert, err := tls.LoadX509KeyPair(req.ClientCertPath, req.ClientKeyPath)
-	if err == nil {
-		roots, err := newCertPool([]string{req.CACertPath})
-		if err == nil {
+	// Figure out the registry address scheme (http or https). If the scheme
+	// was specified explicitly, keep it as-is. Otherwise, default to https
+	// unless the insecure flag was provided.
+	defaultScheme := "https"
+	if req.Insecure {
```
knisbet: Maybe it's a docker thing, but usually Insecure implies turning off TLS certificate chain validation, not changing the scheme to http.

lib/docker/imageservice.go
```diff
@@ -62,6 +66,36 @@ type RegistryConnectionRequest struct {
...
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	roots, err := newCertPool([]string{r.CACertPath})
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	return &tls.Config{
+		Certificates: []tls.Certificate{certificate},
+		MinVersion:   tls.VersionTLS10,
```
knisbet: Min version should be TLS 1.2 or 1.3. TLS 1.3 is modern compatability, which is what we should be using if we fully control the system. Otherwise 1.2 is acceptable for intermediate compatability.

For TLS recommended settings we've been following https://wiki.mozilla.org/Security/Server_Side_TLS and also need to hard code the settings (golang doesn't necessarily choose sensible defaults from what I recall).

lib/docker/imageservice.go
```diff
@@ -336,29 +382,59 @@ func initTransport(req RegistryConnectionRequest) (*http.Transport, string) {
...
+
+	// If the scheme was set explicitly to https, along with the insecure
+	// flag, ignore the certificate error (this is what Docker does too).
+	var err error
+	if req.Insecure && strings.HasPrefix(registryAddress, "https://") {
+		transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+	} else if strings.HasPrefix(registryAddress, "https://") {
+		transport.TLSClientConfig, err = req.TLSClientConfig()
+		if err != nil {
+			log.WithError(err).Debugf("No TLS trust for %s.", req)
```
knisbet: It seems a bit odd to log the error and drop it at Debug level if the trust store cannot be initialized. My preferred approach would be to test if the trust store config has been passed (CACertPath, etc) and hard error if that configuration cannot be used.

lib/docker/imageservice.go
```diff
@@ -336,29 +382,59 @@ func initTransport(req RegistryConnectionRequest) (*http.Transport, string) {
...
-			return transport, fmt.Sprintf("https://%v", req.RegistryAddress)
 		}
 	}
 
-	log.Debugf("No TLS trust for %s.", req)
-	transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
-	return transport, fmt.Sprintf("http://%v", req.RegistryAddress)
+	challengeManager, err := ping(transport, registryAddress)
+	if err != nil {
+		return nil, "", trace.Wrap(err)
```
knisbet: Also add fields for the address being pinged, and any other context.

lib/docker/imageservice.go
r0mant: Yeah, they will overwrite each other. I tried it with multiple levels and it doesn't work unfortunately. At least the OpenShift registry requires you to have just `<namespace>/debian-tall`.

lib/docker/imageservice.go
r0mant: Yeah, I think "prefix" will be a bit clearer.

lib/docker/imageservice.go
r0mant: Yeah, in Docker "insecure" actually kinda means both things:
* Try to use https and ignore the certificate error.
* If https is not available, use http.
I guess we could default to https and require "http://" explicitly for "insecure-insecure" registries.

lib/docker/imageservice.go
r0mant: Yeah, this is a side effect of the preexisting behavior unfortunately - these paths are actually always being passed and set to the default locations inside Gravity clusters so I kept this for now for backwards compatibility. I've updated this to use a more secure TLS config though in case system trust store is used.

PR #1078, opened 2020-02-05
lib/pack/suite: fix dropped error

alrs:
This fixes a dropped error variable in `/lib/pack/suite`.


a-palchikov:
Thank you for your PR but I think I've managed to remove most of the missing error checks in [this](https://github.com/gravitational/gravity/pull/1048) PR so this should not be necessary.

Issue #1044, opened 2020-01-24
Bump Kubernetes in 6.3 and master to 1.17.2


helgi:
Just to keep up with k8s releases https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.17.md#v1172 - a few relevant fixes in there on kubeletes and other similar

eldios:
fixed in https://github.com/gravitational/gravity/pull/1080

eldios:
Used this issue for 6.3.x branch... I'll create a separate one for master/7.x

PR #1080, opened 2020-02-05
Updated K8S and Planet version

eldios:
Updated K8S and Planet version


eldios:
fixes https://github.com/gravitational/gravity/issues/1044

Issue #1079, opened 2020-02-05
Extend gravity reports with cluster status information and etcd dump


a-palchikov:
Currently these bits are missing and would provide a lot more insight into a cluster if captured.

PR #1081, opened 2020-02-06
[5.5.x] Add etcd dump and cluster status to debug reports

a-palchikov:
Updates https://github.com/gravitational/gravity/issues/1079
Requires https://github.com/gravitational/planet/pull/553.


lib/report/k8s.go
```diff
@@ -58,25 +58,33 @@ func KubernetesInfo(ctx context.Context, runner utils.CommandRunner) Collectors
...
+				}).Warn("Failed to query container.")
 				continue
 			}
 			for _, container := range containers {
 				name := fmt.Sprintf("k8s-logs-%v-%v-%v", namespace, pod, container)
 				commands = append(commands, Cmd(name, utils.PlanetCommand(kubectl.Command("logs", pod,
 					"--namespace", namespace,
-					fmt.Sprintf("-c=%v", container)))...),
-				)
+					fmt.Sprintf("-c=%v", container)))...))
```
knisbet: Doesn't cluster-info dump also collect container logs?

a-palchikov: Yes it does and hence I went forward and deleted all ad-hoc captures for kubernetes resources leaving only the `kubectl cluster-info dump` and the captures of previous containers since kubectl is not capturing them.

PR #1084, opened 2020-02-07
Add ability to change advertise IP.

r0mant:
This PR implements ability to change advertise IP of a single-cluster node. This allows users implement scenarios where they want to deliver their pre-installed and pre-configured clusters as a full virtual appliance such as Amazon AMI or VMWare OVF/OVA.

### How it works

It introduces a new operation which I called "reconfigure" (potentially, it can also be used to reconfigure more stuff than just IP/hostname). It comes with its own plan/FSM/etc and is basically a lightweight install that runs on top of the existing cluster state and regenerates all configuration and secrets, then starts all services. Runtime/application phases are not executed b/c everything has been installed during original install (and is kinda the point of this feature).

The hard part was figuring out all the stuff that needs to be repaired on the node in order to accommodate the changed IP address. I've split those in two phases - "pre-cleanup" which is executed right away before reconfiguring anything, and "post-cleanup" which fixes stuff once we have Kubernetes. Things like removing lingering network interfaces, cleaning up old packages in both local and cluster states, recreating pods and so on. I tried to put proper descriptions on respective phase executors to explain what they do and why.

### Implementation notes

I tried to take as many shortcuts as I could to keep it simple. For example, since we're dealing with a single-node, there's no agent and everything runs locally. Also, since "reconfigure" is basically a lightweight "install" operation, it reuses a lot of the installer logic, plan generation and so on.

### Additional

Also introduced a couple of new commands:

```
$ gravity system stop [--disable]
$ gravity system start [--enable]
```

which stop/start our systemd services on the node. For convenience as they're needed for reconfiguration (see the included docs for more info).


lib/install/reconfigure/plan_test.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+}
+
+func (s *ReconfiguratorSuite) verifyPreCleanupPhase(c *check.C, phase storage.OperationPhase) {
+	master := s.suite.Master()
+	storage.DeepComparePhases(c, storage.OperationPhase{
+		ID: phases.PreCleanupPhase,
+		Phases: []storage.OperationPhase{
+			{
+				ID:   fmt.Sprintf("%v%v", phases.PreCleanupPhase, phases.NetworkPhase),
```
a-palchikov: nit: `fmt.Sprint(phases.PreCleanupPhase, phases.NetworkPhase)`

lib/install/reconfigure/plan_test.go
```diff
@@ -0,0 +1,200 @@
...
+			Server: &master,
+		},
+	}, phase)
+}
+
+func (s *ReconfiguratorSuite) verifyConfigurePhase(c *check.C, phase storage.OperationPhase) {
+	storage.DeepComparePhases(c, storage.OperationPhase{
+		ID: installphases.ConfigurePhase,
+		Data: &storage.OperationPhaseData{
+			Install: &storage.InstallOperationData{},
```
a-palchikov: Is nil value an option here?

lib/install/reconfigure/planbuilder.go
```diff
@@ -0,0 +1,130 @@
...
+			Server: &b.Master,
+		},
+	})
+}
+
+// AddPreCleanupPhase adds the phase that does pre-reconfiguration cleanups
+// on the node to the plan.
+func (b *PlanBuilder) AddPreCleanupPhase(plan *storage.OperationPlan) {
+	plan.Phases = append(plan.Phases, storage.OperationPhase{
+		ID:          phases.PreCleanupPhase,
```
a-palchikov: `PreCleanup` does not relay too much meaning. As far as I understand, the split is between things running prior to kubernetes being available and after  - what about splitting around kubernetes? Something like `PreKubernetes` and `PostKubernetes`?

lib/install/reconfigure/state.go
```diff
@@ -0,0 +1,116 @@
...
+	Cluster storage.Site
+	// InstallOperation is the original install operation.
+	InstallOperation storage.SiteOperation
+}
+
+// Server returns the server from the state.
+func (s State) Server() (*storage.Server, error) {
+	servers := s.Cluster.ClusterState.Servers
+	if len(servers) != 1 {
+		return nil, trace.BadParameter("expected 1 server, got: %#v", servers)
```
a-palchikov: nit: I think it's better to have a good formatter even though this looks like logic error which should not be user-facing but still it could help debugging when server list outputs only the relevant bits.

lib/install/reconfigure/state.go
```diff
@@ -0,0 +1,116 @@
...
+// reconfigured.
+func ValidateLocalState(state *State) error {
+	// Only single-node clusters are supported.
+	// TODO(r0mant): Because the above function uses the original site-export
+	// package to build the cluster state, this check can either fail if a
+	// multi-node cluster was shrunk down to 1 node, or will let the operation
+	// through if a 1-node cluster was expanded. Need some way to deduce the
+	// most recent cluster state based on the local state.
+	if len(state.Cluster.ClusterState.Servers) != 1 {
+		return trace.BadParameter("reconfiguration is only supported for single-node clusters, this one appears to have %v nodes",
```
a-palchikov:
```go
log.Field("state", state.Cluster.ClusterState.Servers).Warn("Reconfiguring multi-node cluster.")
return trace.BadParameter("only a single node cluster can be reconfigured") // other bits of the message are not important
```

lib/install/suite.go
```diff
@@ -0,0 +1,861 @@
...
+		Port:  10053,
+	}
+	s.cluster, err = s.services.Operator.CreateSite(
+		ops.NewSiteRequest{
+			AccountID:  account.ID,
+			DomainName: "example.com",
+			AppPackage: appPackage.String(),
+			Provider:   schema.ProviderAWS,
+			DNSConfig:  s.dnsConfig,
+		})
```
a-palchikov: `err` is dropped

lib/install/suite.go
```diff
@@ -0,0 +1,861 @@
...
+	_, err = s.services.Users.CreateClusterAdminAgent(s.cluster.Domain,
+		storage.NewUser(storage.ClusterAdminAgent(s.cluster.Domain), storage.UserSpecV2{
+			AccountID: defaults.SystemAccountID,
+		}))
+	c.Assert(err, check.IsNil)
+	s.adminAgent, err = s.services.Operator.GetClusterAgent(ops.ClusterAgentRequest{
+		AccountID:   account.ID,
+		ClusterName: s.cluster.Domain,
+		Admin:       true,
+	})
```
a-palchikov: `err` is dropped

lib/install/reconfigure/plan.go
```diff
@@ -0,0 +1,96 @@
...
+		ClusterName:   operation.SiteDomain,
+		Servers:       append(builder.Masters, builder.Nodes...),
+		DNSConfig:     cluster.DNSConfig,
+	}
+
+	builder.AddPreCleanupPhase(plan)
+	builder.AddChecksPhase(plan)
+	builder.AddConfigurePhase(plan)
+	builder.AddPullPhase(plan)
+	builder.AddMastersPhase(plan)
```
a-palchikov: should check the error return

tool/gravity/cli/reconfigure.go
```diff
@@ -0,0 +1,152 @@
...
+		ConnectStrategy: strategy,
+		Lifecycle: &installerclient.AutomaticLifecycle{
+			Aborter:            AborterForMode(config.Mode, env),
+			Completer:          InstallerCompleteOperation(env),
+			DebugReportPath:    DebugReportPath(),
+			LocalDebugReporter: InstallerGenerateLocalReport(env),
+		},
+	})
+	if utils.IsContextCancelledError(err) {
+		InstallerCleanup()
```
a-palchikov: this should handle an error

lib/system/environ/uninstall.go
```diff
@@ -131,6 +157,57 @@ func DisableAgentServices(logger log.FieldLogger) error {
...
+	}
+	var errors []error
+	for _, service := range services {
+		printer.PrintStep("Stopping system service %s", service.Package)
+		log := logger.WithField("package", service.Package)
+		err := svm.StopPackageService(service.Package)
+		if err != nil {
+			log.WithError(err).Warn("Failed to stop service.")
+			errors = append(errors, err)
+		} else if disable {
```
a-palchikov: this should be a separate function (same applies for start/enable)

tool/gravity/cli/reconfigure.go
```diff
@@ -0,0 +1,152 @@
...
+// address for single-node clusters.
+func reconfigureCluster(env *localenv.LocalEnvironment, config InstallConfig, confirmed bool) error {
+	// See if there's a cluster at all and whether it's running. The operation
+	// can be only performed when everything's stopped.
+	err := localenv.DetectCluster(env)
+	if err != nil && trace.IsNotFound(err) {
+		return trace.NotFound("The cluster doesn't appear to be installed on this node.")
+	}
+	if err == nil {
+		return trace.NotFound(`The cluster appears to be running on this node. Please stop all services using "gravity system stop" prior to initiating the operation.`)
```
a-palchikov: Why `NotFound` error here? I think `AlreadyExists` fits better (for the lack of `AlreadyRunning`).

tool/gravity/cli/reconfigure.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/gravitational/gravity/lib/localenv"
+	"github.com/gravitational/gravity/lib/system/signals"
+	"github.com/gravitational/gravity/lib/utils"
+	"github.com/gravitational/gravity/lib/utils/cli"
+
+	"github.com/fatih/color"
+	"github.com/gravitational/trace"
+)
+
+const reconfigureMessage = `WARNING: This action will launch the operation of reconfiguring the cluster installed on this node to use a new advertise IP %v.
```
a-palchikov: `Warning: this operation will update the advertise address of this cluster to %q - would you like to proceed?`

tool/gravity/cli/register.go
```diff
@@ -139,6 +139,12 @@ func RegisterCommands(app *kingpin.Application) *Application {
 	g.ResumeCmd.Force = g.ResumeCmd.Flag("force", "Force execution of specified phase.").Bool()
 	g.ResumeCmd.PhaseTimeout = g.ResumeCmd.Flag("timeout", "Phase execution timeout.").Default(defaults.PhaseTimeout).Hidden().Duration()
 
+	g.ReconfigureCmd.CmdClause = g.Command("reconfigure", `Reconfigure the advertise address used by the node. Supported for single-node clusters only. Cluster services must be stopped (e.g. using "gravity system stop") prior to initiating the operation.`)
```
a-palchikov: Can the services be shut down automatically if the user confirms the operation?

tool/gravity/cli/system.go
```diff
@@ -558,6 +559,68 @@ func systemServiceStatus(env *localenv.LocalEnvironment, pkg loc.Locator, servic
 	return nil
 }
 
+const systemStopMessage = `WARNING: This action will stop all Gravity and Kubernetes services on the node.
```
a-palchikov: Do we want to display warning banners for these commands? Could corresponding high-level commands do this automatically?

lib/install/reconfigure/phases/cleanup.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/gravitational/trace"
+	"k8s.io/client-go/kubernetes"
+)
+
+// NewCleanup dispatches the specific cleanup phase to an appropriate executor.
+func NewCleanup(p fsm.ExecutorParams, operator ops.Operator, packages pack.PackageService, client *kubernetes.Clientset) (fsm.PhaseExecutor, error) {
+	switch {
+	case strings.HasPrefix(p.Phase.ID, PreCleanupPhase):
+		switch phaseID := strings.Replace(p.Phase.ID, PreCleanupPhase, "", 1); phaseID {
```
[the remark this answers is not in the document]
r0mant: Yeah, flatter plan structure is probably better in this case, will update.

lib/install/reconfigure/phases/packages.go
```diff
@@ -0,0 +1,93 @@
...
+type packagesExecutor struct {
+	// FieldLogger is used for logging.
+	logrus.FieldLogger
+	// ExecutorParams are common executor parameters.
+	fsm.ExecutorParams
+	// LocalPackages is the node-local package service.
+	LocalPackages pack.PackageService
+}
+
+// Execute removes old configuration & secret packages from the node.
```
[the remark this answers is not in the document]
r0mant: So yeah, currently only config/secret packages are marked with this label - but the way I was thinking, we're removing everything for the old node so if there were any other packages marked it, we'd have to remove them too anyway.

lib/install/reconfigure/phases/pods.go
```diff
@@ -0,0 +1,95 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// NewPods returns executor that removes old Kubernetes pods.
+//
+// All Kubernetes pods have to be recreated when the cluster comes up after
+// reconfiguration to ensure that no old pods are lingering (Kubernetes may
+// get confused after the old node is gone and keep pods in terminating state
+// for a long time) and that they mount proper service tokens.
```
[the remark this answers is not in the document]
r0mant: That's what I thought/hoped for too, but unfortunately it doesn't seem to be the case. In my testing, the Kubernetes would often get "confused" (for the lack of better word) and all the pods for the old node would get stuck in terminating state indefinitely (even after the node is gone) messing up scheduling for the new ones, so I had to resort to recreating all of them.

r0mant:
@a-palchikov @knisbet Think I've addressed all/most of your guys feedback and the feedback I got from the product team. See the latest commit if you're interested in the changes but here's the gist of stuff worth mentioning:

* UX changed a bit. There's no `gravity reconfigure` command anymore - instead, a user uses `gravity start --advertise-addr=xxx` command to start the node with a new IP. Under the hoods it still does the "reconfigure" operation. Also, the stop command is just `gravity stop` now (not a system subcommand).
* No more requirement to have the original installer present on the node.
* Fixed bad progress output that would duplicate timestamps.
* Fixed an issue with installer intermittently failing to connect to the process.
* Originally I used to use the "site-export" package to determine the state of the cluster in the absence of the running cluster, which is problematic for numerous reasons. Now I've changed this to instead have the process periodically sync its cluster state to the local backend and use that.

@knisbet - with IP address not being present on the node - this is actually checked for and if I give `--advertise-addr=1.1.1.1` on my machine, it correctly returns an error. If it still happens for you (or if I misunderstood the issue), might be worth investigating.

tool/gravity/cli/reconfigure.go
```diff
@@ -150,3 +139,33 @@ func newReconfigurator(ctx context.Context, config *install.Config, state *recon
...
+// validateReconfiguration determines if reconfiguration can be launched on this
+// node.
+//
+// If all checks pass, returns information about the existing cluster state such
+// as the cluster object and the original install operation.
+func validateReconfiguration(env *localenv.LocalEnvironment, config InstallConfig) (*reconfigure.State, error) {
+	// The cluster should be installed but not running.
+	err := localenv.DetectCluster(env)
+	if err != nil && trace.IsNotFound(err) {
+		return nil, trace.BadParameter("Cluster doesn't appear to be installed on this node.")
```
a-palchikov: nit: `Cluster` and `node` don't mix well when reading this - how about `Gravity is not installed`?

tool/gravity/cli/register.go
```diff
@@ -139,11 +139,16 @@ func RegisterCommands(app *kingpin.Application) *Application {
 	g.ResumeCmd.Force = g.ResumeCmd.Flag("force", "Force execution of specified phase.").Bool()
 	g.ResumeCmd.PhaseTimeout = g.ResumeCmd.Flag("timeout", "Phase execution timeout.").Default(defaults.PhaseTimeout).Hidden().Duration()
 
-	g.ReconfigureCmd.CmdClause = g.Command("reconfigure", `Reconfigure the advertise address used by the node. Supported for single-node clusters only. Cluster services must be stopped (e.g. using "gravity system stop") prior to initiating the operation.`)
-	g.ReconfigureCmd.Path = g.ReconfigureCmd.Arg("path", "Path to the directory with the unpacked cluster image. Defaults to the current directory.").String()
-	g.ReconfigureCmd.AdvertiseAddr = g.ReconfigureCmd.Flag("advertise-addr", "New advertise address the node will use. Must be present on the node. Will be auto-selected if not specified.").String()
-	g.ReconfigureCmd.FromService = g.ReconfigureCmd.Flag("from-service", "Run in service mode.").Hidden().Bool()
-	g.ReconfigureCmd.Confirm = g.ReconfigureCmd.Flag("confirm", "Suppress confirmation prompt.").Bool()
+	// TODO(r0mant): Add flags --planet and --teleport to stop/start command.
```
a-palchikov: Why do you need these flags? Are they supposed to control individual services outside the context of the reconfiguration operation?

tool/gravity/cli/register.go
```diff
@@ -139,11 +139,16 @@ func RegisterCommands(app *kingpin.Application) *Application {
...
+	g.StopCmd.CmdClause = g.Command("stop", "Stop Gravity services on the node.")
+	g.StopCmd.Confirmed = g.StopCmd.Flag("confirm", "Suppress confirmation prompt.").Bool()
+	g.StopCmd.Disable = g.StopCmd.Flag("disable", "Disable systemd services as well.").Bool()
+
+	g.StartCmd.CmdClause = g.Command("start", "Start Gravity services on the node.")
+	g.StartCmd.AdvertiseAddr = g.StartCmd.Flag("advertise-addr", "New advertise address the node will use. Must be present on the node.").String()
```
a-palchikov: Is this parameter validated?

tool/gravity/cli/stop_start.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+		env.PrintStep("Starting system service %s", pack)
+		err := svm.StartPackageService(pack, false)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+	}
+
+	if enable {
+		env.PrintStep("Gravity services have been enabled and started")
```
a-palchikov: nit: I really feel there's no need to make user differentiate these states - can the services be started and disabled?

a-palchikov:
@r0mant Also, the corresponding enterprise PR seems to be missing.

tool/gravity/cli/register.go
r0mant: This is an old TODO item, forgot to remove.

tool/gravity/cli/register.go
r0mant: Yes, same way as the advertise-addr parameter for gravity install.

docs/6.x/cluster.md
```diff
@@ -1480,39 +1481,34 @@ There are a few restrictions and assumptions about this procedure to keep in min
...
+Stop and optionally disable all Gravity and Kubernetes services on the node:
 
 ```bash
-$ sudo gravity system stop --disable
+$ sudo gravity stop [--disable]
 ```
 
-At this point you can create a machine image (AMI/OVF/OVA/etc). If needed, restart the services on the node after the
-image has been created:
+At this point the machine's snapshot (AMI/OVF/OVA/etc) can be taken. The `--disable` flag can be provided to ensure that
```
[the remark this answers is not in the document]
r0mant: Hm, I suppose we could always disable/enable.

lib/process/process.go
```diff
@@ -1412,6 +1485,8 @@ func (p *Process) initService(ctx context.Context) (err error) {
 		p.startService(p.runApplicationsSynchronizer)
 		p.startService(p.runNodeLabelsReconciler(client))
 
+		go p.runClusterStateSynchronizer(p.context)
```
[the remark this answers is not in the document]
r0mant: Didn't even know we had this.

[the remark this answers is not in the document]
a-palchikov: It also drops the error from the function so we don't know if it started.

docs/6.x/cluster.md
```diff
@@ -1464,6 +1464,54 @@ The command can also output the images in the json or yaml format which can come
 $ gravity registry list --format=json
 ```
 
+## Changing Node Advertise Address
+
+!!! note "Supported version":
+    Changing a node's advertise address is supported starting from Gravity 7.
```
benarent: I think we should fork our docs and put this into 7.0 vs putting it into 6.x. Unless this feature is party back-ported to an earlier version.

docs/6.x/cluster.md
```diff
@@ -1464,6 +1464,54 @@ The command can also output the images in the json or yaml format which can come
...
+use the command:
+
+```bash
+$ sudo gravity start
+```
+
+Once the image has been deployed on a new node, start the cluster providing a new advertise address configuration:
+
+```bash
+$ sudo gravity start --advertise-addr=<new-ip>
```
benarent: Do we support IPv4 and IPv6?

docs/6.x/cluster.md
r0mant: Yeah, we should. Not as a part of this PR though.

docs/6.x/cluster.md
r0mant: Only IPv4.

PR #1089, opened 2020-02-07
Create CODEOWNERS

knisbet:
What does the team think about using code-owners to automatically assign code reviewers?

The owners can be pointed at github teams, but for the short term it's probably just as easy to assign individual reviewers.

https://github.blog/2017-07-06-introducing-code-owners/
https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners


a-palchikov:
I think it's a good idea.

knisbet:
> That's pretty cool. This will also block the PR from being merged until one of us reviews it, correct?

I think we can turn that on yes as a branch protection, so yes.

PR #1103, opened 2020-02-13
[6.3.x] backport fix for helm symlink on host

a-palchikov:
Backports [this](https://github.com/gravitational/gravity/pull/1033/commits/d63601d1a0dfc473071f9354ce010e64abb21329).

### DIFF
--- a/lib/log/log.go
+++ b/lib/log/log.go
@@ -1,0 +1,142 @@
+/*
+Copyright 2019 Gravitational, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


lib/log/log.go
```diff
@@ -0,0 +1,142 @@
...
+package log
+
+import (
+	"io"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Logger defines a subset of the structured logging interface
+type Logger interface {
```
r0mant: Do we need to add this to our lib package?

a-palchikov: Not sure I follow - it is in `lib` already.

r0mant: Yeah, sorry - the question was **why** do we need to add this to our lib package :) Seems like it mostly vendors the logrus interface?

a-palchikov: It is meant as a replacement for `FieldLogger` interface for when one needs to provide FieldLogger-like interface backed by an alternate implementation - for example, [fsm.Logger](https://github.com/gravitational/gravity/blob/version/6.3.x/lib/fsm/logger.go#L32) cannot correctly expose `FieldLogger` interface since it cannot handle the `WithXXX` APIs which [return](https://github.com/sirupsen/logrus/blob/v1.0.6/logrus.go#L113-L115) the concrete `*log.Entry` instead of interface values.

PR #1104, opened 2020-02-13
[6.1.x] convenience commands for gravity-specific services

a-palchikov:
Add low-level convenience commands to work with gravity-specific systemd services:

```sh
$ gravity system service stop planet
$ gravity system service start teleport
$ gravity system service journal planet -r
```
Note, that `planet` / `teleport` in the examples above is a pattern - not a service name. It is matched against a list of (possibly active - depending on the action) package services. It ensures that the pattern is unique - otherwise an error is given.

`system service journal` also accepts an arbitrary list of `journalctl` options (`-r` in this example).

A full package locator (albeit less convenient) is also accepted:
```sh
gravity system service stop gravity__gravitational.io__planet__6.1.14-11505-30-g9d0ccde.service
```

Needs forward port to master.

### DIFF
--- a/lib/systemservice/systemd.go
+++ b/lib/systemservice/systemd.go
@@ -231,7 +231,7 @@ func (s *systemdManager) DisablePackageService(pkg loc.Locator) error {
 
 // IsPackageServiceInstalled checks if the package service is installed
 func (s *systemdManager) IsPackageServiceInstalled(pkg loc.Locator) (bool, error) {
-	units, err := s.ListPackageServices()
+	units, err := s.ListPackageServices(DefaultListServiceOptions)
 	if err != nil {
 		return false, trace.Wrap(err)
 	}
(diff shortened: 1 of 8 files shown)


lib/systemservice/systemd.go
```diff
@@ -244,10 +244,23 @@ func (s *systemdManager) IsPackageServiceInstalled(pkg loc.Locator) (bool, error
...
+		args = append(args, "--all")
+	}
+	if opts.Type != "" {
+		args = append(args, "--type", opts.Type)
+	}
+	if opts.State != "" {
+		args = append(args, "--state", opts.State)
+	}
+	if opts.Pattern != "" {
+		args = append(args, opts.Pattern)
```
r0mant: Should this be quoted maybe?

a-palchikov: I don't believe so - system `exec` will be taking care of that.

PR #1107, opened 2020-02-14
Updated kubernetes to 1.17.3

eldios:
Updated kubernetes to 1.17.3


eldios:
using temp PLANET_TAG for now

PR #1108, opened 2020-02-14
[6.1.x] SELinux support

a-palchikov:
Add SELinux support to Gravtiy.

The policy itself is available in a [separate repository](github.com/gravitational/selinux/).
Policy is embedded in the binary during the build to make bootstrapping automated (esp. important for additional nodes joined to a cluster).

The most complexity lies mostly in the bootstrapping process which is implemented by the new [selinux](https://github.com/gravitational/gravity/blob/dmitri/6.1.x/3224-selinux-container/lib/system/selinux/bootstrap.go) package.
Due to some peculiarities with using stock SELinux policy tools, the bootstrapping process uses transactions for updates (as individual edit operations will be much slower for the amount of updates Gravity needs). The necessary support (specifically filecontext and port configurations) is available in the [schema](https://github.com/gravitational/gravity/tree/dmitri/6.1.x/3224-selinux-container/lib/system/selinux/internal/schema) package.

### Upgrade
Upgrades will only use SELinux if the cluster has it turned on.

### TODO
Take custom ports (from the application manifest) into account when bootstrapping the local ports.

### Limitations
Currently only redhat/CentOS are supported.

Requires https://github.com/gravitational/planet/pull/542 and https://github.com/gravitational/gravity.e/pull/4273.


build.assets/Makefile
```diff
@@ -72,7 +75,14 @@ TELEKUBE_TSH_TAG := v3.2.13
 # Extra flags that may be provided when publishing telekube artifacts (e.g. --insecure)
 TELEKUBE_PUBLISH_FLAGS ?=
 
+GOCACHE_DOCKER_OPTIONS ?=
+GOCACHE ?= $(HOME)/.cache/go-build
```
a-palchikov: This helps tremendously during development and it's off by default (hence should not be an issue on jenkins) so I'm able to opt-in if necessary.

lib/expand/bootstrap.go
```diff
@@ -200,3 +205,17 @@ func getPeerAddrAndToken(ctx operationContext, role string) (peerAddr, token str
 	}
 	return peerAddr, instructions.Token, nil
 }
+
+func patchSELinuxConfig(ctx operationContext) error {
```
a-palchikov: Configuring custom values is tricky on joining nodes - in this case, in order to set up the custom `vxlan` port, the installer needs to bootstrap the policy with enough rules to make it to dial the cluster and, after learning about the overrides - update the port configuration with the new value for the port.

lib/install/phases/bootstrap.go
```diff
@@ -197,6 +198,7 @@ func (p *bootstrapExecutor) configureSystemDirectories() error {
 		filepath.Join(stateDir, "planet", "etcd"),
 		filepath.Join(stateDir, "planet", "registry"),
 		filepath.Join(stateDir, "planet", "docker"),
+		filepath.Join(stateDir, "planet", "kubelet"),
```
a-palchikov: Needed to have additional directories explicitly created in order to correctly label them (see also for `logrange` below).

lib/rpc/deploy.go
```diff
@@ -209,9 +210,11 @@ func deployAgentOnNode(ctx context.Context, req DeployAgentsRequest, node, nodeS
 		WithRetries("%s package unpack %s %s --debug --ops-url=%s --insecure",
 			constants.GravityBin, secretsPackage, secretsHostDir, defaults.GravityServiceURL).
 		IgnoreError("/bin/systemctl stop %s", defaults.GravityRPCAgentServiceName).
-		WithRetries("%s package export --file-mask=%o %s %s --ops-url=%s --insecure",
+		WithRetries("%s package export --file-mask=%o %s %s --ops-url=%s --insecure --file-label=%s",
 			constants.GravityBin, defaults.SharedExecutableMask,
-			req.GravityPackage, gravityHostPath, defaults.GravityServiceURL).
+			req.GravityPackage, gravityHostPath, defaults.GravityServiceURL,
+			selinux.GravityProcessLabel,
```
a-palchikov: Need to have the new binary run with proper SELinux context.

lib/schema/ports.go
```diff
@@ -0,0 +1,76 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package schema
+
+// DefaultPortRanges defines the list of default ports for the cluster
+var DefaultPortRanges = PortRanges{
```
a-palchikov: I had to split ports into groups to support the policy grouping.
Also, certain ports are actually defined in another policy (like `5000` we use for registry - these are grouped in `Reserved`) so it needs to be configurable separately.

lib/system/selinux/bootstrap.go
```diff
@@ -0,0 +1,525 @@
...
+fcontext -a -f a -t gravity_home_t -r 's0' '{{.Path}}/.gravity(/.*)?'
+`))
+
+var unloadScript = template.Must(template.New("selinux-unload").Parse(`
+port -D
+fcontext -D
+`))
+
+// TODO: reflect the actual state of local customizations (incl. custom vxlan port)
+var unloadScript0 = template.Must(template.New("selinux-unload0").Parse(`
```
a-palchikov: This is currently under TODO but I'll be returning to using this at some point.

tool/gravity/cli/commands.go
```diff
@@ -1528,6 +1541,16 @@ type SystemStreamRuntimeJournalCmd struct {
 	*kingpin.CmdClause
 }
 
+// SystemSelinuxBootstrapCmd configures SELinux file contexts and ports on the node
+type SystemSelinuxBootstrapCmd struct {
```
a-palchikov: Since the policy is automatically bootstrapped - there's no way to do this manually with ease. This will allow to generate the local filecontext/port changes script to run along with the policy module.
I might be adding additional command to extract the policy modules as I cannot completely exclude the possibility of something breaking and us needing to load policy manually.

lib/checks/autofix/autofix.go
```diff
@@ -45,7 +45,7 @@ func Fix(ctx context.Context, probes []*agentpb.Probe, progress utils.Progress)
 		if probe.Status != agentpb.Probe_Failed {
 			continue
 		}
-		if err := fixProbe(ctx, probe, progress); err != nil {
+		if err := fixProbe(ctx, probe, progress); err != nil && !trace.IsNotImplemented(err) {
```
r0mant: So if fixProbe returns "not implemented" error, the probe will be added to fixed probes?

lib/expand/bootstrap.go
r0mant: Could you add a comment about what this function does - "patch selinux config" doesn't tell much.

lib/install/phases/bootstrap.go
```diff
@@ -351,6 +358,23 @@ func (*bootstrapExecutor) PostCheck(ctx context.Context) error {
 	return nil
 }
 
+func (p *bootstrapExecutor) applySelinuxFilecontexts(ctx context.Context, stateDir string) error {
```
r0mant: Nit: Selinux -> SELinux, Filecontexts -> FileContexts?

lib/install/phases/bootstrap.go
```diff
@@ -351,6 +358,23 @@ func (*bootstrapExecutor) PostCheck(ctx context.Context) error {
...
+	if !(selinux.GetEnabled() && p.Plan.SELinux) {
+		p.Info("SELinux is disabled.")
+		return nil
+	}
+	out, err := exec.CommandContext(ctx, "restorecon", "-R", "-v", stateDir).CombinedOutput()
```
r0mant: Could you add a comment here about what "restorecon" command does?

lib/ops/opsservice/site.go
```diff
@@ -187,6 +187,10 @@ func (s *site) agentUser() (storage.User, error) {
 	return s.users().GetTelekubeUser(s.agentUserEmail())
 }
 
+func (s *site) selinuxEnabled() (enabled bool) {
```
r0mant: Nit: seLinux?

lib/schema/ports.go
```diff
@@ -0,0 +1,76 @@
...
+		{Protocol: "tcp", From: 2379, To: 2380, Description: "etcd"},
+		{Protocol: "tcp", From: 6443, To: 6443, Description: "kubernetes API server"},
+		{Protocol: "tcp", From: 10248, To: 10255, Description: "kubernetes internal services range"},
+		{Protocol: "tcp", From: 7496, To: 7496, Description: "serf (health check agents) peer to peer"},
+		{Protocol: "tcp", From: 7373, To: 7373, Description: "serf (health check agents) peer to peer"},
+	},
+	Installer: []PortRange{
+		{Protocol: "tcp", From: 61009, To: 61010, Description: "installer ports"},
+		{Protocol: "tcp", From: 61022, To: 61025, Description: "installer ports"},
+		{Protocol: "tcp", From: 61080, To: 61080, Description: "installer ports"},
```
r0mant: Is this actually used anywhere?

lib/schema/ports.go
```diff
@@ -0,0 +1,76 @@
...
+		{Protocol: "tcp", From: 3012, To: 3012, Description: "gravity agent RPC"},
+	},
+	Vxlan: PortRange{
+		Protocol: "udp", From: 8472, To: 8472, Description: "overlay network",
+	},
+	Generic: []PortRange{
+		{Protocol: "tcp", From: 3022, To: 3025, Description: "teleport internal SSH control panel"},
+		{Protocol: "tcp", From: 3080, To: 3080, Description: "teleport Web UI"},
```
r0mant: I don't think this is used either - Teleport web UI is disabled in Gravity.

lib/schema/ports.go
```diff
@@ -0,0 +1,76 @@
...
+		{Protocol: "tcp", From: 3008, To: 3011, Description: "internal gravity services"},
+		{Protocol: "tcp", From: 32009, To: 32009, Description: "gravity Hub control panel"},
+		{Protocol: "tcp", From: 7575, To: 7575, Description: "planet agent RPC"},
+	},
+	Reserved: []PortRange{
+		// Defines as kubernetes_port_t by default
+		{Protocol: "tcp", From: 4001, To: 4001, Description: "etcd"},
+		// Defines as afs3_callback_port_t by default
+		{Protocol: "tcp", From: 7001, To: 7001, Description: "etcd"},
+		// Defines as commplex_main_port_t by default
```
r0mant: Nit: Is "commplex" a typo (for "complex") or not? Also, should it say "Define**d**" here and above?

lib/system/environ/uninstall.go
```diff
@@ -219,17 +237,20 @@ func dockerInfo() (*utils.DockerInfo, error) {
...
 	stateDir, err := state.GetStateDir()
 	if err == nil {
-		dirs = append(dirs, stateDir)
+		paths = append(paths, stateDir)
 	}
-	dirs = append(dirs, state.StateLocatorPaths...)
-	if stateDir, err := state.GravityInstallDir(); err == nil {
-		dirs = append(dirs, stateDir)
+	paths = append(paths, state.StateLocatorPaths...)
+	if utils.Exe.WorkingDir != "/" {
```
r0mant: What does this test for? Not sure I understand. A comment might be warranted.

lib/system/selinux/bootstrap.go
```diff
@@ -0,0 +1,525 @@
...
+	// TODO: remove the policy module.
+	// Note, this is only possible when there're no more gravity-labeled
+	// files/directories.
+}
+
+// IsSystemSupported returns true if the system specified with given ID
+// is supported
+func IsSystemSupported(systemID string) bool {
+	switch systemID {
+	case "centos", "rhel":
```
r0mant: We probably have constants for these IDs somewhere (where we do OS test).

lib/system/selinux/bootstrap.go
```diff
@@ -0,0 +1,525 @@
...
+		return true
+	default:
+		return false
+	}
+}
+
+// Patch executes the patch script with the underlying configuration
+func (r PatchConfig) Patch() error {
```
r0mant: So this seems to be dealing specifically with custom vxlan ports, so maybe call this function so? PatchVxlan. And update the godoc respectively.

tool/gravity/cli/clusterupdate.go
```diff
@@ -392,3 +399,28 @@ Please use the gravity binary from the upgrade installer tarball to execute the
...
+	selinuxEnabled := status.Cluster != nil && status.Cluster.SELinux
+	if !selinuxEnabled {
+		// Nothing to do
+		return nil
+	}
+	return selinux.Bootstrap(selinux.BootstrapConfig{})
+}
+
+func queryClusterStatus() (status *libstatus.Status, err error) {
+	out, err := exec.Command("gravity", "status", "--output=json").CombinedOutput()
```
r0mant: Why not use the API? Or localenv.ClusterOperation - seems like you only need the cluster above?

tool/gravity/cli/config.go
```diff
@@ -286,7 +297,9 @@ func (i *InstallConfig) CheckAndSetDefaults() (err error) {
 	}
 	err = i.validateApplicationDir()
 	if err != nil {
-		return trace.Wrap(err)
+		return trace.Wrap(err, "failed to validate installer directory. "+
+			"Make sure you're running the installer from the directory with the contents "+
```
r0mant: "Make sure you're launching install from the unpacked cluster image directory." <-- IIRC we wanted to get rid of the "installer" terminology.

Also, do you actually have to launch it from that directory? The install command also accepts a path IIRC.

tool/gravity/cli/config.go
```diff
@@ -436,6 +450,27 @@ func (i *InstallConfig) RunLocalChecks() error {
...
+func (i *InstallConfig) BootstrapSELinux(printer utils.Printer) error {
+	if !i.SELinux || i.FromService {
+		return nil
+	}
+	metadata, err := monitoring.GetOSRelease()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if !selinux.IsSystemSupported(metadata.ID) {
+		log.WithField("os", metadata).Info("No SELinux policy for specified system, will skip.")
```
r0mant: Will an error be returned if you attempt to launch it with selinux on an unsupported distro? Or it will silently fallback to "no selinux"?

tool/gravity/cli/selinux.go
```diff
@@ -0,0 +1,96 @@
+/*
+Copyright 2019 Gravitational, Inc.
```
r0mant: 2020

tool/gravity/cli/selinux.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if !isSELinuxAlreadyBootstrapped() {
+		printer.PrintStep("Bootstrapping installer for SELinux")
+		if err := libselinux.Bootstrap(config); err != nil {
+			return trace.Wrap(err)
+		}
+	}
+	if procContext["type"] == libselinux.GravityInstallerProcessContext["type"] {
```
r0mant: What does this check for? A comment would be nice.

tool/gravity/cli/selinux.go
```diff
@@ -0,0 +1,96 @@
...
+	logger.WithField("context", newProcContext).Info("Set process context.")
+	if err := selinux.SetExecLabel(newProcContext.Get()); err != nil {
+		return trace.Wrap(err)
+	}
+	logger.WithField("args", os.Args).Info("Respawn.")
+	cmd := os.Args[0]
+	return syscall.Exec(cmd, os.Args, newRespawnEnviron())
+}
+
+func bootstrapSelinux(env *localenv.LocalEnvironment, path, stateDir string, vxlanPort int) error {
```
r0mant: Nit: SELinux?

tool/gravity/cli/selinux.go
```diff
@@ -0,0 +1,96 @@
...
+	return libselinux.WriteBootstrapScript(f, config)
+}
+
+func isSELinuxAlreadyBootstrapped() bool {
+	_, ok := os.LookupEnv(alreadyBootstrappedEnv)
+	return ok
+}
+
+func newRespawnEnviron() (environ []string) {
+	return append(os.Environ(), fmt.Sprintf("%v=yes", alreadyBootstrappedEnv))
```
r0mant: I wonder why you decided to use environment variable to pass this info, as opposed to cli flag for example.

tool/gravity/cli/status.go
```diff
@@ -284,6 +285,7 @@ func printStatusText(cluster clusterStatus) {
 }
 
 func printClusterStatus(cluster statusapi.Cluster, w io.Writer) {
+	fmt.Fprintf(w, "SELinux support:\t%v\n", formatSELinuxStatus(cluster.SELinux))
```
r0mant: Nit: I would only display this if selinux is enabled. Otherwise it's gonna be noise most of the time.

r0mant (commented):
First pass, gotta look more into lib/selinux package.

lib/checks/autofix/autofix.go
a-palchikov: Hmm - good catch since I wanted to only avoid log spam in the impossible case.

lib/schema/ports.go
a-palchikov: Yes, [it is](https://github.com/gravitational/gravity/blob/dmitri/6.1.x/3224-selinux-container/vendor/github.com/gravitational/teleport/lib/config/configuration.go#L464) - let me know if that's also bogus.

lib/schema/ports.go
a-palchikov: Removed.

lib/schema/ports.go
a-palchikov: No, [it's](https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/kernel/corenetwork.te.in#L138) not a typo.

tool/gravity/cli/clusterupdate.go
a-palchikov: Not really follow about the `localenv.ClusterOperation` - is this a new API?
The reason this does not use any APIs - the use of an API would require connecting to a port which the current executable might not be capable of - hence, it uses the existing binary (which should be correctly labeled on an installed cluster with SELinux previously enabled) to learn whether it should bootstrap.
If you have a better idea how to work around this chicken and egg problem - I'd be more than grateful.

tool/gravity/cli/config.go
a-palchikov: It is the default mode of operation - not sure a user would bother specifying the path on a command line although I could hint at that in the message.

tool/gravity/cli/config.go
a-palchikov: Since the knob is on automatically, I thought it is wrong to fail with an error - it flips it and continues as if it has been disabled.

tool/gravity/cli/selinux.go
a-palchikov: Because it is very simple and local - the command line option would require a lot more changes.

lib/system/environ/uninstall.go
```diff
@@ -219,22 +238,31 @@ func dockerInfo() (*utils.DockerInfo, error) {
...
 	stateDir, err := state.GetStateDir()
 	if err == nil {
-		dirs = append(dirs, stateDir)
+		paths = append(paths, stateDir)
 	}
-	dirs = append(dirs, state.StateLocatorPaths...)
-	if stateDir, err := state.GravityInstallDir(); err == nil {
-		dirs = append(dirs, stateDir)
+	paths = append(paths, state.StateLocatorPaths...)
+	// do not attempt to remove state directory if started with root
```
r0mant: Why?

lib/system/selinux/internal/schema/ports.go
```diff
@@ -0,0 +1,101 @@
...
+			}
+			return trace.Wrap(err)
+		}
+		args = append(args, token)
+	}
+	return r.Parse(args)
+}
+
+// Parse parses port command from the specified arguments
+func (r *PortCommand) Parse(args []string) error {
```
r0mant: Doesn't all this rather belong in cli package?

r0mant (approved):

a-palchikov:
Closing until we've cleared whether we need to backport this.

PR #1110, opened 2020-02-15
re-arrange coredns config generation

knisbet:
re-arrange coredns config generation to be consistent between install and upgrade. The upgrade code that generates coredns config is missing the logic to skip loopback addresses, and also to read systemd-resolved.

Requires backport


r0mant (approved):

lib/install/phases/coredns.go
```diff
@@ -163,8 +145,26 @@ func (r *corednsExecutor) Rollback(context.Context) error {
...
+	// Optionally try and load upstream nameservers from systemd-resolved by reading the compatibility resolv.conf
+	// More Info: https://github.com/gravitational/gravity/issues/606#issuecomment-529171440
+	// TODO(knisbet) is there a better way to pull upstream resolvers directly from systemd?
+	systemdResolvConf, err := systeminfo.ResolvFromFile("/run/systemd/resolve/resolv.conf")
+	if err != nil && !trace.IsNotFound(err) {
+		return "", trace.Wrap(err)
+	}
+
+	config.UpstreamNameservers = mergeUpstreamResolvers(resolvConf, systemdResolvConf)
+	config.Rotate = resolvConf.Rotate || systemdResolvConf.Rotate
```
a-palchikov: This is going to use `systemdResolvConf` even when it's not been initialized (i.e. the file was not read)

a-palchikov (changes requested):
So, all three coredns tests are failing when I run them locally:
first one because apparently I have a different set of nameservers in my resolv.conf:
```
forward . 192.168.178.331 2a02:8XXX:4440:3XX.8:d33:527b:.8XXX:5fb3.8 {
    policy sequential
    health_check 0
}
// it is a mixture of values: expected 1.1.1.1 and 8.8.8.8 while existing 192.168.178.33 and 2a02:8XXX:4440:3XXX:d33:527b:8XXX:5fb3
```
the other two are failing due to different reasons but my point is - the test has to take test input not use the actual `resolv.conf` contents.

lib/install/phases/coredns.go
knisbet: I'm not sure I understand this concern. An uninitialized config will have the rotate bool as false. A boolean or of false will stay false. So whichever config we're able to pick up and merge, if it's set on one it will be set on both. I suppose there is an issue if both /etc/resolv and the systemd resolv both specify upstreams that are configured in a different way, but for now we're just merging the configs, we don't have a good way to dictate which we should be using.

knisbet:
@a-palchikov my bad on the unit tests, I separated them out. I just submitted this quickly after a customer call after my day had ended. Should be fixed now.

lib/install/phases/coredns.go
a-palchikov: My concern is the potential nil pointer access.

knisbet: IC, yea I missed that ResolvFromFile returns nil and not an empty struct.

a-palchikov (approved):

PR #1117, opened 2020-02-19
Update docs/5.x with status history command

bernardjkim:


docs/5.x/cluster.md
```diff
@@ -221,6 +221,51 @@ $ curl -sk https://localhost:7575 | python -m json.tool
 
 In this case the response HTTP status code will be `503 Service Unavailable`.
 
+### Cluster Status History
+
+Running `gravity status history` will display a history of changes to the
```
a-palchikov: nit: the history?

docs/5.x/cluster.md
```diff
@@ -221,6 +221,51 @@ $ curl -sk https://localhost:7575 | python -m json.tool
 
 In this case the response HTTP status code will be `503 Service Unavailable`.
 
+### Cluster Status History
+
+Running `gravity status history` will display a history of changes to the 
+cluster status. 
+
+Example output may looks something like the following:
```
a-palchikov: may look~~s~~

docs/5.x/cluster.md
```diff
@@ -221,6 +221,51 @@ $ curl -sk https://localhost:7575 | python -m json.tool
...
+```
+
+Here's an example of how to view the history remotely via `tsh`:
+
+```bsh
+$ tsh --cluster=production ssh admin@node gravity status history
+```
+
+The `gravity status history` command is an additional tool to help debug issues
+with a Cluster. The `gravity status` command only displays the current status
```
a-palchikov: nit: cluster (non-capitalized)

As a side note, this reads as following: we have `gravity status` which is not useful enough hence we have `gravity status history`. Shouldn't `gravity status` instead have been made useful? I know this is probably too late of a discussion, but including interesting historical developments in `gravity status` so it is useful would've been more helpful (like automatically including the last events leading to `NodeDegraded` if the node is still unhealthy, etc.).
This could be something to look into later.

docs/5.x/faq.md
```diff
@@ -114,3 +114,8 @@ If the installed kernel exposes the option fs.may_detach_mounts we recommend alw
 sysctl -w fs.may_detach_mounts=1
 echo "fs.may_detach_mounts = 1" >> /etc/sysctl.d/10-may_detach_mounts.conf
 ```
+
+## Cluster Status
+
+Gravity provides the `gravity status` command to view [Cluster Status](/cluster/#cluster-status).
+This tool can help identify issues with the Cluster.
```
a-palchikov: nit: cluster (non-capitalized)

bernardjkim: It looks like we are using "Cluster" (capitalized) when referring to a gravity cluster in the docs. Is there a difference when using "Cluster" vs "cluster"?  If not, then its probably better to leave it capitalized for consistency.

a-palchikov: Um - ok, it does look like this - sorry, non issue.

docs/5.x/cluster.md
bernardjkim: Good point. We can leave out the section pointing out the limitations of `gravity status` and just focus on what the new `history` sub-command provides.

a-palchikov (commented):
Just noticed that you updated the docs/5.x directory which seems to be the wrong place - is this specifically for 5.x?

bernardjkim:
@a-palchikov 
Status history feature will be available on `5.5.x`, `6.1.x`, `6.3.x`. I updated the `6.x` docs in #1066 but not `5.x` docs. This PR is just to copy over docs from `6.x` to `5.x`.

a-palchikov (approved):

PR #1125, opened 2020-02-19
Bump docker library version.

UnderGreen:
Bump coordinate library version. Eliminate all "Sirupsen"
related code.


Gopkg.toml
```diff
@@ -150,7 +153,7 @@ ignored = [
 
 [[override]]
   name = "github.com/docker/docker"
-  version = "v1.4.1-4831-g0f5c9d3"
+  revision = "be7ac8be2ae072032a4005e8f232be3fc57e4127"
```
UnderGreen: I took the same revision as in Kubernetes 1.15.x branch.

lib/system/fs.go
```diff
@@ -54,7 +54,7 @@ func GetFilesystem(ctx context.Context, path string, runner utils.CommandRunner)
 // GetFilesystemForPath returns the filesystem type for given path.
 // It does not verify whether the path actually exists
 func GetFilesystemForPath(path string) (fstype string, err error) {
-	mounts, err := mount.GetMounts()
+	mounts, err := mount.GetMounts(mount.ParentsFilter(path))
```
a-palchikov: Does this render the code below (or parts of it) redundant?

a-palchikov (approved):

lib/system/fs.go
UnderGreen: ParentsFilter returns all mountpoints which can be parents for given path. It decreases amount of returned mounts to a few, but we still need logic above to traverse path till we find real mountpoint.

r0mant (approved):
Ooh finally.

PR #1130, opened 2020-02-20
Make sure key store directory is not created unless necessary.

r0mant:
There was an annoying side effect of initializing a local key store (where now-obsolete gravity config file is) which would attempt to create ~/.gravity directory even for read operations, such as when determining whether a user is logged into a cluster or not. This not just created unnecessary dirs/files, but would also lead to the commands failing if user, say, doesn't have permissions to create that directory (e.g. bandwagon when creating a user - that's how I discovered this).

Now the keystore doesn't attempt to create this directory unless necessary. The LocalLogins backend implementation takes care of this on its own.

### DIFF
--- a/lib/utils/fileutils.go
+++ b/lib/utils/fileutils.go
@@ -246,7 +246,7 @@ func EnsureLocalPath(customPath, defaultLocalDir, defaultLocalPath string) (stri
 	return path, nil
 }
 
-// GetLocalPath constructs path to the local gravity config file like decsribed
+// GetLocalPath constructs path to the local gravity config file like described
 // in the EnsureLocalPath above.
 func GetLocalPath(customPath, defaultLocalDir, defaultLocalPath string) (string, error) {
 	if customPath != "" {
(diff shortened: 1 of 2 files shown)


lib/utils/fileutils.go
```diff
@@ -246,7 +243,20 @@ func EnsureLocalPath(customPath string, defaultLocalDir, defaultLocalPath string
 			return "", trace.Wrap(err)
 		}
 	}
-	return customPath, nil
+	return path, nil
+}
+
+// GetLocalPath constructs path to the local gravity config file like decsribed
```
a-palchikov: described

lib/utils/fileutils.go
```diff
@@ -246,7 +243,20 @@ func EnsureLocalPath(customPath string, defaultLocalDir, defaultLocalPath string
...
+// in the EnsureLocalPath above.
+func GetLocalPath(customPath, defaultLocalDir, defaultLocalPath string) (string, error) {
+	if customPath != "" {
+		return customPath, nil
+	}
+	homeDir := os.Getenv(constants.EnvHome)
+	if homeDir == "" {
+		return "", trace.BadParameter("no path provided and environment variable %v is not not set", constants.EnvHome)
```
a-palchikov: not ~~not~~

a-palchikov (approved):

PR #1151, opened 2020-02-22
Migration to Go modules

UnderGreen:


UnderGreen:
Depends on https://github.com/gravitational/gravity.e/pull/4276

a-palchikov:
Something's up with the build/

a-palchikov (approved):

a-palchikov:
Just occurred to me that local builds will fail if the repository is within `GOPATH` - this needs to add the explicit `GO111MODULE=on`.

wadells (changes requested):
Why are we migrating to go modules?  I don't currently see a linked issue, and the commit message for 7824a255517ac482252ecb2bb58fad7c2c8ede17 doesn't discuss the why at all.

If we're doing a 1-for-1 swap of dep -> go modules, I'm not convinced that this PR should have any changes in the vendor directory.  Seeing thousands of changes there suggests we're changing dependency versions at the same time as changing dependency managers which dramatically increases the risk surface area. To counteract this, I'd recommend doing this migration as a two step process, with one commit/PR that converts from dep -> modules with no changes in vendor (if possible).  If changes to dependiencies are necessary for migrating dependncy managers, these necessary changes should be explicitly called out in the commit message and PR. I'd then recommend updating modules and dependencies as needed for keeping current or other work in a second later commit/PR .  In the first dependency manager swap commit, proof that the listed dependencies &  versions in Gopkg.lock and go.sum are identical woudn't go awry.

How have you tested this change?

PR #1152, opened 2020-02-24
Fix tests and forward-port service management sub-commands

a-palchikov:
* Fix chart tests broken with the https://github.com/gravitational/gravity/pull/1072 commit.
 * Forward port low-level system management commands for starting/stopping
gravity-specific services as well as viewing system journal.

Requires https://github.com/gravitational/trace/pull/55.

### DIFF
--- a/lib/app/handler/handler.go
+++ b/lib/app/handler/handler.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"net"
 	"net/http"
 	"strconv"
 	"strings"
@@ -530,15 +529,6 @@ func (h *WebHandler) exportApp(w http.ResponseWriter, req *http.Request,
 	if err := json.NewDecoder(req.Body).Decode(&config); err != nil {
 		return trace.Wrap(err)
 	}
-	_, _, err = net.SplitHostPort(config.RegistryHostPort)
-	if config.RegistryHostPort == "" || err != nil {
-		message := "invalid host:port value"
-		if err != nil {
-			message = err.Error()
-		}
-		return trace.BadParameter("registryHostPort: %v", message)
-	}
-
 	if err = context.applications.ExportApp(app.ExportAppRequest{
 		Package:         *locator,
 		RegistryAddress: config.RegistryHostPort,
--- a/lib/rpc/server/agent.go
+++ b/lib/rpc/server/agent.go
@@ -55,7 +55,7 @@ func (srv *agentServer) Command(req *pb.CommandArgs, stream pb.Agent_CommandServ
 
 // PeerJoin accepts a new peer
 func (srv *agentServer) PeerJoin(ctx context.Context, req *pb.PeerJoinRequest) (*types.Empty, error) {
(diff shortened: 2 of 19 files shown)


lib/app/handler/handler.go
```diff
@@ -530,15 +529,6 @@ func (h *WebHandler) exportApp(w http.ResponseWriter, req *http.Request,
 	if err := json.NewDecoder(req.Body).Decode(&config); err != nil {
 		return trace.Wrap(err)
 	}
-	_, _, err = net.SplitHostPort(config.RegistryHostPort)
```
a-palchikov: I don't believe this should do the validation.
Also, the changes to registry code seem to be allowing for optional HTTP scheme so this was wrong in enforcing the `host:port` form.

lib/rpc/server/agent.go
```diff
@@ -91,14 +91,14 @@ func (srv *agentServer) GetRuntimeConfig(ctx context.Context, _ *types.Empty) (*
...
+	tempDir := os.TempDir()
 	config := &pb.RuntimeConfig{
 		Role:          srv.Role,
 		AdvertiseAddr: srv.Config.Listener.Addr().String(),
-		DockerDevice:  srv.DockerDevice,
 		SystemDevice:  srv.SystemDevice,
 		Mounts:        srv.Mounts,
 		StateDir:      stateDir,
-		TempDir:       srv.TempDir,
+		TempDir:       tempDir,
```
r0mant: Do you need a variable for it? Why not just `TempDir: os.TempDir()`?

r0mant (approved):

lib/rpc/server/agent_test.go
```diff
@@ -455,3 +458,7 @@ func (r *peerStore) remove(peer Peer) {
 	defer r.Unlock()
 	delete(r.peers, peer.Addr())
 }
+
+func isPeerDeniedError(err error) bool {
+	return strings.Contains(err.Error(), "peer not authorized")
```
knisbet: Nit: "peer not authorized" should probably be a constant, or the approach where there is a pre-created error that is returned and used for comparison. var ErrPeerNotAuthorized...

lib/systemservice/systemd.go
```diff
@@ -332,19 +345,19 @@ func (s *systemdManager) UninstallService(req UninstallServiceRequest) error {
 	serviceName := serviceName(req.Name)
 	out, err := invokeSystemctl("stop", serviceName)
 	if err != nil && !IsUnknownServiceError(err) {
-		return trace.Wrap(err, out)
+		return trace.Wrap(err)
```
knisbet: I'm not sure it's better to drop the out string from systemctl? Specifically on error does this not provided more context?

knisbet (approved):

lib/systemservice/systemd.go
a-palchikov: Oh, I did not want that - my bad.

lib/rpc/server/agent_test.go
a-palchikov: I removed the comparison with a string literal.

PR #1156, opened 2020-02-25
Add client/server version information to gravity status.

### LINKED ISSUE #1149: Add gravity version to gravity status output
When running `gravity status`, it should display the version of:

* Version of `gravity` binary that's been invoked.
* Version of `gravity` binary on server side (gravity-site).

r0mant:
A small improvement to add client/server version information to gravity status output - I oftentimes find myself missing it from customer reports and otherwise, and also helpful to know which version gravity-site is running as well. Adds the following line to gravity status output:

```
ubuntu@node-1:~/installer$ sudo gravity status
Cluster name:		heuristicliskov1018
Cluster status:		active
Cluster image:		telekube, version 7.0.0-alpha.2.13
Gravity version:	7.0.0-alpha.2.13 (client) / 7.0.0-alpha.2.13 (server)
...
```

Needs to be backported everywhere. Closes https://github.com/gravitational/gravity/issues/1149.

### DIFF
--- a/lib/ops/opshandler/opshandler.go
+++ b/lib/ops/opshandler/opshandler.go
@@ -124,6 +124,9 @@ func NewWebHandler(cfg WebHandlerConfig) (*WebHandler, error) {
 	// Report portal status
 	h.GET("/portal/v1/status", h.getStatus)
 
+	// Return server version
+	h.GET("/portal/v1/version", h.needsAuth(h.getVersion))
+
 	// Applications API
 	h.GET("/portal/v1/apps", h.needsAuth(h.getApps))
 	h.GET("/portal/v1/gravity", h.needsAuth(h.getGravityBinary))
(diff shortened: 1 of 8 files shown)


lib/ops/opshandler/opshandler.go
```diff
@@ -124,6 +124,9 @@ func NewWebHandler(cfg WebHandlerConfig) (*WebHandler, error) {
 	// Report portal status
 	h.GET("/portal/v1/status", h.getStatus)
 
+	// Return server version
+	h.GET("/portal/v1/version", h.needsAuth(h.getVersion))
```
a-palchikov: Why do you think this needs auth?

a-palchikov (approved):

lib/ops/opshandler/opshandler.go
r0mant: Good question - I remember version information being freely available was once flagged during a security review, so I opted for protecting this endpoint with auth as well.

lib/status/status.go
```diff
@@ -208,6 +214,8 @@ type Cluster struct {
 	Endpoints Endpoints `json:"endpoints"`
 	// Extension is a cluster status extension
 	Extension `json:",inline,omitempty"`
+	// Version is the server version information.
+	Version *modules.Version `json:"version"`
```
a-palchikov: One more thing that occurred to me is that this will omit the client binary version in the status JSON blob captured as part of the debug report - I'd add it as well (`ClientVersion`?)

lib/ops/opshandler/opshandler.go
knisbet: correct, version information is generally considered an information disclosure, since it can be used to target attacks against known vulnerable versions and avoid versions that might be able to detect exploit attempts.

PR #1161, opened 2020-02-26
re-arrange coredns config generation (#1110)

knisbet:
* re-arrange coredns config generation to be consistent between install and upgrade

* separate testable / untestable corefile generation

* protect against nil resolv conf

### DIFF
--- a/lib/install/phases/coredns.go
+++ b/lib/install/phases/coredns.go
@@ -87,27 +87,9 @@ func (r *corednsExecutor) Execute(ctx context.Context) error {
 	r.Progress.NextStep("Configuring CoreDNS")
 	r.Info("Configuring CoreDNS.")
 
-	// Read the resolv.conf from the host doing installation
-	// it will be used for configuring coredns upstream servers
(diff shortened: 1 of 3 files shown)


lib/install/phases/coredns.go
```diff
@@ -163,6 +145,34 @@ func (r *corednsExecutor) Rollback(context.Context) error {
...
+	systemdResolvConf, err := systeminfo.ResolvFromFile("/run/systemd/resolve/resolv.conf")
+	if err != nil && !trace.IsNotFound(err) {
+		return "", trace.Wrap(err)
+	}
+
+	config.UpstreamNameservers = mergeUpstreamResolvers(resolvConf, systemdResolvConf)
+	if resolvConf != nil && resolvConf.Rotate {
+		config.Rotate = true
+	}
+	if systemdResolvConf != nil && systemdResolvConf.Rotate {
```
r0mant: Nit: I'd do `else if` here to avoid an extra check in case the above branch already set it to true.

r0mant (approved):

PR #1168, opened 2020-02-27
Forward port improvements from 6.1.x

a-palchikov:
This PR contains changes I wanted to keep separate from the SELinux PR I'm about to submit.

 * Forward port install/expand improvements from 6.1.x.
 * Remove devicemapper-related code (including handling of docker device).

Requires https://github.com/gravitational/gravity.e/pull/4279.


lib/defaults/defaults.go
```diff
@@ -263,6 +263,18 @@ const (
 	// SatelliteRPCAgentPort is port used by satellite agent to expose its status
 	SatelliteRPCAgentPort = 7575
 
+	// SatelliteRPCAgentPort is port used by satellite agent to expose metrics
```
a-palchikov: I re-arranged the defaults to have constants and variables (and functions) separately - mostly moving things that did not belong in `var` block to `const`.

lib/defaults/defaults.go
```diff
@@ -530,8 +539,6 @@ const (
 
 	// APIPrefix defines the URL prefix for kubernetes-related queries tunneled from a master node
 	APIPrefix = "/k8s"
-	// APIServerPort defines the port of the kubernetes API server
-	APIServerPort = 8080
```
a-palchikov: Unused?

lib/install/client/client.go
```diff
@@ -306,10 +305,6 @@ type Lifecycle interface {
 	Abort(context.Context, *Client) error
 }
 
-func removeSocketFileCommand(socketPath string) (cmd string) {
-	return fmt.Sprintf("-/usr/bin/rm -f %v", socketPath)
```
a-palchikov: this is done automatically when we create the service socket now.

lib/install/engine/cli/cli.go
```diff
@@ -85,9 +85,6 @@ func (r *Engine) Execute(ctx context.Context, installer install.Interface, confi
 }
 
 func (r *Engine) execute(ctx context.Context, installer install.Interface, config install.Config) (err error) {
-	if err := r.validate(ctx, config); err != nil {
```
a-palchikov: not resume friendly - moved elsewhere.

lib/install/fsmspec.go
```diff
@@ -65,9 +65,53 @@ func FSMSpec(config FSMConfig) fsm.FSMSpecFunc {
 
 		case strings.HasPrefix(p.Phase.ID, phases.MastersPhase), strings.HasPrefix(p.Phase.ID, phases.NodesPhase):
 			return phases.NewSystem(p,
-				config.Operator, remote)
+				config.Operator, config.LocalPackages, remote)
 
-		case p.Phase.ID == phases.WaitPhase:
+		case strings.HasPrefix(p.Phase.ID, phases.ExportPhase):
```
a-palchikov: Re-arranged to keep non-trivial and trivial (`case phase.ID`) cases separately.

lib/install/operation.go
```diff
@@ -232,19 +227,6 @@ func (i *Installer) getClusterStatus() (*status.Status, error) {
 	return status, nil
 }
 
-// upsertAdminAgent creates an admin agent for the cluster being installed
-func (i *Installer) upsertAdminAgent(clusterName string) error {
```
a-palchikov: Moved to lib/ops/opsservice to the point of creating the cluster where it's more context-specific and resume-friendly.

lib/log/log.go
```diff
@@ -0,0 +1,142 @@
...
+package log
+
+import (
+	"io"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Logger defines a subset of the structured logging interface
+type Logger interface {
```
a-palchikov: This interface is the basis for re-implementation of loggers (i.e. fsm logger that'd like to look like `logrus.FieldLogger` but cannot due to `FieldLogger` not being totally generic).

lib/utils/cli/cli_test.go
```diff
@@ -61,6 +61,13 @@ func (*S) TestUpdatesCommandLine(c *check.C) {
 				"install", "--token", `"some token"`, `"/path/to/data"`, "--advertise-addr", `"localhost:8080"`,
 			},
 		},
+		{
+			comment:   "Handles negated flags",
+			inputArgs: []string{"install", "--no-selinux", "/path/to/data"},
```
a-palchikov: not really selinux-related but since this is just a test it is not an issue.

lib/install/client/install.go
```diff
@@ -157,6 +142,22 @@ type InstallerStrategy struct {
...
 
+// isServiceFailed returns an error if the service has failed.
+func isServiceFailed(serviceName string) func() error {
+	return func() error {
+		failed, err := service.IsFailed(serviceName)
+		if err == nil && failed {
+			return trace.Errorf("service %q has failed. Check journal log for details.",
+				serviceName)
+		}
+		return nil
```
r0mant: return trace.Wrap(err)?

lib/install/proto/proto.go
```diff
@@ -69,7 +69,12 @@ func KeyToProto(key ops.SiteOperationKey) *OperationKey {
 
 // HasSpecificPhase determines if this request is for a specific phase (other than root)
 func (r *ExecuteRequest) HasSpecificPhase() bool {
-	return r.Phase != nil && !r.Phase.IsResume()
+	return !r.HasResume()
+}
+
+// HasResume determines if this is a request to resume an operation
+func (r *ExecuteRequest) HasResume() bool {
+	return r.Phase == nil || r.Phase.IsResume()
```
r0mant: That's a little weird that "has resume" returns true if the phase is empty. Shouldn't this be `r.Phase != nil && r.Phase.IsResume()`?

lib/install/utils.go
```diff
@@ -379,9 +408,12 @@ func tryInstallBinary(targetPath string, uid, gid int, logger log.FieldLogger) e
 	if err != nil {
 		return trace.Wrap(err, "failed to determine path to binary")
 	}
-	err = os.MkdirAll(filepath.Dir(targetPath), defaults.SharedDirMask)
-	if err != nil {
-		return trace.Wrap(err)
+	dir := filepath.Dir(targetPath)
+	if dir != string(filepath.Separator) {
```
r0mant: What's this logic for?

lib/log/log.go
```diff
@@ -0,0 +1,142 @@
+/*
+Copyright 2019 Gravitational, Inc.
```
r0mant: Is this a mostly copy-paste of logrus interface? If so, we should probably use their header.

lib/ops/opsservice/configure.go
```diff
@@ -230,7 +229,7 @@ func (s *site) configureExpandPackages(ctx context.Context, opCtx *operationCont
 			masterParams.sniHost = trustedCluster.GetSNIHost()
 		}
 		err = s.configurePlanetMasterSecrets(opCtx, masterParams)
-		if err != nil {
+		if err != nil && !trace.IsAlreadyExists(err) {
```
r0mant: Hmm so here and in all other places where you added this check (which is a bunch) - how can this happen? Don't the generated config and secret package names include timestamp?

I also wonder if silently swallowing this error may be misleading - if you request to generate the packages, shouldn't we rather upsert them than ignore if they already exist?

lib/ops/opsservice/install.go
```diff
@@ -897,7 +890,7 @@ func (s *site) newProvisioningToken(operation ops.SiteOperation) (token string,
 		tokenRequest.Expires = s.clock().UtcNow().Add(defaults.InstallTokenTTL)
 	}
 	_, err = s.users().CreateProvisioningToken(tokenRequest)
-	if err != nil {
+	if err != nil && !trace.IsAlreadyExists(err) {
```
r0mant: Why ignore this here too? Shouldn't the caller be making this decision?

r0mant (commented):
Looks good overall but I feel a little uneasy about all these new `&& !trace.IsAlreadyExists(err)` when configuring packages / tokens / etc., for a couple of reasons:

* They're completely invisible so nobody will have any idea that it happened.
* They're hardcoded on server which doesn't give the caller opportunity to handle them (i.e. ignore / do upsert, etc.) which may lead to unexpected behaviors.

lib/install/proto/proto.go
a-palchikov: Resuming is starting the operation from scratch and skipping the phases already marked as completed. When starting from scratch - the phase is empty.

lib/ops/opsservice/configure.go
a-palchikov: Yes, they contain the timestamp which is pre-generated - to avoid generating the new package each time the function is re-run. If this generated a new package each time it was invoked, we could never properly generate a plan for an operation.

Do you have a suggestion on how to solve this issue in context of resumption - i.e. when a phase fails for some reason and then needs to re-execute the same steps incl. the generation of packages?

lib/log/log.go
a-palchikov: Technically, it's not a copy - it's been adopted to mirror the interface but does not duplicate the original - not sure this needs the license copy-paste.

lib/ops/opsservice/install.go
a-palchikov: Yes, moved to caller but in general all these `IsAlreadyExists` checks are mitigating the non-transactional nature of the backend operation.

lib/install/proto/proto.go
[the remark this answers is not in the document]
a-palchikov: Actually, you're both correct - I double-checked and the resume operation always uses the root phase ID so this will be changed to match the expectation.

r0mant (approved):
Discussed offline and settled that it should be safe to ignore already exists errors for re-entrancy, will add logs though.

PR #1170, opened 2020-02-28
lib/systemservice: remove unused systemdUnit.servicePath()

alrs:
This removes an unused function from `lib/systemservice`.

### DIFF
--- a/lib/systemservice/systemd.go
+++ b/lib/systemservice/systemd.go
@@ -123,10 +123,6 @@ func (u *systemdUnit) serviceName() string {
 		systemdServiceDelimiter) + systemdUnitFileSuffix
 }
 
-func (u *systemdUnit) servicePath() string {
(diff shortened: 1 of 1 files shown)


wadells (approved):
LGTM,  The last use of this function was removed in 2016, before Gravity was OSS.  The commit in which it was removed is in our private prehistory:

https://github.com/gravitational/gravity.e/commit/5a10606e18a367c77677140d631c90445891e285

And the code that replaced it:

https://github.com/gravitational/gravity/blob/3041f5adea706cf0601cd5268076b8539aa6d7b9/lib/systemservice/systemd.go#L161

Unfortunately, Our Gravity Jenkins job is currently broken with PRs from forks.  Once we fix that (or work around it) this is good to take.  No promises as to when, but that is on our roadmap as I prefer the fork workflow.

Thanks for your contribution  @alrs!

Issue #1065, opened 2020-02-03
Failure to deploy agents does not prevent upgrade from starting


r0mant:
**Describe the bug**

If upgrade agent fails to deploy on a node, the upgrade operation still starts and then fails when trying to execute a phase on that node.

**To Reproduce**

**Expected behavior**

- [x] Upgrade operation does not start if not all agents deployed successfully.
- [x] Agent deployment is more resilient to network failures (add retries).

**Logs**

```
2020-01-21T16:17:44Z WARN [RPC:DEPLO] logname:  cmd:rm -rf /var/lib/gravity/site/update/agent/secrets node:10.18.2.87:3022 stream:stderr utils/logging.go:94
2020-01-21T16:17:44Z WARN [RPC:DEPLO] "no login name\n" cmd:rm -rf /var/lib/gravity/site/update/agent/secrets node:10.18.2.87:3022 stream:stderr utils/logging.go:94
2020-01-21T16:17:44Z WARN [RPC:DEPLO] "logname: no login name\n" cmd:rm -rf /var/lib/gravity/site/update/agent/secrets node:10.18.2.91:3022 stream:stderr utils/logging.go:94
2020-01-21T16:17:44Z WARN [RPC:DEPLO] Subcommand failed, sequence interrupted. command:{rm -rf /var/lib/gravity/site/update/agent/secrets true map[PATH:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/writable/bin] false} error:[
ERROR REPORT:
Original Error: *errors.errorString EOF
Stack Trace:
    /gopath/src/github.com/gravitational/gravity/lib/utils/ssh.go:220 github.com/gravitational/gravity/lib/utils.SSHRunAndParse
    /gopath/src/github.com/gravitational/gravity/lib/utils/ssh.go:167 github.com/gravitational/gravity/lib/utils.(*sshCommands).run
    /gopath/src/github.com/gravitational/gravity/lib/utils/ssh.go:135 github.com/gravitational/gravity/lib/utils.(*sshCommands).Run
    /gopath/src/github.com/gravitational/gravity/lib/rpc/deploy.go:221 github.com/gravitational/gravity/lib/rpc.deployAgentOnNode
    /gopath/src/github.com/gravitational/gravity/lib/rpc/deploy.go:135 github.com/gravitational/gravity/lib/rpc.DeployAgents.func1
[...]

PR #1174, opened 2020-02-29
[5.5] Upgrade agents deployment improvements

r0mant:
This PR addresses a number of issues with upgrade agents deployment we discovered during a recent support call and adds a few improvements/niceties as well. So, in no particular order:

* Make sure the "main" upgrade agent that runs automatic operation does not commence the operation until all other agents are available.
* Make deployment of a single agent more resilient by retrying on networking errors. I've tested this by playing with iptables ability to randomly drop packets.
* Make sure that if the upgrade operation fails to initialize (e.g. to deploy one of agents), all agents are shut down. Previously they were kept lingering even though the operation was canceled.
* Various CLI and logging polish: make output more consistent and also a progress message is now printed when an agent is successfully deployed.
* Upgrade trace library to take advantage of improved output of wrapped errors I did a couple of months ago.

Needs forward-ports everywhere. Closes https://github.com/gravitational/gravity/issues/1065.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -962,12 +962,12 @@
   revision = "cdbeea01bfdb0053a9defb1a1f0bfe5c89bc4c2f"
 
 [[projects]]
-  digest = "1:418247f700939c99867b42639d985686ba229fd9f8ee9e119c7cc3ffde6e0fe3"
+  digest = "1:3afdc8c5bad39bda75ede9cd18bf590c385ff9a28c742f7330e3d3e19c685c47"
   name = "github.com/gravitational/trace"
   packages = ["."]
   pruneopts = "UT"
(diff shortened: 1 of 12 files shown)


a-palchikov (approved):

a-palchikov:
I believe the changes in trace broke a test:
```
FAIL: fs_test.go:37: S.TestFetchesFilesystem

fs_test.go:74:
    c.Assert(err, ErrorMatches, testCase.err, comment)
... error string = "" +
...     "failed to determine filesystem type on /dev/foo: error\n" +
...     "\terror"
... regex string = "error, failed to determine filesystem type on /dev/foo: error"
... lsblk fails

OOPS: 0 passed, 1 FAILED
--- FAIL: TestSystem (0.01s)
FAIL
FAIL	github.com/gravitational/gravity/lib/system	0.144s
```

PR #1176, opened 2020-03-02
[5.5.x] fix version script to carefully consider commits since last tag

a-palchikov:
Update the version script to avoid adding 0 commits since last tag to fix forming versions like '5.5.37-0'.
The last 5.5.x release is building to an invalid bucket because of this.

Needs forward ports.

### DIFF
--- a/version.sh
+++ b/version.sh
@@ -1,15 +1,18 @@
 #!/bin/bash
+set -Euo pipefail
 
-# this versioning algo:
-# keeps tag as is in case if this version is an equal match
-# otherwise adds .<number of commits since last tag>
+# versioning algorithm:
+# keep tag as is in case if this version is an equal match or number of commits are 0
+# otherwise add .<number of commits since last tag> or -<number of commits since last tag> (depending on the prerelease info)
 
 SHORT_TAG=`git describe --abbrev=0 --tags`
 LONG_TAG=`git describe --tags`
 COMMIT_WITH_LAST_TAG=`git show-ref --tags --dereference | grep "refs/tags/${SHORT_TAG}^{}" | awk '{print $1}'`
-COMMITS_SINCE_LAST_TAG=`git rev-list  ${COMMIT_WITH_LAST_TAG}..HEAD --count`
+COMMITS_SINCE_LAST_TAG=`git rev-list ${COMMIT_WITH_LAST_TAG}..HEAD --count`
 
 if [[ "$LONG_TAG" == "$SHORT_TAG" ]] ; then
+    echo "$SHORT_TAG"
+elif [[ "$COMMITS_SINCE_LAST_TAG" == "0" ]] ; then
     echo "$SHORT_TAG"
 elif [[ "$SHORT_TAG" != *-* ]] ; then
     echo "$SHORT_TAG-${COMMITS_SINCE_LAST_TAG}"


UnderGreen (approved):

wadells:
I was just running into this.

I suggest:  As soon as we're 1 commit past the latest tag, the patch version bumps, and we move to alpha.

e.g. instead of 6 commits after `5.5.37` being `5.5.37-6`, it becomes `5.5.38-6`

This is because hyphens are used to designate pre-release versions:

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST NOT include leading zeroes. Pre-release versions have a lower precedence than the associated normal version. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92.

https://semver.org/#spec-item-9

In particular: "Pre-release versions have a lower precedence than the associated normal version." means that

`5.5.37` < `5.5.38-6` < `5.5.38`

whereas we currently do something like:

`5.5.37` > `5.5.37-6` < `5.5.38`

Which means Gravity identifies that 5.5.37-6 to 5.5.37 is not a valid upgrade scenario, and fails.  I believe this is the underlying cause of failures like this:

https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&minLogLevel=0&timestamp=2020-03-02T19:30:55.974000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%22dee502ee-6ffd-4214-b322-ad6624ff2ec7%22%0Alabels.__suite__%3D%224e0bc64a-7bf2-4658-91d2-ff3f669999d7%22%0Aseverity%3E%3DINFO&dateRangeEnd=2020-03-02T19:30:51.565Z&interval=PT1H&dateRangeUnbound=backwardInTime

wadells:
Alternative candidate implementation here: #1179

FWIW: The reason we weren't seeing version.sh work correctly on 5.5.x is because the most recent 5.5.x tags are not annotated tags, just lightweight tags.  This means the following line:

https://github.com/gravitational/gravity/blob/20c2bc373ae33f0a401d5eed78ba9d9f8e1256be/version.sh#L9

won't properly resolve these tags to commits, because lightweight tags don't have a corresponding dereference (which is what `grep ...^{}` is looking for). 291b703c is an easy fix from #1179

We need to standardized on annotated tags (in part because all signed tags are a subset of annotated tags, and we probably need signed tags for compliance reasons).  This will mean giving a couple releases annotated tags:

```
walt@work:~/git/gravity$ git show-ref --tags --dereference | cut -d' ' -f 2- | cut -d'^' -f1 | uniq -c | grep -P '^\s+1'
      1 refs/tags/5.5.29
      1 refs/tags/5.5.30
      1 refs/tags/5.5.36
      1 refs/tags/5.5.37
```

New tags will point to the same commit of course, but it will be a different form of tag. I know we've maybe had some issues in the past with rewriting tags?  I definitely want to proceed with caution.

wadells:
I believe this is superseded by  #1179.  Are you comfortable closing it @a-palchikov or is there value in continuing this line of development?

PR #1178, opened 2020-03-03
Migration to go modules

UnderGreen:


a-palchikov (approved):
lgtm but did you verify the dependencies were mapped 1:1?

UnderGreen:
@a-palchikov Almost everything is mapped 1:1 except of kingpin. I mentioned about it in slack, not sure should I update it here. Probably will be better to fix the problem with kingpin in master and rebase after that.
And a lot of deleted files. It looks like `go` processing them different than `dep`. And few files were chmoded from 755 to 644, maybe it is just my default git behaviour.

build.assets/Dockerfile
```diff
@@ -1,24 +1,16 @@
-FROM quay.io/gravitational/debian-venti:go1.12.9-stretch
+FROM quay.io/gravitational/debian-venti:go1.13.8-stretch
```
wadells: Why is 13.8 needed in the buildbox for this upgrade?

build.assets/Dockerfile
```diff
@@ -40,11 +32,16 @@ RUN (mkdir -p /opt/protoc && \
...
 
-ENV PROTO_INCLUDE "/usr/local/include":"/src":"${GRPC_GATEWAY_ROOT}/third_party/googleapis":"${GOGOPROTO_ROOT}/gogoproto"
+ARG USER_ID=1000
+RUN adduser jenkins --uid=$USER_ID --disabled-password --system
+RUN (mkdir -p /src/github.com/gravitational/gravity && \
+     chown -R jenkins /gopath && \
+     chown -R jenkins /src && \ 
+     mkdir -p /.cache && \
+     chmod 777 /.cache && \
+     chmod 777 /tmp)
```
wadells: Won't putting these RUN directives after the clones result in a bunch of superfulous layer churn and ever so slightly slower builds?  I'd expect because these are relatively static, they should be before the dynamic `git clone` layers, where they were originally.

wadells (changes requested):
With the exception of the kingpin pseudo-version, the code looks reasonable. That is my only blocking issue, the rest are just questions to help me build my understanding.

I've seen [several](https://github.com/gravitational/gravity/pull/1095) [recent](https://github.com/gravitational/gravity/pull/1200) [PRs](https://github.com/gravitational/gravity/pull/1151) from you without any context provided.  As a reminder: https://gravitational.slab.com/posts/practices-12mko276#good-pull-request

Please make sure your commit messages address the **_why_** of each change. It makes it easier for peers to understand what is going on and give good feedback.  Context of why a change was made is also invaluable to maintainers after you're gone.

And at least when I review the PR, it'll save a bunch of "why?" questions inline.

Perhaps all the above is a bit strong, given this is a draft PR between two of your branches.  I have a suspicion it is more about trying out builds on jenkins than requesting peer feedback.  I'm working on a  jenkins "gravity try build" job exactly for this use case (as desperately need that functionality too).  If you're interested, let me know and I'll accelerate the GA of that work.

wadells:
> And a lot of deleted files. It looks like go processing them different than dep. And few files were chmoded from 755 to 644, maybe it is just my default git behaviour.

Likely.  I agree with you that mode bit churn or some file removals isn't a big concern.  Likely slightly different tooling between dep and native modules.

go.mod
```diff
@@ -54,18 +54,14 @@ require (
...
-	github.com/gravitational/bolt v1.3.2-gravitational // indirect
 	github.com/gravitational/certificate-transparency-go v0.0.0-20180803094710-99d8352410cb // indirect
-	github.com/gravitational/cfssl v0.0.0-20180619163912-4b8305b36ad0 // indirect
 	github.com/gravitational/configure v0.0.0-20180808141939-c3428bd84c23
 	github.com/gravitational/coordinate v0.0.0-20180225144834-2bc9a83f6fe2
 	github.com/gravitational/form v0.0.0-20151109031454-c4048f792f70
-	github.com/gravitational/go-oidc v0.0.1 // indirect
 	github.com/gravitational/go-vhost v0.0.0-20171024163855-94d0c42e3263
-	github.com/gravitational/kingpin v2.1.10+incompatible // indirect
+	github.com/gravitational/kingpin v0.0.0-00010101000000-000000000000 // indirect
```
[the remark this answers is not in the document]
UnderGreen: This is what go automatically adds in go.mod. You can't delete it, with any new go action(build, test, mod) this line will be added again. But in next commit, I will get rid of it.

build.assets/Dockerfile
```diff
@@ -40,11 +32,16 @@ RUN (mkdir -p /opt/protoc && \
      git clone https://github.com/gogo/protobuf --branch ${GOGO_PROTO_TAG} /src/github.com/gogo/protobuf && cd /src/github.com/gogo/protobuf && make install && \
      git clone https://github.com/grpc-ecosystem/grpc-gateway --branch ${GRPC_GATEWAY_TAG} /src/github.com/grpc-ecosystem/grpc-gateway && cd /src/github.com/grpc-ecosystem/grpc-gateway && pwd && go install ./protoc-gen-grpc-gateway)
 
-ENV PROTO_INCLUDE "/usr/local/include":"/src":"${GRPC_GATEWAY_ROOT}/third_party/googleapis":"${GOGOPROTO_ROOT}/gogoproto"
+ARG USER_ID=1000
```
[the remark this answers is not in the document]
UnderGreen: That's a good point. I believe 995 was chosen randomly. Because it is buildbox container you can build it with any user who has access to default GOMOD/GOPATH directories. 
Yes, for future improvements like mounting GOMOD directory from the host to cache builds it will be better to use the user from the host.
@a-palchikov if you have any thoughts why 996 UID was chosen I can change it back.

build.assets/Dockerfile
UnderGreen: Because of go modules, resolving dependencies in 1.13 works much better and with 1.12 as far as I remember I had huge problems with it.

build.assets/Dockerfile
UnderGreen: Hmm, I didn't think about that. I moved all that code because of `chown -R jenkins /gopath`, otherwise after building grpc few lines above everything inside gopath is owned by root. I will implement the fix for that situation differently. Thanks for pointing on that.

UnderGreen:
@wadells, @a-palchikov I've addressed your review comments and force-pushed new commits with detailed information about changes in each commit.

wadells (approved):
LGTM.

Issue #1189, opened 2020-03-03
Gravity binary should be included in downloads


r0mant:
Currently, our downloads include only `tele` and `tsh` binaries. They should also include `gravity` binary.

Things to consider:

- [x] Open-source downloads.
- [ ] Enterprise downloads.

wadells:
I was wondering about this. :laughing:

PR #1193, opened 2020-03-03
Add gravity to release tarball.

r0mant:
Add `gravity` binary to the release tarball users download from our Downloads page.

I also added it to the enterprise installer script - I realize this won't actually have effect until we somehow manage to upgrade our get.gravitational.io, but I figured I'd add it there anyway.

About https://github.com/gravitational/gravity/issues/1189.

### DIFF
--- a/build.assets/README.md
+++ b/build.assets/README.md
@@ -40,6 +40,8 @@ What are these binaries?
   Gravity clusters and installed applications.
 * `tsh` is a tool to remotely connect to clusters created from the images.
   tsh supports both SSH and Kubernetes API.
+* `terraform-provider-gravity` is a Terraform plugin used to support management
+  of open-source Gravity clusters.
 
 See the quick start to learn how to use these tools:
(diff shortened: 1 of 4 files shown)


wadells (approved):
LGTM.

wadells:
Should `terraform-provider-gravity` also be included too?  It seems to be treated in a similar manner elsewhere in our Makefiles:

https://github.com/gravitational/gravity/blob/229a0f96774a4f052bfd3062c79bd24a9568c249/Makefile#L203

I'm not familiar enough with the use of these different binaries to be sure, just pattern matching after spending lots of time looking at the Makefiles.

a-palchikov (approved):

knisbet:
I'm not sure I understand why this is changing?

r0mant:
We need to add gravity binary to our release tarballs because our new quickstart cli flows are going to use it for certain things (like, installing apps into existing clusters) and otherwise users have no idea where to get it from.

PR #1179, opened 2020-03-03
[5.5.x] Adjust patch versioning strategy.

wadells:
After discovering that upgrading from `5.5.37` to the latest build of `5.5.x` was not possible and discussing with @r0mant, I believe that our patch version should increment as soon as we're one commit past a tagged release.  As discussed at https://semver.org/#spec-item-9, the ordering would now be:

5.5.37 < 5.5.38-dev.10 < 5.5.38

As opposed to our previous

5.5.37-10 < 5.5.37 < 5.5.38

Where 5.5.37-10 is intended to denote 10 commits after 5.5.37.

If you're hesitant about the `dev` pre-release tag or the `+sha` build metadata (now removed from this PR), I specifically factored these out into ac0c996  and 4bf85f0, such that we can pick and choose.  The commit messages have context on why I chose to add each of these.

**Notes:** 

The first commit in this stack fixes a latent bug we had where we were incorrectly handling lightweight tags as opposed to annotated tags.  This bug was the root cause of #1176.

[...]

PR #1195, opened 2020-03-04
[5.5] extend upgrade matrix

wadells:
**Attention:** This work is dependent on #1179.  The `5.5.37` -> `5.5.x-latest` upgrade scenario is expected to fail without that work.  Please go review that first, if it isn't yet merged.

This PR adds several new upgrade scenarios to our 5.5.x branch, increasing our coverage for key gravity versions deployed in the field.  This gets us most of the value of https://github.com/gravitational/robotest/issues/155 without requiring any substantial development.

Testing done:

I successfully ran a version of this patch rebased on top of #1179 here:

https://jenkins.gravitational.io/view/Robotest/job/gravity-try-build/21/

Note: This work increases PR build time to almost 2 hours!  This sucks.  I'd like to keep PR builds under an hour, if possible.  We may look at bumping concurrency further:

https://github.com/gravitational/gravity/blob/20c2bc373ae33f0a401d5eed78ba9d9f8e1256be/Jenkinsfile#L33-L35

I haven't found the priority to look into that yet though.  I suspect 4 simultaneous tests may be tuned to the capacity of our GCE account.  Do you know where 4 comes from @a-palchikov (or anyone else)?

### DIFF
--- a/build.assets/robotest_run_nightly.sh
+++ b/build.assets/robotest_run_nightly.sh
@@ -5,10 +5,21 @@ readonly UPGRADE_FROM_DIR=${1:-$(pwd)/../upgrade_from}
 
 DOCKER_STORAGE_DRIVERS="overlay2"
 
+# UPGRADE_MAP maps gravity version -> list of OS releases to upgrade from
 declare -A UPGRADE_MAP
-# gravity version -> list of OS releases to test upgrades on
+# latest patch release on compatible LTS, keep these up to date
+UPGRADE_MAP[5.5.37]="centos:7 ubuntu:16"
 UPGRADE_MAP[5.2.16]="centos:7 ubuntu:16"
+
+# via intermediate upgrade
 UPGRADE_MAP[5.0.35]="debian:9 ubuntu:16"
+
+# important versions in the field
+UPGRADE_MAP[5.2.12]="ubuntu:16"
+UPGRADE_MAP[5.5.19]="ubuntu:16"
+UPGRADE_MAP[5.5.20]="ubuntu:16"
+UPGRADE_MAP[5.5.28]="ubuntu:16"
+UPGRADE_MAP[5.5.36]="ubuntu:16"
 
 readonly GET_GRAVITATIONAL_IO_APIKEY=${GET_GRAVITATIONAL_IO_APIKEY:?API key for distribution Ops Center required}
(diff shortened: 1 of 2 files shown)


r0mant (approved):

wadells:
Force pushed to rebase these changes on top of #1179 in 5.5.x.  We should be good to go with this PR in the morning (if @a-palchikov doesn't request changes).

build.assets/robotest_run_nightly.sh
```diff
@@ -5,11 +5,22 @@ readonly UPGRADE_FROM_DIR=${1:-$(pwd)/../upgrade_from}
 
 DOCKER_STORAGE_DRIVERS="overlay2"
 
+# UPGRADE_MAP maps gravity version -> list of OS releases to upgrade from
 declare -A UPGRADE_MAP
-# gravity version -> list of OS releases to test upgrades on
-UPGRADE_MAP[5.2.15]="centos:7 ubuntu:16"
+# latest patch release on compatible LTS, keep these up to date
+UPGRADE_MAP[5.5.37]="centos:7 ubuntu:16"
```
a-palchikov: Why thing that was popped up a while ago is keeping these versions in sync. I think if we moved them to the makefile closer to where the versions are defined / queried, it would be much easier to handle.
Something along the lines of:
```
VERSIONS_LTS_TO_LATEST = 5.5.37 5.2.15
VERSIONS_TO_LATEST = 5.2.12 5.5.19 5.5.20 5.5.28 5.5.36
```
Also, we need to set up nightly if not yet done for `5.5.x`.

a-palchikov (approved):

build.assets/robotest_run_nightly.sh
wadells: > Also, we need to set up nightly if not yet done for 5.5.x.

Excellent point.  This does not yet exist.  I'll set it up.

> One thing that popped up a while ago is keeping these versions in sync. I think if we moved them to the makefile closer to where the versions are defined / queried, it would be much easier to handle.
> Something along the lines of:
> 
> ```
> VERSIONS_LTS_TO_LATEST = 5.5.37 5.2.15
> VERSIONS_TO_LATEST = 5.2.12 5.5.19 5.5.20 5.5.28 5.5.36
> ```

I've filed #1199 to track this work.

PR #1197, opened 2020-03-04
Default to tele version during tele pull.

### LINKED ISSUE #1181: tele pull downloads alpha versions by default
If you run

```
$ tele pull gravity
```

the latest alpha version of the image will be pulled.

Instead, tele should default to its own version just like `tele build` does i.e. if `tele` is of version 6.1.1, then `tele pull gravity` should pull the base image version 6.1.1.

r0mant:
Make sure that when a user does `tele pull gravity`, it defaults to the version of tele itself, same behavior as with tele build. Otherwise it defaults to the latest version and pulls an alpha (it will still pull an alpha if tele itself is alpha, but if tele is, say, 7.0.0, it will pull that).

Closes https://github.com/gravitational/gravity/issues/1181.

### DIFF
--- a/lib/hub/hub.go
+++ b/lib/hub/hub.go
@@ -199,7 +199,7 @@ func (h *s3Hub) Download(f *os.File, locator loc.Locator) (err error) {
 	if err != nil {
 		err := utils.ConvertS3Error(err)
 		if trace.IsNotFound(err) {
-			return trace.NotFound("application %v:%v not found in %v, use 'tele ls' to see available applications",
+			return trace.NotFound("image %v:%v not found in %v, use 'tele ls' to see available images",
 				locator.Name, locator.Version, h.Bucket)
 		}
 		return trace.Wrap(err)
--- a/lib/hub/hub.go
+++ b/lib/hub/hub.go
@@ -199,7 +199,7 @@ func (h *s3Hub) Download(f *os.File, locator loc.Locator) (err error) {
 	if err != nil {
 		err := utils.ConvertS3Error(err)
 		if trace.IsNotFound(err) {
-			return trace.NotFound("image %v:%v not found in %v, use 'tele ls' to see available images",
+			return trace.NotFound("image %v:%v not found in %v, use 'tele ls -a' to see available images",
 				locator.Name, locator.Version, h.Bucket)
 		}
 		return trace.Wrap(err)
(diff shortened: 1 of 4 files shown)


lib/loc/make.go
```diff
@@ -30,13 +30,23 @@ import (
 //  - if it's in the 'name:ver' format, returns locator with system repo (systemrepo/name:ver)
 //  - if it's in the 'name' format, returns locator with system repo and latest meta-version (systemrepo/name:0.0.0+latest)
 func MakeLocator(app string) (*Locator, error) {
+	return MakeLocatorWithDefault(app, "")
+}
+
+// MakeLocatorWithDefault is like MakeLocator but uses the provided default
```
a-palchikov: nit: I think MakeXXX is non-consistent with the rest of the constructors - would be more consistent to use `NewLocatorWihDefault` (i.e. prefix with `New` for new APIs)

lib/loc/make.go
```diff
@@ -48,6 +58,5 @@ func MakeLocator(app string) (*Locator, error) {
 		}
 		return NewLocator(defaults.SystemAccountOrg, parts[0], version)
 	}
-	return nil, trace.BadParameter(
-		"invalid app name format: %v, should be: 'repo/name:ver' or 'name:ver' or 'name'", app)
+	return nil, trace.BadParameter("invalid app name format: %v, should be: 'repo/name:ver' or 'name:ver' or 'name'", app)
```
a-palchikov: it should be `package` instead of `app` and I would place the placeholder in quotes to emphasize the possibly empty input.
Also, why use truncated words in an error message? `repository/name:version`

a-palchikov (approved):

lib/hub/hub.go
```diff
@@ -199,7 +199,7 @@ func (h *s3Hub) Download(f *os.File, locator loc.Locator) (err error) {
 	if err != nil {
 		err := utils.ConvertS3Error(err)
 		if trace.IsNotFound(err) {
-			return trace.NotFound("application %v:%v not found in %v, use 'tele ls' to see available applications",
+			return trace.NotFound("image %v:%v not found in %v, use 'tele ls' to see available images",
```
knisbet: I think technically `tele ls -a` these days to see available images, unless tele ls has been fixed to list the latest on each release.

knisbet (approved):

PR #1200, opened 2020-03-04
Forward-port improvements in terraform-dev from version/5.5.x

UnderGreen:
### Purpose

- Fix the issue with resolving DNS requests inside VMs. With terraform-libvirt plugin [v0.5.2](https://github.com/dmacvicar/terraform-provider-libvirt/tree/v0.5.2) DNS enabled option was added and the default value is false.
- Move all cloud-init code from terraform file into dedicated to improve readability.
- Use one root disk to simplify setup. Usage of the dedicated disk for `/tmp` directory(which was used to unpack installer tarball) is deprecated.
- Set minimum requirements for terraform to `0.12.x`.
- Change default OS for VM to Ubuntu 18.04.

### DIFF
--- a/terraform-dev/cloudinit.cfg
+++ b/terraform-dev/cloudinit.cfg
@@ -1,0 +1,44 @@
+#cloud-config
+packages: [python, curl, htop, iotop, lsof, ltrace, mc, net-tools, strace, tcpdump, telnet, vim, wget, ntp, traceroute, bash-completion]
+ssh_authorized_keys: ["${file("ssh/key.pub")}"]
+manage_resolv_conf: true
+resolv_conf:
+  nameservers: ['8.8.4.4', '8.8.8.8', '1.1.1.1']
+  options:
+    rotate: true
+bootcmd:
+- echo ${ip_address} ${hostname} >> /etc/hosts
+hostname: "${hostname}"
+write_files:
+- content: "br_netfilter"
+  path: /etc/modules-load.d/br_netfilter.conf
+- content: "ebtables"
+  path: /etc/modules-load.d/ebtables.conf
+- content: "overlay"
+  path: /etc/modules-load.d/overlay.conf
+- content: |
+    ip_tables
+    iptable_nat
+    iptable_filter
+  path: /etc/modules-load.d/iptables.conf
+- content: |
+    net.bridge.bridge-nf-call-arptables=1
+    net.bridge.bridge-nf-call-ip6tables=1
+    net.bridge.bridge-nf-call-iptables=1
+  path: /etc/sysctl.d/10-br-netfilter.conf
+- content: |
+    net.ipv4.ip_forward=1
+  path: /etc/sysctl.d/10-ipv4-forwarding-on.conf
+- content: |
+    fs.may_detach_mounts=1
+  path: /etc/sysctl.d/10-fs-may-detach-mounts.conf
+runcmd:
+- 'modprobe overlay'
+- 'modprobe br_netfilter'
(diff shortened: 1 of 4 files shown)


UnderGreen:
In the master branch, terraform-dev is unusable because there is a problem with DNS. I fixed that in 5.5.x a while ago, just need to forward-port.

a-palchikov (approved):

wadells:
I'd like to review this PR, as I'm getting to know a lot more about terraform. However, without knowing where the changes came from (e.g. which PR or commit they were ported from) or why they were made, I don't know that I can give constructive feedback.

`git cherry-pick -x` will leave a paper trail of where a commit came from. This becomes increasingly important in an org with lots of ports.  Unfortunately, git tracks diffs/ports terribly compared to other VCS where diffs are first class citizens.

UnderGreen:
Improvements are splitted in a few PR requests. Let me provide a better description here instead of pointing to previsous PR without any description.

terraform-dev/cloudinit.cfg
```diff
@@ -0,0 +1,44 @@
...
+manage_resolv_conf: true
+resolv_conf:
+  nameservers: ['8.8.4.4', '8.8.8.8', '1.1.1.1']
+  options:
+    rotate: true
+bootcmd:
+- echo ${ip_address} ${hostname} >> /etc/hosts
+hostname: "${hostname}"
+write_files:
+- content: "br_netfilter"
```
knisbet: Most active releases should actually set and load modules, and if they don't it's an issue with the release. I would consider dropping the cloud-config that sets this, so that problems with the modules get surfaced. The downside is when trying to use an old/broken release it will keep popping up.

knisbet (approved):

terraform-dev/cloudinit.cfg
knisbet: Actually, I'm not sure if br_netfilter works correctly on 5.5, but I think the other modules should/do.

terraform-dev/install.yaml
```diff
@@ -65,23 +65,23 @@
...
-        --token=token
+        --token=longtoken
       {% else %}
-      /tmp/gravity join \
+      {{ ansible_env.HOME }}/gravity join \
         {{hostvars[groups['all'][0]]['ansible_default_ipv4']['address']}} \
         --advertise-addr={{hostvars[inventory_hostname]['ansible_default_ipv4']['address']}} \
         --token=longtoken
       {% endif %}
   - name: Clean tmp
```
wadells: You might update this to `Clean installer` or similar.

terraform-dev/install.yaml
```diff
@@ -65,23 +65,23 @@
...
     file:
       state: absent
-      path: "/tmp/installer"
+      path: "{{ ansible_env.HOME }}/installer"
   - name: Clean tmp gravity
```
wadells: Likewise: "Clean gravity" as this is no longer in `tmp`

wadells (approved):
LGTM.

terraform-dev/cloudinit.cfg
UnderGreen: I personally use terraform-dev a lot to test upgrades from 5.0.x and I need those modules to be enabled.
I believe we should test modules autoload in gravity PRs by disabling the same code in robotest terraform scripts. @wadells

wadells: I filed https://github.com/gravitational/robotest/issues/184 to track testing our module autoload.

PR #1213, opened 2020-03-06
[5.5.x] add a sub-command to rotate RPC credentials on a cluster

a-palchikov:
This PR addresses the issue with RPC credentials rotation by introducing a new `system` sub-command:
```sh
usage: gravity system rotate-rpc-creds [<flags>]

Renew cluster RPC credentials

Flags:
      --log-file="/var/log/gravity-install.log"  
```
It is capable of verifying as well as generating new credentials package to avoid user-unfriendly and  fragile workarounds.
It is supposed to work even in case of controller Pod not running.

### DIFF
--- a/lib/localenv/clusterenv.go
+++ b/lib/localenv/clusterenv.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/gravitational/gravity/lib/app"
 	"github.com/gravitational/gravity/lib/app/service"
+	"github.com/gravitational/gravity/lib/blob"
 	libcluster "github.com/gravitational/gravity/lib/blob/cluster"
 	"github.com/gravitational/gravity/lib/blob/fs"
 	"github.com/gravitational/gravity/lib/httplib"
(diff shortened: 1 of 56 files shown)


lib/localenv/clusterenv.go
```diff
@@ -107,12 +112,43 @@ func newClusterEnvironment(config clusterEnvironmentConfig) (*ClusterEnvironment
...
+	}
+
+	server, err := ops.FindLocalServer(cluster.ClusterState)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	objects, err := libcluster.New(libcluster.Config{
+		Local:       localObjects,
+		WriteFactor: 1,
```
r0mant: Is this going to work with multi-node clusters? The blob will be available on 1 node and when other gravity-sites are coming up they won't be able to find it. Will they be able to replicate it?

r0mant (commented):
I wonder if this command is going to work with multi-node clusters. I remember running into issues before when trying to use our etcd-backed blob store on a single node, other nodes were getting some sort of blob-related errors.

lib/localenv/clusterenv.go
a-palchikov: As far as understand, this is how the active cluster controller works but without the replication loop. As I see this environment's blob service working - it needs to write the package file and update the backend accordingly - as if this was a single node cluster.
Then, the cluster controller should pick it up and distribute it.
I'll have to test this though.

tool/gravity/cli/rotate.go
```diff
@@ -38,16 +39,47 @@ import (
...
+		if archive == nil {
+			trace.NotFound("no RPC credentials found. Run the command again without --dry-run to generate.")
+		}
+		if err := rpc.ValidateCredentials(archive, now); err != nil {
+			return trace.Wrap(err)
+		}
+		env.Println("Nothing to do.")
+		return nil
+	}
+	if !o.force {
```
knisbet: I don't understand why we want/need force logic here? If a user requests rotation of the rpc credentials, why not just rotate them regardless of whether they're valid or not?

lib/rpc/tls.go
```diff
@@ -295,3 +307,26 @@ func upsertPackage(packages pack.PackageService, pkg loc.Locator, archive utils.
...
 	return trace.Wrap(err)
 }
+
+func validateCertificateExpiration(pemBytes []byte, now time.Time) error {
+	const tolerance = 30 * time.Second
+	cert, err := tlsca.ParseCertificatePEM(pemBytes)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if now.Add(-tolerance).Before(cert.NotBefore) {
```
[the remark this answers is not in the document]
a-palchikov: Not sure I completely follow your suggestion - the NotBefore of the certificate is backdated by an hour at generation time and you mean that this code should simply validate against `time.Now()`?

a-palchikov:
So it seems I have to reset this PR to unblock jenkins. Will close this and recreate.

Issue #1064, opened 2020-02-03
Add pre-upgrade version checks to prevent unsupported upgrade paths


r0mant:
We've had a few scenarios where a customer would attempt an upgrade for an unsupported path (e.g. direct 5.0 to 5.5 w/o intermediate runtime) which successfully launched only to fail midway requiring our assistance in rollback.

Gravity should check versions of installed cluster and upgrade and refuse to launch upgrades for unsupported paths. This should be implemented/ported to all current versions.

wadells:
I'd love it if the logic for calculating supported upgrades was modular, such that other tooling (robotest) could use the same logic instead of re-implementing this particular wheel.

wadells:
Also, #940 should be considered when designing these checks, as the work here is likely to increase the number of upgrades that can't be tried at all.

wadells:
> This should be implemented/ported to all current versions.

The more I think about this feature, the less convinced I am that we need to port it to all active branches.  Should this only go into master?

Here's my thinking:  The customers on 5.5 and 6.1 who would hit this likely already have and now understand our supported upgrade policy (via trial by fire).  It is new customers who are more likely to stumble on this, and they'll almost certainly choose to spin up on the most recent release instead of one of the older LTS releases.

Furthermore do we backport features?  I was under the impression that LTS got security patch ports, and maybe a feature that is critical to a paying customer.  Neither this bug and nor the PR #1232 description persuade me that upgrade validation falls into either of those buckets.

wadells:
https://github.com/gravitational/gravity/issues/1735 is a concrete check that we should implement.

Issue #940, opened 2019-12-06
Ability to force upgrade regardless of version


adamlamar:
When gravity detects that a new cluster image is older than the existing cluster image version, it will refuse to upgrade. This seems like the right behavior in general.

In some circumstances, especially when testing or changing artifact names, it is useful to force an upgrade even though gravity may consider it a downgrade based solely on the semantic versioning scheme.

Request: create a `--force` or `--ignore-version-check` option to allow `./upgrade` to proceed anyway.

shiranp:
+1

PR #1232, opened 2020-03-12
(5.5) Check for unsupported runtime upgrade paths.

r0mant:
This PR adds a pre-upgrade check that validates that the upgrade path between the installed runtime and the new runtime exists. About https://github.com/gravitational/gravity/issues/1064.

Upgrade path is considered valid if one of these is true:
* Direct upgrade is supported.
* Upgrade via intermediate runtimes is supported and there are suitable intermediate runtimes.

Versions where direct/intermediate upgrades are supported from are kept in the code so we'll need to be updating them when releasing new major/minor versions which is fine since it doesn't happen too often.

Things that would be nice to have but a bit out of scope for this so postponing them for later:
* Having some gravity command that would print supported upgrade paths, incl. in machine-readable format (json/yaml) so 3rd party tools (like robotest) can parse it.
* Have a force flag to upgrade to potentially skip this check, however I'm not very sure about usefulness of this one.

Needs forward-ports to 6.1, 6.3 and master (minus the intermediate runtimes part since intermediate upgrades haven't been forward-ported yet).


lib/ops/opsservice/versions.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+	}
+	return nil
+}
+
+// checkRuntimeUpgradePath checks that upgrade path between two provided runtimes exists.
+//
+// An upgrade path between runtimes is considered valid if:
+//
+//  - Direct upgrade is supported b/w old and new versions, OR
```
a-palchikov: nit: is it just me or `between` is more readable than `b/w`?

lib/ops/opsservice/versions.go
```diff
@@ -0,0 +1,203 @@
...
+			"from": req.FromRuntime,
+			"to":   req.ToRuntime,
+		}).Info("Validated runtime upgrade path: direct.")
+		return nil
+	}
+	// There's no direct upgrade path b/w the versions, see if there's
+	// an upgrade path with intermediate hops.
+	intermediateVersions := req.supportsUpgradeVia(*versionFrom)
+	if len(intermediateVersions) == 0 {
+		return trace.BadParameter(unsupportedError, req.FromRuntime, req.ToRuntime)
```
a-palchikov: `unsupportedErrorTmpl` to hint that it has format placeholders?

lib/ops/opsservice/versions.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	// Verify required intermediate versions exist.
+	var runtimes []loc.Locator
```
a-palchikov:
```go
return validateIntermediateVersionsExist(req, intermediateVersions)
```
instead of the comment.

lib/ops/opsservice/versions.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+		versionLabel := e.RuntimeLabels[pack.PurposeRuntimeUpgrade]
+		if versionLabel == "" {
+			return false
+		}
+		version, err := semver.NewVersion(versionLabel)
+		if err != nil {
+			return false
+		}
+		if loc.GreaterPatch(*version, *intermediateVersion) {
```
a-palchikov:
```go
return loc.GreaterPatch(*version, *intermediateVersion)
```

lib/ops/opsservice/versions_test.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	for _, test := range tests {
+		err := checkRuntimeUpgradePath(checkRuntimeUpgradePathRequest{
+			FromRuntime:           test.fromRuntime,
+			ToRuntime:             currentRuntime,
+			DirectUpgradeVersions: directUpgradeVersions,
+			UpgradeViaVersions:    upgradeViaVersions,
+			Packages:              s.packages,
+		})
+		if test.unsupported {
```
a-palchikov: I'd use specific error value instead of the boolean flag when asserting this fact - it is better to be explicit.

lib/ops/opsservice/versions.go
```diff
@@ -0,0 +1,203 @@
...
+	UpgradeViaVersions = map[*semver.Version]Versions{
+		// Upgrades from 5.0 are possible via 5.2.15 and later 5.2 releases.
+		semver.New("5.0.0"): Versions{semver.New("5.2.15")},
+	}
+)
+
+// checkRuntimeUpgradePathRequest is a request to validate upgrade path b/w runtimes.
+type checkRuntimeUpgradePathRequest struct {
+	// FromRuntime is the currently installed runtime version.
+	FromRuntime loc.Locator
```
a-palchikov: nit: why are these Exported? I don't see where it's required

a-palchikov (approved):

a-palchikov (approved):

wadells (commented):
The code looks reasonable to me.  I have a couple high level remarks:

1) Is there a way to override this check and try the upgrade anyhow?  This was specifically requested in #940.  If we don't plan to support that, we should probably let those contributors know.  Or if we do plan to (but not now) a note to them wouldn't go awry.

2) This code and commit messages don't at all address why these upgrade scenarios are unsupported.  I'd ask that you leave a comment or some documentation/code addressing why.  My understanding of is two primary causes:

* The combination of which kubernetes we bundle, and supported [kubernetes skew](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-version-skew), best described at https://gravitational.com/gravity/docs/cluster/#direct-upgrades-from-older-lts-versions
* We don't test unsupported upgrades.
* Are there other reasons?

3) What is the UX of this to the end user?  Can you give an example of how the error messages bubble up, when e.g. attempting an unsupported upgrade via CLI?

r0mant:
Yeah, good call on the upgrade policy comment, I added it. That ticket you linked to is actually specifically requesting downgrades which is probably not something we wanna do (short-term at least) as it would require some significant redesign. As for the errors, in case of unsupported upgrade path, the users will see errors like shown here: https://github.com/gravitational/gravity/blob/39633651fe4367ebccd2fb11d10e3e9ebf49eac1/lib/ops/opsservice/versions.go#L243-L256.

PR #1253, opened 2020-03-19
Relax fio test requirements a bit.

r0mant:
I think the fio test might be a little bit too strict right now and there's also some potential unknowns related to executing fio binary on people's machines so relax the test requirements a bit so this check doesn't get in the way too much:

* Bump allowed fsync latency to 50ms.
* Only consider the check failed if we got a successful test result back and its results are unsatisfactory. If fio binary failed to execute for some reason, log it and keep going (there's fallback to dd check anyway).

### DIFF
--- a/lib/checks/disks.go
+++ b/lib/checks/disks.go
@@ -18,7 +18,9 @@ package checks
 
 import (
 	"context"
+	"fmt"
 	"path/filepath"
+	"strings"
 
 	"github.com/gravitational/gravity/lib/defaults"
 	"github.com/gravitational/gravity/lib/network/validation/proto"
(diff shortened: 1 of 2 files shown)


lib/checks/disks.go
```diff
@@ -81,16 +83,32 @@ func fioEtcdJob(filename string) *proto.CheckDisksRequest {
...
+}
+
+// Error returns all errors encountered during fio disk test.
+func (e *fioTestError) Error() string {
+	return strings.Join(e.messages, ", ")
+}
+
+// isFioTestError returns true if the provided error is the fio disk test error.
+func isFioTestError(err error) bool {
+	_, ok := err.(*fioTestError)
```
a-palchikov: `trace.Unwrap(err).(*fioTestError)`

a-palchikov (approved):

PR #1255, opened 2020-03-19
Fix flag propagation to service

a-palchikov:
Fix implementation of `cli.CommandArgs` to properly replace flags (i.e. when updating the value on existing command line).


lib/utils/cli/cli.go
```diff
@@ -32,65 +32,158 @@ type CommandArgs struct {
...
 		}
 	}
-	outputCommand := ctx.SelectedCommand.FullCommand()
-	return append([]string{outputCommand}, args...), nil
+	for _, flag := range r.FlagsToAdd {
+		if _, ok := flag.(arg); ok {
+			// Args have been filtered above
+			continue
+		}
+		if _, exists := seen[flag.Name()]; !exists {
```
r0mant: Could this have been done as a part of the above loop, without needing this second one? Something like:

```go
for _, flag := range r.FlagsToAdd {
  if arg, ok := flag.(arg); ok {
    // Collect positional arguments that aren't yet present in the resulting command-line.
    if _, exists := seen[arg.name]; !exists {
      args = append(args, arg)
    }
  } else {
    // Collect flags that aren't yet present in the resulting command-line.
    if _, exists := seen[flag.Name()]; !exists {
      elems = append(elems, flag)
    }
  }
}
```

lib/utils/cli/cli.go
```diff
@@ -32,65 +32,158 @@ type CommandArgs struct {
...
+	return r.name
 }
 
-// Flag represents a single command-line flag.
-type Flag struct {
-	// Name is the flag name.
-	Name string
-	// Value is the flag value.
-	Value string
+// Formats returns this flag formatted for command line
```
r0mant: Nit: "this _argument_".

lib/utils/cli/cli.go
```diff
@@ -32,65 +32,158 @@ type CommandArgs struct {
...
+	return []string{fmt.Sprint(strconv.Quote(r.value))}
 }
 
-// NewFlag creates a new flag.
+type arg struct {
+	name  string
+	value string
+}
+
+// Flag represents a command line option
```
r0mant: Nit: I'd just say "a command-line _flag_".

r0mant (approved):

PR #1263, opened 2020-03-20
Revert back to explicit bootstrapping of SELinux configuration

### LINKED ISSUE #1262: Robotest upgrade failure
Version: `7.0.0-rc.2.4`
OS: CentOS 7.x/rhel 7.x

Upgrading fails at `upload` step with the error message:
```
open /path/to/installer/directory/gravity.db: permission denied
```

a-palchikov:
Revert back to explicit bootstrapping of SELinux configuration per operation instead of early bootstrapping for install/join/upgrade.

The change does not matter much for installs/joins but in case of upgrades it allows to avoid unnecessary bootstrapping if the node does not have SELinux support turned on.
Additionally, it avoid the pitfall of having the labeled installer binary and SELinux support off - if that were the case, the binary running in confined context will fail to access resources it does not have access to (incl. access to the state directory which is labeled with a base type instead of the type expected by the installer SELinux domain).

A note on installer operation during upgrade: as previously implemented but discarded, the installer will attempt to determine the state of SELinux support by querying cluster state using the existing gravity binary which aims to avoid the risk of running into a permission denied at the early stage prior to configuring the SELinux configuration for the installer.

Requires https://github.com/gravitational/gravity.e/pull/4283.
Fixes https://github.com/gravitational/gravity/issues/1262.

### DIFF
--- a/tool/gravity/cli/config.go
+++ b/tool/gravity/cli/config.go
@@ -504,7 +504,9 @@ func (i *InstallConfig) RunLocalChecks() error {
 
 // BootstrapSELinux configures SELinux on a node prior to installation
 func (i *InstallConfig) BootstrapSELinux(ctx context.Context, printer utils.Printer) error {
-	if !i.SELinux || i.FromService || i.Mode == constants.InstallModeInteractive || i.Remote {
+	logger := log.WithField(trace.Component, "selinux")
+	if !i.shouldBootstrapSELinux() {
+		logger.Info("SELinux disabled with configuration.")
 		return nil
 	}
 	metadata, err := monitoring.GetOSRelease()
@@ -512,16 +514,20 @@ func (i *InstallConfig) BootstrapSELinux(ctx context.Context, printer utils.Prin
 		return trace.Wrap(err)
 	}
 	if !selinux.GetEnabled() {
-		i.WithField(trace.Component, "selinux").Info("SELinux not enabled on host.")
+		logger.Info("SELinux not enabled on host.")
 		i.SELinux = false
 	} else if !libselinux.IsSystemSupported(metadata.ID) {
-		i.WithField("id", metadata.ID).Info("Distribution not supported.")
+		logger.WithField("id", metadata.ID).Info("Distribution not supported.")
 		i.SELinux = false
 	}
 	return BootstrapSELinuxAndRespawn(ctx, libselinux.BootstrapConfig{
 		StateDir: i.StateDir,
 		OS:       metadata,
 	}, printer)
+}
+
+func (i *InstallConfig) shouldBootstrapSELinux() bool {
+	return i.SELinux && !(i.FromService || i.Mode == constants.InstallModeInteractive || i.Remote)
 }
 
 func (i *InstallConfig) validateApplicationDir() error {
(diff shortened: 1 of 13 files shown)


tool/gravity/cli/config.go
```diff
@@ -499,6 +502,28 @@ func (i *InstallConfig) RunLocalChecks() error {
 	}))
 }
 
+// BootstrapSELinux configures SELinux on a node prior to installation
+func (i *InstallConfig) BootstrapSELinux(ctx context.Context, printer utils.Printer) error {
+	if !i.SELinux || i.FromService || i.Mode == constants.InstallModeInteractive || i.Remote {
```
r0mant: Nit: `if i.shouldBootstrapSELinux() {`?

tool/gravity/cli/config.go
```diff
@@ -499,6 +502,28 @@ func (i *InstallConfig) RunLocalChecks() error {
...
+	}
+	metadata, err := monitoring.GetOSRelease()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if !selinux.GetEnabled() {
+		i.WithField(trace.Component, "selinux").Info("SELinux not enabled on host.")
+		i.SELinux = false
+	} else if !libselinux.IsSystemSupported(metadata.ID) {
+		i.WithField("id", metadata.ID).Info("Distribution not supported.")
```
r0mant: Should this also use "component: selinux" field? Otherwise the logs will just say "distribution not supported" without any indication this is only SELinux-specific.

tool/gravity/cli/config.go
```diff
@@ -829,6 +854,29 @@ func (j *JoinConfig) GetRuntimeConfig() proto.RuntimeConfig {
 	}
 }
 
+// bootstrapSELinux configures SELinux on a node prior to join
+func (j *JoinConfig) bootstrapSELinux(ctx context.Context, printer utils.Printer) error {
+	if !j.SELinux || j.FromService {
```
r0mant: Nit: `j.shouldBootstrapSELinux()`.

tool/gravity/cli/config.go
```diff
@@ -829,6 +854,29 @@ func (j *JoinConfig) GetRuntimeConfig() proto.RuntimeConfig {
...
+	metadata, err := monitoring.GetOSRelease()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	logger := log.WithField(trace.Component, "selinux")
+	if !selinux.GetEnabled() {
+		logger.Info("SELinux not enabled on host.")
+		j.SELinux = false
+	} else if !libselinux.IsSystemSupported(metadata.ID) {
+		logger.WithField("id", metadata.ID).Info("Distribution not supported.")
```
r0mant: Add "selinux" trace component.

tool/gravity/cli/config.go
```diff
@@ -874,6 +923,28 @@ func (r *autojoinConfig) checkAndSetDefaults() error {
 	return nil
 }
 
+// bootstrapSELinux configures SELinux on a node prior to join
+func (j *autojoinConfig) bootstrapSELinux(ctx context.Context, printer utils.Printer) error {
```
r0mant: Same two comments as above really.

r0mant (approved):

PR #1267, opened 2020-03-23
[6.1.x] workaround initial install DNS config for hooks

a-palchikov:
Use DNS configuration from command line for initial install operation environment so that the application service can use proper name resolution to execute and track application hooks.

Updates github.com/gravitational/gravity/issues/1265.

### DIFF
--- a/tool/gravity/cli/utils.go
+++ b/tool/gravity/cli/utils.go
@@ -76,7 +76,24 @@ func (g *Application) NewInstallEnv() (env *localenv.LocalEnvironment, err error
 	} else {
 		stateDir = filepath.Join(stateDir, defaults.LocalDir)
 	}
-	return g.getEnv(stateDir)
+	return g.getEnvWithArgs(localenv.LocalEnvironmentArgs{
+		StateDir:         stateDir,
+		Insecure:         *g.Insecure,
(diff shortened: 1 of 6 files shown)


tool/gravity/cli/utils.go
```diff
@@ -76,7 +76,24 @@ func (g *Application) NewInstallEnv() (env *localenv.LocalEnvironment, err error
...
+		// Use DNS configuration from installer command line.
+		// TODO(dmitri): setting this will only be useful for the install operation
+		// as in this case the DNS coniguration will first be set in local state during
+		// boostrapping step and the application service that is created based on this
+		// setting would have otherwise pointed to the legacy DNS configuration which is
+		// incorrect.
+		// This is rather a workaround - proper solution will be more involved and will have
+		// the application service using the kubernetes client (and hence the DNS config
+		// to resolve the names) only for hooks.
+		DNS:      localenv.DNSConfig(g.InstallCmd.DNSConfig()),
```
r0mant: Will/should this work when resuming install as well?

a-palchikov: yes - since the `resume` command is client-side proxy to the service that always runs with all configuration options available even after a restart.

r0mant (approved):

PR #1285, opened 2020-03-26
Wait for operator.

r0mant:
My previous attempt to fix the issue with waiting for process listeners to come up by embedding "wait" function into the process itself has some deficiencies as it turns out. In particular, process itself can't really reliably call it's own API in a general use-case b/c it doesn't know how clients are supposed to reach it: for example in installer use-case the advertise address is the IP address which works fine but when installing a hub, the "pack" address (which I used before) points to the public advertise hostname of the hub.

This PR instead moves the logic of checking the operator status to the caller - I integrated it into RemoteEnvironment instead which I think suits its purpose cause it knows the address of the installer it talks to, and the installer calls this method explicitly.

### DIFF
--- a/lib/localenv/remoteenv.go
+++ b/lib/localenv/remoteenv.go
@@ -17,11 +17,13 @@ limitations under the License.
 package localenv
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/gravitational/gravity/lib/app"
 	"github.com/gravitational/gravity/lib/app/client"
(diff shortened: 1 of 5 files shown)


lib/localenv/remoteenv.go
```diff
@@ -165,6 +167,22 @@ func (w *RemoteEnvironment) LoginWizard(addr, token string) (entry *storage.Logi
 	})
 }
 
+// WaitForOperator blocks until the configured operator becomes available or timeout expires.
```
a-palchikov: or context expires.

a-palchikov (approved):

PR #1292, opened 2020-03-26
[report] add audit log capture to help diagnose SELinux denials

a-palchikov:
This PR adds support for capturing SElinux denials from the host's audit log enriched with details specific to the installer.
The rules are only added for the duration (or part) of the operation to avoid introducing debugging state to the host. The rule captures only subset of syscalls responsible for creating/removing the filesystem resources as well as basic network operations.

This will hopefully aid in troubleshooting failed installs/joins on SELinux-enabled hosts.

Requires https://github.com/gravitational/planet/pull/589.


lib/report/system.go
```diff
@@ -158,3 +161,27 @@ func fetchEtc(name string) CollectorFunc {
...
+//
+// For example, to see the SELinux denials generated for the gravity domain
+// as a list of allow rules:
+//
+//  gunzip -c audit.log.gz | ausearch --subject gravity_t | audit2allow
+func auditLog() Collector {
+	var subjects []string
+	for _, domain := range auditlog.Domains {
+		subjects = append(subjects,
+			fmt.Sprint("/sbin/ausearch --success no --message all --raw --start yesterday --end today ", "--subject ", domain, " ;"),
```
r0mant: Just curious how this time filter works. Will it collect the entire "yesterday"? Otherwise, if "today" has just started, there may not be enough data to work with.

lib/system/auditlog/auditlog.go
```diff
@@ -0,0 +1,105 @@
+/*
+Copyright 2018 Gravitational, Inc.
```
r0mant: 2020

lib/system/auditlog/auditlog.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// New returns a new instance of the logger
+func New() *Auditlog {
+	return &Auditlog{
+		log: log.New(logrus.WithField(trace.Component, "auditlog")),
+	}
+}
+
+// AddDefaultRules adds default audit rules for the all known domains
```
r0mant: Nit: just "for all known domains" I think.

lib/system/auditlog/auditlog.go
```diff
@@ -0,0 +1,105 @@
...
+	cmd.Stdout = w
+	cmd.Stderr = w
+	logger.Info("Set up audit rule.")
+	if err := cmd.Run(); err != nil {
+		return trace.Wrap(err, "failed to set up audit rule for process")
+	}
+	return nil
+}
+
+// RemoveRules removes audit rules previously with this package
```
r0mant: ... previously _configured_ (or _set up_) with this package?

tool/gravity/cli/install.go
```diff
@@ -725,6 +726,9 @@ func join(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, confi
 
 // TerminationHandler implements the default interrupt handler for the installer service
 func TerminationHandler(interrupt *signals.InterruptHandler, printer utils.Printer) {
+	audit := auditlog.New()
+	audit.AddDefaultRules()   //nolint:errcheck
+	defer audit.RemoveRules() //nolint:errcheck
```
r0mant: This can potentially not execute if the installer terminates in a some kind of ungraceful way. So I'm curious, what is the downside (if any) of these rules not being removed if that happens?

r0mant (approved):

lib/report/system.go
```diff
@@ -158,3 +161,27 @@ func fetchEtc(name string) CollectorFunc {
...
+		)
+	}
+	script := fmt.Sprintf(`
+#!/bin/bash
```
knisbet: Is bash guarenteed to be installed on all the distros we support? I imagine it is, but might be safer to make this work with /bin/sh?

tool/gravity/cli/report.go
```diff
@@ -19,23 +19,31 @@ package cli
...
+		w = f
+	}
 	config := report.Config{
 		Filters:    filters,
 		Compressed: compressed,
 		Packages:   env.Packages,
 	}
-	err := report.Collect(context.TODO(), config, w)
-	return trace.Wrap(err)
+	return report.Collect(context.TODO(), config, w)
```
knisbet: is there a reason this is no longer being wrapped?

knisbet (approved):

tool/gravity/cli/install.go
a-palchikov: I haven't done exhaustive testing but batching and grouping fields and syscalls seems to be the recommended way to reduce kernel churn and limit the results.

Not sure what we can do else to counter the ungraceful termination other than explicitly considering or ignoring the rest of the signals that have termination as the default behavior and can be handled.

lib/report/system.go
a-palchikov: I'll replace `today` with `now` although both seem to mean the same thing. Unless this is not what you meant.

r0mant (approved):

r0mant (approved):

PR #1293, opened 2020-03-26
[5.5.x] Add nethealth resource and checker

bernardjkim:
- Bump monitoring-app to 5.5.12. This adds nethealth resources.
- Bump planet to 5.5.36. This adds nethealth-checker to planet agents.
- Bump satellite to 5.5.11.
- Bump trace to 1.1.10.

Requires https://github.com/gravitational/gravity.e/pull/4286.


a-palchikov:
Hmm, the failure looks like maybe `http.NewRequestWithContext` is not available with the version of Go we use. 
Seems to be my bad for suggesting it. Maybe it's better to change it to use:
```go
req, err := http.NewRequest(http.GET, url, &buf)
// handle error
req = req.WithContext(ctx)
```
as a fallback?

a-palchikov:
@bernardjkim I updated the tests since there was a minor change in semantics. I wonder if we needed to bump the minor version instead of the patch in trace though. Although the test use-case was also fragile.

r0mant (approved):

a-palchikov (approved):

PR #1294, opened 2020-03-26
SELinux documentation

a-palchikov:
Add high-level overview of SELinux support, changes to the manifest and CLI.

### DIFF
--- a/docs/7.x/installation.md
+++ b/docs/7.x/installation.md
@@ -121,6 +121,7 @@ Flag               | Description
 `--dns-zone`         | _(Optional)_ Specify an upstream server for the given DNS zone within the Cluster. Accepts `<zone>/<nameserver>` format where `<nameserver>` can be either `<ip>` or `<ip>:<port>`. Can be specified multiple times.
 `--vxlan-port`       | _(Optional)_ Specify custom overlay network port. Default is `8472`.
 `--remote` | _(Optional)_ Excludes this node from the Cluster, i.e. allows to bootstrap the Cluster from a developer's laptop, for example. In this case the Kubernetes master will be chosen randomly.
+`--selinux` | _(Optional)_ Turns on SELinux support. Defaults to `true` on supported OS distributions. Can be explicitly turned off with `--no-selinux`.
 
 The `gravity join` command accepts the following arguments:
 
@@ -134,6 +135,7 @@ Flag               | Description
 `--state-dir`      | _(Optional)_ Directory where all Gravity system data will be kept on this node. Defaults to `/var/lib/gravity`.
 `--service-uid`    | _(Optional)_ Service user ID (numeric). See [Service User](pack.md#service-user) for details. A user named `planet` is created automatically if unspecified.
 `--service-gid`    | _(Optional)_ Service group ID (numeric). See [Service User](pack.md#service-user) for details. A group named `planet` is created automatically if unspecified.
+`--selinux` | _(Optional)_ Turns on SELinux support. Defaults to `true` on supported OS distributions. Can be explicitly turned off with `--no-selinux`.
 
 ### Environment Variables
 
--- a/docs/7.x/installation.md
+++ b/docs/7.x/installation.md
@@ -135,7 +135,7 @@ Flag               | Description
 `--state-dir`      | _(Optional)_ Directory where all Gravity system data will be kept on this node. Defaults to `/var/lib/gravity`.
 `--service-uid`    | _(Optional)_ Service user ID (numeric). See [Service User](pack.md#service-user) for details. A user named `planet` is created automatically if unspecified.
 `--service-gid`    | _(Optional)_ Service group ID (numeric). See [Service User](pack.md#service-user) for details. A group named `planet` is created automatically if unspecified.
-`--selinux` | _(Optional)_ Turns on SELinux support. Defaults to `true` on supported OS distributions. Can be explicitly turned off with `--no-selinux`.
+`--selinux` | _(Optional)_ Turns on SELinux support. Defaults to `false`. See [SELinux](selinux.md) for more details.
 
 ### Environment Variables
 
--- a/docs/7.x/installation.md
+++ b/docs/7.x/installation.md
@@ -121,7 +121,7 @@ Flag               | Description
 `--dns-zone`         | _(Optional)_ Specify an upstream server for the given DNS zone within the Cluster. Accepts `<zone>/<nameserver>` format where `<nameserver>` can be either `<ip>` or `<ip>:<port>`. Can be specified multiple times.
 `--vxlan-port`       | _(Optional)_ Specify custom overlay network port. Default is `8472`.
 `--remote` | _(Optional)_ Excludes this node from the Cluster, i.e. allows to bootstrap the Cluster from a developer's laptop, for example. In this case the Kubernetes master will be chosen randomly.
-`--selinux` | _(Optional)_ Turns on SELinux support. Defaults to `true` on supported OS distributions. Can be explicitly turned off with `--no-selinux`.
+`--selinux` | _(Optional)_ Turns on SELinux support. Defaults to `false`. See [SELinux](selinux.md) for more details.
 
 The `gravity join` command accepts the following arguments:
 
--- a/docs/7.x/pack.md
+++ b/docs/7.x/pack.md
@@ -468,7 +468,7 @@ nodeProfiles:
           # If left empty or unspecified, the installer will label the directory
           # using the default file label.
           # Special label value of "none" explicitly turns off labeling for the directory
-          label: "SELinux label"
+          seLinuxLabel: "system_u:object_r:my_file_type_t:s0"
 
       # This setting makes sure specified devices from host are made available
       # inside Gravity container
--- a/docs/7.x/pack.md
+++ b/docs/7.x/pack.md
@@ -462,6 +462,13 @@ nodeProfiles:
           # Recursive defines a recursive mount, i.e. all submounts under specified path
           # are also mounted at the corresponding location in the targetPath subtree
           recursive: false
+          # SELinux label to use for the mount.
+          # If a label is given, the installer will label the directory
+          # using the specified label.
+          # If left empty or unspecified, the installer will label the directory
+          # using the default file label.
+          # Special label value of "none" explicitly turns off labeling for the directory
+          label: "SELinux label"
 
       # This setting makes sure specified devices from host are made available
       # inside Gravity container
--- a/docs/7.x/selinux.md
+++ b/docs/7.x/selinux.md
@@ -1,23 +1,13 @@
 # SELinux
 
 Starting with version 7, Gravity comes with SELinux support.
-It is not a requirement to have SELinux enabled, but whenever the installer detects that
-it runs on a SELinux-enabled host, it will automatically turn on SELinux support.
-
-When operating with SELinux support on, the following changes:
-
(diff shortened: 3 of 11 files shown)


docs/7.x/pack.md
```diff
@@ -462,6 +462,13 @@ nodeProfiles:
           # Recursive defines a recursive mount, i.e. all submounts under specified path
           # are also mounted at the corresponding location in the targetPath subtree
           recursive: false
+          # SELinux label to use for the mount.
+          # If a label is given, the installer will label the directory
+          # using the specified label.
+          # If left empty or unspecified, the installer will label the directory
+          # using the default file label.
+          # Special label value of "none" explicitly turns off labeling for the directory
+          label: "SELinux label"
```
r0mant: I didn't notice this when reviewing the code, but I think that "label" might be a too generic property of a node profile is this is SELinux specific. Maybe rename it to "seLinuxLabel" to make it explicit?

docs/7.x/selinux.md
```diff
@@ -0,0 +1,214 @@
+# SELinux
+
+Starting with version 7, Gravity comes with SELinux support.
```
r0mant: Not sure this needs to be mentioned since we have 7.0 specific docs now.

Also, weren't this going to be backported?

r0mant (approved):

docs/7.x/installation.md
```diff
@@ -121,6 +121,7 @@ Flag               | Description
 `--dns-zone`         | _(Optional)_ Specify an upstream server for the given DNS zone within the Cluster. Accepts `<zone>/<nameserver>` format where `<nameserver>` can be either `<ip>` or `<ip>:<port>`. Can be specified multiple times.
 `--vxlan-port`       | _(Optional)_ Specify custom overlay network port. Default is `8472`.
 `--remote` | _(Optional)_ Excludes this node from the Cluster, i.e. allows to bootstrap the Cluster from a developer's laptop, for example. In this case the Kubernetes master will be chosen randomly.
+`--selinux` | _(Optional)_ Turns on SELinux support. Defaults to `true` on supported OS distributions. Can be explicitly turned off with `--no-selinux`.
```
benarent: It would be good to list which OS distro support selinux

docs/7.x/selinux.md
```diff
@@ -0,0 +1,265 @@
...
+
+## Upgrades
+
+The upgrade will automatically determine whether SELinux support is on on the cluster nodes and will install and configure
+the new policy individually on each node.
+
+
+## Kubernetes
+
+Kubernetes comes with SELinux support in the form of `seLinuxOptions` inside the `securityContext`.
```
benarent: Is this a specific K8s version, or just the latest?

docs/7.x/installation.md
benarent: (agh, I See it's below )

benarent (changes requested):
This looks comprehensive. I've only a few notes. I've not tried running this, maybe @jon-can or @stevenGravy can try it as part of their test plan. 

One last things would be to add this to the Docs Nav.

https://github.com/gravitational/gravity/blob/master/docs/7.x.yaml#L37

docs/7.x/selinux.md
a-palchikov: The idea was to reference the version used in the latest Gravity version (`1.17.x` here) - I'll provide a visible reference to documentation.

r0mant (approved):

docs/7.x/selinux.md
```diff
@@ -0,0 +1,266 @@
...
+$ runcon --role sysadm_r --type sysadm_t ./gravity install --selinux ...
+```
+
+Gravity supports SELinux users `sysadm_u` and `unconfined_u` (and their corresponding roles) out of the box.
+If you need to install and manage Gravity clusters using a different user/role, a custom user policy is required.
+
+
+## Installation
+
+Installer will use SELinux if all of these conditions are met:
```
wadells: Perhaps:

> `gravity install` will use...

> "The installer will use..."

A cold sentence start with "Installer will" feels odd.

docs/7.x/selinux.md
```diff
@@ -0,0 +1,266 @@
...
+
+  * the host has SELinux enabled
+  * installer has SELinux support for the host OS distribution (see [Supported OS Distributions](#os-distribution-support) below)
```
wadells: Ditto. "the installer"

docs/7.x/selinux.md
```diff
@@ -0,0 +1,266 @@
...
+  * SELinux support has been turned on with `--selinux`
+
+Installer does the following as the first step when running on a host with the above conditions met:
```
wadells: "The installer"

docs/7.x/selinux.md
```diff
@@ -0,0 +1,266 @@
...
+    volumes:
+      - path: /var/data
+        targetPath: /var/data
+        seLinuxLabel: none
+```
+
+
+!!! warning "Performance"
+  Relabeling of directories with a large number of files/sub-directories can be time-consuming.
+  This will be improved in future releases.
```
wadells: I would avoid making promises about future work in the docs.  We can't predict what our priorities will be.

docs/7.x/selinux.md
```diff
@@ -0,0 +1,266 @@
...
+
+
+## Custom SELinux policies
+
+It is not yet possible to bundle a custom SELinux policy in the cluster image. If you have custom SELinux policy and want
+to use it for your Kubernetes workloads, you'll need to make sure to load the policy on each node where necessary prior to
+installing the cluster image.
+
+It is planned to add support for custom policies in a future version.
```
wadells: Ditto: I would avoid making promises about future work in the docs.

If you do want to leave folks a pointer, please link the appropriate github issue tracking this feature, to help set users expectations appropriately.

docs/7.x/selinux.md
```diff
@@ -0,0 +1,266 @@
...
+  * SELinux support has been turned on with `--selinux`
+
+Installer does the following as the first step when running on a host with the above conditions met:
+
+  * loads the Gravity SELinux policy module
+  * creates local port bindings for Gravity and Kubernetes-specific ports
+  * creates local file contexts for paths used during the operation
+  * creates local file contexts for custom state directory (if it has been overridden with `--state-dir`)
+
+Additional SELinux configuration might happen later as part of execution of the operation plan.
```
wadells: Which additional configuration?  This offers a tantalizing hint that something may happen, but provides no useful information about what or how someone reading the docs could learn more.

docs/7.x/selinux.md
```diff
@@ -0,0 +1,266 @@
...
+ Bootstrapping installer for SELinux
+ ...
+```
+
+SELinux support is managed per-node.
+
+## Upgrades
+
+The upgrade will automatically determine whether SELinux support is on on the cluster nodes and will install and configure
+the new policy individually on each node.
```
wadells: Do we have a test scenario for this yet?  I should probably get one added (at the least to my robotest backlog) if not.

lib/schema/manifest.go
```diff
@@ -869,10 +869,10 @@ type Volume struct {
 	Mode string `json:"mode,omitempty"`
 	// Recursive means that all mount points inside this mount should also be mounted
 	Recursive bool `json:"recursive,omitempty"`
-	// Label specifies the SELinux label.
+	// SELinuxLabel specifies the SELinux label.
 	// If left unspecified, the default.ContainerFileLabel will be used to label the directory.
-	// If a special value NoLabel is specified - no labeling is performed
-	Label string `json:"label,omitempty"`
+	// If a special value SELinuxLabelNone is specified - no labeling is performed
+	SELinuxLabel string `json:"seLinuxLabel,omitempty"`
```
wadells: Per your & @r0mant's advice to me in https://github.com/gravitational/robotest/pull/193#discussion_r398268442, should this be a pointer to a string to distinguish between the unset and "none" case?

I assume you chose to use a string flag for a reason, and I'd like to understand why/how this is different than the uid/gid settings case.

wadells (approved):
All I have is a couple wording suggestions and some questions for my own edification.

Otherwise, this looks great!

r0mant:
@a-palchikov @benarent @wadells Can we merge and deploy this now? Customers starting to notice the absence of these docs.

docs/7.x/installation.md
```diff
@@ -134,6 +135,7 @@ Flag               | Description
 `--state-dir`      | _(Optional)_ Directory where all Gravity system data will be kept on this node. Defaults to `/var/lib/gravity`.
 `--service-uid`    | _(Optional)_ Service user ID (numeric). See [Service User](pack.md#service-user) for details. A user named `planet` is created automatically if unspecified.
 `--service-gid`    | _(Optional)_ Service group ID (numeric). See [Service User](pack.md#service-user) for details. A group named `planet` is created automatically if unspecified.
+`--selinux` | _(Optional)_ Turns on SELinux support. Defaults to `true` on supported OS distributions. Can be explicitly turned off with `--no-selinux`.
```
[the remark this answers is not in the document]
a-palchikov: That's a lie - I forgot to change the default for joins it seems - thanks for pointing it out!

It's a bit complicated with defaults for SELinux and I decided to keep the setting per-node, so joins would need to be augmented with the flag to turn SELinux support on.

docs/7.x/selinux.md
```diff
@@ -0,0 +1,266 @@
...
+$ sudo --role sysadm_r --login
+```
+
+Alternatively, directly run the installer using the role `sysadm_r` and type `sysadm_t`:
+```sh
+$ runcon --role sysadm_r --type sysadm_t ./gravity install --selinux ...
+```
+
+Gravity supports SELinux users `sysadm_u` and `unconfined_u` (and their corresponding roles) out of the box.
+If you need to install and manage Gravity clusters using a different user/role, a custom user policy is required.
```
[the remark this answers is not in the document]
a-palchikov: Thing is, custom user policies are not yet supported (it's hinted at below in the Custom Policies) section. This will be implemented later.

docs/7.x/selinux.md
a-palchikov: Good point - will expand on the additional configuration.

docs/7.x/selinux.md
a-palchikov: There was one implicit test for installs before I flipped the support flag to default to false initially.
There's a way to introduce this to the robotests again though w/o disturbing the rest of the versions - via an environment variable. I'll create an issue for robotest to do this.

lib/schema/manifest.go
a-palchikov: In case of uid/gid which were represented as integers and the fact that `0` is a valid value for both it made it ambiguous to tell between unset and `0`.
In case of strings, the empty value (or unset) cannot be taken for a valid value. Also "none" being a special value vs empty is a design decision meaning no labeling vs default label.

a-palchikov:
I will update the docs as per @wadells suggestions.

wadells:
> @a-palchikov @benarent @wadells Can we merge and deploy this now or is there anything else to address? Customers starting to notice the absence of these docs.

LGTM!

docs/7.x/selinux.md
wadells: I filed https://github.com/gravitational/robotest/issues/201 to track this.  Please add any ideas about implementations details there.  I'd love to hear more detail on your remark:

> There's a way to introduce this to the robotests again though w/o disturbing the rest of the versions - via an environment variable

PR #1296, opened 2020-03-27
Tweak top-level Makefile for targeted builds

a-palchikov:
Move selinux dependency closer to $(BINARIES) target in top-level makefile to enable targeted builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -597,9 +597,7 @@ tele-mac: flags
 # goinstall builds and installs gravity locally
 #
 .PHONY: goinstall
-goinstall: remove-temp-files compile
-	mkdir -p $(GRAVITY_BUILDDIR)
-	mkdir -p $(TF_PROVIDER_DIR)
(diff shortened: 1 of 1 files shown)


r0mant (approved):

Makefile
```diff
@@ -597,9 +597,7 @@ tele-mac: flags
 # goinstall builds and installs gravity locally
 #
 .PHONY: goinstall
-goinstall: remove-temp-files compile
-	mkdir -p $(GRAVITY_BUILDDIR)
-	mkdir -p $(TF_PROVIDER_DIR)
+goinstall: remove-temp-files compile | $(TF_PROVIDER_DIR) $(GRAVITY_BUILDDIR)
```
wadells: This is slick!  I haven't seen order-only prerequistes used before.  But reading through https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html makes me appreciate how they're used with directories.  I definitely could use this in some of my Make work.

wadells (approved):

Issue #1077, opened 2020-02-05
[5.5.x] improve clean up logic for failed expand operations


a-palchikov:
Improve logic for when the expand operation fails or when the agent is explicitly terminated.
Also, provide a set of sub-commands to deal with orphaned operations in a safe manner.

helgi:
Seems like this will solve part of https://github.com/gravitational/gravity/issues/900 for me :) and perhaps part of https://github.com/gravitational/gravity/issues/897 as well

aelkugia:
I had created #1127 but will close as redundant. A customer ran into this issue -

**What happened:**

For expands, the presence of pending previous expand operations can be a blocker. In order to be able to start a new expand operation, the pending ones need to be removed from the queue (in etcd). Currently, we have been providing customers at times with a script to do so.

**What should happen:**

There should be a dedicated troubleshooting command to remove errored operations. A potential auto-fix/auto-detect and recover from state would lead to an even more improved user experience.

**Example script used:**

```
#!/bin/bash
set -Eeou pipefail
function die {
  >&2 echo $@
  exit 1
}
if [ "$EUID" -ne 0 ]
  then echo "Please run the script as root"
  exit
fi
status=$(gravity status --output=json)
cluster_name=$(echo "${status}" | jq -r ".cluster.domain")
operation_ids=$(echo "${status}" | jq -c -r '.cluster.active_operations[]? | select(.type=="operation_expand") | .id')
if [ "" = "${operation_ids}" ]; then
  die "No active expand operation found"
fi
for id in ${operation_ids}; do
  gravity enter -- --notty /usr/bin/etcdctl -- rm --recursive /gravity/local/sites/${cluster_name}/ops/${id}
done
gravity status-reset
```

a-palchikov:
* extended debug logging (incl. gRPC logs)
 * add expiration to controller reconnect loop
 * attempt to properly fail operation in case of abort (add context HTTP APIs if possible)
 * (tentative) streaming logs from agent
 * automatically fail pending operations that do not make progress
 * complicate logic of operation state transitioning (if join a worker node, cluster can transition to active on error) - same for shrink.

helgi:
Will this address https://github.com/gravitational/gravity/issues/1036 ?

a-palchikov:
Partially yes - once the changes are forward-ported to 6.2.x.

helgi:
Has this been forward ported?

a-palchikov:
[...]

PR #1298, opened 2020-03-27
[5.5.x] expand improvements

a-palchikov:
* Debug logging includes gRPC logs by default
* Controller reconnect loop (running inside the active gravity-site Pod) expires the peer connections after a certain timeout
* `plan complete` can now fail the expand operation from any controller node
* Joining agent loop will properly fail the operation is if exits with an error. Also, agent report API is now context-aware to be able to cancel/abort operation if network communication is one-way - i.e. joining node -> controller node but not the other way around),

Requires https://github.com/gravitational/gravity.e/pull/4287.
Updates https://github.com/gravitational/gravity/issues/1077.

### DIFF
--- a/lib/expand/fsm.go
+++ b/lib/expand/fsm.go
@@ -198,26 +198,16 @@ func (e *fsmEngine) RunCommand(ctx context.Context, runner rpc.RemoteRunner, nod
 // Complete is called to mark operation complete
 func (e *fsmEngine) Complete(fsmErr error) error {
 	plan, err := e.GetPlan()
-	if err != nil {
+	if err != nil && !trace.IsNotFound(err) {
 		return trace.Wrap(err)
 	}
-	if fsm.IsCompleted(plan) {
-		err = ops.CompleteOperation(e.OperationKey, e.Operator)
-	} else {
-		var message string
-		if fsmErr != nil {
-			message = trace.Unwrap(fsmErr).Error()
-		}
-		err = ops.FailOperation(e.OperationKey, e.Operator, message)
+	if fsmErr == nil {
+		fsmErr = trace.Errorf("completed manually")
 	}
-	if err != nil {
-		return trace.Wrap(err)
+	if err == nil {
+		return fsm.CompleteOperation(plan, e.Operator, fsmErr.Error())
 	}
-	e.WithFields(logrus.Fields{
-		constants.FieldSuccess: fsm.IsCompleted(plan),
-		constants.FieldError:   fsmErr,
-	}).Debug("Marked operation complete.")
-	return nil
+	return ops.FailOperation(fsm.OperationKey(*plan), e.Operator, fsmErr.Error())
 }
 
 // UpdateProgress reports operation progress to the cluster's operator
--- a/lib/expand/fsm.go
+++ b/lib/expand/fsm.go
@@ -198,16 +198,13 @@ func (e *fsmEngine) RunCommand(ctx context.Context, runner rpc.RemoteRunner, nod
 // Complete is called to mark operation complete
 func (e *fsmEngine) Complete(fsmErr error) error {
 	plan, err := e.GetPlan()
-	if err != nil && !trace.IsNotFound(err) {
+	if err != nil {
 		return trace.Wrap(err)
 	}
 	if fsmErr == nil {
 		fsmErr = trace.Errorf("completed manually")
 	}
-	if err == nil {
-		return fsm.CompleteOperation(plan, e.Operator, fsmErr.Error())
-	}
-	return ops.FailOperation(fsm.OperationKey(*plan), e.Operator, fsmErr.Error())
+	return fsm.CompleteOrFailOperation(plan, e.Operator, fsmErr.Error())
 }
 
 // UpdateProgress reports operation progress to the cluster's operator
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -620,7 +620,7 @@ func (p *Peer) waitForAgents(ctx operationContext) error {
 			if tm.IsZero() {
 				return trace.ConnectionProblem(nil, "timed out waiting for agents to join")
 			}
-			report, err := ctx.Operator.GetSiteExpandOperationAgentReport(ctx.Operation.Key())
+			report, err := ctx.Operator.GetSiteExpandOperationAgentReport(p.Context, ctx.Operation.Key())
 			if err != nil {
 				log.WithField("err", err).Warn("Failed to query operation report.")
 				continue
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/gravitational/gravity/lib/app"
@@ -144,6 +145,7 @@ type Peer struct {
 	agentDoneCh <-chan struct{}
 	// agent is this peer's RPC agent
 	agent *rpcserver.PeerServer
+	wg    sync.WaitGroup
 }
 
 // NewPeer returns new cluster peer client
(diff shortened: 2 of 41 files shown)


lib/expand/fsm.go
```diff
@@ -198,26 +198,16 @@ func (e *fsmEngine) RunCommand(ctx context.Context, runner rpc.RemoteRunner, nod
...
-		return trace.Wrap(err)
+	if err == nil {
+		return fsm.CompleteOperation(plan, e.Operator, fsmErr.Error())
 	}
-	e.WithFields(logrus.Fields{
-		constants.FieldSuccess: fsm.IsCompleted(plan),
-		constants.FieldError:   fsmErr,
-	}).Debug("Marked operation complete.")
-	return nil
+	return ops.FailOperation(fsm.OperationKey(*plan), e.Operator, fsmErr.Error())
```
r0mant: Won't plan be nil if NotFound happens above and this will panic?

lib/expand/join.go
```diff
@@ -504,6 +507,29 @@ func (p *Peer) getAgent(opCtx operationContext) (*rpcserver.PeerServer, error) {
...
+		defer cancel()
+		if p.agent != nil {
+			p.Info("Stopping peer.")
+			if err := p.agent.Stop(stopCtx); err != nil {
+				p.WithError(err).Error("Failed to stop peer.")
+			}
+		}
+		if ctx != nil && ctx.Operation.ID != "" {
+			if err2 := ops.FailOperation(ctx.Operation.Key(), ctx.Operator, err.Error()); err2 != nil {
+				p.WithError(err2).Error("Failed to mark the operation as failed.")
```
r0mant: Add more fields to the log such as operation ID? Might be useful for log entries above too.

lib/expand/join.go
```diff
@@ -504,6 +507,29 @@ func (p *Peer) getAgent(opCtx operationContext) (*rpcserver.PeerServer, error) {
...
+		p.WithError(err).Warn("Peer is exiting with error.")
+		stopCtx, cancel := context.WithTimeout(context.Background(), defaults.AgentStopTimeout)
+		defer cancel()
+		if p.agent != nil {
+			p.Info("Stopping peer.")
+			if err := p.agent.Stop(stopCtx); err != nil {
+				p.WithError(err).Error("Failed to stop peer.")
+			}
+		}
+		if ctx != nil && ctx.Operation.ID != "" {
```
r0mant: Should this also be executed during install operation, or should we check specifically for expand here?

lib/fsm/utils.go
```diff
@@ -216,6 +218,25 @@ func ClusterKey(plan storage.OperationPlan) ops.SiteKey {
 	}
 }
 
+// CompleteOperation completes the operation given by the plan in the specified operator.
+// planErr optionally specifies the error to record in the failed message
+func CompleteOperation(plan *storage.OperationPlan, operator ops.Operator, planErr string) (err error) {
```
r0mant: I'd consider renaming this to CompleteOrFailOperation. When I was looking at where it's called above I initially assumed it always sets operation to "completed".

lib/ops/opsservice/install.go
```diff
@@ -128,16 +128,19 @@ func (s *site) createInstallExpandOperation(operationType, operationInitialState
 	}
 
 	tokenType := storage.ProvisioningTokenTypeInstall
+	expires := s.clock().UtcNow().Add(defaults.InstallTokenTTL)
 	if op.Type == ops.OperationExpand {
 		tokenType = storage.ProvisioningTokenTypeExpand
+		// Do not expire join tokens
+		expires = time.Time{}
```
r0mant: Won't they keep accumulating for every join operation then? As an easy way to clean them up, we could loop over all expand tokens before creating an expand operation and delete the ones which are not for an in-progress expands.

tool/gravity/cli/install.go
```diff
@@ -743,6 +746,29 @@ func completeJoinPlan(localEnv, joinEnv *localenv.LocalEnvironment, operation *o
 	return joinFSM.Complete(trace.Errorf("completed manually"))
 }
 
+func completeJoinPlanFromAnotherNode(localEnv *localenv.LocalEnvironment, operation *ops.SiteOperation) error {
+	clusterEnv, err := localEnv.NewClusterEnvironment()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if operation == nil {
+		operation, err = ops.GetExpandOperation(clusterEnv.Backend)
```
r0mant: Which one is it going to return? Just wondering what might happen if multiple expands are running. Maybe always require --operation-id when completing expands?

tool/gravity/cli/install.go
```diff
@@ -743,6 +746,29 @@ func completeJoinPlan(localEnv, joinEnv *localenv.LocalEnvironment, operation *o
 	return joinFSM.Complete(trace.Errorf("completed manually"))
 }
 
+func completeJoinPlanFromAnotherNode(localEnv *localenv.LocalEnvironment, operation *ops.SiteOperation) error {
```
r0mant: I'd maybe call it "completeJoinPlanFrom*Existing*Node" and add a comment explaining what this is for.

lib/expand/fsm.go
a-palchikov: Good catch - I don't really need this change after I've moved the completion code to another function.

lib/expand/join.go
a-palchikov: I don't see a reason to exclude install here.

lib/ops/opsservice/install.go
a-palchikov: Do we want to clean them up? AFAIC, there's no reliable way to tell whether the operation is complete if it's failed, hence we don't remove the operation but merely mark it as failed - why remove the token, it is as good a history as the operation itself?

lib/ops/opsservice/operationgroup.go
```diff
@@ -257,17 +257,30 @@ func (g *operationGroup) onSiteOperationComplete(key ops.SiteOperationKey) error
...
 	}
 
-	site, err := g.operator.openSite(g.siteKey)
+	cluster, err := g.operator.openSite(g.siteKey)
 	if err != nil {
 		return trace.Wrap(err)
 	}
 
+	if operation.IsCompleted() {
+		token, err := cluster.users().GetOperationProvisioningToken(key.SiteDomain, key.OperationID)
```
r0mant: Nit: I'd move this block into a separate method (`cleanOperationToken`) and maybe do it on a best-effort basis i.e. do not fail if this fails for some reason?

lib/storage/keyval/tokens.go
```diff
@@ -81,7 +81,7 @@ func (b *backend) GetOperationProvisioningToken(clusterName, operationID string)
 			return t, nil
 		}
 	}
-	return nil, trace.Wrap(err)
+	return nil, trace.NotFound("not provisioning token for cluster %v and operation %v", clusterName, operationID)
```
r0mant: not -> no?

tool/gravity/cli/operation.go
```diff
@@ -99,7 +99,10 @@ func completeOperationPlan(localEnv, updateEnv, joinEnv *localenv.LocalEnvironme
 		// There's only one install operation
 		return completeInstallPlan(localEnv, op)
 	case ops.OperationExpand:
-		return completeJoinPlan(localEnv, joinEnv, op)
+		if op == nil {
+			return trace.BadParameter("operation ID must be specified")
```
r0mant: Nit: Maybe require this only if multiple operations are in progress? I.e. above find all active operations, and if there're more than 1 and operationID is empty then bounce, otherwise proceed as before.

r0mant (approved):

tool/gravity/cli/operation.go
a-palchikov: That code wrong anyways - I've restructured operation handling so it's uniform for all operation types.

Issue #1279, opened 2020-03-25
Planet container in inconsistent state


a-palchikov:
On multiple occasions we saw running planet container in an inconsistet state:
 * planet (service) user not matching the one from host. For example, the host planet user might have ID `997` and belong to a planet group with ID `994` but inside the container the user would default to planet with stock UID:GID of `1000`:`1000`.
 * `/etc/resolv.conf` is empty inside the container although failure to reflect the host changes are fatal (i.e. the `start` command should be failing if there are errors reflecting the host state)

The latest occurrence is during the upgrade from `5.5.28` -> `5.5.38` where on one of the control nodes the planet container appeared in such state during `system-upgrade` step.

Restart of the container fixes the inconsistency.

We should consider changing upgrade tests to use a custom service user.

wadells:
Are there any logs or snippets of the symptoms that I could look at?  I'm trying to determine the observed impact of this bug, to make sure the test case I'm working on exercises it.

For example is it:

1) Everything in the cluster seems to be working ok after the upgrade, but planet is running under the wrong uid/gid?  No internal conflict, but planet might conflict with another service/user on the system specific to a customer.
2) Planet fails to X because of uid/gid mismatch.  What is X in the logs or `gravity status` output?

a-palchikov:
If the planet container starts with the wrong uid/gid combo, services that run in the context of the service user will fail to read the files.
As an example (also taken from the failed upgrade), is the etcd service (running as service user) that is unable to read its startup configuration (TLS certs/keys). In this case, the automatic upgrade process fails after a timeout.

r0mant:
Closed in https://github.com/gravitational/gravity/pull/1301 (5.5), https://github.com/gravitational/gravity/pull/1306 (6.1), https://github.com/gravitational/gravity/pull/1309 (6.3) and https://github.com/gravitational/gravity/pull/1310 (7.0).

PR #1301, opened 2020-03-30
(5.5) Unpack packages during bootstrap upgrade phase.

r0mant:
This PR attempts to resolve https://github.com/gravitational/gravity/issues/1279. While I couldn't reproduce the _exact_ scenario, I got a sort of similar behavior to the one described by repeatedly executing/rolling back the system-upgrade phase and watching the changes to /etc/passwd inside planet's rootfs.

As it turns out, some of the code that performs upgrade actually re-unpacks the planet package in a way that overrides its rootfs (in /var/lib/gravity/local/packages/unpacked/...) even if it had already been unpacked before. Both @a-palchikov and I came to a conclusion that there may be a race between this and the "gravity package command" command that actually starts the planet and also unpacks the package.

So, this PR does the following:

* Explicitly unpack new packages during upgrade's bootstrap phase. This is the same behavior as during install actually, and it makes sure that when gravity starts installing the package, it is already unpacked.
* Update the "updater" code to not unpack the package if it's already been unpacked (which it will be, since we unpack it at bootstrap) by calling a corresponding PackageService method rather than using the one from pack/utils.go which always overwrites it.

Closes https://github.com/gravitational/gravity/issues/1279. Should be forward-ported.

### DIFF
--- a/tool/gravity/cli/system.go
+++ b/tool/gravity/cli/system.go
@@ -1276,7 +1276,7 @@ func unpack(p *localpack.PackageServer, loc loc.Locator) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	return trace.Wrap(pack.Unpack(p, loc, path, nil))
+	return p.Unpack(loc, path)
 }
 
 func maybeConvertLegacyPlanetConfigPackage(configPackage loc.Locator) (*loc.Locator, error) {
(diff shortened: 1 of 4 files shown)


tool/gravity/cli/system.go
```diff
@@ -1276,7 +1276,7 @@ func unpack(p *localpack.PackageServer, loc loc.Locator) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	return trace.Wrap(pack.Unpack(p, loc, path, nil))
+	return p.Unpack(loc, path)
```
r0mant: I'm not sure why this is still here (in cli package) since upgrade uses lib/update/system now but decided to update it here as well just in case. IIRC we kept this to support legacy "gravity system update" command, which we should ideally rewrite to use lib/ too.

a-palchikov (approved):

PR #1312, opened 2020-04-01
Add custom planet uid & gid to PR upgrade tests

wadells:
Although I've never been able to repro the exact behavior, this is
trying to catch failures like issue #1279 by upgrading with a custom
planet service account.

Removed files are cleanup of dead code that would have been touched by the upgrade3lts -> upgrade rename.

### Testing Done

[This jenkins job](https://jenkins.gravitational.io/view/Robotest/job/Robotest-run-gce/158/console) shows an upgrade test run with [the uid-gid container tag and the service_uid and service_gid options set](https://jenkins.gravitational.io/view/Robotest/job/Robotest-run-gce/158/parameters/).

I also commented out the docker invcation at the end of `robotest_run_suite.sh`, gave it some needed variables and generated the following suite (newlines added for readability):

```
resize={"to":3,"flavor":"one","nodes":1,"role":"node","state_dir":"/var/lib/telekube","os":"ubuntu:18","storage_driver":"overlay2"}
upgrade={"flavor":"three","nodes":3,"role":"node","service_uid":994,"service_gid":994,"os":"ubuntu:18","storage_driver":"overlay2","from":"/telekube_6.3.6.tar"}
upgrade={"flavor":"three","nodes":3,"role":"node","service_uid":994,"service_gid":994,"os":"ubuntu:16","storage_driver":"overlay2","from":"/telekube_6.1.0.tar"}
upgrade={"flavor":"three","nodes":3,"role":"node","service_uid":994,"service_gid":994,"os":"ubuntu:18","storage_driver":"overlay2","from":"/telekube_6.1.18.tar"}
install={"flavor":"three","nodes":3,"role":"node","os":"redhat:7","storage_driver":"overlay2"}
install={"installer_url":"/installer/opscenter.tar","nodes":1,"flavor":"standalone","role":"node","os":"ubuntu:18","ops_advertise_addr":"example.com:443"}
```


a-palchikov (approved):

Jenkinsfile
```diff
@@ -33,7 +33,7 @@ properties([
            defaultValue: '1',
            description: 'How many times to repeat each test.'),
     string(name: 'ROBOTEST_VERSION',
-           defaultValue: 'stable-gce',
+           defaultValue: 'uid-gid',
```
r0mant, line 36: I'm assuming this is going to change back to stable tag at some point?

build.assets/robotest_run_suite.sh
```diff
@@ -44,9 +44,10 @@ function build_upgrade_step {
   local release=${2:?$usage}
   local storage_driver=${3:?$usage}
   local cluster_size=${4:?$usage}
+  local service_opts='"service_uid":994,"service_gid":994' # see issue #1279
```
r0mant, line 47: A total nitpick, but can we please use different numbers for uid and gid? Failure scenarios keep flowing in my head where we'd somehow mix uid/gid in the code and it won't surface if they're the same. Guess I'm a bit paranoid :)

r0mant (approved):
Nice :+1:

Jenkinsfile
wadells: I probably won't change it back to a "stable tag", because the named feature tag pattern we're currently using is atypical.  If you're concerned about stability, both stable-gce and uid-gid are releases off robotest's master branch, and I believe uid-gid is a direct descendant of stable-gce (though I haven't proved it yet).

I'd like to return robotest to semver release tagging/naming but doing so was not higher priority than this work. I chose not to fix semver for this particular task because our current semver is 1.0.217 or something silly -- with several backwards incompatible changes (e.g. not attempting to maintain AWS and Azure support). I'll cut a 2.0 when I make a bit of priority to revisit our Robotest release process.

r0mant: Yeah, for sure, I just meant to use a some kind of "proper" tag eventually because "uid-gid" is sort of like "dima" :)

build.assets/robotest_run_suite.sh
wadells: Changed to uid:997 gid:994 in a02d745321c709d343af0d0e5a10f9f46bc013d3.  I haven't explicitly tested this beforehand, but I'll keep an eye on the PR build for verification.

Jenkinsfile
wadells: I've filed https://github.com/gravitational/robotest/issues/200 to track this.

PR #1317, opened 2020-04-02
Add tele helm build command

### LINKED ISSUE #1184: Implement tele helm build command
tele should support building application images out of Helm charts:

- [ ] The command that builds an app image should be called `tele helm build`.
- [ ] Should be able to build local Helm chart.
- [ ] Should be able to build remote Helm chart.

### LINKED ISSUE #1185: Add support for building cluster image out of helm charts
tele should support building cluster images directly out of helm charts, without having to create a manifest.

- [ ] tele build can build cluster image from local chart.
- [ ] tele build can build cluster image from remote chart.
- [ ] tele build supports `--with-base-image` flag to specify which base image to use.

### LINKED ISSUE #1186: Generate default hooks for helm charts
When building a cluster image from a chart, Gravity should generate default hooks unless the user provided them in the manifest file:

- [ ] Default install hook that calls `helm install` on the user's chart.
- [ ] Default upgrade hook that calls `helm upgrade`.
- [ ] Default rollback hook that calls `helm rollback`.

### LINKED ISSUE #1188: Add support for gravity helm commands
Currently, the set of subcommands that deal with application images are called `gravity app`.

We need to add a `gravity helm` alias to them so the commands would be `gravity helm install`, `gravity helm upgrade`, etc.

r0mant:
Before this PR `tele build` was used to build both cluster (when given a manifest file) and application (when given a Helm chart) images. Now, according to our new designed CLI flows, they are split:

* `tele build` builds cluster images either from manifests or charts.
* `tele helm build` (or `tele app build` alias) builds application images from charts.

In addition, when `tele build` creates a cluster image from a Helm chart, it pre-generates all hooks so the application is installed together with the cluster.

Implementation-wise, to keep things simple I split the monolithic builder we have into separate application and cluster builders, and the old "builder" became an "engine" which contains common parts of the two. There are undoubtedly opportunities for refactoring this further but I didn't want to spend more time on this now.

Other things that are worth mentioning:

* `gravity app` commands are also now aliased to `gravity helm`.
* Add `--with-base-image` flag to `tele build` that allows to select base image on the CLI.

Closes https://github.com/gravitational/gravity/issues/1184, closes https://github.com/gravitational/gravity/issues/1185, closes https://github.com/gravitational/gravity/issues/1186, closes https://github.com/gravitational/gravity/issues/1188.

This should go into 7.1 and as such not be merged until we tag 7.0.0.


wadells:
How did I end up authoring https://github.com/gravitational/gravity/pull/1317/commits/6b860823ea3674550cf677a63de4b84ae3824c89 ? I don't remember doing that. :laughing:

r0mant:
@wadells lol no idea, I was wondering that as well. I did rebase/fixup locally sometime before pushing, maybe that caused that somehow.

lib/builder/source.go
```diff
@@ -0,0 +1,153 @@
...
+// Dir returns directory where manifest is located.
+func (s *clusterImageSourceManifest) Dir() string {
+	return filepath.Dir(s.manifestPath)
+}
+
+// Manifest returns parsed cluster image manifest.
+func (s *clusterImageSourceManifest) Manifest() (*schema.Manifest, error) {
+	manifest, err := schema.ParseManifest(s.manifestPath)
+	if err != nil {
+		log.Error(trace.DebugReport(err))
```
a-palchikov, line 113: That's not going to look good in logs.

lib/builder/source.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil, trace.BadParameter("could not parse manifest file:\n%v",
+			trace.Unwrap(err)) // show original parsing error
```
a-palchikov, line 115: Why not wrap the original error?

lib/builder/source.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	return manifest, nil
+}
+
+const (
+	pathError = `Path provided to tele build command should be one of the following:
+* Cluster image manifest file path.
+* Directory that contains cluster image manifest file named "app.yaml".
```
a-palchikov, line 150: nit: replace `app.yaml` with `defaults.ManifestFileName`

lib/builder/util.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	_, err = client.Version()
+	if err != nil {
+		log.Error(trace.DebugReport(err))
+		return trace.BadParameter(
+			`Docker does not seem to be available on this machine.
+To resolve the issue:
+  * Install it (https://docs.docker.com/engine/installation/).
```
a-palchikov, line 72: nit: replace `it` with Docker

lib/builder/util.go
```diff
@@ -0,0 +1,124 @@
...
+	teleVersion, err := semver.NewVersion(version.Get().Version)
+	if err != nil {
+		return trace.Wrap(err, "failed to determine tele version")
+	}
+	if !versionsCompatible(*teleVersion, *runtimeVersion) {
+		return trace.BadParameter(
+			`Version of this tele binary (%[1]v) is not compatible with the base image version specified in the manifest (%[2]v).
+To resolve the issue you can do one of the following:
+  * Download tele binary of the same version as the specified base image (%[2]v) and use it to build the image.
+  * Specify base image version compatible with this tele in the manifest file: "baseImage: gravity:%[1]v".
```
a-palchikov, line 90: This does not provide enough information for user to know what compatible base image could be.

lib/builder/util.go
```diff
@@ -0,0 +1,124 @@
...
+//   2. Runtime version is not greater than tele version.
+func versionsCompatible(teleVer, runtimeVer semver.Version) bool {
+	return teleVer.Major == runtimeVer.Major &&
+		teleVer.Minor == runtimeVer.Minor &&
+		!teleVer.LessThan(runtimeVer)
+}
+
+// ensureCacheDir makes sure that the default local cache directory for
+// the provided Gravity Hub exists.
+func ensureCacheDir(hubURL string) (string, error) {
```
a-palchikov, line 113: nit: could use `dir` as a named return

tool/tele/cli/commands.go
```diff
@@ -50,11 +54,45 @@ type VersionCmd struct {
...
+	// VendorIgnorePatterns if file pattern to ignore when searching for images
+	VendorIgnorePatterns *[]string
+	// SetImages rewrites images to specified versions
+	SetImages *loc.DockerImages
+	// Parallel defines the number of tasks to execute concurrently
+	Parallel *int
+	// Quiet allows to suppress console output
+	Quiet *bool
+	// Verbose enables more detailed build output.
+	Verbose *bool
```
a-palchikov, line 82: Why not use `--debug` for this? `tele helm --debug build --verbose` would look confusing.

a-palchikov (approved):

lib/app/hooks/configure.go
```diff
@@ -408,20 +408,16 @@ TMPDIR={{.StateDir}} {{.StateDir}}/gravity --state-dir={{.StateDir}} app unpack
 	--service-uid={{.ServiceUser.UID}} \
 	--insecure --ops-url=$ops_url \
 	{{.Package}} {{.ResourcesDir}}
-{{if .HelmValues}}
```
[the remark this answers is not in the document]
r0mant: Not really, this is just to make sure that values file is always present inside hooks even if it's empty. This allows to always pass it to helm commands without checking whether it exists or not.

PR #1327, opened 2020-04-06
Fix jenkins config lag & try build support

### LINKED ISSUE #1315: Changes to Jenkins build parameters take one build to queisce
If there are preexisting builds (e.g. nightly builds or 2nd+ PR build) and the Jenkins parameters defined in Jenkinsfile or Jenkinsfile.nightly change, the changes do not affect the to the next build.

### Context
I noticed this when first when my preeemptible changes (#1290 ) landed, but did not affect the [next build](https://jenkins.gravitational.io/job/Gravity-Nightly/348/).  I dismissed this first time it as a fluke.

I then noticed this behavior again when I updated the robotest version in PR #1312 and the [next build](https://jenkins.gravitational.io/job/Gravity-Nightly/353/console) failed because it wasn't using the new container, resulting in 

```
suite_test.go:82: failed to parse args: no such function: "upgrade"
```

Which stems from the backwards incompatible change in https://github.com/gravitational/robotest/pull/190, and the fact that [nightly 353 did not pick up the new uid-gid tag](https://jenkins.gravitational.io/job/Gravity-Nightly/353/parameters/).

The parameters I'm talking about are here: 

https://github.com/gravitational/gravity/blob/c3823b8ff7b176d38f6a5b4d50f14e6b8b890438/Jenkinsfile#L12-L44
https://github.com/gravitational/gravity/blob/c3823b8ff7b176d38f6a5b4d50f14e6b8b890438/Jenkinsfile.nightly#L55-L87

These do not apply to the first build after a change, because as Job parameters, they exist as configuration state within the Jenkins instance.  They're then set to the old defaults when `propegateParamsToEnv` is run:

https://github.com/gravitational/gravity/blob/c3823b8ff7b176d38f6a5b4d50f14e6b8b890438/Jenkinsfile.nightly#L2-L8

Which results to line 4 evaluating false.  This means the Jenkins job config is changed but not the local environment variables, which are set to the previously configured defaults.  With the job config changed, the next build will pick up the changes -- or in the case of PR builds, the first build will always pick up all changes.

wadells:
This change set brings two substantial edits: 

* Remove Robotest Jenkins parameters for #1315 
* Add a flag to skip Jenkins parameter definition to better support try builds

and one piece of cleanup:

* Remove unused AWS config.

See the commit messages for more context as to why I made each change.

Fixes #1315 and unblocks backporting #1312 (at least it will prevent 1312 backports from breaking other developers in-flight PRs).

### Testing Done
[Try Build 40](https://jenkins.gravitational.io/view/Robotest/job/gravity-try-build/40/) tests the PR Jenkinsfile & script edits.
[Try Build 41](https://jenkins.gravitational.io/view/Robotest/job/gravity-try-build/41/) tests the Nightly Jenkinsfile & script edits.

Although these are still running at the time of posting, I've gotten several previously green PR tests (38 & 39) and I will check results to be sure there aren't any Nightly regressions before merging -- the PR build will show there aren't any PR regressions.  I didn't want to wait another day to get feedback, as this work is already several days into development.

### Notes
With this merge, I'm getting close to having reliably Gravity Try Builds ready for GA.  I've been included work on them for my own purposes because I need a way t test nightly & PR edits outside of the official runs.  But I believe they'll be useful to other Gravity developers as well, as a way to pre-vet code.  For example PRs that are for testing like https://github.com/gravitational/gravity/pull/1242 can be replaced by try builds. The remaining work is largely UI/UX to make sure it is easy for Gravity developers to kick off try builds.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,8 +10,11 @@ def propagateParamsToEnv() {
 properties([
   disableConcurrentBuilds(),
   parameters([
+    // WARNING: changing parameters will not affect the next build, only the following one
+    // see issue #1315 or https://stackoverflow.com/questions/46680573/ -- 2020-04 walt
+
+    // For choice parameters, the first choice will be the default.
     choice(choices: ["run", "skip"].join("\n"),
-           // defaultValue is not applicable to choices. The first choice will be the default.
            description: 'Run or skip robotest system wide tests.',
            name: 'RUN_ROBOTEST'),
     choice(choices: ["true", "false"].join("\n"),
@@ -23,24 +26,6 @@ properties([
     choice(choices: ["true", "false"].join("\n"),
            description: 'Abort all tests upon first failure.',
            name: 'FAIL_FAST'),
-    choice(choices: ["gce"].join("\n"),
-           description: 'Cloud provider to deploy to.',
-           name: 'DEPLOY_TO'),
-    string(name: 'PARALLEL_TESTS',
-           defaultValue: '4',
-           description: 'Number of parallel tests to run.'),
-    string(name: 'REPEAT_TESTS',
-           defaultValue: '1',
-           description: 'How many times to repeat each test.'),
-    string(name: 'ROBOTEST_VERSION',
-           defaultValue: 'uid-gid',
-           description: 'Robotest tag to use.'),
-    choice(choices: ["false", "true"].join("\n"),
-           description: 'Whether to use preemptible VMs.',
-           name: 'GCE_PREEMPTIBLE'),
-    choice(choices: ["custom-4-8192", "custom-8-8192"].join("\n"),
-           description: 'VM type to use.',
-           name: 'GCE_VM'),
   ]),
 ])
 
@@ -90,7 +75,7 @@ timestamps {
                 [$class: 'FileBinding', credentialsId:'OPS_SSH_KEY', variable: 'SSH_KEY'],
                 [$class: 'FileBinding', credentialsId:'OPS_SSH_PUB', variable: 'SSH_PUB'],
                 ]) {
-                  sh "make -C e robotest-run-suite ROBOTEST_VERSION=$ROBOTEST_VERSION"
+                  sh 'make -C e robotest-run-suite'
             }
           }else {
             echo 'skipped system tests'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,27 +7,31 @@ def propagateParamsToEnv() {
   }
 }
 
-properties([
-  disableConcurrentBuilds(),
-  parameters([
-    // WARNING: changing parameters will not affect the next build, only the following one
-    // see issue #1315 or https://stackoverflow.com/questions/46680573/ -- 2020-04 walt
-
-    // For choice parameters, the first choice will be the default.
-    choice(choices: ["run", "skip"].join("\n"),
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -83,24 +82,14 @@ timestamps {
           }
         },
         robotest : {
-          if (params.RUN_ROBOTEST == 'run') {
+          if (env.RUN_ROBOTEST == 'run') {
             withCredentials([
-                [
-                  $class: 'UsernamePasswordMultiBinding',
```
a-palchikov, line 89: Hmm, I believe robotest is using these to pull static assets like prebuilt installers from s3 - is this not necessary anymore?

build.assets/robotest_run_nightly.sh
```diff
@@ -25,15 +25,23 @@ readonly ROBOTEST_SCRIPT=$(mktemp -d)/runsuite.sh
 
 # number of environment variables are expected to be set
 # see https://github.com/gravitational/robotest/blob/master/suite/README.md
-export ROBOTEST_VERSION=${ROBOTEST_VERSION:-stable-gce}
+export ROBOTEST_VERSION=${ROBOTEST_VERSION:-uid-gid}
```
webvictim, line 28: Is the use of this variable correct? Looks like it might contain a `uid:gid` pair rather than a semver.

build.assets/robotest_run_suite.sh
```diff
@@ -22,15 +22,23 @@ readonly ROBOTEST_SCRIPT=$(mktemp -d)/runsuite.sh
 
 # number of environment variables are expected to be set
 # see https://github.com/gravitational/robotest/blob/master/suite/README.md
-export ROBOTEST_VERSION=${ROBOTEST_VERSION:-stable-gce}
+export ROBOTEST_VERSION=${ROBOTEST_VERSION:-uid-gid}
```
webvictim, line 25: `uid-gid` vs semver

webvictim (commented):
LGTM on the whole, but I don't know a massive amount about the implementation.

Jenkinsfile
a-palchikov: Ah - just noticed https://github.com/gravitational/robotest/issues/202 that should address this - nvm.

a-palchikov (approved):

Jenkinsfile
wadells: We currently pull everything from get.gravitational.io for the nightly and PR runs.  You're correct that if we were to use S3 artifacts, we'd need these creds again.

It sounds like we may be moving back to AWS anyhow soon enough, at which point I'll have a patch to cutover.  These just weren't used currently.

build.assets/robotest_run_suite.sh
wadells: I agree.  I've got a bit more work to do on the Robotest side before I cut back to semver.  This is tracked in https://github.com/gravitational/robotest/issues/200

build.assets/robotest_run_nightly.sh
wadells: Yeah, it is a poorly named tag in quay.io https://quay.io/repository/gravitational/robotest-suite?tab=tags.  As mentioned in your other comment, I should be going to semver shortly after a 2 year hiatus for the project.

PR #1328, opened 2020-04-07
Add ability to view operations in cli.

### LINKED ISSUE #897: Show more information on operations in gravity status
When running `gravity status` to see what operations are in progress, and detecting if anything is stuck, often causes frustration related to how little information a given operation ID has.

f.ex. given an operation (join) how can I know what node it matches in the worker list below; I might see 5 offline nodes while they are in the join process. It seem digging through `etcdctl` is the best option to get furhter info.

Ideal scenario is at least show the machine identifier used elsewhere in `gravity status` as part of the operations view at the top and perhaps ability to get more information on an operation (`gravity operation view <id>` maybe?) that can almost just be a json dump of what's in etcd?

### LINKED ISSUE #900: Add gravity CLI commands to work with operations
Adding view (dumping the JSON object f.ex.) and delete (to deal with stuck operations) to remove the need for operators to drop down into `etcd` to resolve problems.

r0mant:
Also, better operation details in gravity status.

Better show than tell:

```
$ ./gravity resource get operations
ID                                       Description                         State         Created
--                                       -----------                         -----         -------
2af46ef1-7c6c-44c4-8058-18eee7f2a6e8     Upgrade to version 5.5.40-dev.6     Completed     Thu Apr  2 23:02 UTC
60078163-4a2b-4578-a05c-720d05618ffc     Upgrade to version 5.5.39           Completed     Mon Mar 30 18:45 UTC
ce830126-1b82-4736-b04b-a11b5cd3336e     Install on 3 nodes                  Completed     Fri Mar 27 19:19 UTC
```

```
$ ./gravity status
Cluster name:		lucidknuth3007
...
Last completed operation:
    * Upgrade to version 5.5.40-dev.6
      ID:		2af46ef1-7c6c-44c4-8058-18eee7f2a6e8
      Started:		Thu Apr  2 23:02 UTC (4 days ago)
      Completed:	Thu Apr  2 23:11 UTC (4 days ago)
...
```

Closes https://github.com/gravitational/gravity/issues/897, closes https://github.com/gravitational/gravity/issues/900.

### DIFF
--- a/lib/storage/operation.go
+++ b/lib/storage/operation.go
@@ -1,0 +1,354 @@
+/*
+Copyright 2020 Gravitational, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package storage
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/gravitational/teleport/lib/services"
+	"github.com/gravitational/teleport/lib/utils"
+
+	"github.com/gravitational/trace"
+	"github.com/jonboulle/clockwork"
+)
+
+// Operation represents a single cluster operation.
+type Operation interface {
+	// Resource provides common resource methods.
+	services.Resource
+	// CheckAndSetDefaults validates the object and sets defaults.
+	CheckAndSetDefaults() error
+	// GetType returns the operation type.
+	GetType() string
+	// GetCreates returns the operation created timestamp.
+	GetCreated() time.Time
+	// GetUpdated returns the operation last updated timestamp.
+	GetUpdated() time.Time
+	// GetState returns the operation state.
+	GetState() string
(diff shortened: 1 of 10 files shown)


helgi:
Awesome! This is super useful. Would it be possible to get `delete` as well? Described it in #900 - It reduces the need greatly to need to drop down into `etcd` to clear up a stuck operation (it if ever happens) in a timely manner, and the CLI here could provide a much safer way to do it. Can, of course, be added later :)

a-palchikov:
@helgi `delete` is not necessary once it's possible to transition any given operation into failed state manually (this is what I'm working on). Deleting will also destroy operation history which should ideally be preserved.
When an operation is transitioned to failed state, the cluster state is unblocked accordingly (if it's the last active operation). We are also considering allowing certain operations even on a blocked cluster (for example, expands with a worker node if the cluster is currently blocked only on other worker expansions).

a-palchikov (approved):

lib/storage/operation.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+// GetOperationSchema returns a cluster operation schema.
+func GetOperationSchema() string {
+	return fmt.Sprintf(services.V2SchemaTemplate, services.MetadataSchema,
+		OperationSpecV2Schema, "")
+}
+
+// OperationSpecV2Schema is the operation json schema.
+var OperationSpecV2Schema = fmt.Sprintf(`{
```
a-palchikov, line 289:
```go
../lib/storage/operation.go:289: Sprintf format has invalid argument index
```
I believe this should be `%[1]v`

helgi:
@a-palchikov ah great, thanks for the update. Am on board with that approach and it makes much more sense :)

PR #1343, opened 2020-04-08
Clarify End of Life for non-LTS releases

wadells:
This changeset documents our soon to be official :crossed_fingers:  release support policy and refactors how we display supported & EoL'd releases to use two different tables.  I deemphasized notes related to unsupported releases as well.

Non-LTS EoL'd releases now all have a "Supported Until" date determined these via the `published_at` field for the subsequent x.x.0 release here: https://api.github.com/repos/gravitational/gravity/releases

### Testing Done
I ran `make run-docs` and got the following screenshot:
Screenshot_2020-04-08 Releases - Gravitational Gravity

### Original Remarks
I've now added a table split per Ben's advice.  The following is context from the PR description that started of that discussion.

I'm considering a more prominent division between supported & EOL'd releases.  Currently we have the `*` marker and a `Supported Until` in the past.  We may consider any of the following:

1. Different color for EOL'd rows
2. Two seperate tables, one for supported lines of development and one for EOL'd ones
3. Everything in the same table, but EOL hidden/collapsed until some sort of toggle is punched.

I lean towards option 2 as it doesn't require any css or javascript work and keeps things simple for future maintainers.  If you are in favor of any of the above, let me know and I can add it to this PR.

### DIFF
--- a/docs/7.x/changelog.md
+++ b/docs/7.x/changelog.md
@@ -8,11 +8,21 @@
 | 6.1.20        | Yes | March 31st, 2020     | November 10th, 2021  | 1.15.11            | 3.2.12           |
 | 5.5.40        | Yes | April 3rd, 2020      | September 7th, 2020  | 1.13.11            | 3.0.6-gravity    |
 
+Gravity offers one Long Term Support (LTS) release for every 2nd Kubernetes
+minor version, allowing for seamless upgrades per Kubernetes
+[supported version skew policy](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-version-skew).
+LTS Gravity versions are supported with security and bug fixes for two years
+after the x.x.0 release date.
+
+Non-LTS (regular) versions of Gravity have the latest features and offer
+more current versions of Kubernetes. Regular versions of Gravity are supported
+with security and bug fixes until the subsequent Gravity release is published.
+
 ## Unsupported Releases
 
 These releases are past their End of Life date, and no longer receive security
-fixes and features. [Paying Gravity customers](https://gravitational.com/gravity/demo/)
-may conditionally receive extended support.
+and bug fixes. [Gravity customers](https://gravitational.com/gravity/demo/) can
+extend updates past EOL through customer agreements if required.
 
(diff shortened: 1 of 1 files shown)


benarent:
+1 for splitting it into two tables.

wadells:
> +1 for splitting it into two tables.

Split in fbe96b6.

docs/7.x/changelog.md
```diff
@@ -1,41 +1,34 @@
...
-    expired, please refer to the [Gravity Manual Certificates Renewal](https://gravitational.zendesk.com/hc/en-us/articles/360000755967-Telekube-Manual-Certificates-Renewal)
-    article in our Help Center.
+## Unsupported Releases
 
-!!! note "Unsupported releases"
-    Releases marked with `*` in the table above are no longer supported and
-    do not receive updates and bugfixes.
+These releases are past their End of Life date, and no longer receive security
+fixes and features. [Paying Gravity customers](https://gravitational.com/gravity/demo/)
+may conditionally receive extended support.
```
wadells, line 15: Do we want upsells like this here?  I added it for two reasons:

1) To assuage current customers that we'll almost certainly cut further patch releases for "technically unsupported" branches.
2) To encourage OSS users needing extended support to convert.

I'm open to better language.  Thoughts @benarent, @stevenGravy, or @sw4ti16?

r0mant (approved):
We can probably also collapse the second table by default, will make the page even cleaner. There should be a markdown option to do that:

```
<details><summary>Click here to view an example IAM policy.</summary>
// collapsed content
</details>
```

wadells:
> We can probably also collapse the second table by default, will make the page even cleaner. There should be a markdown option to do that:
> 
> ```
> <details><summary>Click here to view an example IAM policy.</summary>
> // collapsed content
> </details>
> ```

Unfortunately mkdocs (and markdown more generally) doesn't like markdown-inside-html-inside-markdown, resulting in the following:

Screenshot_2020-04-08 Releases - Gravitational Gravity

a-palchikov (approved):

PR #1347, opened 2020-04-09
[6.3.x] forward port changes/improvements from 6.1.x

a-palchikov:
This PR forward-ports improvements from 6.1.x. Will be the basis for the following expand improvement forward port.

Requires https://github.com/gravitational/gravity.e/pull/4293.


.gitignore
```diff
@@ -1,25 +1,18 @@
 .vscode
 .vagrant
 .DS_Store
-build
```
a-palchikov, line 4: A handful of changes here but ultimately to fix the problem with arbitrary `build` directories under vendor not being inlcuded.

lib/localenv/clusterenv.go
```diff
@@ -41,14 +41,12 @@ import (
 // with all services initialized
 func (r *LocalEnvironment) NewClusterEnvironment(opts ...ClusterEnvironmentOption) (*ClusterEnvironment, error) {
 	client, _, err := httplib.GetClusterKubeClient(r.DNS.Addr())
-	if err != nil {
-		log.Errorf("Failed to create Kubernetes client: %v.",
-			trace.DebugReport(err))
+	if err != nil && !trace.IsNotFound(err) {
```
a-palchikov, line 44: Otherwise, the logs get spam for reasons that kubeconfig files are not available (which is true only after the installation).

lib/localenv/clusterenv.go
```diff
@@ -41,14 +41,12 @@ import (
...
+		log.WithError(err).Warn("Failed to create Kubernetes client.")
 	}
 	auditLog, err := r.AuditLog(context.TODO())
-	if err != nil {
-		log.Warnf("Failed to create audit log: %v.",
-			trace.DebugReport(err))
+	if err != nil && !trace.IsNotFound(err) {
```
a-palchikov, line 48: Ditto here but for being unable to find etcd key which is again the case when installing.
Not sure if there's a chance this will ignore a "good" error - need more information in the error value.

lib/report/system.go
```diff
@@ -117,16 +110,16 @@ func systemFileLogs() Collectors {
...
+		Script("gravity-install.log", fmt.Sprintf(template, defaults.GravityUserLogPath)),
+		Script("gravity-local-system.log", fmt.Sprintf(template, defaults.GravitySystemLogFile)),
+		Script("gravity-local-install.log", fmt.Sprintf(template, defaults.GravityUserLogFile)),
 	}
 }
 
 // planetLogs fetches planet syslog messages as well as the fresh journal entries
 func planetLogs() Collectors {
 	return Collectors{
-		// Fetch planet syslog messages as a tarball
```
a-palchikov, line 128: Only worked in 5.5.x

r0mant (approved):

PR #1348, opened 2020-04-09
[6.3.x] expand improvements

a-palchikov:
This is an indirect forward port of [these changes](https://github.com/gravitational/gravity/pull/1298).

When it came to adding context support, I only opted to add it to APIs that mattered in expand case - we will need a good pass over the Operator APIs to add this support universally.

### DIFF
--- a/lib/defaults/defaults.go
+++ b/lib/defaults/defaults.go
@@ -221,10 +221,6 @@ const (
 	// to reconnect to the peer
 	AgentGroupPeerReconnectTimeout = 15 * time.Minute
 
-	// EnvPeerConnectTimeout is the environment variable that overrides the value of the
-	// agent timeout for the validating connect
-	EnvPeerConnectTimeout = "GRAVITY_PEER_CONNECT_TIMEOUT"
-
 	// GravityPackagePrefix defines base prefix of gravity package
 	GravityPackagePrefix = "gravitational.io/gravity"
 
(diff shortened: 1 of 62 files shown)


lib/expand/join.go
```diff
@@ -1053,16 +1076,16 @@ func (p *Peer) emitAuditEvent(ctx operationContext) error {
 	return nil
 }
 
-func (p *Peer) operationContext() (*operationContext, error) {
+func (p *Peer) operationContext(ctx context.Context) (*operationContext, error) {
```
a-palchikov, line 1079: That's a bug that prevented the agent to leave the cluster when aborted as it would receive on `p.ctx.Done()` immediately and exit with an error before.

tool/gravity/cli/install.go
```diff
@@ -735,6 +763,7 @@ var InterruptSignals = signals.WithSignals(
 	os.Interrupt,
 	syscall.SIGTERM,
 	syscall.SIGQUIT,
+	syscall.SIGHUP,
```
a-palchikov, line 766: This should cover the missing case of SSH session closing and terminating the group.

lib/defaults/defaults.go
```diff
@@ -1118,6 +1122,13 @@ var (
 	// agents to form a cluster before commencing the operation
 	AgentWaitTimeout = 5 * time.Minute
 
+	// GenerateDebugReportTimeout specifies the maximum amount of time to wait for
+	// cluster-wide debug report generation.
+	// TODO(dmitri): this is trickier to handle correctly - ideally, this should much shorter
+	// and for the initial data wait (until there's data coming on the conn).
+	// There should be no timeout for the actual report receive operation.
+	GenerateDebugReportTimeout = 10 * time.Minute
```
r0mant, line 1130: Hm, does this apply to reports generated with `gravity report` CLI command as well? If so, I think this might basically render it broken because in our support cases report generation takes a very long time, might be longer than 10 minutes for sure.

r0mant (approved):
Just a question about report timeout.

lib/defaults/defaults.go
a-palchikov: No, it's used in `gravity install` for automatically generated crash reports atm. There's no good solution so far (I believe as far as go1.13+) as 'there's no way to have server respond with a timeout for sends without the global write timeout (which would be wrong for downloads).

PR #1358, opened 2020-04-10
[7.0.x] forward port changes from 6.3.x and 5.5.x

a-palchikov:
This PR catches up with 6.3.x and 5.5.x branches and ports the following additionally:
https://github.com/gravitational/gravity/pull/326
https://github.com/gravitational/gravity/pull/1081
https://github.com/gravitational/gravity/pull/1043
https://github.com/gravitational/gravity/pull/1093
https://github.com/gravitational/gravity/pull/1278

Requires https://github.com/gravitational/gravity.e/pull/4294.


lib/ops/opsservice/agents.go
```diff
@@ -588,16 +599,29 @@ func (r *agentGroup) add(p rpcserver.Peer, hostname string) {
...
 }
 
 // hasPeer determines whether the group already has a peer with the specified
-// hostname but a different address
+// address and hostname
 func (r *agentGroup) hasPeer(addr, hostname string) bool {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	for existingAddr, existingHostname := range r.hostnames {
+		if existingHostname == hostname && existingAddr == addr {
```
r0mant, line 615: Hmm, looking at how this is used, won't this allow the case where you launch "gravity join" twice on the same host?

tool/gravity/cli/system.go
```diff
@@ -47,6 +51,38 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+func systemClusterInfo(env *localenv.LocalEnvironment) error {
```
r0mant, line 54: What's this used for?

a-palchikov: This wraps `kubectl cluster-info dump --output-directory=/path/to/dir` [here](https://github.com/gravitational/gravity/blob/dmitri/7.0.x/forward-port-6.3.x/lib/report/k8s.go#L41)

lib/ops/opsservice/agents.go
a-palchikov: Yes, but it should not be possible since the port will be in-use by only one?

a-palchikov:
Superseded by #1361

Issue #1346, opened 2020-04-09
[BUG] Upgrade doesn't adhere to K8s version skew policy


r0mant:
**Describe the bug**

When the upgrade runs, it follows the following plan when upgrading the first master (say, node-1):

1. Leader is moved to another master: from node-1 to node-2.
2. First master node-1 is upgraded.
3. Leader is moved back to the first master: from node-2 to node-1.

Between steps 2 and 3 we have a situation when a new kubelet/kube-proxy/etc. on node-1 are talking to old apiserver on node-2 before the leader moved back.

This is unsupported scenario according to Kubernetes [version skew policy](https://kubernetes.io/docs/setup/release/version-skew-policy/):

> kubelet must not be newer than kube-apiserver, and may be up to two minor versions older.

In some cases it results in kubelet being unable to report back to apiserver (for example, it tries to register new labels which old apiserver doesn't support) which leads to node degraded status and the upgrade is unable to proceed b/c we wait for a healthy status after master upgrade.

This for example happens when upgrading from 6.2 to 6.3 on a cluster with AWS integration b/c some kubelet labels changed from `failure-domain.beta.kubernetes.io` to `topology.kubernetes.io` and when new kubelet attempts to set them the old apiserver doesn't allow it.

We need to come up with a way to perform upgrade so the new apiserver becomes active right away.

**To Reproduce**

Just run upgrade that includes K8s version change e.g. b/w 6.2 and 6.3.

**Expected behavior**

Upgrade should follow K8s version skew policy.

**Logs**

**Environment (please complete the following information):**
 - OS [e.g. Redhat 7.4]:
 - Gravity [e.g. 5.5.4]:
 - Platform [e.g. Vmware, AWS]:

**Additional context**

PR #1410, opened 2020-04-20
reorder election and health checks

knisbet:
Updates #1346 

This PR slightly reorganizes the upgrade process for a node to be more compatible with the kubernetes version skew policy. As outlined in #1346 the way we rolling restart planet previously allows an older kubelet, kube-proxy, etc to try and work with an older version of the API. This doesn't fundamentally change that behaviour, but instead continues the upgrade to the point where the API fails over to the new version, before health checking the upgrade.

Previous (on multi-node cluster):
1. Upgrade planet
2. Health check node
3. Failover apiserver to new version on upgraded planet

New Behaviour:
1. Upgrade planet
2. Wait for node to be online, but don't health check
3. Failover apiserver to upgraded node / version
4. Healthcheck the node after failover

The item to consider, is to block the upgrade while planet starts, I'm currently using connectivity to planet-agent for this heuristic. So poll the status endpoint of planet until we get some sort of response, but actually wait for health later after we do the failover. Not sure if anyone has any better ideas for heuristics on waiting for planet to start before proceeding.


lib/update/cluster/builder_test.go
```diff
@@ -489,6 +473,28 @@ func (r *params) leaderMasterPhase() storage.OperationPhase {
 				},
 				Requires: []string{t("/masters/%v/drain")},
 			},
+			{
+				ID:          t("/masters/%[1]v/elect"),
```
a-palchikov, line 477: nit: not sure why it was using a positional placeholder with just a single value:
```go
"/masters/%v/elect"
```

lib/update/system/system.go
```diff
@@ -883,31 +882,14 @@ func ensureServiceRunning(servicePackage loc.Locator) error {
 	return trace.Wrap(err)
 }
 
-func waitNodeStatus(ctx context.Context) (err error) {
+// waitForAgentRunning blocks until able to successfully connect to the local agent, but ignores whether the node is
+// healthy
+func waitForAgentRunning(ctx context.Context) error {
```
a-palchikov, line 887: Is this used anywhere? Looks like the code moved to `lib/status`.

a-palchikov (approved):

lib/update/cluster/builder.go
```diff
@@ -317,24 +317,40 @@ func (r phaseBuilder) masters(leadMaster storage.UpdateServer, otherMasters []st
...
+
+		// election - force failover to first upgraded master
+		electionChanges := electionChanges{
+			description: fmt.Sprintf("Make node %q Kubernetes leader", leadMaster.Hostname),
+			enable:      serversToStorage(leadMaster),
+			disable:     serversToStorage(otherMasters...),
+		}
+
+		node.AddSequential(r.commonNode(leadMaster, leadMaster, supportsTaints,
+			waitsForEndpoints(len(otherMasters) == 0), electionChanges)...)
```
r0mant, line 337: Nit: Since you moved this under `if len(otherMasters) != 0`, here and below you can just pass bool flags which will improve readability a little bit, i.e. `waitForEndpoints(false)` here and `waitForEndpoints(true)` below.

r0mant (approved):

PR #1426, opened 2020-04-22
Added note on oidc re auth0

stevenGravy:
Note on turning off OIDC Conformant settings that prevents auth0 connectors from populating claims.  Otherwise a warning will show like "Error validating callback: unable to map claims to role for connector:" in the log.

### DIFF
--- a/docs/7.x/config.md
+++ b/docs/7.x/config.md
@@ -13,7 +13,7 @@ Resource Name             | Resource Description
 `role`                    | User role (Enterprise only)
 `user`                    | Cluster user
 `token`                   | User tokens such as API keys
-`logforwarder`            | Forwarding logs to a remote rsyslog server
+`logforwarder`            | Forwarding logs to a remote rsyslog server!
 `trusted_cluster`         | Managing access to remote Gravity Hubs (Enterprise Only)
 `endpoints`               | Gravity Hub endpoints for user and Cluster traffic (Enterprise Only)
 `cluster_auth_preference` | Cluster authentication settings such as second-factor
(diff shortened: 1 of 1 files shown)


docs/7.x/config.md
```diff
@@ -404,6 +404,8 @@ spec:
     - {% raw %}{claim: "roles", value: "gravitational/admins", roles: ["@teleadmin"]}{% endraw %}
 ```
 
+!!! note For Auth0 the OIDC Conformant setting should be off in Advanced Setting -> OAuth or Claims will not populate properly
```
r0mant, line 407: Nit: Could you add a dot at the end of the sentence please?

Also, shouldn't the note be formatted like this:

```
!!! note:
  <note text>
```

Disregard if your way is displayed properly as well.

r0mant (approved):

docs/7.x/config.md
stevenGravy: think it would still work but modified to make it uniform with other notes in the doc

docs/7.x/config.md
```diff
@@ -13,7 +13,7 @@ Resource Name             | Resource Description
 `role`                    | User role (Enterprise only)
 `user`                    | Cluster user
 `token`                   | User tokens such as API keys
-`logforwarder`            | Forwarding logs to a remote rsyslog server
+`logforwarder`            | Forwarding logs to a remote rsyslog server!
```
a-palchikov, line 16: Why an exclamation mark here?

stevenGravy: that was an error, thanks for catching @a-palchikov !  it has been removed

a-palchikov (approved):

PR #1427, opened 2020-04-22
[5.5.x] use encryption key for serf

a-palchikov:
Uses install operation ID (as a simple UID) as an encryption key for the serf cluster to prevent unrelated nodes from joining the cluster.

Requires https://github.com/gravitational/planet/pull/605.
Fixes https://github.com/gravitational/gravity.e/issues/3372.
Design doc: https://docs.google.com/document/d/10VV-RrGdf7V3AKHAhCaCnuCuMBr2zDXKyaZjGi5Pahk/edit#.


r0mant (commented):
Sorry, this is without looking at the code much yet - but how are upgrades handled? I.e. during rolling upgrade a new serf client starts with the encryption key - won't it fail to join the cluster that's been setup without encryption?

a-palchikov:
Moved discussion regarding upgrade to the attached google doc.

a-palchikov:
Closing until further notice.

PR #1430, opened 2020-04-22
[5.5.x] wait for agent to leave cluster

a-palchikov:
During expand, after operation completion, wait for agent to exit the cluster so the cluster controller does not re-attempt to reconnect.
Also, do not remove the expand token as the agent will have to validated when it leaves.

### DIFF
--- a/lib/rpc/server/agent.go
+++ b/lib/rpc/server/agent.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/gravitational/gravity/lib/defaults"
 	pb "github.com/gravitational/gravity/lib/rpc/proto"
 	"github.com/gravitational/gravity/lib/state"
 	"github.com/gravitational/gravity/lib/storage"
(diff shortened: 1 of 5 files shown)


knisbet (approved):

lib/rpc/server/agent.go
```diff
@@ -136,7 +137,15 @@ func (srv *agentServer) GetCurrentTime(ctx context.Context, _ *types.Empty) (*ty
 // Shutdown requests agent to shut down
 func (srv *agentServer) Shutdown(ctx context.Context, _ *types.Empty) (*types.Empty, error) {
 	srv.Info("Shutdown.")
-	go srv.Stop(ctx)
+	go func() {
+		// Create a separate context from the parent one since the parent
+		// context is canceled once the handler has returned
+		ctx, cancel := context.WithTimeout(context.Background(), defaults.ShutdownTimeout)
+		if err := srv.Stop(ctx); err != nil {
+			srv.Warnf("Failed to shutdown: %v.", err)
```
r0mant, line 145: srv.WithError(err).Warn()?

r0mant (approved):
Do not forget to restore the join token ttl as we discussed pls.

lib/rpc/server/agent.go
a-palchikov: This is not critical and will likely be spam.

PR #1434, opened 2020-04-23
Fix some join token issues

r0mant:
Fix a couple of issues surfaced after recent expand improvements:

* Bring back ttl for the join token created for an expand operation as some logic depends on it.
* Dynamically determine teleport auth tokens when the process starts instead of hardcoding them in teleport configuration.

Refs https://github.com/gravitational/gravity/issues/1445.

### DIFF
--- a/lib/ops/opsservice/install.go
+++ b/lib/ops/opsservice/install.go
@@ -885,6 +885,10 @@ func (s *site) newProvisioningToken(operation ops.SiteOperation) (token string, 
 		OperationID: operation.ID,
 		UserEmail:   agentUser.GetName(),
 	}
+	if operation.Type == ops.OperationExpand {
+		// Set a TTL for expand provisioning token.
+		tokenRequest.Expires = s.clock().UtcNow().Add(24 * time.Hour)
+	}
 	_, err = s.users().CreateProvisioningToken(tokenRequest)
 	if err != nil && !trace.IsAlreadyExists(err) {
 		log.WithError(err).Warn("Failed to create provisioning token.")
--- a/lib/ops/opsservice/configure.go
+++ b/lib/ops/opsservice/configure.go
@@ -1109,16 +1109,9 @@ func (s *site) getTeleportMasterConfig(ctx *operationContext, configPackage loc.
 	fileConf.AdvertiseIP = advertiseIP.String()
 	fileConf.Global.NodeName = master.FQDN(s.domainName)
 
-	joinToken, err := s.service.GetExpandToken(s.key)
-	if err != nil {
-		return nil, trace.Wrap(err)
-	}
-
 	// turn on auth service
 	fileConf.Auth.EnabledFlag = "yes"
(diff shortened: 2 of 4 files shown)


lib/ops/opsservice/install.go
```diff
@@ -885,6 +885,10 @@ func (s *site) newProvisioningToken(operation ops.SiteOperation) (token string,
 		OperationID: operation.ID,
 		UserEmail:   agentUser.GetName(),
 	}
+	if operation.Type == ops.OperationExpand {
+		// Set a TTL for expand provisioning token.
+		tokenRequest.Expires = s.clock().UtcNow().Add(24 * time.Hour)
```
a-palchikov, line 890: Can elaborate on why the TTL is required?

r0mant: Yeah, when we generate Teleport node configuration (during install or join or upgrade), it uses a join token to authenticate with auth server: https://github.com/gravitational/gravity/blob/6.3.12/lib/ops/opsservice/configure.go#L1183.

The GetExpandToken function picks any non-expiring join token: https://github.com/gravitational/gravity/blob/6.3.12/lib/ops/opsservice/service.go#L440.

Since we removed TTL here, it means that this new token can now also be returned by GetExpandToken - i.e. new Teleport node can try to authenticate using it, but auth servers know nothing about this new token unless they are restarted (since they initialize static tokens upon startup).

I considered multiple ways to fix this (e.g. introduce another type of token specifically for teleport nodes instead of using expand tokens), but thought that just bringing the ttl back for now would be less risky.

a-palchikov: Hmm, in this case we could introduce a new expand token type for the static (i.e. the one used for teleport auth) token or the operation expand token (i.e. the one used by agents to form the operation agent group) - either way, this could be a simple type change or am I missing anything?

r0mant: @a-palchikov Yeah, that's exactly the alternate approach I described in my last paragraph - have another token type specific to teleport nodes. I can certainly rework this if you feel strongly about it (although as I recall there wasn't a terribly good reason _not_ to have it - I think join prov tokens were meant to expire after some time by design)- pretty much the sole reason I opted to restore the ttl here is to reduce the risk of another regression in a patch release. Last couple 6.3 releases have been rough.

What do you think about this:

* Here, in 6.3, we just restore ttl for these tokens - as I was writing the above, I think we actually made a mistake removing it. Not just because this bug unsurfaced, but these expand tokens were originally meant to be temporary, accumulating these tokens for each operation is no good.
* When porting this to 7.0/master, in addition to also restoring the ttl, I will also make a change to use specific "teleport" tokens for teleport nodes - the approach we discussed above.

This way we'll get this fixed in 6.3 with least blood and reduce the risk, and newer version will get a "proper" fix.

a-palchikov (approved):

vli63127:
Hi, I'm also having issues to join the cluster while the cluster status shows: expanding. Do you have a workaround without upgrading the version? Thanks.

PR #1441, opened 2020-04-23
Update vendored helm to 2.15.2.

r0mant:


Gopkg.toml
```diff
@@ -101,7 +101,7 @@ ignored = [
 
 [[constraint]]
   name = "k8s.io/helm"
-  branch = "release-2.15"
+  version = "=v2.15.2"
```
a-palchikov, line 104: I wonder why the branch does not work - isn't it supposed to point to latest `2.15.x` tag?

a-palchikov (approved):

knisbet (approved):

Gopkg.toml
r0mant: As discussed offline, it does work, I just prefer to pin specific version to avoid surprises with accidental version updates.

Issue #1445, opened 2020-04-24
[BUG] Teleport on joined nodes fails to join cluster


r0mant:
**Describe the bug**

There was recently implemented a behavioral change that removed a TTL from the join token generated during an expand operation. As innocent as it may seem, it resulted into a pretty bad regression causing Teleport on all new nodes joined to cluster running this version failing to join the cluster.

This means that further upgrade attempts of such cluster will fail b/c Teleport is not accessible on those nodes.

The comment on this PR gives more details as to why this happened:

https://github.com/gravitational/gravity/pull/1434#discussion_r413674898

**To Reproduce**

1. Install 5.5.40.
2. Join a node.
3. Observe Teleport on the new node failing to join.

**Expected behavior**

- [x] Teleport node should join fine.
  - [x] 5.5 (release in 5.5.41)
  - [x] 6.1 (wasn't affected)
  - [x] 6.3 (released in 6.3.13)
  - [x] 7.0 (released in 7.0.3)
  - [x] master
- [x] If Teleport node fails to join, the join operation should also fail instead of succeeding.
- [x] Provide a way to recover affected clusters somehow.
- [x] gravity status should detect that Teleport on a node is down.

**Logs**

**Environment (please complete the following information):**
 - OS [e.g. Redhat 7.4]:
 - Gravity [e.g. 5.5.4]: 5.5.40 and other versions where we removed TTL from join token
 - Platform [e.g. Vmware, AWS]:

**Additional context**

PR #1446, opened 2020-04-24
Add internal commands to fix teleport tokens.

r0mant:
### Description

This PR adds a few internal commands that should help us recover clusters that may have been affected by https://github.com/gravitational/gravity/issues/1445, in case we encounter any.

The background of the issue is Teleport on the nodes joined to a 5.5.40 cluster fails to join the cluster which means further upgrades will fail (and node isn't accessible via Teleport).

It adds the following commands.

#### Display teleport config from package

```
$ gravity system teleport show-config --package=<config-package>
```

#### Update auth token used by Teleport node

```
$ gravity system teleport set-node-token --package=<config-package> --token=token
```

#### Update auth tokens used by Teleport auth server

```
$ gravity system teleport set-master-tokens --package=<config-package> --token=token1 --token=token2
```

### Testing done

First I've done the following setup to reproduce the issue:

* Install single-node 5.5.40 cluster and join a node to it.
* Verify Teleport on the joined node isn't able to join due to bad token.

Then I fixed the Teleport node:

* Used above set-node-token command to update config to use join token.
* Restarted Teleport service and verified it joined successfully.

Also tested the Teleport master part:

* Set different node token via set-node-token command.
* Cleared /var/lib/gravity/teleport, restarted and verified it's failing to join.
* Used set-master-tokens to update auth token on master to include the one I used above.
* Restarted gravity-site and verified the node also joined now.

### DIFF
--- a/tool/gravity/cli/teleport.go
+++ b/tool/gravity/cli/teleport.go
@@ -21,9 +21,11 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/gravitational/gravity/lib/loc"
 	"github.com/gravitational/gravity/lib/localenv"
 	"github.com/gravitational/gravity/lib/pack"
+	"github.com/gravitational/gravity/lib/process"
 
 	"github.com/gravitational/teleport/lib/config"
 	"github.com/gravitational/teleport/lib/defaults"
@@ -66,7 +68,7 @@ func updateTeleportMasterTokens(env *localenv.LocalEnvironment, packageName stri
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	fmt.Println("Teleport master auth token updated. Please restart gravity-site using 'kubectl -nkube-system delete pods -lapp=gravity-site'.")
+	fmt.Println("Teleport master auth token updated. Please restart gravity-site using 'kubectl -nkube-system delete pods -lapp=gravity-site'")
 	return nil
 }
 
@@ -84,7 +86,7 @@ func updateTeleportNodeToken(env *localenv.LocalEnvironment, packageName, token 
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	fmt.Println("Teleport node auth token updated. Please restart Teleport service using 'sudo systemctl restart *teleport*'.")
+	fmt.Println("Teleport node auth token updated. Please restart Teleport service using 'sudo systemctl restart *teleport*'")
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


tool/gravity/cli/teleport.go
```diff
@@ -0,0 +1,156 @@
...
+func showTeleportConfig(env *localenv.LocalEnvironment, packageName string) error {
+	locators, err := getTeleportLocators(env, packageName)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	config, err := readTeleportFileConfig(locators.configReader)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	configBytes, err := yaml.Marshal(config)
```
a-palchikov, line 44: Can we add `--output=<json|yaml|text>` also here or it does not make sense?

tool/gravity/cli/teleport.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	for _, token := range tokens {
+		fileConfig.Auth.StaticTokens = append(fileConfig.Auth.StaticTokens, config.StaticToken(
+			fmt.Sprintf("node:%v", token)))
+	}
+	err = saveTeleportFileConfig(env.Packages, fileConfig, locators.teleportLocator, locators.configLocator, locators.configEnvelope.RuntimeLabels)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	fmt.Println("Teleport master auth token updated. Please restart gravity-site using 'kubectl -nkube-system delete pods -lapp=gravity-site'.")
```
a-palchikov, line 69: https://github.com/gravitational/gravity/blob/version/5.5.x/tool/gravity/cli/rotate.go#L90

tool/gravity/cli/teleport.go
```diff
@@ -0,0 +1,156 @@
...
+	fileConfig, err := readTeleportFileConfig(locators.configReader)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	fileConfig.AuthToken = token
+	err = saveTeleportFileConfig(env.Packages, fileConfig, locators.teleportLocator, locators.configLocator, locators.configEnvelope.RuntimeLabels)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	fmt.Println("Teleport node auth token updated. Please restart Teleport service using 'sudo systemctl restart *teleport*'.")
```
a-palchikov, line 87: ditto - see above.

tool/gravity/cli/teleport.go
r0mant: It already prints it as a yaml.

tool/gravity/cli/teleport.go
r0mant: I'd rather keep this atomic, so these methods are only for updating config. Users can restart teleport/gravity-site themselves. I was planning on writing a KB article on how to use these if needed.

a-palchikov (approved):

PR #1450, opened 2020-04-24
Add pull request template.

r0mant:
As discussed today, add pull request template. Should we add anything else to it?

Testing on my own repo it also looks like this should only go to the master branch, so version branches don't require it.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,11 @@
+## Description
+<!--Provide high-level overview of what the PR is for.-->
+
+## Implementation
+<!--Add any relevant implementation details that might help the reviewers.-->
+
+## Linked tickets
+<!--Mention Github issues that are closed or referenced by this PR.-->
+
+## Testing done
+<!--Explain what kind of testing this PR underwent.-->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,11 +1,40 @@
 ## Description
-<!--Provide high-level overview of what the PR is for.-->
+<!--Provide high-level overview of what the change is for.-->
+
+## Type of change
+<!--Select all options appropriate for this change.-->
+
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] Regression fix (non-breaking change which fixes a regression)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- [ ] Internal change (not necessarily a bug fix or a new feature)
+- [ ] This change has a user-facing impact
+
+## Linked tickets and other PRs
+<!--Keep only relevant items.-->
+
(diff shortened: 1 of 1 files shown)


knisbet:
I would consider structuring the template more as a checklist or including checklists as part of the request to help communicate the status of a particular PR. May also help with creating a standard that easier to evolve over time.


Example:
- https://github.com/coreos/flannel/blob/master/.github/PULL_REQUEST_TEMPLATE.md

Here's some references for the type of information captured in other OSS projects:
- https://github.com/kubernetes/kubernetes/blob/master/.github/PULL_REQUEST_TEMPLATE.md
- https://embeddedartistry.com/blog/2017/08/04/a-github-pull-request-template-for-your-projects/

.github/pull_request_template.md
```diff
@@ -0,0 +1,11 @@
+## Description
+<!--Provide high-level overview of what the PR is for.-->
+
+## Implementation
+<!--Add any relevant implementation details that might help the reviewers.-->
```
wadells, line 5: For my edification, could you link to an PR that is a good example of providing implementation details?  I don't presently understand what exactly should go in this section (or if it should be present on most/all PRs)

wadells (commented):
> Testing on my own repo it also looks like this should only go to the master branch, so version branches don't require it.

Under the theory of "Show, don't tell" will you please link to a PR from this template in your repo?

r0mant:
@knisbet @wadells Updated to include more info and a checklist. Here's an example of a PR created using this template: https://github.com/r0mant/emacs.d/pull/5. Note that the PR was not made to a master branch (roman/test1<--roman/test2) but still uses the template like I mentioned.

.github/pull_request_template.md
r0mant: Sure, this is basically to have a chance to explain any implementation details that you think may not be immediately obvious to the reviewer, or help understand the high-level approach to implementation of larger changesets, etc. For example, look at https://github.com/gravitational/gravity/pull/1317 (implementation section is sort of woven into the description there) or https://github.com/gravitational/gravity/pull/1084 or https://github.com/gravitational/gravity/pull/898, etc.

wadells (approved):
Overall, I think this feels a bit "heavy".  However, I also I think we should merge it and give it a try for a week or two.  We can always tweak it later to add missing parts or remove sections that aren't regularly providing value.

r0mant:
Here's an example of "real" PR using this template: https://github.com/gravitational/gravity/pull/1453. I agree it feels a bit heavy. Maybe instead of checkboxes we could use bullet points in some sections. Or utilize Github labels for some parts instead?

knisbet:
I think for "Type of change" should also be remove those that don't apply.

a-palchikov (approved):

UnderGreen (approved):

bernardjkim (approved):

PR #1453, opened 2020-04-24
Fix join token issue.

r0mant:
## Description

This PR addresses the recent join token related regression described in https://github.com/gravitational/gravity/issues/1445 in branch 7.0.

It does 2 things:

* Brings back the TTL for a join token.
* Instead of hardcoding auth tokens in Teleport config during install/upgrade, Teleport auth server (gravity-site) now reads them from the backend during startup.

## Type of change

- [x] Regression fix (non-breaking change which fixes a regression)

* Refs #1445
* Ports #1434

## TODOs

- [x] Self-review the change
- [x] Perform manual testing

## Implementation

Forward-ported the same changes from 6.3.

## Testing done

Install a single node cluster:

```
node-1:~/installer$ sudo ./gravity install --advertise-addr=192.168.99.102
```

Add a node to it:

```
node-2:~$ sudo ./gravity join 192.168.99.102 --token=bf75c3d3015132bca18286001de50e4f --role=knode --advertise-addr=192.168.99.103
```

Verify Teleport on the new node connects successfully:

```
node-2:~$ journalctl -u *teleport* -f
..
Apr 24 22:09:16 node-2 gravity[5707]: INFO [PROC:1]    The new service has started successfully. Starting syncing rotation status with period 1m30s. service/connect.go:379
```

### DIFF
--- a/lib/ops/opsservice/install.go
+++ b/lib/ops/opsservice/install.go
@@ -886,6 +886,10 @@ func (s *site) newProvisioningToken(operation ops.SiteOperation) (token string, 
 		OperationID: operation.ID,
 		UserEmail:   agentUser.GetName(),
 	}
+	if operation.Type == ops.OperationExpand {
+		// Set a TTL for expand provisioning token.
+		tokenRequest.Expires = s.clock().UtcNow().Add(24 * time.Hour)
+	}
 	_, err = s.users().CreateProvisioningToken(tokenRequest)
 	if err != nil && !trace.IsAlreadyExists(err) {
 		log.WithError(err).Warn("Failed to create provisioning token.")
--- a/lib/ops/opsservice/install.go
+++ b/lib/ops/opsservice/install.go
@@ -887,7 +887,11 @@ func (s *site) newProvisioningToken(operation ops.SiteOperation) (token string, 
 		UserEmail:   agentUser.GetName(),
 	}
 	if operation.Type == ops.OperationExpand {
-		// Set a TTL for expand provisioning token.
+		// TODO(r0mant): Due to current implementation this TTL is required to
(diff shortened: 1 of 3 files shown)


lib/ops/opsservice/install.go
```diff
@@ -886,6 +886,10 @@ func (s *site) newProvisioningToken(operation ops.SiteOperation) (token string,
 		OperationID: operation.ID,
 		UserEmail:   agentUser.GetName(),
 	}
+	if operation.Type == ops.OperationExpand {
+		// Set a TTL for expand provisioning token.
```
knisbet, line 890: Nit, comment on why, not what. So in the comment, link to the issue or problem or regression, and say the TTL is set because this is required by teleport, etc, otherwise it would be very easy to come along and make the same mistake again.

lib/ops/opsservice/install.go
```diff
@@ -886,6 +886,10 @@ func (s *site) newProvisioningToken(operation ops.SiteOperation) (token string,
...
+		tokenRequest.Expires = s.clock().UtcNow().Add(24 * time.Hour)
```
knisbet, line 891: I'm curious as well on why 24 hours? I'm not all that familiar with this code, so without tracing it all through, what happens after the 24 hours?

knisbet (approved):

lib/ops/opsservice/install.go
r0mant: Nothing, really. Before we removed this the ttl was 1 hour which was a bit short b/c if expand operation was for some reason created but not started/completed within 1 hour, the token would become invalid. 24 hours just gives more breathing room.

Issue #1469, opened 2020-04-28
Cluster controller fails to pull packages


a-palchikov:
During normal operation, each `gravity-site` Pod runs a blob synchronizer that periodically pulls new blobs it does not have locally.
Due to the way the synchronzier is implemented, it will stop pulling at the first failed attempt to pull.
One of the issues it might encounter is a stale peer - i.e. a master node that was previously part of the cluster but left leaving its peer state intact. If a file object (blob) was only available on that peer, it will prevent the controller loop from advancing and depending on the order of blobs might prevent it from pulling all new blobs.

PR #1472, opened 2020-04-29
[6.3.x] clean up object peers and lost objects during upgrade

a-palchikov:
## Description

 * Implements stale object peer cleanup as a preliminary step to using the cluster package service.
 * Changes the implementation of the cluster controller's package blob synchronizer loop to work on best-effort basis - i.e. attempting to fetch as many blobs as possible even if some blobs are not accessible.
 * Adds a step to the shrink operation that removes the server's peer reference from the database.

## Type of change

* Updates https://github.com/gravitational/gravity/issues/1469

## Testing done

Implemented unit tests for the new upgrade step that cleans up peer state.
Tested on a two-node cluster with manually injected invalid peer state (peer + object that only had the invalid peer in its peer list).

## Additional Information
A somewhat better solution would probably be adding the explicit time-to-live to peer records, which will mean:
  * nodes won't have to clean up upon leaving
  * no need for redundant `last_heartbeat` attribute
  * package blob synchronizer loop will be to consider peers as accessible based on record existence
  * each node will have to have an explicit heartbeat loop to bump TTL

### DIFF
--- a/lib/blob/cluster/cluster.go
+++ b/lib/blob/cluster/cluster.go
@@ -232,7 +232,7 @@ func (c *cluster) fetchNewObjects() error {
 		c.Infof("Found missing object %v.", hash)
 		err = c.fetchObject(hash)
 		if err != nil {
-			c.Warningf("Failed to fetch object(%v) %v.", hash, trace.DebugReport(err))
+			c.WithError(err).Warn("Failed to fetch object(%v).", hash)
 			return trace.Wrap(err)
 		}
 	}
@@ -244,7 +244,7 @@ func (c *cluster) fetchObject(hash string) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	peers, err := c.getPeers(peerIDs)
+	peers, err := c.getPeers(peerIDs...)
 	if err != nil {
 		return trace.Wrap(err)
 	}
@@ -256,13 +256,13 @@ func (c *cluster) fetchObject(hash string) error {
 	for _, p := range peers {
 		objects, err := c.getObjects(p)
 		if err != nil {
-			c.Errorf("Failure to fetch %v from %v: %v.", hash, p, trace.DebugReport(err))
+			c.WithError(err).Errorf("Failure to fetch %v from %v.", hash, p)
 			errors = append(errors, err)
 			continue
 		}
 		f, err := objects.OpenBLOB(hash)
 		if err != nil {
-			c.Errorf("Failure to fetch %v from %v: %v.", hash, p, trace.DebugReport(err))
+			c.WithError(err).Errorf("Failure to fetch %v from %v.", hash, p)
 			errors = append(errors, err)
 			continue
 		}
@@ -285,8 +285,7 @@ func (c *cluster) fetchObject(hash string) error {
 
 // WriteBLOB writes object to the storage, returns object envelope
 func (c *cluster) WriteBLOB(data io.Reader) (*blob.Envelope, error) {
-	// get peers
-	peers, err := c.getPeers(nil)
+	peers, err := c.getPeers()
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
@@ -302,7 +301,7 @@ func (c *cluster) WriteBLOB(data io.Reader) (*blob.Envelope, error) {
 
 	successWrites := []string{c.ID}
 	if len(successWrites) >= c.WriteFactor {
-		c.Debugf("Got enough success writes for %v %v.", envelope.SHA512, successWrites)
+		c.Debugf("Got enough success writes (%v) for %v.", successWrites, envelope.SHA512)
 		err := c.Backend.UpsertObjectPeers(envelope.SHA512, successWrites, 0)
 		if err != nil {
 			return nil, trace.Wrap(err, "failed to write object metadata %v", err)
@@ -326,19 +325,19 @@ func (c *cluster) WriteBLOB(data io.Reader) (*blob.Envelope, error) {
 		}
 		peerClient, err := c.GetPeer(p)
 		if err != nil {
-			c.Infof("%v returned error: %v", p, err)
+			c.Warnf("failed to create clien to peer %v: %v", p, err)
 			errors = append(errors, err)
 			continue
 		}
 		_, err = peerClient.WriteBLOB(f)
 		if err != nil {
-			c.Infof("%v returned error: %v", p, err)
+			c.Warnf("%v returned error: %v", p, err)
 			errors = append(errors, err)
 			continue
 		}
 		successWrites = append(successWrites, p.ID)
 		if len(successWrites) >= c.WriteFactor {
-			c.Debugf("Got enough success writes for %v %v.", envelope.SHA512, successWrites)
+			c.Debugf("Got enough success writes (%v) for %v.", successWrites, envelope.SHA512)
 			err := c.Backend.UpsertObjectPeers(envelope.SHA512, successWrites, 0)
 			if err != nil {
 				return nil, trace.Wrap(err, "failed to write object metadata %v", err)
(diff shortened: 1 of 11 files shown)


lib/blob/cluster/cluster.go
```diff
@@ -214,6 +214,10 @@ func (c *cluster) purgeDeletedObjects() error {
 	return nil
 }
 
+// fetchNewObjects downloads the objects that has been recorded in the database
```
r0mant, line 217: have been

lib/blob/cluster/cluster.go
```diff
@@ -228,23 +232,24 @@ func (c *cluster) fetchNewObjects() error {
...
 		}
 	}
+	var errors []error
 	for _, hash := range missingObjects {
 		c.Infof("Found missing object %v.", hash)
 		err = c.fetchObject(hash)
 		if err != nil {
-			c.Warningf("Failed to fetch object(%v) %v.", hash, trace.DebugReport(err))
-			return trace.Wrap(err)
+			c.WithError(err).Warnf("Failed to fetch object(%v).", hash)
```
r0mant, line 240: Nit: Use field for the hash? `c.WithError(err).WithField("hash", hash).Warn()`

Also - do you know if there's some other information about the missing object available here (in addition to hash) that we can add to the log as well to make it more useful in determining which package this refers to? Countless times we've seen "failed to fetch object(qweasdzxc)" in the logs without a slightest idea of what package this is. It's probably not available here as blob is low level but I'm just wondering.

lib/blob/cluster/cluster.go
```diff
@@ -256,13 +261,13 @@ func (c *cluster) fetchObject(hash string) error {
 	for _, p := range peers {
 		objects, err := c.getObjects(p)
 		if err != nil {
-			c.Errorf("Failure to fetch %v from %v: %v.", hash, p, trace.DebugReport(err))
+			c.WithError(err).Errorf("Failure to fetch %v from %v.", hash, p)
```
r0mant, line 264: Nit: Not a big deal but I'd consider updating this and other logs to use fields for hash/peer.

lib/blob/cluster/cluster.go
```diff
@@ -326,19 +330,19 @@ func (c *cluster) WriteBLOB(data io.Reader) (*blob.Envelope, error) {
 		}
 		peerClient, err := c.GetPeer(p)
 		if err != nil {
-			c.Infof("%v returned error: %v", p, err)
+			c.Warnf("failed to create clien to peer %v: %v", p, err)
```
r0mant, line 333: **F**ailed, clien**t**

lib/blob/cluster/cluster.go
```diff
@@ -347,7 +351,10 @@ func (c *cluster) WriteBLOB(data io.Reader) (*blob.Envelope, error) {
 		}
 	}
 
-	return nil, trace.Wrap(trace.NewAggregate(errors...), "not enough successfull writes")
+	if len(errors) == 0 {
+		return nil, trace.NotFound("not enough peers")
+	}
+	return nil, trace.Wrap(trace.NewAggregate(errors...), "not enough successful writes")
```
r0mant, line 357: Include how many successful writes we got in the message? I think in our case it will always be 0 though (since we set write factor to 1)?

lib/update/cluster/phases/init.go
```diff
@@ -491,3 +503,126 @@ func (p *updatePhaseInit) removeConfiguredPackages() error {
...
+	var errors []error
+	for _, hash := range objects {
+		peers, err := r.p.GetObjectPeers(hash)
+		if err != nil && !trace.IsNotFound(err) {
+			return trace.Wrap(err)
+		}
+		if !r.invalidPeers(peers...) {
+			continue
+		}
+		r.log.WithField("hash", hash).Info("Remove lost object.")
```
r0mant, line 536: I'd say "Remove object without peers" and log the peers as well.

lib/update/cluster/phases/init.go
```diff
@@ -491,3 +503,126 @@ func (p *updatePhaseInit) removeConfiguredPackages() error {
...
+	}
+	if err := r.removeInvalidObjectPeers(); err != nil {
+		return trace.Wrap(err)
+	}
+	return nil
+}
+
+// removeLostObjects removes the object hashes from the state database
+// that only have peers that are no longer in the cluster
+func (r peerCleaner) removeLostObjects() error {
```
r0mant, line 522: Is there a chance it will nuke something it shouldn't? Is this a change that's necessary to recover some cluster? I'd be wary to deal directly with blobs in the object store esp. on every upgrade. What if there's a blob on a master node that doesn't have valid peers but is still used somehow?

lib/ops/opsservice/shrink.go
```diff
@@ -307,8 +307,16 @@ func (s *site) shrinkOperationStart(ctx *operationContext) (err error) {
 		if !force {
 			return trace.Wrap(err, "failed to remove the node from the cluster")
 		}
-		ctx.Warningf("Failed to remove node %q from the cluster, force continue: %v.",
-			serverName, trace.DebugReport(err))
+		ctx.WithError(err).Warnf("Failed to remove node %q from the cluster, force continue.",
+			serverName)
+	}
+
+	if err = s.removeObjectPeer(server.ObjectPeerID()); err != nil && !trace.IsNotFound(err) {
```
r0mant, line 314: Isn't gravity-site still running at this point (when removing a master)? So there's a chance it will just resync the object and put itself as a peer again?

r0mant (commented):
Blob changes make sense, but I feel a little uneasy about removing blobs directly from the store during every upgrade.

lib/blob/cluster/cluster.go
a-palchikov: No, but we can add `metadata` field so there's a way to trace a blob to a specific package (anywhere in the cluster).

lib/update/cluster/phases/init.go
a-palchikov: That's a good point and it's something I've been thinking with the alternative approach of adding TTL to a peer. We can treat the peer list in the object record as sort of history and ignore peers that aren't valid and, by extension, not attempting to pull objects with all invalid or empty peer list.

lib/update/cluster/phases/init.go
a-palchikov: I opted out of object deletion.

lib/blob/cluster/cluster.go
```diff
@@ -429,9 +466,13 @@ func (c *cluster) getPeers(ids []string) ([]storage.Peer, error) {
 			if !matchPeer(ids, p.ID) {
 				continue
 			}
-			// if it's last heartbeat is older than the acceptance time frame
+			// if its last heartbeat is older than the acceptance time frame
 			if c.Clock.Now().UTC().Sub(p.LastHeartbeat) > missedWindow {
-				c.Infof("Excluding %v, missed heartbeat window %v, last heartbeat: %v.", p.ID, missedWindow, p.LastHeartbeat)
+				c.WithFields(logrus.Fields{
+					"peer":           p.String(),
+					"time-window":         missedWindow,
```
r0mant, line 473: Nit: Indentation on this line is messed up.

r0mant (approved):

lib/blob/cluster/cluster.go
```diff
@@ -272,21 +295,21 @@ func (c *cluster) fetchObject(hash string) error {
 			errors = append(errors, err)
 			continue
 		}
-		c.Infof("Successfully fetched %v from %v.", envelope, p)
+		logger.WithField("package", envelope).Error("Successfully fetched object from peer.")
```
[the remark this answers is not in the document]
a-palchikov: My bad.

PR #1477, opened 2020-04-29
Add Teleport node status checks

r0mant:
## Description

This PR implements a few changes to better handle Teleport node failures and provide more visibility in their statuses. The lack of these things recently resulted in a [regression](https://github.com/gravitational/gravity/issues/1445) that went unnoticed for some time.

* Add additional step to "gravity join" operation to check that Teleport node on the joined node has successfully registered with the cluster, and fail operation if it hasn't.
* Related to the above, I found out that "gravity join" would exit with 0 return code even in case of a failed join. I wonder if this is the reason our "resize" robotest scenarios rarely fail 😄Fixed that too.
* Add Teleport status to node statuses in "gravity status" output. Note, it does not affect the cluster status at the moment and serves informational purposes. Here's how it looks:

```
Cluster nodes:
    Masters:
        * node-1 (192.168.99.102, node)
            Status:	healthy
            Teleport:	connected
    Nodes:
        * node-2 (192.168.99.103, knode)
            Status:	healthy
            Teleport:	connected
```

Note that I considered just displaying it similar to how we display failed probes, but decided this way would give better visibility - for example, a node may be offline but Teleport may be still running on it (and thus, it's still accessible via it).

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1445.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

### Make sure join fails if Teleport node doesn't connect

```
// Launch join operation
ubuntu@node-2:~$ sudo ./gravity join 192.168.99.102 --advertise-addr=192.168.99.103 --token=f11415da6889 --role=knode

// In another tab, break some Teleport-specific directory
ubuntu@node-2:~$ sudo chattr +i /var/lib/gravity/teleport/log

// Make sure join fails and exits with a non-successful return code
Wed Apr 29 22:17:00 UTC	Wait for the Teleport node to join cluster
Wed Apr 29 22:19:01 UTC	Operation failure: failed to execute phase "/wait"
	retry exceeded 2m0s
		Teleport on node(addr=192.168.99.103, hostname=node-2, role=knode, cluster_role=node) hasn't registered yet
[ERROR]: failed to join the cluster
ubuntu@node-2:~$ echo $?
255
```

### Make sure gravity status detects down Teleport nodes

```
// Stop Teleport node on one node
ubuntu@node-2:~$ sudo systemctl stop gravity__gravitational.io__teleport__3.0.5.service

// Call gravity status and make sure Teleport is shown as disconnected
ubuntu@node-1:~$ gravity status
...
Cluster nodes:
    Masters:
        * node-1 (192.168.99.102, node)
            Status:	healthy
            Teleport:	connected
    Nodes:
        * node-2 (192.168.99.103, knode)
            Status:	healthy
            Teleport:	disconnected
```

## Additional information

Needs forward-ports to 6.1, 7.0 and master.

### DIFF
--- a/lib/expand/builder.go
+++ b/lib/expand/builder.go
@@ -235,6 +235,15 @@ func (b *planBuilder) AddWaitPhase(plan *storage.OperationPlan) {
 				},
 				Requires: []string{WaitPlanetPhase},
 			},
+			{
+				ID:          WaitTeleportPhase,
+				Description: "Wait for the Teleport node to join cluster",
+				Data: &storage.OperationPhaseData{
+					Server:     &b.JoiningNode,
+					ExecServer: &b.JoiningNode,
+				},
+				Requires: []string{WaitPlanetPhase},
+			},
 		},
 	})
 }
--- a/lib/expand/fsmspec.go
+++ b/lib/expand/fsmspec.go
@@ -94,6 +94,10 @@ func FSMSpec(config FSMConfig) fsm.FSMSpecFunc {
 			return phases.NewWaitK8s(p,
 				config.Operator)
 
+		case strings.HasPrefix(p.Phase.ID, WaitTeleportPhase):
+			return phases.NewWaitTeleport(p,
+				config.Operator)
+
 		case strings.HasPrefix(p.Phase.ID, PostHookPhase):
 			return installphases.NewHook(p,
 				config.Operator,
@@ -125,6 +129,8 @@ const (
 	WaitPlanetPhase = "/wait/planet"
 	// WaitK8sPhase waits for joining node to register with Kubernetes
 	WaitK8sPhase = "/wait/k8s"
+	// WaitTeleportPhase waits for Teleport on the joining node to join the cluster
+	WaitTeleportPhase = "/wait/teleport"
 	// PostHookPhase runs post-expand application hook
 	PostHookPhase = "/postHook"
 	// ElectPhase enables leader election on master node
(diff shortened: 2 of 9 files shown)


knisbet:
I'm not sure I agree with adding teleport status as a separate status line:

```
    Nodes:
        * node-2 (192.168.99.103, knode)
            Status:	healthy
            Teleport:	disconnected
```

It's one of those UX things, where we need to look at how we train our customers to check status. How are they supposed to interpret a Healthy node where teleport is disconnected, and react to that piece of information? Will they internalize what this means and how it impacts the system, when it doesn't alert the customer in the same pattern as our other alerts.

knisbet (approved):

a-palchikov (approved):

r0mant:
@knisbet cc @a-palchikov That's fair, as I mentioned in the ticket description I first considered a more "traditional" approach where it would look like this:

Screen Shot 2020-04-29 at 12 25 56 PM

Why I opted for the one implemented, is for a couple of reasons (also mentioned above):

* Teleport is a bit separate component that does not really affect the "health" of the system in the same way as other probes do which is why I thought that just shoving it among other probes may not be the best fit.
* Since it's separate, gravity status can now give insight into the case where, for example, the node is offline (planet stopped) but teleport is connected - in which case it can still be used to teleport onto the node to perform maintenance.

What I'm personally not a fan of is that maybe saying "Teleport" might be revealing a bit of implementation details. We can obviously always tweak this, but maybe something like "SSH Status" or "Access Status" or "Remote Access" would be more appropriate? E.g.

```
Cluster nodes:
    Masters:
        * node-1 (192.168.99.102, node)
            Status:	healthy
            SSH Status:	online
```

Issue #1468, opened 2020-04-28
[BUG] gravity upgrade --resume without upgrade operations results in panic


r0mant:
**Describe the bug**

Per customer report:

> We have been reported an internal bug where cluster upgrade from 5.5.40 to 5.5.41 failed with a crash on gravity upgrade command. On first try upgrading, It failed with “The cluster is in degraded state so uploading a new cluster image is prohibited. Then, changed to the ./installer directory, ran ./gravity status to confirm the cluster is still reporting healthy. And then ran ./gravity upgrade --resume . The following SIGSEGV error returned immediately.

This is partly a user error as there's no upgrade operation yet so the command should be just "./gravity upgrade" without --resume flag but the command should not produce a panic.

**To Reproduce**

**Expected behavior**

It should not crash and return an appropriate error.

**Logs**

```
# ./gravity upgrade --resume
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x267cc07]
goroutine 1 [running]:
github.com/gravitational/gravity/lib/localenv.(*LocalEnvironment).SelectOpsCenter(0x0, 0x0, 0x0, 0xc000a8db38, 0x29fdeae, 0x345d900, 0x0)
	/gopath/src/github.com/gravitational/gravity/lib/localenv/localenv.go:284 +0x37
github.com/gravitational/gravity/lib/localenv.(*LocalEnvironment).CurrentLogin(0x0, 0xc0000996c0, 0x0, 0x1)
	/gopath/src/github.com/gravitational/gravity/lib/localenv/localenv.go:395 +0x37
github.com/gravitational/gravity/lib/localenv.(*LocalEnvironment).CurrentOperator(0x0, 0xc000a8dba0, 0x1, 0x1, 0x11, 0x2e981327, 0xed50cb375)
	/gopath/src/github.com/gravitational/gravity/lib/localenv/localenv.go:404 +0x43
github.com/gravitational/gravity/tool/gravity/cli.executeJoinPhase(0xc000732be0, 0x0, 0x356f876, 0x1, 0x0, 0x0, 0x0, 0x34630b8a000, 0x0, 0xc000712150, ...)
	/gopath/src/github.com/gravitational/gravity/tool/gravity/cli/install.go:464 +0x8a
[...]

PR #1479, opened 2020-04-30
[5.5.x] avoid crashes when working with operational state

a-palchikov:
## Description

This PR restructures the way operational environments are created by doing away with command comparisons in favor of creating the environments where they're actually needed. This alleviates the problems with command-specific logic to decide which environment to create for which command early in the process.
Additionally, this backports the change necessary to be able to complete an expand operation from another cluster node.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1468.
* Requires https://github.com/gravitational/gravity.e/pull/4295.
* Ports https://github.com/gravitational/gravity/pull/1347 (partially).

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

Upgraded a 2-node cluster using the branch.
Attempted the `gravity upgrade --resume` with an active expand operation:
```
$ ./gravity upgrade --resume
no active update operation found
```

### DIFF
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -433,7 +433,8 @@ func (p *Peer) tryConnect() (op *operationContext, err error) {
 			return op, trace.Wrap(err)
 		}
 		if trace.IsCompareFailed(err) {
-			p.sendMessage("Waiting for another operation to finish at %v", addr)
+			p.WithError(err).Info("Failed to create expand operation.")
+			p.sendMessage("Waiting to create expand operation at %v: %v", addr, err)
 		}
 	}
 	return op, trace.Wrap(err)
--- a/tool/gravity/cli/clusterupdate.go
+++ b/tool/gravity/cli/clusterupdate.go
@@ -98,6 +98,9 @@ func newClusterUpdater(
 func executeUpdatePhase(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, params PhaseParams) error {
 	operation, err := getActiveOperation(env, environ, params.OperationID)
 	if err != nil {
+		if trace.IsNotFound(err) {
+			return trace.NotFound("no active update operation found")
+		}
 		return trace.Wrap(err)
 	}
 	if operation.Type != ops.OperationUpdate {
--- a/tool/gravity/cli/clusterupdate.go
+++ b/tool/gravity/cli/clusterupdate.go
@@ -95,11 +95,23 @@ func newClusterUpdater(
 	return updater, nil
 }
 
-func executeUpdatePhase(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, params PhaseParams, operation ops.SiteOperation) error {
+func executeUpdatePhase(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, params PhaseParams) error {
+	operation, err := getActiveOperation(env, environ, params.OperationID)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if operation.Type != ops.OperationUpdate {
+		return trace.NotFound("no active update operation found")
+	}
+	return executeUpdatePhaseForOperation(env, environ, params, *operation)
+}
+
+func executeUpdatePhaseForOperation(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, params PhaseParams, operation ops.SiteOperation) error {
 	updateEnv, err := environ.NewUpdateEnv()
 	if err != nil {
 		return trace.Wrap(err)
 	}
+	defer updateEnv.Close()
 	updater, err := getClusterUpdater(env, updateEnv, operation, params.SkipVersionCheck)
 	if err != nil {
 		return trace.Wrap(err)
@@ -109,11 +121,12 @@ func executeUpdatePhase(env *localenv.LocalEnvironment, environ LocalEnvironment
 	return trace.Wrap(err)
 }
 
-func rollbackUpdatePhase(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, params PhaseParams, operation ops.SiteOperation) error {
+func rollbackUpdatePhaseForOperation(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, params PhaseParams, operation ops.SiteOperation) error {
 	updateEnv, err := environ.NewUpdateEnv()
 	if err != nil {
 		return trace.Wrap(err)
 	}
+	defer updateEnv.Close()
 	updater, err := getClusterUpdater(env, updateEnv, operation, params.SkipVersionCheck)
 	if err != nil {
 		return trace.Wrap(err)
@@ -123,11 +136,12 @@ func rollbackUpdatePhase(env *localenv.LocalEnvironment, environ LocalEnvironmen
 	return trace.Wrap(err)
 }
 
-func completeUpdatePlan(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, operation ops.SiteOperation) error {
+func completeUpdatePlanForOperation(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, operation ops.SiteOperation) error {
 	updateEnv, err := environ.NewUpdateEnv()
 	if err != nil {
 		return trace.Wrap(err)
 	}
+	defer updateEnv.Close()
 	updater, err := getClusterUpdater(env, updateEnv, operation, true)
 	if err != nil {
 		return trace.Wrap(err)
--- a/tool/gravity/cli/clusterupdate.go
+++ b/tool/gravity/cli/clusterupdate.go
@@ -147,7 +147,13 @@ func completeUpdatePlanForOperation(env *localenv.LocalEnvironment, environ Loca
 		return trace.Wrap(err)
 	}
 	defer updater.Close()
-	return trace.Wrap(updater.Complete(nil))
+	if err := updater.Complete(nil); err != nil {
+		return trace.Wrap(err)
+	}
+	if err := updater.Activate(); err != nil {
+		return trace.Wrap(err)
+	}
+	return nil
 }
 
 func getClusterUpdater(localEnv, updateEnv *localenv.LocalEnvironment, operation ops.SiteOperation, noValidateVersion bool) (*update.Updater, error) {
--- a/tool/gravity/cli/clusterupdate.go
+++ b/tool/gravity/cli/clusterupdate.go
@@ -55,13 +55,11 @@ func updateCheck(env *localenv.LocalEnvironment, updatePackage string) error {
 }
 
 func updateTrigger(
-	localEnv *localenv.LocalEnvironment,
-	updateEnv *localenv.LocalEnvironment,
+	localEnv, updateEnv *localenv.LocalEnvironment,
 	updatePackage string,
 	manual, noValidateVersion bool,
 ) error {
-	ctx := context.TODO()
-	updater, err := newClusterUpdater(ctx, localEnv, updateEnv, updatePackage, manual, noValidateVersion)
+	updater, err := newClusterUpdater(context.TODO(), localEnv, updateEnv, updatePackage, manual, noValidateVersion)
 	if err != nil {
 		return trace.Wrap(err)
 	}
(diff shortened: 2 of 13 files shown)


lib/expand/join.go
```diff
@@ -433,7 +433,8 @@ func (p *Peer) tryConnect() (op *operationContext, err error) {
 			return op, trace.Wrap(err)
 		}
 		if trace.IsCompareFailed(err) {
-			p.sendMessage("Waiting for another operation to finish at %v", addr)
+			p.WithError(err).Info("Failed to create expand operation.")
+			p.sendMessage("Waiting to create expand operation at %v: %v", addr, err)
```
a-palchikov, line 437: The old message was misleading as there multiple possible reasons for this error including cluster being degraded.

tool/gravity/cli/operation.go
```diff
@@ -210,27 +231,17 @@ func (r *backendOperations) List(localEnv, updateEnv, joinEnv *localenv.LocalEnv
...
-		}
-		if wizardEnv != nil && wizardEnv.Operator != nil {
-			op, err := ops.GetWizardOperation(wizardEnv.Operator)
-			if err == nil {
-				log.Debug("Fetched install operation from wizard environment.")
-				r.operations[op.ID] = (ops.SiteOperation)(*op)
-			} else {
-				log.WithError(err).Warn("Failed to query install operation.")
-			}
+		if err := r.listInstallOperation(); err != nil {
```
a-palchikov, line 243: Moved it to a separate function.

tool/gravity/cli/system.go
```diff
@@ -1120,7 +1120,7 @@ func systemUninstall(env *localenv.LocalEnvironment, confirmed bool) error {
 	// remove all files and directories gravity might have created on the system
 	for _, path := range append(state.StateLocatorPaths, defaults.ModulesPath, defaults.SysctlPath, defaults.GravityEphemeralDir) {
 		// errors are expected since some of them may not exist
-		if err := os.Remove(path); err == nil {
+		if err := os.RemoveAll(path); err == nil {
```
a-palchikov, line 1123: This was not cleaning anything up.

tool/gravity/cli/clusterupdate.go
```diff
@@ -97,7 +95,11 @@ func newClusterUpdater(
 	return updater, nil
 }
 
-func executeUpdatePhase(env, updateEnv *localenv.LocalEnvironment, params PhaseParams, operation ops.SiteOperation) error {
+func executeUpdatePhase(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, params PhaseParams, operation ops.SiteOperation) error {
+	updateEnv, err := environ.NewUpdateEnv()
+	if err != nil {
+		return trace.Wrap(err)
+	}
```
r0mant, line 102: defer updateEnv.Close()? And in the 2 places below as well.

tool/gravity/cli/operation.go
```diff
@@ -116,24 +117,44 @@ func completeOperationPlan(localEnv, updateEnv, joinEnv *localenv.LocalEnvironme
...
 	case ops.OperationUpdateRuntimeEnviron:
-		return completeEnvironPlan(localEnv, updateEnv, op)
+		err = completeEnvironPlan(localEnv, environ, op)
 	case ops.OperationUpdateConfig:
-		return completeConfigPlan(localEnv, updateEnv, op)
+		err = completeConfigPlan(localEnv, environ, op)
 	default:
 		return trace.BadParameter("operation type %q does not support plan completion", op.Type)
 	}
+	if op.Type != ops.OperationInstall && trace.IsNotFound(err) {
```
r0mant, line 132: Is this special use-case mostly to be able to complete join (or any operation?) from another node?

tool/gravity/cli/operation.go
```diff
@@ -210,27 +231,17 @@ func (r *backendOperations) List(localEnv, updateEnv, joinEnv *localenv.LocalEnv
 			log.WithError(err).Warn("Failed to query cluster operations.")
 		}
 	}
-	if updateEnv != nil {
-		r.getOperationAndUpdateCache(updateEnv.Backend, log.WithField("context", "update"))
+	if err := r.listUpdateOperation(environ); err != nil && !trace.IsNotFound(err) {
```
r0mant, line 234: Hm, I might be misunderstanding the logic here, but the way it was implemented before IIRC we would pass either updateEnv or joinEnv here - one of those would be nil depending on which command is run - and hence we would only query either update or join operation. Now, we query both so backendOperations may contain both upgrade and join operation if both backends are present on the node. Would that confuse anything?

tool/gravity/cli/operation.go
```diff
@@ -249,7 +260,49 @@ func (r *backendOperations) init(clusterBackend storage.Backend) error {
...
+		// Timeout error means the directory is used by the active installer process
+		// which means, it's the installer environment, not joining node's
+		return nil
+	}
+	defer env.Close()
+	r.getOperationAndUpdateCache(env.Backend, log.WithField("context", "expand"))
+	return nil
+}
+
+func (r *backendOperations) listInstallOperation() error {
```
r0mant, line 292: Nit: This method doesn't return anything so we can do away with the return parameter.

tool/gravity/cli/operation.go
```diff
@@ -265,6 +318,16 @@ func (r *backendOperations) getOperationAndUpdateCache(backend storage.Backend,
 }
 
 func (r backendOperations) isActiveInstallOperation() bool {
+	// Bail out if there's an operation from a local backend and we failed to query
```
r0mant, line 321: TBH I'm not sure I follow this comment - could it be reworded somehow maybe? To explain what "r.clusterOperation" is, what "r.operations" are and why this condition means this is not an active install operation.

tool/gravity/cli/plan.go
```diff
@@ -64,9 +64,15 @@ func initUpdateOperationPlan(localEnv, updateEnv *localenv.LocalEnvironment) err
...
 }
 
-func displayOperationPlan(localEnv, updateEnv, joinEnv *localenv.LocalEnvironment, operationID string, format constants.Format) error {
-	operations, err := getLastOperation(localEnv, updateEnv, joinEnv, operationID)
+func displayOperationPlan(localEnv *localenv.LocalEnvironment, environ LocalEnvironmentFactory, operationID string, format constants.Format) error {
+	operations, err := getLastOperation(localEnv, environ, operationID)
 	if err != nil {
+		if trace.IsNotFound(err) {
+			return trace.NotFound(`no operation found.
+This usually means that the installation has failed to start or was not started.
```
r0mant, line 72: Nit: Can this also mean that "join" operation has failed to start or was not started?

tool/gravity/cli/plan.go
```diff
@@ -82,18 +88,23 @@ func displayOperationPlan(localEnv, updateEnv, joinEnv *localenv.LocalEnvironmen
...
+		return displayUpdateOperationPlan(localEnv, environ, op.Key(), format)
 	case ops.OperationUpdateConfig:
-		return displayUpdateOperationPlan(localEnv, updateEnv, op.Key(), format)
+		return displayUpdateOperationPlan(localEnv, environ, op.Key(), format)
 	case ops.OperationGarbageCollect:
 		return displayClusterOperationPlan(localEnv, op.Key(), format)
 	default:
 		return trace.BadParameter("unknown operation type %q", op.Type)
 	}
+	if err != nil && trace.IsNotFound(err) {
```
r0mant, line 103: What is this fallback for? Seems like the switch above covers all operation types?

tool/gravity/cli/plan.go
```diff
@@ -109,7 +120,11 @@ func displayClusterOperationPlan(env *localenv.LocalEnvironment, opKey ops.SiteO
 	return trace.Wrap(err)
 }
 
-func displayUpdateOperationPlan(localEnv, updateEnv *localenv.LocalEnvironment, opKey ops.SiteOperationKey, format constants.Format) error {
+func displayUpdateOperationPlan(localEnv *localenv.LocalEnvironment, environ LocalEnvironmentFactory, opKey ops.SiteOperationKey, format constants.Format) error {
+	updateEnv, err := environ.NewUpdateEnv()
+	if err != nil {
+		return trace.Wrap(err)
+	}
```
r0mant, line 127: defer updateEnv.Close()?

tool/gravity/cli/plan.go
```diff
@@ -231,4 +228,21 @@ func tryReconcilePlan(ctx context.Context, localEnv, updateEnv *localenv.LocalEn
...
 }
 
-const recoveryModeWarning = "Failed to retrieve plan from etcd, showing cached plan. If etcd went down as a result of a system upgrade, you can perform a rollback phase. Run 'gravity plan --repair' when etcd connection is restored.\n"
+func getPlanFromWizard(opKey ops.SiteOperationKey) (*storage.OperationPlan, error) {
+	wizardEnv, err := localenv.NewRemoteEnvironment()
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	if wizardEnv.Operator == nil {
+		return nil, trace.NotFound("no operation plan")
```
r0mant, line 237: Nit: I think the message here is a bit misleading as we didn't actually check for plan. I'd rather say why the operator is nil here.

tool/gravity/cli/run.go
```diff
@@ -362,36 +353,58 @@ func Execute(g *Application, cmd string, extraArgs []string) error {
...
+		defer updateEnv.Close()
 		return initUpdateOperationPlan(localEnv, updateEnv)
 	case g.UpgradeCmd.FullCommand():
 		if *g.UpgradeCmd.Resume {
 			*g.UpgradeCmd.Phase = fsm.RootPhase
 			*g.UpgradeCmd.Force = false
 		}
 		if *g.UpgradeCmd.Phase != "" {
-			return executePhase(localEnv, updateEnv, joinEnv,
+			op, err := getActiveOperation(localEnv, g, "")
```
r0mant, line 380: Nit: Can this logic be moved under "executeUpdatePhase"? Seems a bit weird to have this in run.go.

tool/gravity/cli/operation.go
a-palchikov: Yes.

tool/gravity/cli/operation.go
a-palchikov: No, prior to that both `updateEnv` and `joinEnv` could be active at the same time as the `plan` command is ambiguous so it worked similar to this before. 
One thing that to account for is the residual state that might have been left behind as the `leave` (or more specifically `system uninstall`) command would not remove the join state directory (only relevant in 5.5.x though) and since the expand operation was never updated to completion in this local store, it will popup as an unfinished (hence active) and maybe even unrelated to the current cluster.
All because of such disparate state management.

tool/gravity/cli/plan.go
a-palchikov: Possible - I'll update the message.

tool/gravity/cli/plan.go
a-palchikov: Looks like a backport gone slightly wrong - I'll fix that.

tool/gravity/cli/plan.go
a-palchikov: That's a good question - I have no idea how the operator could be nil after a successful call to `NewRemoveEnvironment`.

tool/gravity/cli/operation.go
a-palchikov: I'll update the struct responsible for operation collection.

r0mant (approved):

tool/gravity/cli/plan.go
```diff
@@ -64,9 +64,15 @@ func initUpdateOperationPlan(localEnv, updateEnv *localenv.LocalEnvironment) err
...
 
-func displayOperationPlan(localEnv, updateEnv, joinEnv *localenv.LocalEnvironment, operationID string, format constants.Format) error {
-	operations, err := getLastOperation(localEnv, updateEnv, joinEnv, operationID)
+func displayOperationPlan(localEnv *localenv.LocalEnvironment, environ LocalEnvironmentFactory, operationID string, format constants.Format) error {
+	operations, err := getLastOperation(localEnv, environ, operationID)
 	if err != nil {
+		if trace.IsNotFound(err) {
+			return trace.NotFound(`no operation found.
+This usually means that the installation/join operation has failed to start or was not started.
+Clean up the node with 'gravity leave' if necessary and start the installation/join with either 'gravity install' or 'gravity join'.
```
knisbet, line 73: How is a user supposed to know if running gravity leave is necessary?

knisbet (approved):

tool/gravity/cli/plan.go
a-palchikov: I'll have the code check and output the appropriate message.

PR #1481, opened 2020-04-30
Forward-port teleport tokens fixes.

r0mant:
## Description

This PR forward-ports fixes related to teleport node token regression (https://github.com/gravitational/gravity/issues/1445) as well as some extra stuff.

* Same as in other branches, have gravity-site read auth tokens upon startup instead of hardcoding them in configuration.
* Use a new provisioning token type, "teleport", for teleport nodes authentication. But also fallback to "expand" tokens for backwards compatibility.
* Initialize this "teleport" prov. token during install/upgrade.
* Also forward port (from https://github.com/gravitational/gravity/pull/1446) the command that can show teleport configuration from a package. I thought that the other two commands may not really be useful anymore.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1445.

* Ports https://github.com/gravitational/gravity/pull/1453, https://github.com/gravitational/gravity/pull/1446.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

### Fresh install scenario
- [x] Install cluster using this branch.
- [x] Join another node.
- [x] Verify Teleport joins ok and uses Teleport provisioning token.

### Upgrade scenario
- [x] Install 7.0.4 cluster.
- [x] Upgrade to this branch.
- [x] Verify Teleport provisioning token was created.
- [x] Join another node.
- [x] Verify Teleport joins ok and uses Teleport provisioning token.

### DIFF
--- a/lib/ops/opsservice/configure.go
+++ b/lib/ops/opsservice/configure.go
@@ -1118,16 +1118,9 @@ func (s *site) getTeleportMasterConfig(ctx *operationContext, configPackage loc.
 	fileConf.AdvertiseIP = advertiseIP.String()
 	fileConf.Global.NodeName = master.FQDN(s.domainName)
 
-	joinToken, err := s.service.GetExpandToken(s.key)
-	if err != nil {
-		return nil, trace.Wrap(err)
-	}
-
 	// turn on auth service
 	fileConf.Auth.EnabledFlag = "yes"
 	fileConf.Auth.ClusterName = telecfg.ClusterName(s.domainName)
-	fileConf.Auth.StaticTokens = telecfg.StaticTokens{
-		telecfg.StaticToken(fmt.Sprintf("node:%v", joinToken.Token))}
 
 	// turn on proxy and Kubernetes integration
 	fileConf.Proxy.EnabledFlag = "yes"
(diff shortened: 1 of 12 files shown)


lib/ops/opsservice/configure.go
```diff
@@ -1187,12 +1180,28 @@ func toObject(in interface{}) (map[string]interface{}, error) {
...
+			return t.Token, nil
+		}
 	}
+	// Fallback to expand tokens for backwards compatibility.
+	for _, t := range tokens {
+		if t.IsExpand() && t.IsPersistent() {
+			return t.Token, nil
+		}
+	}
+	return "", trace.NotFound("could not find persistent teleport or expand token among %v",
```
a-palchikov, line 1200: Is this user facing error? What could be added to help user in search for the culprit?

lib/update/cluster/phases/init.go
```diff
@@ -211,6 +214,40 @@ func (p *updatePhaseInit) Execute(context.Context) error {
...
+	cluster, err := p.Backend.GetLocalSite(defaults.SystemAccountID)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	allTokens, err := p.Backend.GetSiteProvisioningTokens(cluster.Domain)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	for _, t := range allTokens {
+		if t.IsTeleport() && t.IsPersistent() {
```
a-palchikov, line 229: Can a teleport token be not persistent?

a-palchikov (approved):

lib/ops/opsservice/configure.go
r0mant: It's not meant to be seen by a user. This would mean a some sort of bug or a programming error on our part as these tokens are always supposed to exist, not sure what we can add here to explain that they might be seeing this due to a bug :)

lib/update/cluster/phases/init.go
r0mant: Not as implemented currently, but I'd like to avoid the situation if somebody (even me) uses teleport expiring tokens for something in future and we'll end up in the same situation as we did last week.

Issue #1112, opened 2020-02-17
Consider removing or loosening the same OS requirement


a-palchikov:
This has been brought up multiple times and gets in the way without clear benefits.

If there're benefits - we should document them clearly.
If there aren't any - we should drop the requirement given that:
  * manifest lists all available OS distribution requirements for a node
  * OS distribution is one of the [officially supported](https://gravitational.com/gravity/docs/requirements/#linux-distributions)

There maybe legitimate requirements to build the environment based on type of the workload which might include a specialized distribution or version.

r0mant:
https://github.com/gravitational/gravity/issues/1913 duplicate.

PR #1484, opened 2020-05-01
[6.3.x] loosen same OS requirement

a-palchikov:
## Description

This PR loosens the same OS requirement preflight check with the following semantics:

  * OS distributions are the same if both name and versions are exact match
  * OS distributions are compatible (and check succeeds) if name is an exact match and versions match on the prefix up to the first '.'
  * otherwise, OS distributions are not compatible

Also, this has been brought up multiple times (frequently in the form of asking whether it's possible to suppress just the OS check).

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1112.

## TODOs

- [x] Self-review the change
- [x] Write tests
- [ ] Perform manual testing
- [ ] Write documentation
- [ ] Address review feedback

## Testing done

Unit-tested change.

### DIFF
--- a/lib/checks/checks.go
+++ b/lib/checks/checks.go
@@ -756,28 +756,56 @@ func checkRAM(info ServerInfo, ram schema.RAM) error {
 	return nil
 }
 
-// checkSameOS makes sure all servers have the same OS/version
+// checkSameOS verifies the OS distribution requirement for the specified set of servers.
+// The check will pass if all nodes in the cluster are based on the same OS distribution and major version.
+// Variance in minor/patch versions is acceptable.
 func checkSameOS(servers []Server) error {
-	osToNodes := make(map[string][]string)
+	// distros maps distribnution name to list of versions
+	distros := make(map[string][]string)
 	for _, server := range servers {
-		os := systeminfo.OS(server.GetOS()).Name()
-		osToNodes[os] = append(osToNodes[os], fmt.Sprintf("%v (%v)",
-			server.ServerInfo.GetHostname(), server.AdvertiseAddr))
+		info := server.GetOS()
+		distros[info.ID] = append(distros[info.ID], info.Version)
 	}
-
-	if len(osToNodes) > 1 {
-		var formatted []string
-		for os, nodes := range osToNodes {
-			formatted = append(formatted, fmt.Sprintf(
-				"%v: %v", os, strings.Join(nodes, ", ")))
+	if len(distros) != 1 {
+		return trace.BadParameter("servers have different OS distributions: %v", formatKeysAsList(distros))
+	}
+	// Version verification is purposedly simply and will compare the prefixes
+	// upto to either the first '.' or eol
+	for _, versions := range distros {
+		if !verifyCommonVersionPrefix(versions...) {
+			return trace.BadParameter("servers have different OS versions: %v", formatAsList(distros))
 		}
-		return trace.BadParameter(
-			"servers have different OSes/versions:\n%v",
-			strings.Join(formatted, "\n"))
 	}
-
-	log.Infof("Servers passed check for the same OS: %v.", osToNodes)
+	log.Info("Servers passed check for the same OS.")
 	return nil
+}
+
+func verifyCommonVersionPrefix(versions ...string) bool {
+	prefixes := make(map[string]struct{})
+	for _, v := range versions {
+		index := strings.Index(v, ".")
+		if index == -1 {
+			index = len(v)
+		}
+		prefixes[v[:index]] = struct{}{}
+	}
+	return len(prefixes) == 1
+}
+
+func formatAsList(m map[string][]string) (result []string) {
+	result = make([]string, 0, len(m))
+	for k, v := range m {
+		result = append(result, fmt.Sprintf("%v (%q)", k, v))
+	}
+	return result
+}
+
+func formatKeysAsList(m map[string][]string) (result []string) {
+	result = make([]string, 0, len(m))
+	for k := range m {
+		result = append(result, k)
+	}
+	return result
 }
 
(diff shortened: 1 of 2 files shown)


lib/checks/checks_test.go
```diff
@@ -118,41 +118,80 @@ func (s *ChecksSuite) TestTime(c *check.C) {
...
+		{
+			comment: "no match on distribution",
+			servers: []Server{
+				newServer("node-1", "centos", "7"),
+				newServer("node-2", "rhel", "7"),
+			},
+			err: "servers have different OS distributions.*",
+		},
+		// {
+		// 	comment: "no match on bogus input? (identical to previous behavior)",
```
a-palchikov, line 167: Should this case be also supported? This seems unlikely as the `VERSION_ID` is [described](https://www.freedesktop.org/software/systemd/man/os-release.html#VERSION_ID=) as containing mostly numeric input.

helgi:
just a thought, for the rare customer (not myself) that still needs/want the strict validation maybe that could be something they can set in the app.yaml or as part of `upgrade` as a flag.

For me I would want to allow CentOS 6 to 7 to 8 upgrade (generally); at least historically I did not see a reason. Even I wouldn't mind going CentOS -> Ubuntu. That wouldn't work for everyone of course. Future enhancement maybe

a-palchikov:
It seems that a lot more flexibility is wanted. I wonder if an additional condition for the OS requirement is possible to either say that only strict matching is required (`=7`, `=8` for the version and something for the distribution - to pin a specific distribution) to something very loose to say that any distribution/version combination is allowed (from the list of specified).

helgi:
Yeah, that's along the lines I was thinking. `app.yaml` would still need to define the OS/version as before.

Looking at https://github.com/gravitational/gravity/blob/master/assets/telekube/resources/app.yaml#L75-L91 as an example I would probably say this list means you can upgrade between any of those. Then the `=8` can be added to further restrict it, and if a user only wanted `8.0` and `8.1` but not `8.2` upgradability for the given built then they can do `=8.0` and `=8.1`, or even just the fact a more detailed version was used is enough.

Still pondering this how to best do it in a way where you can list multiple OS like the file above without wanting to allow fully upgradability between them all. That's a bit more complex without adding a new field `os_upgrade` or `os_install` to make a distinction between install and upgrade

lib/checks/checks.go
```diff
@@ -756,28 +756,56 @@ func checkRAM(info ServerInfo, ram schema.RAM) error {
 	return nil
 }
 
-// checkSameOS makes sure all servers have the same OS/version
+// checkSameOS verifies the OS distribution requirement for the specified set of servers.
+// The check will pass if all nodes in the cluster are based on the same OS distribution and major version.
+// Variance in minor/patch versions is acceptable.
 func checkSameOS(servers []Server) error {
-	osToNodes := make(map[string][]string)
+	// distros maps distribnution name to list of versions
```
r0mant, line 763: distribution

lib/checks/checks.go
```diff
@@ -756,28 +756,56 @@ func checkRAM(info ServerInfo, ram schema.RAM) error {
...
-			server.ServerInfo.GetHostname(), server.AdvertiseAddr))
+		info := server.GetOS()
+		distros[info.ID] = append(distros[info.ID], info.Version)
 	}
+	if len(distros) != 1 {
+		return trace.BadParameter("servers have different OS distributions: %v", formatKeysAsList(distros))
+	}
+	// Version verification is purposedly simply and will compare the prefixes
+	// upto to either the first '.' or eol
+	for _, versions := range distros {
```
r0mant, line 774: Nit: Isn't `len(distros) == 1` here so the loop isn't really necessary?

lib/checks/checks.go
```diff
@@ -756,28 +756,56 @@ func checkRAM(info ServerInfo, ram schema.RAM) error {
...
+		if !verifyCommonVersionPrefix(versions...) {
+			return trace.BadParameter("servers have different OS versions: %v", formatAsList(distros))
+		}
+	}
+	log.Info("Servers passed check for the same OS.")
```
r0mant, line 779: Nit: Add collected OS versions to the message?

lib/checks/checks.go
```diff
@@ -756,28 +756,56 @@ func checkRAM(info ServerInfo, ram schema.RAM) error {
...
 
-	if len(osToNodes) > 1 {
-		var formatted []string
-		for os, nodes := range osToNodes {
-			formatted = append(formatted, fmt.Sprintf(
-				"%v: %v", os, strings.Join(nodes, ", ")))
+func verifyCommonVersionPrefix(versions ...string) bool {
+	prefixes := make(map[string]struct{})
+	for _, v := range versions {
+		index := strings.Index(v, ".")
```
r0mant, line 786: It might be just a question of preference of course, but why mess with indices/lengths/maps/etc. if all we need to do is compare first elements of strings split by a dot?

```go
func verifyCommonVersionPrefix(versions ...string) bool {
  if len(versions) <= 1 {
    return true
  }
  for i := 0; i < len(versions)-1; i += 1 {
    if strings.Split(versions[i], ".") != strings.Split(versions[i+1], ".") {
      return false
    }
  }
  return true
}
```

lib/checks/checks_test.go
r0mant: I like how the docs are precise, "mostly numeric" :) I think it's entirely possible system-release (or os-release) file can be malformed. But in this case it'll likely fail due to required version mismatch? So the OS "homogeneity" test doesn't necessarily need to account for it.

r0mant (approved):

lib/checks/checks.go
a-palchikov: That's one way to get at the first item w/o knowing the key

PR #1495, opened 2020-05-04
Dmitri/1471 upgrade resume

a-palchikov:
## Type of change

* Refs https://github.com/gravitational/gravity/issues/1471.

* Ports https://github.com/gravitational/gravity/pull/1479.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

### error for an unstarted install operation
```
$ ./gravity plan resume
...
Mon May  4 17:48:34 UTC	Connecting to installer
[ERROR]: failed to find installer service. Use 'gravity install' to start new installation or 'gravity join' to join the existing cluster.
	no service unit file in /home/centos/.gravity
```

### error for resuming on an existing cluster
```
# ./gravity plan resume
[ERROR]: no active operation found
```

### explicit resumption of the upgrade operation (also when another operation is in progress)
```
$ ./gravity upgrade --resume
[ERROR]: no active update operation found
```

### DIFF
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -831,7 +831,11 @@ func (p *Peer) tryConnect(operationID string) (ctx *operationContext, err error)
 		}
 		if trace.IsCompareFailed(err) {
 			p.WithError(err).Info("Failed to create expand operation.")
-			p.printStep("Waiting to create expand operation at %v: %v", addr, err)
+			if utils.IsClusterDegradedError(err) {
+				p.printStep("Waiting to cluster to become healthy")
+			} else {
+				p.printStep("Waiting to another operation to complete at %v", addr)
+			}
 		}
 	}
 	return ctx, trace.Wrap(err)
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -830,7 +830,8 @@ func (p *Peer) tryConnect(operationID string) (ctx *operationContext, err error)
 			return nil, utils.Abort(err)
 		}
 		if trace.IsCompareFailed(err) {
-			p.printStep("Waiting for another operation to finish at %v", addr)
+			p.WithError(err).Info("Failed to create expand operation.")
+			p.printStep("Waiting to create expand operation at %v: %v", addr, err)
 		}
 	}
 	return ctx, trace.Wrap(err)
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -830,11 +830,11 @@ func (p *Peer) tryConnect(operationID string) (ctx *operationContext, err error)
 			return nil, utils.Abort(err)
 		}
 		if trace.IsCompareFailed(err) {
-			p.WithError(err).Info("Failed to create expand operation.")
+			p.Warnf("Waiting for precondition to create expand operation: %v.", err)
 			if utils.IsClusterDegradedError(err) {
-				p.printStep("Waiting to cluster to become healthy")
+				p.printStep("Cluster is degraded, waiting for it to become healthy")
 			} else {
-				p.printStep("Waiting to another operation to complete at %v", addr)
+				p.printStep("Waiting for another operation to complete at %v", addr)
 			}
 		}
 	}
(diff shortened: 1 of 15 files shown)


tool/gravity/cli/operation.go
```diff
@@ -197,59 +206,59 @@ func completeClusterOperationPlan(localEnv *localenv.LocalEnvironment, operation
...
+}
+
+func getActiveOperation(localEnv *localenv.LocalEnvironment, environ LocalEnvironmentFactory, operationID string) (*clusterOperation, error) {
+	operation, err := getLastOperation(localEnv, environ, operationID)
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
-	return op, nil
+	if operation.IsCompleted() {
+		return nil, operationNotFound{message: "no active operation found", existing: true}
```
r0mant, line 250: With this change, will we still be able to see a plan for a completed operation?

a-palchikov: When displaying the plan, this function is not used - `getLastOperation` is used instead.

tool/gravity/cli/operation.go
```diff
@@ -197,59 +206,59 @@ func completeClusterOperationPlan(localEnv *localenv.LocalEnvironment, operation
...
-	if len(operations) != 1 {
-		log.Infof("Multiple operations found: \n%v\n, please specify operation with --operation-id.\n"+
-			"Displaying the most recent operation.",
-			oplist(operations))
+	firstOperation := operations[len(operations)-1]
+	if firstOperation.Type == ops.OperationInstall && firstOperation.IsCompleted() {
+		return nil
 	}
-	return &operations[0], nil
+	return trace.NotFound("no install operation found")
```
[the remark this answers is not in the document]
a-palchikov: This function is not used - I'll remove it.

lib/expand/join.go
```diff
@@ -830,7 +830,8 @@ func (p *Peer) tryConnect(operationID string) (ctx *operationContext, err error)
 			return nil, utils.Abort(err)
 		}
 		if trace.IsCompareFailed(err) {
-			p.printStep("Waiting for another operation to finish at %v", addr)
+			p.WithError(err).Info("Failed to create expand operation.")
+			p.printStep("Waiting to create expand operation at %v: %v", addr, err)
```
knisbet, line 834: So I'm not sure how this displays when encountered, but if the error text sounds like an error I could see it causing confusion on how to proceed, even though it's likely just waiting for the cluster to complete it's operation.

knisbet (approved):

lib/expand/join.go
a-palchikov: Make sense - I'll provide a couple of deviations:
```
Waiting for cluster to become healthy
```
when the cluster is degraded and
```
Waiting for another operation to complete
```
for other cases (`cannot create operation for master with another node joining`, `cannot create operation for node while master is joining`, `join capacity is exceeded`)

lib/expand/join.go
```diff
@@ -830,7 +830,12 @@ func (p *Peer) tryConnect(operationID string) (ctx *operationContext, err error)
 			return nil, utils.Abort(err)
 		}
 		if trace.IsCompareFailed(err) {
-			p.printStep("Waiting for another operation to finish at %v", addr)
+			p.WithError(err).Info("Failed to create expand operation.")
```
r0mant, line 833: Nit: In a similar vein to what @knisbet suggested here before, maybe make this log message a bit less scary? Something like, "waiting on a precondition to create expand operation". Also, do we need full error trace here?

lib/expand/join.go
```diff
@@ -830,7 +830,12 @@ func (p *Peer) tryConnect(operationID string) (ctx *operationContext, err error)
...
+			if utils.IsClusterDegradedError(err) {
+				p.printStep("Waiting to cluster to become healthy")
```
r0mant, line 835: for cluster

Also, maybe elaborate a little bit, like "Cluster is currently degraded, waiting for it to become healthy"

lib/expand/join.go
```diff
@@ -830,7 +830,12 @@ func (p *Peer) tryConnect(operationID string) (ctx *operationContext, err error)
...
+			} else {
+				p.printStep("Waiting to another operation to complete at %v", addr)
```
r0mant, line 837: for another operation

r0mant (approved):

Issue #1503, opened 2020-05-06
[BUG] Can't set role_map on trusted cluster resource


r0mant:
**Describe the bug**

Currently, when creating a trusted cluster resource, Gravity always fills in the "roles" field with the default value if it's empty. This leads to an invalid trusted cluster configuration when trying to set "role_map" - because these two are mutually exclusive, if you set "role_map" and leave "roles" empty, Gravity will fill it out.

Gravity should only fill out "roles" if "role_map" is not provided.

**To Reproduce**

Create a trusted cluster with role_map e.g.

```yaml
spec:
  ...
  role_map:
  - remote: "*"
    local: "admin"
```

and observe errors in the logs and tsh/kubectl not working.

**Expected behavior**

Roles field should not be populated if role_map is present.

**Logs**

**Environment (please complete the following information):**
 - OS [e.g. Redhat 7.4]:
 - Gravity [e.g. 5.5.4]:
 - Platform [e.g. Vmware, AWS]:

**Additional context**

PR #1504, opened 2020-05-06
(6.3) Fix issue with trusted cluster role map

r0mant:
## Description

Currently Gravity will always auto-fill `roles` field on the trusted cluster resource if it's not set. It should not be populated if `role_map` is set because they are mutually exclusive.

## Type of change

* Closes https://github.com/gravitational/gravity/issues/1503.

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

Use trusted cluster resource:

```yaml
kind: trusted_cluster
version: v2
metadata:
  name: hub.gravitational.io
spec:
  enabled: true
  role_map:
  - remote: "viewer"
    local: ["viewer"]
  token: <snip>
  tunnel_addr: hub.gravitational.io:3024
  web_proxy_addr: hub.gravitational.io:32009
```

Before the change:

```yaml
ubuntu@node-2:~$ sudo gravity resource get trustedcluster hub.gravitational.io --format=yaml
kind: trusted_cluster
version: v2
metadata:
  name: hub.gravitational.io
spec:
  ...
  role_map:
  - local:
    - viewer
    remote: viewer
  roles:
  - '@teleadmin'
  ...
```

After the change:

```yaml
ubuntu@node-2:~$ sudo gravity resource get trustedcluster hub.gravitational.io --format=yaml
kind: trusted_cluster
version: v2
metadata:
  name: hub.gravitational.io
spec:
  ...
  role_map:
  - local:
    - viewer
    remote: viewer
  ...
```

## Additional information

Needs forward-ports to 7.0 and master.

### DIFF
--- a/lib/storage/trustedcluster.go
+++ b/lib/storage/trustedcluster.go
@@ -123,7 +123,7 @@ type TrustedClusterSpecV2 struct {
 	SNIHost string `json:"sni_host"`
 	// Roles is a list of roles that users will be assuming when connecting to
 	// this cluster
-	Roles []string `json:"roles"`
+	Roles []string `json:"roles,omitempty"`
 	// RoleMap specifies role mappings to remote roles
 	RoleMap teleservices.RoleMap `json:"role_map,omitempty"`
 	// PullUpdates indicates whether the trusted cluster should pull updates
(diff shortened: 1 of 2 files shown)


lib/storage/trustedcluster.go
```diff
@@ -323,13 +323,19 @@ func (c *TrustedClusterV2) CheckAndSetDefaults() error {
 	if c.Spec.ReverseTunnelAddress == "" {
 		return trace.BadParameter("tunnel_addr can't be empty")
 	}
+	if len(c.Spec.Roles) != 0 && len(c.Spec.RoleMap) != 0 {
+		return trace.BadParameter("either roles or role_map should be set, not both")
```
a-palchikov, line 327: Would it hurt if this reset the roles when a role map has been specified or there's no implicit precedence to decide which to use over which?

lib/storage/trustedcluster_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"gopkg.in/check.v1"
+)
+
+type TrustedClusterSuite struct{}
+
+var _ = check.Suite(&TrustedClusterSuite{})
+
+// TestTrustedCluster verifies basic trusted cluster resource parsing and
+// default field values.
+func (s *TrustedClusterSuite) TestTrustedCluster(c *check.C) {
```
a-palchikov, line 33: `TestTrustedClusterUsesRoleDefaults`

lib/storage/trustedcluster_test.go
```diff
@@ -0,0 +1,113 @@
...
+			},
+			{
+				Remote: "developer",
+				Local:  []string{"developer", "viewer"},
+			},
+		},
+	}))
+}
+
+// TestTrustedClusterRoles makes sure roles and role_map can't be both set.
```
a-palchikov, line 93: What about explicit roles configuration?

a-palchikov (approved):

lib/storage/trustedcluster.go
r0mant: Yeah, I would not make assumptions for the user to avoid surprises.

lib/storage/trustedcluster_test.go
r0mant: Added another test for this.

a-palchikov (approved):

Issue #1523, opened 2020-05-08
Bogus expand oprations resulting from a succesful install


a-palchikov:
## Current Result

Sometimes the installation will fail to properly shut down the installer service on joining nodes leading 
to bogus expand operations created as the services restart.

## Expected Result

There should be no expand operations created as a result of a successful installation.

Gravity Version: 7.0.x

Issue #1529, opened 2020-05-08
Crash report collection hangs long time on error


a-palchikov:
## Current Result
After an operation fails and the client collects the debug report, it hangs on waiting for the service to stop. The service will be restarted in case of an error, and due to a regression, the client will attempt to wait for service to stop before giving up after a timeout.

## Expected Result
Client exits immediately after collecting debug report.

Gravity Version:  7.0.x, 6.3.x

PR #1537, opened 2020-05-10
[7.0.x] service stop conditions

a-palchikov:
## Description
This PR fixes the following issues with the install/expand code:
  * only wait for service to stop on terminal conditions (i.e. aborted or completed installation) but not on transient stops like a failure
  * initiate operation in expand case only explicitly - on client request. This avoids accidental operations after completion.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1523, https://github.com/gravitational/gravity/issues/1529.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done
 * Installation on a 3-node cluster.
 * Installation on a 3-node cluster with simulated pre-flight failures to test resumption logic.
 * Repeated shrink/expand of a single node.

### DIFF
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -83,7 +83,7 @@ func NewPeer(config PeerConfig) (*Peer, error) {
 		closeC:    make(chan closeResponse),
 		connectC:  make(chan connectResult, 1),
 	}
-	peer.startConnectLoop()
+	// peer.startConnectLoop()
 	peer.startStatusLoop()
 	peer.startExecuteLoop()
 	peer.startReconnectWatchLoop()
@@ -116,6 +116,9 @@ type Peer struct {
 	execDoneC chan install.ExecResult
 	// wg is a wait group used to ensure completion of internal processes
 	wg sync.WaitGroup
+	// connectOnce enables the execute loop to start the connect loop
+	// only on the first execute request
+	connectOnce sync.Once
 }
 
 // Run runs the peer operation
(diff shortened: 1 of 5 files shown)


lib/expand/join.go
```diff
@@ -462,6 +452,7 @@ func (p *Peer) submit(req *installpb.ExecuteRequest) bool {
 // execute executes either the complete operation or a single phase specified with req
 func (p *Peer) execute(req *installpb.ExecuteRequest) (dispatcher.Status, error) {
 	p.WithField("req", req).Info("Execute.")
+	p.startConnectLoop()
```
r0mant, line 455: What was the problem with having this in NewPeer? It's a bit confusing now that some peer routines are started there and this one here.

lib/expand/join.go
```diff
@@ -142,6 +146,7 @@ func (p *Peer) Run(listener net.Listener) error {
 			p.WithError(err2).Warn("Failed to mark operation as failed.")
 		}
 	}
+	p.WithField("exit-error", err).Info("Exit with error.")
```
r0mant, line 149: Error or Warn?

lib/install/install.go
```diff
@@ -84,6 +84,7 @@ func (i *Installer) Run(listener net.Listener) error {
 	}()
 	err := <-i.errC
 	i.stop()
+	i.WithField("exit-error", err).Info("Exit with error.")
```
r0mant, line 87: Error or warn?

lib/expand/join.go
a-palchikov: Sorry - I've been meaning to add proper explanations.
The problem with connect loop starting in  `NewPeer` is that connect loop in its current form is a potentially mutating operation - i.e. it implicitly creates a cluster operation after successfully connecting to the controller. Since `NewPeer` is on the path of agent initialization, this can lead to undesired state if not properly controlled (which it wasn't when the agent could restart after a successfully completed operation resulting in bogus new and empty expand operations).

It also seemed to be clearer to move this behavior inside the part of code that handles client request to perform the operation thus making it explicit.

lib/expand/join.go
a-palchikov: It is more informational than it is an error or warning. I might even remove it as debugging aid.

r0mant (approved):

r0mant:
@a-palchikov I also have a couple of questions/notes regarding this PR:

* This is supposed to fix https://github.com/gravitational/gravity/issues/1523, right? Just double-checking since it's not mentioned in the description.
* According to the practices we established recently, once we've merged this to mitigate the issue, what do we do to make sure this does not regress again? Whatever it is, we should do it right away and not consider regression over until then.
* Related to the above, do you know why robotest didn't detect this issue? We're testing 3-node install and if the cluster stays "expanding" after install robotest should have caught it...

a-palchikov:
@r0mant - it ws mentioned in the PR description but had a typo in the reference URL - it mentioned 1532 instead of 1523, so yes - it's meant as an aid for that.
Regarding the regression testing - it did come up during testing as we saw just not reliably - i.e. it is somewhat racy. I'm not sure how to go about testing racy bugs - first thing that comes to mind is by increasing the number of test runs of the same scenario to increase the chances of something like that triggering.
We still have a lot to explore regarding the use and benefits of automatic fault injection which will help uncover a lot more bugs than we do with exercising the happy path.

wadells:
> Related to the above, do you know why robotest didn't detect this issue? We're testing 3-node install and if the cluster stays "expanding" after install robotest should have caught it...

> Regarding the regression testing - it did come up during testing as we saw just not reliably - i.e. it is somewhat racy. 

I agree with @a-palchikov. We did hit this ([example logs](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&minLogLevel=0&timestamp=2020-05-13T00:08:51.568000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%22776556d1-2a52-4168-a29d-17ff8aa5511d%22%0Alabels.__suite__%3D%22248439f5-9cfc-459d-b5eb-143b5d7c44ba%22&dateRangeEnd=2020-05-13T00:08:48.292Z&interval=PT1H&scrollTimestamp=2020-05-12T20:53:49.127493095Z&dateRangeUnbound=backwardInTime)) but it is racy.  However, the race doesn't matter for just the install test, because we don't currently check status in a way that would catch this.  In particular:

https://github.com/gravitational/robotest/blob/ba7f76b4b7d05f9994bb6fff7b40838214157503/suite/sanity/install.go#L59

maps to:

https://github.com/gravitational/robotest/blob/f27079bfac827d7fc272547377ac2c550a0a8856/infra/gravity/node_commands.go#L287-L304

Which only checks against degraded.

I just realized this at the end of the day yesterday, and will get a ticket & fix out.

Issue #1539, opened 2020-05-11
Expand operation sometimes non-resumable


a-palchikov:
If the expand operation fails after the teleport service has been installed, i.e.:
```
[ERROR]: failed to execute phase "/etcd" client: etcd cluster is unavailable or misconfigured; error #0: client: endpoint https://10.138.0.54:2379 exceeded header timeout ; error #1: client: endpoint https://10.138.0.50:2379 exceeded header timeout (client: etcd cluster is unavailable or misconfigured; error #0: client: endpoint https://10.138.0.54:2379 exceeded header timeout ; error #1: client: endpoint https://10.138.0.50:2379 exceeded header timeout )
```
the resumption will result in another error:
```
[ERROR]: rpc error: code = Unknown desc = One of existing servers already has hostname "robotest-8168c627-node-0": ["robotest-8168c627-node-1" "robotest-8168c627-node-2" "robotest-8168c627-node-0"]. gravity-agent.service: Main process exited, code=exited, status=255
gravity-agent.service: Failed with result 'exit-code'
```

PR #1542, opened 2020-05-11
[7.0.x] expand resume

a-palchikov:
## Description
This PR refines the logic of the controller to validate whether an agent represents a valid server in the context of an ongoing operation - otherwise, expanding breaks at various points into the operation. As the last example - immediately after failing to add an etcd member (due to a transient connection error), the agent was not able to rejoin on the grounds of there already being a server with the same hostname.
Instead, the controller will first check whether the server is part of the currently active operation (by comparing the server's advertise address against the operation server state).

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1539

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

 * Install a 3-node cluster.
 * Remove a node and re-add simulating a failure on `etcd` step. Make sure it can be resumed with a correct binary.

While testing, I ran into an issue with etcd and was wondering if it was just me or anyone else ran into it as well. Basically one of the test scenarios is removing and re-introducing a node into the cluster. If the issue is time-related at all - the remove/join would be several minutes apart tops. When joining the node, the etcd would frequently panic on data inconsistency. Here's a snippet from the etcd log that captures the panic:
```
ay 14 11:02:53 dmitri-centos-node-2 etcd[235]: raft2020/05/14 11:02:52 INFO: 3830d8d1ab5f5cd0 switched to configuration voters=()
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: raft2020/05/14 11:02:52 INFO: 3830d8d1ab5f5cd0 became follower at term 0
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: raft2020/05/14 11:02:52 INFO: newRaft 3830d8d1ab5f5cd0 [peers: [], term: 0, commit: 0, applied: 0, lastindex: 0, lastterm: 0]
May 14 11:02:52 dmitri-centos-node-2 etcd[235]: starting member 3830d8d1ab5f5cd0 in cluster 10c90c1dbfb0f41d
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: simple token is not cryptographically signed
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started HTTP pipelining with peer 66d74b9b9be85064
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started HTTP pipelining with peer 6c1ddcb80fda6234
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: starting peer 66d74b9b9be85064...
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started HTTP pipelining with peer 66d74b9b9be85064
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started streaming with peer 66d74b9b9be85064 (writer)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started streaming with peer 66d74b9b9be85064 (writer)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started peer 66d74b9b9be85064
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: added peer 66d74b9b9be85064
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: starting peer 6c1ddcb80fda6234...
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started streaming with peer 66d74b9b9be85064 (stream MsgApp v2 reader)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started HTTP pipelining with peer 6c1ddcb80fda6234
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started streaming with peer 66d74b9b9be85064 (stream Message reader)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started streaming with peer 6c1ddcb80fda6234 (writer)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started streaming with peer 6c1ddcb80fda6234 (writer)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started peer 6c1ddcb80fda6234
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: added peer 6c1ddcb80fda6234
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: starting server... [version: 3.4.7, cluster version: to_be_decided]
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started streaming with peer 6c1ddcb80fda6234 (stream MsgApp v2 reader)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: started streaming with peer 6c1ddcb80fda6234 (stream Message reader)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: peer 66d74b9b9be85064 became active
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: established a TCP streaming connection with peer 66d74b9b9be85064 (stream MsgApp v2 reader)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: ClientTLS: cert = /var/state/etcd.cert, key = /var/state/etcd.key, trusted-ca = /var/state/root.cert, client-cert-auth = true, crl-file = 
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: raft2020/05/14 11:02:53 INFO: 3830d8d1ab5f5cd0 [term: 0] received a MsgHeartbeat message with higher term from 66d74b9b9be85064 [term: 101]
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: raft2020/05/14 11:02:53 INFO: 3830d8d1ab5f5cd0 became follower at term 101
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: raft2020/05/14 11:02:53 tocommit(83262) is out of range [lastIndex(0)]. Was the raft log corrupted, truncated, or lost?
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: listening for peers on 10.128.0.104:2380
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: panic: tocommit(83262) is out of range [lastIndex(0)]. Was the raft log corrupted, truncated, or lost?
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: goroutine 162 [running]:
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: log.(*Logger).Panicf(0xc0001f6230, 0x10cbbbf, 0x5d, 0xc0012891a0, 0x2, 0x2)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]:         /usr/local/go/src/log/log.go:219 +0xc1
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: go.etcd.io/etcd/raft.(*DefaultLogger).Panicf(0x1aa7080, 0x10cbbbf, 0x5d, 0xc0012891a0, 0x2, 0x2)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]:         /tmp/etcd-release-3.4.7/etcd/release/etcd/raft/logger.go:127 +0x60
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: go.etcd.io/etcd/raft.(*raftLog).commitTo(0xc0002b8070, 0x1453e)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]:         /tmp/etcd-release-3.4.7/etcd/release/etcd/raft/log.go:203 +0x131
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: go.etcd.io/etcd/raft.(*raft).handleHeartbeat(0xc0000fa780, 0x8, 0x3830d8d1ab5f5cd0, 0x66d74b9b9be85064, 0x65, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]:         /tmp/etcd-release-3.4.7/etcd/release/etcd/raft/raft.go:1396 +0x54
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: go.etcd.io/etcd/raft.stepFollower(0xc0000fa780, 0x8, 0x3830d8d1ab5f5cd0, 0x66d74b9b9be85064, 0x65, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]:         /tmp/etcd-release-3.4.7/etcd/release/etcd/raft/raft.go:1341 +0x480
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: go.etcd.io/etcd/raft.(*raft).Step(0xc0000fa780, 0x8, 0x3830d8d1ab5f5cd0, 0x66d74b9b9be85064, 0x65, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]:         /tmp/etcd-release-3.4.7/etcd/release/etcd/raft/raft.go:984 +0x1113
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: go.etcd.io/etcd/raft.(*node).run(0xc000346840)
May 14 11:02:53 dmitri-centos-node-2 etcd[235]:         /tmp/etcd-release-3.4.7/etcd/release/etcd/raft/node.go:352 +0xab6
May 14 11:02:53 dmitri-centos-node-2 etcd[235]: created by go.etcd.io/etcd/raft.RestartNode
May 14 11:02:53 dmitri-centos-node-2 etcd[235]:         /tmp/etcd-release-3.4.7/etcd/release/etcd/raft/node.go:240 +0x33c
May 14 11:02:53 dmitri-centos-node-2 systemd[1]: etcd.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
May 14 11:02:53 dmitri-centos-node-2 systemd[1]: etcd.service: Failed with result 'exit-code'.
May 14 11:02:53 dmitri-centos-node-2 systemd[1]: Failed to start Etcd Service.
```

The behavior with the change now is continuously retry the add etcd peer step which in this case would be failing with `cluster has not leader` until timeout.
To resolve, I had to manually remove the member from another node:
```
$ etcdctl member remove <failing-member-id>
```
remove the data directory on the joining node:
```
$ rm -r /var/lib/gravity/planet/etcd/v<etcd-version>
```
and rerun the step:
```
$ ./gravity resume
```

### DIFF
--- a/build.assets/robotest_run_suite.sh
+++ b/build.assets/robotest_run_suite.sh
@@ -7,7 +7,7 @@ readonly UPGRADE_FROM_DIR=${1:-$(pwd)/../upgrade_from}
 declare -A UPGRADE_MAP
 
 # latest patch release on this branch, keep this up to date
-UPGRADE_MAP[7.0.3]="ubuntu:18"
+UPGRADE_MAP[7.0.5]="ubuntu:18"
 
 # latest patch release on compatible LTS, keep this up to date
 UPGRADE_MAP[6.1.24]="ubuntu:18"
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -146,8 +146,8 @@ func (p *Peer) Run(listener net.Listener) error {
 				p.WithError(err).Warn("Failed to leave cluster.")
 			}
 		} else if !installpb.IsCompletedError(err) {
-			if err2 := p.fail(err.Error()); err2 != nil {
-				p.WithError(err2).Warn("Failed to mark operation as failed.")
+			if err := p.fail(err.Error()); err != nil {
+				p.WithError(err).Warn("Failed to mark operation as failed.")
 			}
 		}
 	}
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -76,12 +76,13 @@ func NewPeer(config PeerConfig) (*Peer, error) {
 		server:     server,
 		dispatcher: dispatcher,
 		// Account for agent exit or agent reconnect failure
-		errC:      make(chan error, 3),
-		exitC:     make(chan error, 1),
-		execC:     make(chan *installpb.ExecuteRequest),
-		execDoneC: make(chan install.ExecResult, 1),
-		closeC:    make(chan closeResponse),
-		connectC:  make(chan connectResult, 1),
+		errC:        make(chan error, 3),
+		exitC:       make(chan error, 1),
+		execC:       make(chan *installpb.ExecuteRequest),
+		execDoneC:   make(chan install.ExecResult, 1),
+		closeC:      make(chan closeResponse),
+		connectC:    make(chan connectResult, 1),
+		connectingC: make(chan struct{}),
 	}
 	peer.startExecuteLoop()
 	peer.startReconnectWatchLoop()
@@ -95,13 +96,10 @@ type Peer struct {
 	ctx context.Context
 	// cancel cancels internal operation
 	cancel context.CancelFunc
-	// errC is signaled when either the service or agent is aborted
+	// errC is signaled when agent aborts or exits
 	errC chan error
 	// exitC is signaled when the service exits
 	exitC chan error
-	// connectC receives the results of connecting to either wizard
-	// or cluster controller
-	connectC chan connectResult
 	// server is the gRPC installer server
 	server     *server.Server
 	dispatcher dispatcher.EventDispatcher
@@ -114,6 +112,11 @@ type Peer struct {
 	execDoneC chan install.ExecResult
 	// wg is a wait group used to ensure completion of internal processes
 	wg sync.WaitGroup
+	// connectC receives the results of connecting to either wizard
+	// or cluster controller
+	connectC chan connectResult
+	// connectingC is closed once the connect loop starts running
+	connectingC chan struct{}
 	// connectOnce enables the execute loop to start the connect loop
 	// only on the first execute request
 	connectOnce sync.Once
@@ -135,18 +138,19 @@ func (p *Peer) Run(listener net.Listener) error {
 	}
 	// Stopping is on best-effort basis, the client will be trying to stop the service
 	// if notified
-	p.stop()
-	if installpb.IsAbortedError(err) {
-		if err := p.leave(); err != nil {
-			p.WithError(err).Warn("Failed to leave cluster.")
-		}
-	}
-	if err != nil && !installpb.IsCompletedError(err) {
-		if err2 := p.fail(err.Error()); err2 != nil {
-			p.WithError(err2).Warn("Failed to mark operation as failed.")
-		}
-	}
 	p.WithField("exit-error", err).Info("Exit with error.")
+	p.stop()
+	if err != nil {
+		if installpb.IsAbortedError(err) {
+			if err := p.leave(); err != nil {
+				p.WithError(err).Warn("Failed to leave cluster.")
+			}
+		} else if !installpb.IsCompletedError(err) {
+			if err2 := p.fail(err.Error()); err2 != nil {
+				p.WithError(err2).Warn("Failed to mark operation as failed.")
+			}
+		}
+	}
 	return installpb.WrapServiceError(err)
 }
 
@@ -188,7 +192,7 @@ func (p *Peer) Execute(req *installpb.ExecuteRequest, stream installpb.Agent_Exe
 			if result.CompletionEvent != nil {
 				err := stream.Send(result.CompletionEvent.AsProgressResponse())
 				if err != nil {
-					return trace.Wrap(err)
+					return err
 				}
 			}
 			return nil
@@ -339,6 +343,7 @@ func (c *PeerConfig) CheckAndSetDefaults() (err error) {
 
 func (p *Peer) startConnectLoop() {
 	p.connectOnce.Do(func() {
+		close(p.connectingC)
 		p.wg.Add(1)
 		go func() {
 			defer p.wg.Done()
@@ -797,11 +802,11 @@ func (p *Peer) stop() {
 }
 
 func (p *Peer) shutdownAgent(ctx context.Context) error {
-	opCtx, err := p.operationContext(ctx)
+	opCtx, err := p.maybeOperationContext(ctx)
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	if opCtx.agent == nil {
+	if opCtx == nil || opCtx.agent == nil {
 		return nil
 	}
 	p.Info("Stop peer agent.")
@@ -996,9 +1001,13 @@ func (p *Peer) fail(message string) error {
 }
 
 func (p *Peer) failOperation(ctx context.Context, message string) error {
-	opCtx, err := p.operationContext(ctx)
+	opCtx, err := p.maybeOperationContext(ctx)
 	if err != nil {
 		return trace.Wrap(err)
+	}
+	if opCtx == nil {
+		// No operation to fail
+		return nil
 	}
 	return ops.FailOperation(ctx, opCtx.Operation.Key(), opCtx.Operator, message)
 }
@@ -1085,6 +1094,15 @@ func (p *Peer) emitAuditEvent(ctx operationContext) error {
 	events.Emit(p.ctx, ctx.Operator, events.OperationExpandStart,
 		events.FieldsForOperation(*operation))
 	return nil
+}
+
+func (p *Peer) maybeOperationContext(ctx context.Context) (*operationContext, error) {
+	select {
+	case <-p.connectingC:
+		return p.operationContext(ctx)
+	default:
+		return nil, nil
+	}
 }
 
 func (p *Peer) operationContext(ctx context.Context) (*operationContext, error) {
--- a/lib/expand/phases/etcd.go
+++ b/lib/expand/phases/etcd.go
@@ -79,6 +79,7 @@ func NewEtcd(p fsm.ExecutorParams, operator ops.Operator, runner rpc.AgentReposi
 		Runner:         runner,
 		Master:         *p.Phase.Data.Master,
 		ExecutorParams: p,
+		etcdPeerURL:    p.Phase.Data.Server.EtcdPeerURL(),
 	}, nil
 }
 
@@ -93,6 +94,8 @@ type etcdExecutor struct {
 	Master storage.Server
 	// ExecutorParams is common executor params
 	fsm.ExecutorParams
+	// etcdPeerURL specifies this node's etcd peer URL
+	etcdPeerURL string
 }
 
 // Execute adds the joining node to the cluster's etcd cluster
@@ -147,10 +150,10 @@ func (p *etcdExecutor) addEtcdMember(ctx context.Context) (member *etcd.Member, 
 			return trace.Wrap(err)
 		}
 		if p.hasSelfAsMember(peers) {
-			p.Infof("Etcd peer %v already exists.", p.Phase.Data.Server.EtcdPeerURL())
+			p.WithField("peerURL", p.etcdPeerURL).Info("Node is already etcd peer.")
 			return nil
 		}
-		member, err = p.Etcd.Add(ctx, p.Phase.Data.Server.EtcdPeerURL())
+		member, err = p.Etcd.Add(ctx, p.etcdPeerURL)
 		return trace.Wrap(err)
 	})
 	if err != nil {
@@ -160,13 +163,11 @@ func (p *etcdExecutor) addEtcdMember(ctx context.Context) (member *etcd.Member, 
 }
 
 func (p *etcdExecutor) hasSelfAsMember(peers []etcd.Member) bool {
-	peerURL := p.Phase.Data.Server.EtcdPeerURL()
 	for _, peer := range peers {
-		if len(peer.PeerURLs) != 1 {
-			continue
-		}
-		if peer.PeerURLs[0] == peerURL {
-			return true
+		for _, peerURL := range peer.PeerURLs {
+			if peerURL == p.etcdPeerURL {
+				return true
+			}
 		}
 	}
 	return false
--- a/lib/expand/phases/etcd.go
+++ b/lib/expand/phases/etcd.go
@@ -145,8 +145,11 @@ func (p *etcdExecutor) addEtcdMember(ctx context.Context) (member *etcd.Member, 
 	err = utils.RetryTransient(ctx, boff, func() error {
 		var err error
 		member, err = p.Etcd.Add(ctx, p.Phase.Data.Server.EtcdPeerURL())
-		if err != nil && !isMemberAlreadyExistsError(err) {
-			return trace.Wrap(err)
+		if err != nil {
+			if !isMemberAlreadyExistsError(err) {
+				return trace.Wrap(err)
+			}
+			p.Infof("Etcd peer %v already exists.", p.Phase.Data.Server.EtcdPeerURL())
 		}
 		return nil
 	})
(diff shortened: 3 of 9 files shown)


lib/install/client/install.go
```diff
@@ -146,12 +146,14 @@ type InstallerStrategy struct {
...
 
-// isServiceFailed returns an error if the service has failed.
-func isServiceFailed(serviceName string) func() error {
+// shouldReconnectService returns an error if the service has failed or has been uninstalled
+func shouldReconnectService(serviceName string) func() error {
 	return func() error {
-		err := service.IsFailed(serviceName)
+		err := service.IsStatus(serviceName,
+			systemservice.ServiceStatusFailed,
+			systemservice.ServiceStatusUnknown)
```
a-palchikov, line 154: Added the unknown status to catch uninstall events since the dialer runs the connect loop separate from the service uninstall logic and this way will cancel connect attempts at the end of the completed operation faster.

lib/ops/opsservice/agents.go
```diff
@@ -581,6 +566,47 @@ func (r *AgentPeerStore) addGroup(key ops.SiteOperationKey) (*agentGroup, error)
...
+	op, err := r.backend.GetSiteOperation(token.SiteDomain, token.OperationID)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	operation := (ops.SiteOperation)(*op)
+	if operation.IsCompleted() {
+		return trace.BadParameter("operation %v is already completed", token.OperationID)
+	}
+	serverAddr := utils.ExtractHost(addr)
+	if (storage.Servers)(op.Servers).FindByIP(serverAddr) == nil {
```
a-palchikov, line 579: This test (checking only the advertised address)  is simple on purpose - more validations are run after this check has completed. It serves to decouple the teleport membership test which would not work if the node already had teleport set up.

lib/ops/opsservice/agents.go
```diff
@@ -581,6 +566,47 @@ func (r *AgentPeerStore) addGroup(key ops.SiteOperationKey) (*agentGroup, error)
 	return agentGroup, nil
 }
 
+func (r *AgentPeerStore) isPartOfActiveOperation(addr string, token storage.ProvisioningToken) error {
+	op, err := r.backend.GetSiteOperation(token.SiteDomain, token.OperationID)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	operation := (ops.SiteOperation)(*op)
+	if operation.IsCompleted() {
```
r0mant, line 575: Is this going to play well with the fact that we always set install operation to completed when transferring it to etcd even if install is still going?

lib/ops/opsservice/agents.go
```diff
@@ -442,47 +442,32 @@ func (r *AgentPeerStore) authenticatePeer(token string) (*storage.ProvisioningTo
...
-
-	var existingServers []string
-	for _, server := range servers {
-		hostname := server.GetLabels()[ops.Hostname]
-		if hostname == "" {
-			log.Warnf("Server hostname is empty: %+v.", server)
-			continue
+func (r *AgentPeerStore) checkHostname(ctx context.Context, group *agentGroup, addr, hostname string, token storage.ProvisioningToken) error {
+	if err := r.isPartOfActiveOperation(addr, token); err != nil {
+		r.WithError(err).Warn("Failed to check whether the server is part of the active operation.")
```
r0mant, line 463: Nit: This will say "failed to check" even if the check actually succeeded but returned NotFound (which is an expected scenario).

lib/ops/opsservice/agents.go
```diff
@@ -581,6 +566,47 @@ func (r *AgentPeerStore) addGroup(key ops.SiteOperationKey) (*agentGroup, error)
...
 }
 
+func (r *AgentPeerStore) isPartOfActiveOperation(addr string, token storage.ProvisioningToken) error {
+	op, err := r.backend.GetSiteOperation(token.SiteDomain, token.OperationID)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	operation := (ops.SiteOperation)(*op)
+	if operation.IsCompleted() {
+		return trace.BadParameter("operation %v is already completed", token.OperationID)
```
r0mant, line 576: Nit: Specify which kind of operation?

lib/ops/opsservice/agents.go
r0mant, line 579: Nit: I would just implement `FindByIP` on the `op.Servers` too (even if it would just do the same type conversion).

lib/install/client/install.go
r0mant: Would not this cause a sort of race condition issue where say a service is not being uninstalled, but rather not started yet? Not sure if this makes sense, I guess it depends on when "unknown" is actually being returned, just trying to think through various edge cases.

lib/expand/phases/etcd.go
```diff
@@ -137,6 +139,23 @@ func (p *etcdExecutor) Rollback(ctx context.Context) error {
 	return nil
 }
 
+func (p *etcdExecutor) addEtcdMember(ctx context.Context) (member *etcd.Member, err error) {
+	boff := backoff.NewExponentialBackOff()
+	boff.MaxElapsedTime = defaults.TransientErrorTimeout
+	err = utils.RetryTransient(ctx, boff, func() error {
+		var err error
+		member, err = p.Etcd.Add(ctx, p.Phase.Data.Server.EtcdPeerURL())
+		if err != nil && !isMemberAlreadyExistsError(err) {
```
r0mant, line 148: How about we at least log "already exists" case?

lib/ops/opsservice/agents.go
r0mant, line 579: This is only used for install/expand, right? Because IIRC not all operations populate Servers field.

lib/install/client/install.go
a-palchikov: Once the service has been installed, it cannot be in `unknown` state but I agree it is also very vague - I'll see if I can fetch more details to make a better judgement whether the service is still installed.

a-palchikov: On another thought there is not much merit in considering the additional status. I'll go for explicitly canceling the dialing loop when necessary with no changes to this logic.

build.assets/robotest_run_suite.sh
```diff
@@ -7,7 +7,7 @@ readonly UPGRADE_FROM_DIR=${1:-$(pwd)/../upgrade_from}
 declare -A UPGRADE_MAP
 
 # latest patch release on this branch, keep this up to date
-UPGRADE_MAP[7.0.3]="ubuntu:18"
+UPGRADE_MAP[7.0.5]="ubuntu:18"
```
a-palchikov, line 10: Bumped per comment.

lib/expand/join.go
```diff
@@ -114,6 +112,11 @@ type Peer struct {
 	execDoneC chan install.ExecResult
 	// wg is a wait group used to ensure completion of internal processes
 	wg sync.WaitGroup
+	// connectC receives the results of connecting to either wizard
+	// or cluster controller
+	connectC chan connectResult
+	// connectingC is closed once the connect loop starts running
```
a-palchikov, line 118: I introduced this knob to handle the case with now delayed connect loop and possibility of agent using the operation context which is not available in this case. The operation context is required to know the operation details to be able to set the operation into failed state on exiting with error and with the operation not completed.
Since now, the service starts but does not automatically connect to either the installer or cluster controller unless explicitly requested by the client (effectively by one of  `./gravity join` or `./gravity resume`), the operation context is empty.
If you were to immediately stop the service in this state (i.e. with no connect loop running), it would get stuck waiting on the context to become available. To workaround this, I notify that the operation context should be available with this channel by closing it. While it is not closed, the service will not attempt to use the operation context.

lib/install/client/install.go
```diff
@@ -146,16 +146,19 @@ type InstallerStrategy struct {
...
 
-// isServiceFailed returns an error if the service has failed.
-func isServiceFailed(serviceName string) func() error {
+// shouldReconnectService returns a function that determines whether the client should continue
+// reconnecting to the service given with serviceName
+func shouldReconnectService(serviceName string) func() error {
 	return func() error {
-		err := service.IsFailed(serviceName)
+		err := service.IsStatus(serviceName,
+			systemservice.ServiceStatusFailed)
```
a-palchikov, line 154: Removed the `unknown` state from the list as it is really uncertain whether selecting on this active state is worth it or would make the selector (`shouldReconnectService`) racy in any way.

lib/install/utils.go
```diff
@@ -381,19 +380,7 @@ type ExecResult struct {
...
-			detail := trace.UserMessage(err.OrigError())
-			if detail != userMessage {
-				userMessage = fmt.Sprintf("%v (%v)", userMessage, detail)
-			}
-		}
-		return userMessage
-	default:
-		return trace.UserMessage(err)
-	}
+	return trace.UserMessage(err)
```
a-palchikov, line 383: Removed this as it would end up duplicating the error message here:
```go
userMessage = fmt.Sprintf("%v (%v)", userMessage, detail)
```
This [commit](https://github.com/gravitational/trace/pull/56/commits/1ad036c6ae5843db826b0b95a5aaba07bf8fbd65) in trace should render this unnecessary.

knisbet (approved):

lib/expand/phases/etcd.go
```diff
@@ -283,3 +305,9 @@ func opKey(plan storage.OperationPlan) ops.SiteOperationKey {
 		OperationID: plan.OperationID,
 	}
 }
+
+func isMemberAlreadyExistsError(err error) bool {
+	// https://github.com/etcd-io/etcd/blob/v3.4.7/etcdserver/api/membership/errors.go#L27
+	const errPeerURLExists = "membership: peerURL exists"
```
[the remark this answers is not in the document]
a-palchikov: Yes - thanks for heads up. I changed the implementation to list the members instead of hardcoding the error.

lib/expand/phases/etcd.go
```diff
@@ -137,6 +138,40 @@ func (p *etcdExecutor) Rollback(ctx context.Context) error {
...
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	return member, nil
+}
+
+func (p *etcdExecutor) hasSelfAsMember(peers []etcd.Member) bool {
+	peerURL := p.Phase.Data.Server.EtcdPeerURL()
+	for _, peer := range peers {
+		if len(peer.PeerURLs) != 1 {
```
r0mant, line 165: Why not just iterate over all peer URLs?

lib/expand/phases/etcd.go
```diff
@@ -137,6 +138,40 @@ func (p *etcdExecutor) Rollback(ctx context.Context) error {
...
+func (p *etcdExecutor) addEtcdMember(ctx context.Context) (member *etcd.Member, err error) {
+	boff := backoff.NewExponentialBackOff()
+	boff.MaxElapsedTime = defaults.TransientErrorTimeout
+	err = utils.RetryTransient(ctx, boff, func() error {
+		peers, err := p.Etcd.List(ctx)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		if p.hasSelfAsMember(peers) {
+			p.Infof("Etcd peer %v already exists.", p.Phase.Data.Server.EtcdPeerURL())
```
r0mant, line 150: Nit: Use field for peer URL?

lib/expand/join.go
```diff
@@ -135,18 +138,19 @@ func (p *Peer) Run(listener net.Listener) error {
...
-	if err != nil && !installpb.IsCompletedError(err) {
-		if err2 := p.fail(err.Error()); err2 != nil {
-			p.WithError(err2).Warn("Failed to mark operation as failed.")
+	if err != nil {
+		if installpb.IsAbortedError(err) {
+			if err := p.leave(); err != nil {
+				p.WithError(err).Warn("Failed to leave cluster.")
+			}
+		} else if !installpb.IsCompletedError(err) {
+			if err2 := p.fail(err.Error()); err2 != nil {
```
r0mant, line 149: Nit: I think you don't necessarily need to have to use "err2" as it won't overshadow the original err until after p.fail returns, so `err := p.fail(err.Error()); err != nil` should work fine too.

r0mant (approved):

PR #1550, opened 2020-05-12
[master] implement container scanning

knisbet:
## Description

This implements internal changes to gravity tooling, to be able to push all container artifacts from a release to a docker repository for scanning purposes. This is meant to work in conjunction with a repository that is able to scan and alert on vulnerabilities within those containers.

Also fixed are a couple glitches in our registry code when working with public registries, to ensure we correctly indicate which repositories and actions we need permissions for, to allow the registry to correctly embed those permissions within our oauth token. 

## Type of change

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
    - [x] tested quay.io
    - [x] tested gcr
    - [x] tested registry:2 with basic auth
    - [ ] openshift (@r0mant not sure how you tested openshift previously??)
- [ ] Reconfigure build system to pass scanning repository variables to builds
- [ ] Address review feedback

## Implementation

1. Currently, I've wired it in a way that publishes the containers for scanning when publishing a gravity release. This is a little less than ideal, as ideally we'd be scanning regularly, such as every day, and staging the security fixes like normal PRs. Eventually we'll do this, but to get scanning up and running in the short term, it was chosen to wire the scanning into the publishing pipeline.

2. For scanning purposes, I want to push all the images to the same repository, that is pre-configured for scanning. This means to preserve the original repository information I had to get a bit clever and embed this information within the tags, along with the build version. This way, using just the scan results, we're able to work our way backwards and identify the gravity version and images that contain the vulnerability. As an example quay.io/org/example:1.0.0 in gravity 2.0.0 will be rewritten to quay.io/gravitational/gravity-scan: 2.0.0_org_example_1.0.0.

3. The existing Sync function on the imageService performs very similarly to the new SyncForScanning function, however, I'm not convinced the two functions can be cleanly merged, which is why I left the separate but very similar implementation.

4. Within the imageservice, I also ended up writing a wrapper for dockers token handler. The underlying token handler doesn't appear to support adding additional scopes after initialization (not needed for something like `docker push <image>`). So this wrapper exists to re-initialize dockers token handler, if the imageservice starts interacting with a repository that wasn't previously included. 

5. Because this code is intended for internal use only, the CLI options have been marked hidden, and should have no UX impact.

## Testing done

I did my best to test against public repositories, as well as the existing docker registry code.

As an example: 
`TELE_COPY_TO_USER=<user> TELE_COPY_TO_PASS=<pass> make scan-artifacts` to scan with quay.io.

GCR requires some work to use token based authentication: https://cloud.google.com/container-registry/docs/advanced-authentication#token


lib/docker/imageservice.go
```diff
@@ -295,6 +310,86 @@ func (r *imageService) Sync(ctx context.Context, dir string, progress utils.Prin
 	return installedTags, nil
 }
 
+// SyncForScanning syncs all images in a local directory, with a remote registry that is capable of scanning and
+// reporting vulnerabilities in those images. All images and tags are written into a designated repository, with the
+// image tag rewritten to help identify the image.
+// Example of gravity 7.0.0:
+//   quay.io/gravitational/debian-tall:0.0.1 -> quay.io/gravitational/gravity-scan:7.0.0_gravitational_debian-tall_0.0.1
```
a-palchikov, line 317: If I'm not mistaken, the distribution name reference format allows [multiple name components](https://github.com/docker/distribution/blob/v2.7.1/reference/regexp.go#L53) so it could be possible to wrap the scan image under the scan sub-group:
```
quay.io/gravitational/gravity-scan/7.0.0/debian-tall:0.0.1
```
Could be easier to work with.

lib/docker/imageservice.go
```diff
@@ -364,7 +460,7 @@ func newCertPool(CAFiles []string) (*x509.CertPool, error) {
 	return certPool, nil
 }
 
-func initTransport(req RegistryConnectionRequest) (http.RoundTripper, string, error) {
+func initTransport(req RegistryConnectionRequest) (http.RoundTripper, string, *multiScopeTokenHandler, error) {
```
a-palchikov, line 463: Frankly, I would put the result into some registryConnectResponse:
```go
type registryConnectResponse struct {
	transport    http.Transport
	registrAddr  string
	tokenHandler *multiScopeTokenHandler
}
```
or returned a partially initialized `remoteStore` - since it's what `initTransport` is for in the first place.

lib/docker/imageservice.go
```diff
@@ -646,3 +747,84 @@ func (s *remoteStore) updateRepo(ctx context.Context, remote, local distribution
...
+// to be added to an existing client. This works by re-instantiating the wrapped token handler if there
+// are additional scopes (repositories) that need to be added to the oauth token.
+type multiScopeTokenHandler struct {
+	tokenHandler registryauth.AuthenticationHandler
+	scopes       []registryauth.RepositoryScope
+	transport    http.RoundTripper
+	credentials  credentials
+}
+
+func (th *multiScopeTokenHandler) Scheme() string {
```
a-palchikov, line 761: is this implementing some external interface - would be good to know why APIs are exported?

lib/docker/imageservice.go
```diff
@@ -646,3 +747,84 @@ func (s *remoteStore) updateRepo(ctx context.Context, remote, local distribution
...
+			Transport:     th.transport,
+			Credentials:   th.credentials,
+			Scopes:        scopes,
+			OfflineAccess: false,
+			ClientID:      "gravity",
+		},
+	)
+}
+
+func mergeRepositoryScopeActions(left, right registryauth.RepositoryScope) (changed bool, scope registryauth.RepositoryScope) {
```
a-palchikov, line 808: You can do away with additional flags:
```go
func mergeRepositoryScopeActions(left, right registryauth.RepositoryScope) (changed bool, scope registryauth.RepositoryScope) {
	var newActions []string
L:
	for _, rightAction := range right.Actions {
		for _, leftAction := range left.Actions {
			if rightAction == leftAction {
				continue L
			}
		}
		newActions = append(newActions, rightAction)
	}

	if len(newActions) != 0 {
		return true, registryauth.RepositoryScope{
			Repository: left.Repository,
			Class:      left.Class,
			Actions:    append(left.Actions, newActions...),
		}
	}
	return false, left
}
```

lib/docker/imageservice_test.go
```diff
@@ -82,6 +83,94 @@ func (r *ImageServiceSuite) TestListsRepos(c *C) {
 	c.Assert(repos, DeepEquals, []string{"a", "b", "c", "d", "e"})
 }
 
+func (r *ImageServiceSuite) TestMergingScopeActions(c *C) {
+	cases := []struct {
+		scopes   []registryauth.RepositoryScope
+		expected []registryauth.RepositoryScope
+	}{
+		{
+			// merging multiple actions into the same scope
```
a-palchikov, line 92: can you turn the comment into code so failing tests can be easily identified:
```go
comment: "merging multiple actions into the same scope",
...
comment := check.Commentf(tc.comment)
c.Assert(th.scopes, DeepEquals, tt.expected, comment)
```

lib/docker/imageservice.go
knisbet: While allowed by standard, my understanding is most public registries only allow one layer to the naming scheme. So <base>/<org>/<image> (quay.io/gravitational/example) so we would only be able to support this if we were using our own registry.

lib/docker/imageservice.go
knisbet: yes, it's an internal interface within docker.

lib/docker/imageservice_test.go
knisbet: This particular asset library doesn't support comments on DeepEquals (unexpected number of args)

Makefile
```diff
@@ -68,6 +68,15 @@ DEV_PLANET ?=
 OS := $(shell uname | tr '[:upper:]' '[:lower:]')
 ARCH := $(shell uname -m)
 
+# Image Vulnerability Scanning
+# The following variables are used to copy all docker images from a cluster image to a docker repository 
+# that is able to scan and report on those images
+TELE_COPY_TO_REGISTRY ?= quay.io/gravitational
```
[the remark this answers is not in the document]
knisbet: Yea, I was a bit on the fence, although the reason we're wiring it up is to scan the images, realistically what it's actually doing is copying the image to a registry.

lib/app/service/docker.go
```diff
@@ -137,7 +137,12 @@ func (r *layerExporter) pushCmd(name, tag string) error {
 		Tag:  tag,
 	}
 	r.Infof("Pushing %v.", opts)
-	return r.dockerClient.PushImage(opts, dockerapi.AuthConfiguration{})
+	// Workaround a registry issue after updating go-dockerclient, set the password field to an invalid value so the
+	// auth headers are set.
+	// https://github.com/moby/moby/issues/10983
```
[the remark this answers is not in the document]
knisbet: It's something that's seen within go-dockerclient which is the first approach I tried at pushing these images, when talking to unauthenticated registries like the one we setup for image vendoring with some combination of vendoring updates. Even though the registry is unauthenticated, it will deny access if the auth header is missing. 

It's not really needed for this PR, but I figured I should leave it in because someone will inevitably hit this problem in the future.

lib/app/service/sync.go
```diff
@@ -133,6 +136,13 @@ func SyncApp(ctx context.Context, req SyncRequest) error {
 
 	log.Infof("Syncing %v.", req.Package)
 
+	if req.ScanConfig != nil {
+		if err = req.ImageService.SyncForScanning(ctx, *req.ScanConfig, syncPath, req.Progress); err != nil {
```
r0mant, line 140: I think it would be cleaner if you had a separate implementation of ImageService that would reimplement "Sync" method so it would properly sync images for scanning, and just pass it here from the caller - then this method won't need any updating at all, and no new ScanConfig parameters would be needed, so the SyncApp would be completely agnostic to what kind of syncing it should do and just use the passed in image service.

lib/docker/api.go
```diff
@@ -51,6 +52,10 @@ type ImageService interface {
 	// Returns the list of images synced
 	Sync(ctx context.Context, dir string, progress utils.Printer) ([]TagSpec, error)
 
+	// SyncForScanning syncs all images in a local directory, with a remote registry that is capable of scanning and
+	// reporting vulnerabilities in those images
+	SyncForScanning(ctx context.Context, conf ScanningConfig, dir string, progress utils.Printer) (err error)
```
r0mant, line 57: This interface extension also wouldn't be needed if we just provide an alternative implementation of ImageService.

lib/docker/imageservice.go
```diff
@@ -400,49 +409,51 @@ func initTransport(req RegistryConnectionRequest) (http.RoundTripper, string, er
...
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
 
 	return &remoteStore{
-		FieldLogger: log.WithField("remote registry", registryAddr),
-		addr:        registryAddr,
-		transport:   transport,
-		registry:    registry,
+		FieldLogger:  log.WithField("registry_addr", registryAddr),
```
a-palchikov, line 452: nit: `registry-addr`?

lib/docker/imageservice_test.go
a-palchikov: hmm, works for me:
```go
func (r *ImageServiceSuite) TestMergingScopeActions(c *C) {
	cases := []struct {
		comment  string
		scopes   []registryauth.RepositoryScope
		expected []registryauth.RepositoryScope
	}{
		{
			comment: "merging multiple actions into the same scope",
			scopes: []registryauth.RepositoryScope{
				{
					Repository: "a",
					Class:      "a",
					Actions:    []string{"push"},
				},
				{
					Repository: "a",
					Class:      "a",
					Actions:    []string{"pull"},
				},
			},
			expected: []registryauth.RepositoryScope{
				{
					Repository: "a",
					Class:      "a",
					Actions:    []string{"push", "pull"},
				},
			},
		},
		{
			comment: "more that one scope added to the list without merging",
			scopes: []registryauth.RepositoryScope{
				{
					Repository: "a",
					Class:      "a",
					Actions:    []string{"push"},
				},
				{
					Repository: "b",
					Class:      "b",
					Actions:    []string{"pull"},
				},
			},
			expected: []registryauth.RepositoryScope{
				{
					Repository: "a",
					Class:      "a",
					Actions:    []string{"push"},
				},
				{
					Repository: "b",
					Class:      "b",
					Actions:    []string{"pull"},
				},
			},
		},
		{
			comment: "adding the same scope more than once should only include it once",
			scopes: []registryauth.RepositoryScope{
				{
					Repository: "a",
					Class:      "a",
					Actions:    []string{"push"},
				},
				{
					Repository: "a",
					Class:      "a",
					Actions:    []string{"push"},
				},
			},
			expected: []registryauth.RepositoryScope{
				{
					Repository: "a",
					Class:      "a",
					Actions:    []string{"push"},
				},
			},
		},
	}

	for _, tt := range cases {
		comment := Commentf(tt.comment)
		th := &multiScopeTokenHandler{}
		for _, scope := range tt.scopes {
			th.AddScope(scope)
		}

		c.Assert(th.scopes, DeepEquals, tt.expected, comment)
	}
}
```

lib/docker/scanningimageservice.go
```diff
@@ -0,0 +1,148 @@
...
+	registryauth "github.com/docker/distribution/registry/client/auth"
+	"github.com/gravitational/gravity/lib/utils"
+	"github.com/gravitational/trace"
+)
+
+// ScanConfig represents configuration used to push images to a docker repository configured for scanning those
+// images for vulnerabilities or other problems
+type ScanConfig struct {
+	RemoteRepository string
+	TagPrefix        string
```
a-palchikov, line 32: Would be good to have the prefix field documented.

a-palchikov (approved):

lib/docker/imageservice_test.go
knisbet: Ok, it would seem to be a mistake on my end, the library produces an unexpected error if the comment is passed as a string, indicating the wrong number of arguments, which I incorrectly assumed to be an issue with the library not allowing a comment.

a-palchikov (approved):

PR #1551, opened 2020-05-12
[6.1.x] loosen same OS requirement

a-palchikov:
## Description

This PR loosens the same OS requirement preflight check with the following semantics:

  * OS distributions are the same if both name and versions are exact match
  * OS distributions are compatible (and check succeeds) if name is an exact match and versions match on the prefix up to the first '.'
  * otherwise, OS distributions are not compatible

Also, this has been brought up multiple times (frequently in the form of asking whether it's possible to suppress just the OS check).

## Type of change

Backport of https://github.com/gravitational/gravity/pull/1484.

Refs https://github.com/gravitational/gravity/issues/1112.

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

Unit-tested change.
****

### DIFF
--- a/lib/checks/checks.go
+++ b/lib/checks/checks.go
@@ -756,28 +756,58 @@ func checkRAM(info ServerInfo, ram schema.RAM) error {
 	return nil
 }
 
-// checkSameOS makes sure all servers have the same OS/version
+// checkSameOS verifies the OS distribution requirement for the specified set of servers.
+// The check will pass if all nodes in the cluster are based on the same OS distribution and major version.
+// Variance in minor/patch versions is acceptable.
 func checkSameOS(servers []Server) error {
-	osToNodes := make(map[string][]string)
+	// distros maps distribution name to list of versions
+	distros := make(map[string][]string)
 	for _, server := range servers {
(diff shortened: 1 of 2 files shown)


lib/checks/checks.go
```diff
@@ -756,28 +756,58 @@ func checkRAM(info ServerInfo, ram schema.RAM) error {
...
-		os := systeminfo.OS(server.GetOS()).Name()
-		osToNodes[os] = append(osToNodes[os], fmt.Sprintf("%v (%v)",
-			server.ServerInfo.GetHostname(), server.AdvertiseAddr))
+		info := server.GetOS()
+		distros[info.ID] = append(distros[info.ID], info.Version)
 	}
+	if len(distros) != 1 {
+		return trace.BadParameter("servers have different OS distributions: %v", formatKeysAsList(distros))
+	}
+	// Version verification is purposely simply and will compare the prefixes
```
r0mant, line 772: simply -> simple

r0mant (approved):

knisbet (approved):

PR #1560, opened 2020-05-13
Create directory for monitoring data

### LINKED ISSUE #1549: [BUG] Monitoring data is not preserved b/w system upgrades
**Describe the bug**

Metrics collected by our monitoring app are reset after each system upgrade. This is true for both the old InfluxDB-based stack and the new Prometheus-based one:

* InfluxDB keeps data in [/var/lib/data](https://github.com/gravitational/monitoring-app/blob/version/5.5.x/resources/influxdb.yaml#L101) inside planet which is not mounted from host.
* Prometheus currently uses emptyDir volume for its data.

For container metrics this is not a huge issue since they are recreated during upgrade and their metrics get reset anyway, but this also makes it impossible to look at historic data for node metrics.

**To Reproduce**

Perform an upgrade where a system upgrade is involved and look at the Monitoring tab.

**Expected behavior**

- [ ] InfluxDB data persists b/w system upgrades.
- [ ] Prometheus data persists b/w system upgrades and pod restarts.

Bonus point:

- [ ] Is there an easy way to migrate data during an upgrade? For Prometheus we're probably out of luck since it's in the pod's ephemeral storage, but InfluxDB keeps it on disk.

 - Gravity [e.g. 5.5.4]: all

UnderGreen:
## Description
 - Create directory for monitoring data to make them persistent during planet upgrade.
 - Use service-uid/gid from config instead of defaults while set permissions for directories.

## Type of change

Requires migration implementation in monitoring application, which is in progress.

* Closes #1549 for pre-Promethues branch

## Testing done

- [ ] Self-review the change
- [ ] Write tests
- [X] Perform manual testing
- [ ] Write documentation
- [] Address review feedback

### DIFF
--- a/lib/install/phases/bootstrap.go
+++ b/lib/install/phases/bootstrap.go
@@ -203,6 +203,10 @@ func (p *bootstrapExecutor) configureSystemDirectories() error {
 		filepath.Join(stateDir, "secrets"),
 		filepath.Join(stateDir, "backup"),
 	}
+	if p.isMonitoringEnabled() {
+		mkdirList = append(mkdirList, filepath.Join(stateDir, "monitoring"))
+	}
+
 	for _, dir := range mkdirList {
 		p.Infof("Creating system directory %v.", dir)
 		err := os.MkdirAll(dir, defaults.SharedDirMask)
@@ -227,13 +231,17 @@ func (p *bootstrapExecutor) configureSystemDirectories() error {
 		filepath.Join(stateDir, "secrets"),
 		filepath.Join(stateDir, "backup"),
 	}
+	if p.isMonitoringEnabled() {
+		chownList = append(chownList, filepath.Join(stateDir, "monitoring"))
+	}
+
 	for _, dir := range chownList {
 		p.Infof("Setting ownership on system directory %v to %v:%v.",
 			dir, p.ServiceUser.UID, p.ServiceUser.GID)
 		out, err := exec.Command("chown", "-R", fmt.Sprintf("%v:%v",
 			p.ServiceUser.UID, p.ServiceUser.GID), dir).CombinedOutput()
 		if err != nil {
-			return trace.Wrap(err, "failed to chmod %v: %s", dir, out)
+			return trace.Wrap(err, "failed to chown %v: %s", dir, out)
 		}
 	}
 	chmodList := []string{
@@ -278,10 +286,10 @@ func (p *bootstrapExecutor) configureApplicationVolumes() error {
 		uid, gid := mount.UID, mount.GID
 		if !existingDir {
 			if uid == nil {
-				uid = utils.IntPtr(defaults.ServiceUID)
+				uid = utils.IntPtr(p.ServiceUser.UID)
 			}
 			if gid == nil {
-				gid = utils.IntPtr(defaults.ServiceGID)
+				gid = utils.IntPtr(p.ServiceUser.GID)
 			}
 		}
 		// Only chown directories/files if necessary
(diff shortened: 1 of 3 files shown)


lib/install/phases/bootstrap.go
```diff
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"strconv"
 
+	"github.com/davecgh/go-spew/spew"
```
a-palchikov, line 27: nit: move to next import group

lib/install/phases/bootstrap.go
```diff
@@ -203,6 +204,11 @@ func (p *bootstrapExecutor) configureSystemDirectories() error {
 		filepath.Join(stateDir, "secrets"),
 		filepath.Join(stateDir, "backup"),
 	}
+	p.Info(spew.Sdump(p.Application))
```
a-palchikov, line 207: Debugging aid or necessary?

lib/install/phases/bootstrap.go
```diff
@@ -203,6 +204,11 @@ func (p *bootstrapExecutor) configureSystemDirectories() error {
...
+	if p.Application.Manifest.Extensions != nil && p.Application.Manifest.Extensions.Monitoring != nil && !p.Application.Manifest.Extensions.Monitoring.Disabled {
```
a-palchikov, line 208: Does it need to be conditionally enabled? What happens if we always create/chown this directory?

UnderGreen: Just you know, to make things not hardcoded. Users could ask why do I have empty monitoring directories on all servers even if I disabled monitoring extension.

lib/install/phases/bootstrap.go
UnderGreen: PR was an in draft stage :).
Now it is open for review, can you reviet it again? @a-palchikov

lib/install/phases/bootstrap.go
```diff
@@ -363,6 +371,13 @@ func (p *bootstrapExecutor) configureServiceUser() error {
 	return nil
 }
 
+func (p *bootstrapExecutor) isMonitoringEnabled() bool {
+	if p.Application.Manifest.Extensions != nil && p.Application.Manifest.Extensions.Monitoring != nil && p.Application.Manifest.Extensions.Monitoring.Disabled {
```
r0mant, line 375: To be honest, I had the same question as @a-palchikov. Why not simplify things and just always create it? /var/lib/gravity is our state directory and we control what we put in it, users should not be concerned with what directories we need there.

r0mant (commented):
Also, what about upgrades? IIRC during upgrade directories created in a different place.

lib/install/phases/bootstrap.go
UnderGreen: Ok, changed it to be always created.

lib/install/phases/bootstrap.go
```diff
@@ -202,7 +202,9 @@ func (p *bootstrapExecutor) configureSystemDirectories() error {
 		filepath.Join(stateDir, "site", "packages", "tmp"),
 		filepath.Join(stateDir, "secrets"),
 		filepath.Join(stateDir, "backup"),
+		filepath.Join(stateDir, "monitoring"),
```
r0mant, line 205: I think we also need to add it here: https://github.com/gravitational/gravity/blob/version/5.5.x/lib/ops/opsservice/deploy.go#L88. This is where directories get created/updated during an upgrade.

r0mant (approved):

PR #1568, opened 2020-05-14
Child directories for influxdb and kapacitor

UnderGreen:
## Description

Follow up for https://github.com/gravitational/gravity/pull/1560. I just realized we need subdirectories with service-uid permissions, otherwise for hostPath Kubelet will created them with root:root permissions.

## Type of change

* Closes #, refs #

* Requires #

* Ports #

## Testing done

- [X] Self-review the change
- [ ] Write tests
- [X] Perform manual testing
- [ ] Write documentation
- [ ] Address review feedback

### DIFF
--- a/lib/ops/opsservice/deploy.go
+++ b/lib/ops/opsservice/deploy.go
@@ -117,7 +117,8 @@ func remoteDirectories(operation ops.SiteOperation, server *ProvisionedServer, m
 		server.InGravity("planet", "state"),
 		server.InGravity("site"),
 		server.InGravity("backup"),
-		server.InGravity("monitoring"),
+		server.InGravity("monitoring", "influxdb"),
+		server.InGravity("monitoring", "kapacitor"),
 	}
 
 	chmodList := []string{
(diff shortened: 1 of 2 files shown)


lib/ops/opsservice/deploy.go
```diff
@@ -117,7 +117,8 @@ func remoteDirectories(operation ops.SiteOperation, server *ProvisionedServer, m
 		server.InGravity("planet", "state"),
 		server.InGravity("site"),
 		server.InGravity("backup"),
-		server.InGravity("monitoring"),
+		server.InGravity("monitoring", "influxdb"),
```
a-palchikov, line 120: Don't you need to create these first?
Also, I believe this should the other way around - create these directories (i.e. list them in `directories`) and leave just the root directory (`monitoring`) here - it will be `chown`ed recursively.

a-palchikov (changes requested):

lib/ops/opsservice/deploy.go
UnderGreen: Yes, I wanted to add it into directories array. :facepalm:

a-palchikov (approved):

PR #1597, opened 2020-05-21
[5.5.x] improved operation handling

a-palchikov:
## Description

This PR updates the operation handling to make it more user-friendly and expected.
It changes the following:
  * adds `-prune` to serf command line to remove a node as this ensures that the member is evicted immediately
  * reverts the listing of all operations (and not just the last of any specific type) - otherwise operation-id targeted queries might fail if the id refers to an operation further down in operation stack
  * `plan display` explicitly handles invalid operations
  * `plan complete` handles operations even without plan - in this case the operation is explicitly marked as failed
  * uninstall possibly active agent prior to triggering the expand operation. This is done after verifying the package state so it is obvious that the agent is a leftover from the previous unfinished cleanup.

## Type of change

* Bug fix (non-breaking change which fixes an issue)
* Regression fix (non-breaking change which fixes a regression)
* This change has a user-facing impact

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Additional information

Displaying plan for operation that does not support plan:
```
Cluster name:		dev.test
Cluster status:		active
Application:		telekube, version 5.5.46-dev.5
Gravity version:	5.5.46-dev.5 (client) / 5.5.46-dev.5 (server)
Join token:		schmoken
Periodic updates:	Not Configured
Remote support:		Not Configured
Last completed operation:
    * operation_shrink (7347bda1-ba40-475e-b074-5b37bfbf8a6b)
      started:		Thu May 21 19:11 UTC (59 minutes ago)
      completed:	Thu May 21 19:13 UTC (57 minutes ago)
...
$ gravity plan --operation-id=7347bda1-ba40-475e-b074-5b37bfbf8a6b
[ERROR]: cannot display plan for "shrink" operation as it does not support plans
```
Display plan for operation that failed to initialize properly:
```
Cluster name:		dev.test
Cluster status:		expanding
Application:		telekube, version 5.5.46-dev.5
Gravity version:	5.5.46-dev.5 (client) / 5.5.46-dev.5 (server)
Join token:		schmoken
Periodic updates:	Not Configured
Remote support:		Not Configured
Active operations:
    * operation_expand (32c26b8d-08df-4709-ba56-f5ba024808ff)
      started:	Thu May 21 20:13 UTC (5 seconds ago)
      Operation has been created, 0% complete
...
$ gravity plan
[...]

PR #1603, opened 2020-05-25
[7.0.x] restore improperly merged PR changes

a-palchikov:
## Description

Backport changes from https://github.com/gravitational/gravity/pull/1524 that were lost when the PR was merged into a sub-branch instead of the main branch.
Also ports changes from https://github.com/gravitational/gravity/pull/1597.

## Type of change

* Ports https://github.com/gravitational/gravity/pull/1597
* Ports https://github.com/gravitational/gravity/pull/1524

## TODOs

- [x] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback

## Testing done

Will run through a set of basic tests: install/expand/shrink. Will run simulated error tests for expand/install.

### DIFF
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -850,7 +850,12 @@ func (p *Peer) tryConnect(operationID string) (ctx *operationContext, err error)
 			return nil, utils.Abort(err)
 		}
 		if trace.IsCompareFailed(err) {
-			p.printStep("Waiting for another operation to finish at %v", addr)
+			p.Warnf("Waiting for precondition to create expand operation: %v.", err)
+			if utils.IsClusterDegradedError(err) {
+				p.printStep("Cluster is degraded, waiting for it to become healthy")
+			} else {
+				p.printStep("Waiting for another operation to complete at %v", addr)
+			}
 		}
 	}
 	return ctx, trace.Wrap(err)
--- a/lib/expand/join.go
+++ b/lib/expand/join.go
@@ -1028,7 +1028,7 @@ func (p *Peer) createShrinkOperation(ctx context.Context) error {
 		return trace.Wrap(err)
 	}
 	if len(operation.Servers) == 0 {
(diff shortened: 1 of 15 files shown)


knisbet (approved):

lib/expand/join.go
```diff
@@ -1022,6 +1027,10 @@ func (p *Peer) createShrinkOperation(ctx context.Context) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
+	if len(operation.Servers) == 0 {
+		p.WithField("operation", operation.String()).Warn("Failed to create shrink for invalid operation.")
```
r0mant, line 1031: Nit: I think we might never see this log message but still, instead of saying "for invalid operation" I'd say explicitly that servers are missing from the operation state.

tool/gravity/cli/operation.go
```diff
@@ -393,6 +361,24 @@ func (r *backendOperations) listInstallOperation() error {
 	return trace.NotFound("no operation found")
 }
 
+func (r *backendOperations) updateOperationInCache(getter operationGetter, logger logrus.FieldLogger) {
```
r0mant, line 364: Probably not for this PR, but I really think we should try to move all operation handling/displaying logic somewhere to lib/ and cover with tests. It's become pretty non-trivial logic.

r0mant (approved):
Probably not

tool/gravity/cli/operation.go
a-palchikov: I already attempted it at some point but abandoned due to a non-trivial amount of work. I agree though and will reattempt in a later PR.

lib/expand/join.go
a-palchikov: Oh, you can see the message if you abort the expand operation that gets stuck half-way - i.e. if the controller cannot reach the joining node to properly accept the agent.

PR #1606, opened 2020-05-26
[7.0.x] implement container scanning (#1550)

knisbet:
## Description

This implements internal changes to gravity tooling, to be able to push all container artifacts from a release to a docker repository for scanning purposes. This is meant to work in conjunction with a repository that is able to scan and alert on vulnerabilities within those containers.

Also fixed are a couple glitches in our registry code when working with public registries, to ensure we correctly indicate which repositories and actions we need permissions for, to allow the registry to correctly embed those permissions within our oauth token. 

## Type of change

Backport #1550 

## TODOs

- [x] Self-review the change
- [x] Retest the cherry-pick after removal of conflicts

## Implementation

1. Currently, I've wired it in a way that publishes the containers for scanning when publishing a gravity release. This is a little less than ideal, as ideally we'd be scanning regularly, such as every day, and staging the security fixes like normal PRs. Eventually we'll do this, but to get scanning up and running in the short term, it was chosen to wire the scanning into the publishing pipeline.

2. For scanning purposes, I want to push all the images to the same repository, that is pre-configured for scanning. This means to preserve the original repository information I had to get a bit clever and embed this information within the tags, along with the build version. This way, using just the scan results, we're able to work our way backwards and identify the gravity version and images that contain the vulnerability. As an example quay.io/org/example:1.0.0 in gravity 2.0.0 will be rewritten to quay.io/gravitational/gravity-scan: 2.0.0_org_example_1.0.0.

3. The existing Sync function on the imageService performs very similarly to the new SyncForScanning function, however, I'm not convinced the two functions can be cleanly merged, which is why I left the separate but very similar implementation.

4. Within the imageservice, I also ended up writing a wrapper for dockers token handler. The underlying token handler doesn't appear to support adding additional scopes after initialization (not needed for something like `docker push <image>`). So this wrapper exists to re-initialize dockers token handler, if the imageservice starts interacting with a repository that wasn't previously included. 

5. Because this code is intended for internal use only, the CLI options have been marked hidden, and should have no UX impact.

## Testing done

Re-test local build and push to quay.io.

As an example: 
`TELE_COPY_TO_USER=<user> TELE_COPY_TO_PASS=<pass> make scan-artifacts` to scan with quay.io.

GCR requires some work to use token based authentication: https://cloud.google.com/container-registry/docs/advanced-authentication#token


lib/docker/scanningimageservice.go
```diff
@@ -0,0 +1,152 @@
...
+// NewScanningImageService creates an image service that rewrites image paths to a
+// repository used for scanning those images.
+func NewScanningImageService(req RegistryConnectionRequest, conf ScanConfig) (ImageService, error) {
+	base, err := NewImageService(req)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	return &scanningImageService{
+		imageService: base.(*imageService),
```
a-palchikov, line 48: See [here](https://github.com/gravitational/gravity/pull/1608#discussion_r430274388)

a-palchikov (approved):

r0mant (approved):

Issue #1623, opened 2020-05-28
Revise RPC credentials handling


a-palchikov:
### Current Result

There are two types of RPC credentials:
  * installs/expands/leaves - stored in the system repository with a fixed age of 10yrs/20yrs (CA)
  * upgrades - stored in the cluster-specific package repository with a fixed age of 1yrs/5yrs (CA) but with rotation with each operation

### Expected Result
  * Unify both types of credentials - i.e. use either one consistently
  * Extend the cluster controller with the ability to verify and automatically rotate RPC credentials

Design doc: https://docs.google.com/document/d/1KVWEnLVXoExWHrq_Hm4gAJcU1ECiuQrj7k0F6rIzRJ0/edit#

r0mant:
@a-palchikov Looks like with these changes (https://github.com/gravitational/gravity/pull/1629) the rollback of /init phase screws up rpc creds package completely:

```
ubuntu@node-1:~/upgrade$ sudo ./gravity plan execute --phase=/init
Wed Jun  3 21:10:13 UTC	Executing "/init/node-1" locally
Wed Jun  3 21:10:15 UTC	Executing phase "/init" finished in 2 seconds
ubuntu@node-1:~/upgrade$
ubuntu@node-1:~/upgrade$
ubuntu@node-1:~/upgrade$ sudo ./gravity plan rollback --phase=/init
Wed Jun  3 21:10:19 UTC	Rolling back "/init/node-1" locally
Wed Jun  3 21:10:20 UTC	Rolling back phase "/init" finished in 1 second
ubuntu@node-1:~/upgrade$
ubuntu@node-1:~/upgrade$
ubuntu@node-1:~/upgrade$ sudo ./gravity plan execute --phase=/init
Wed Jun  3 21:10:25 UTC	Executing "/init/node-1" locally
Wed Jun  3 21:10:25 UTC	Executing phase "/init" finished in now
[ERROR]: failed to update RPC credentials
	failed to read package gravitational.io/rpcagent-secrets:0.0.1
		object not found
```

r0mant:
Done in 5.5, forward-ports are tracked in https://github.com/gravitational/gravity/issues/1740.

PR #1629, opened 2020-05-29
[5.5.x] rotate RPC credentials on cluster upgrade

a-palchikov:
## Description

* Rotate RPC credentials used for install/expand/leave operations on each cluster upgrade operation.
* Revise "system rotate-rpc-creds" to be able to use it on any affected `5.x` or `6.x` cluster and not necessarily based on `5.5.47`.

## Type of change

* Updates https://github.com/gravitational/gravity/issues/1623

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Write documentation
- [ ] Address review feedback

## Testing done

Tested on `5.5.38` and `6.1.27`:
 * installed 3-node cluster
 * rotated the RPC certificates using the binary from this branch
 * removed a node
 * added the node back to cluster

The binary does not work on 7.x clusters due to changes in the manifest.


tool/gravity/cli/clusterupdate.go
```diff
@@ -60,6 +60,9 @@ func updateTrigger(
 	updatePackage string,
 	manual, noValidateVersion bool,
 ) error {
+	if err := updateRotateCertificates(localEnv); err != nil {
```
r0mant, line 63: Just curious, any reason this is not done as a part of the "init" upgrade phase?

r0mant (commented):
Regarding "testing done" - have you also tried performing an upgrade and verifying that credentials are rotated and that join/remove can be performed afterwards?

a-palchikov:
@r0mant yes, I tested upgrades `5.5.38` -> `5.5.47-dev.2`.

tool/gravity/cli/clusterupdate.go
a-palchikov: No reason - I did not think about it, will move it there.

tool/gravity/cli/clusterupdate.go
```diff
@@ -96,6 +99,18 @@ func newClusterUpdater(
 	return updater, nil
 }
 
+// updateRotateCertificates rotates the RPC credentials used for install/expand/leave operations
+func updateRotateCertificates(env *localenv.LocalEnvironment) error {
+	clusterPackages, err := env.ClusterPackages()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if _, err := rpc.UpsertCredentials(clusterPackages); err != nil {
```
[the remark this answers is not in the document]
a-palchikov: I'll move that to the init phase and log there.

lib/update/cluster/phases/init.go
```diff
@@ -194,34 +194,64 @@ func (p *updatePhaseInit) Rollback(ctx context.Context) error {
...
-	} else {
-		p.Infof("Initialized RPC credentials: %v.", pkg)
+func (p *updatePhaseInit) backupRPCCredentials() error {
+	p.Info("Backup RPC credentials")
+	env, rc, err := rpc.LoadCredentialsData(p.Packages)
+	if err != nil {
+		return trace.Wrap(err)
 	}
+	defer rc.Close()
+	_, err = p.Packages.CreatePackage(rpcBackupPackage, rc, pack.WithLabels(env.RuntimeLabels))
```
r0mant, line 237: Shouldn't this be "upsert"? Otherwise on the second upgrade this will fail because the package will already be there?

a-palchikov: I've already made a change to assign it the corresponding operation label. Ideally, packages marked for the specific operation need to be purged post-operation. I'll make it into upsert though.

r0mant: Yeah I think upsert is better anyway just so the phase is reentrable e.g. if reexecuted multiple times.

r0mant (approved):

r0mant (commented):
@a-palchikov I have just realized that /init phase also executes on each server. Which means that credentials will be rotated multiple times and also each following package backup will override previous.

r0mant (approved):
Actually, I take my last comment back. It looks like it works as expected because there is a part of the init phase that is executed only on leader. I have also verified the following scenario: upgrade to this version, verify credentials have been rotated, verify remove, verify join. So this should be good to go. Going to merge once CI passes.

PR #1642, opened 2020-06-01
[6.3.x] use local image's application version for upgrade

a-palchikov:
## Description

Default to the version of the application from local cluster image for upgrades instead of using the highest available in the cluster store.

## Type of change

* Bug fix (non-breaking change which fixes an issue)

* This change has a user-facing impact

## Linked tickets and other PRs

* Refs https://github.com/gravitational/gravity/issues/1639

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

Tested upgrade trigger on a 6.1.28 3-node cluster by uploading 7.1.x version first and then upgrading from this branch.

PR #1643, opened 2020-06-01
Use local image's application version for upgrade

a-palchikov:
## Description

* Default to the version of the application from local cluster image for upgrades.
* Update latest planet to include a version fix.
* Update versions of (non-)LTS base versions used in robotest.

## Type of change

* Ports https://github.com/gravitational/gravity/pull/1642
* Requires https://github.com/gravitational/planet/pull/659.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

### DIFF
--- a/build.assets/robotest_run_suite.sh
+++ b/build.assets/robotest_run_suite.sh
@@ -12,7 +12,7 @@ UPGRADE_MAP[6.1.28]="ubuntu:18"
 # latest patch release on supported non-LTS version, keep this up to date
 UPGRADE_MAP[6.3.18]="ubuntu:18"
 
-# latest patch release on supported LTS version, keep this up to date
+# latest patch release on 7.0.x branch, keep this up to date
 UPGRADE_MAP[7.0.6]="ubuntu:18"
 
 # important versions in the field, these are static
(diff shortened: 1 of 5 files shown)


build.assets/robotest_run_suite.sh
```diff
@@ -7,10 +7,13 @@ readonly UPGRADE_FROM_DIR=${1:-$(pwd)/../upgrade_from}
...
 # latest patch release on compatible LTS, keep this up to date
-UPGRADE_MAP[6.1.18]="ubuntu:18"
+UPGRADE_MAP[6.1.28]="ubuntu:18"
 
 # latest patch release on supported non-LTS version, keep this up to date
-UPGRADE_MAP[6.3.6]="ubuntu:18"
+UPGRADE_MAP[6.3.18]="ubuntu:18"
+
+# latest patch release on supported LTS version, keep this up to date
+UPGRADE_MAP[7.0.6]="ubuntu:18"
```
wadells, line 16: Is 7.0 LTS?  It seems like this comment should be left out and 7.0.6 should fall under "supported non-LTS version" along with 6.3.

r0mant (commented):
How is this going to work if `gravity upgrade` is launched outside the installer tarball (which is a valid use-case if, say, the new version was downloaded to the cluster using automatic updates or uploaded earlier)?

a-palchikov:
The upcoming intermediate upgrades port will supersede this.

Issue #1639, opened 2020-06-01
Consistently use the version of the local image application for upgrades


a-palchikov:
## Current Result
If a higher version application was previously uploaded into the cluster package storage, upgrading a lower version uses the highest available version although that contradicts the expectation that upgrading from the installer directory should fall back to the local image's application version.

## Expected Result
`gravity upgrade` should default to the local image's application version instead of the one available in the cluster package store.

PR #1646, opened 2020-06-02
Robotest 2.0.0 Nightly Staging

wadells:
## Description
This PR introduces 3 changes:

1) Robotest 2.0.0-rc.1 is added to nightly
2) With 2.0.0, a shrink test is added to nightly
3) I removed Gravity explicitly specifying storage_driver and state_dir as both have sane defaults.

I'm intentionally not adding these changes to the PR build yet because:

1) Robotest in PR builds has been suffering from other instability recently  (#1639 and related work, I believe).  I don't want to introduce possibly confounding factors there until things settle down.
2) I'd like to get https://github.com/gravitational/robotest/pull/233 into 2.0, but I also don't want to further delay some sort of vetting on master.

## Type of change

Introduces the work from https://github.com/gravitational/robotest/pull/229, amongst many others.

This will help catch issues like https://github.com/gravitational/gravity/issues/1445, as well as more general shrink related issues.

## TODOs
- [x] Perform manual testing
- [x] Write automated tests (N/A - this is automated testing)
- [x] Write documentation (N/A)
- [x] Self-review the change
- [ ] Address review feedback

## Testing done
These changes were individually vetted in Robotest PRs 206-234.  This PR itself is testing/staging before introducing robotest 2.0.0 more widely (PR builds, all active release branches).

I would have tested this as a try build, but https://github.com/gravitational/ops/issues/142 makes it tedious to run anything other than the root `Jenkinsfile`.

### DIFF
--- a/build.assets/robotest_run_nightly.sh
+++ b/build.assets/robotest_run_nightly.sh
@@ -25,7 +25,7 @@ readonly ROBOTEST_SCRIPT=$(mktemp -d)/runsuite.sh
 
 # number of environment variables are expected to be set
 # see https://github.com/gravitational/robotest/blob/master/suite/README.md
-export ROBOTEST_VERSION=${ROBOTEST_VERSION:-uid-gid}
+export ROBOTEST_VERSION=${ROBOTEST_VERSION:-2.0.0-rc.1}
 export ROBOTEST_REPO=quay.io/gravitational/robotest-suite:$ROBOTEST_VERSION
 export WAIT_FOR_INSTALLER=true
 export INSTALLER_URL=$GRAVITY_BUILDDIR/telekube.tar
@@ -47,6 +47,7 @@ function build_resize_suite {
   cat <<EOF
  resize={"to":3,"flavor":"one","nodes":1,"role":"node","state_dir":"/var/lib/telekube","os":"ubuntu:18","storage_driver":"overlay2"}
  resize={"to":6,"flavor":"three","nodes":3,"role":"node","state_dir":"/var/lib/telekube","os":"ubuntu:18","storage_driver":"overlay2"}
+ shrink={"nodes":3,"flavor":"three","role":"node","os":"redhat:7"}
 EOF
 }
 


build.assets/robotest_run_nightly.sh
```diff
@@ -45,20 +43,20 @@ export REPEAT_TESTS=${REPEAT_TESTS:-1}
 
 function build_resize_suite {
   cat <<EOF
- resize={"to":3,"flavor":"one","nodes":1,"role":"node","state_dir":"/var/lib/telekube","os":"ubuntu:18","storage_driver":"overlay2"}
- resize={"to":6,"flavor":"three","nodes":3,"role":"node","state_dir":"/var/lib/telekube","os":"ubuntu:18","storage_driver":"overlay2"}
```
r0mant, line 49: Actually, `"state_dir": "/var/lib/telekube"` was there on purpose. It is meant to test custom state directory (the default one is /var/lib/gravity, this parameter made sure to use /var/lib/telekube instead).

wadells: I removed all of the changes in c4e7834 as these were opportunistic cleanup.

@r0mant: Your remark raises many questions, potentially to be improved in the future:

1) Why isn't `state_dir` specified on any other tests? Is `state_dir` relevant to expand testing in particular?
2) Do key customers use a customized `state_dir`? if so, what value they use, and in combination with what other options?

Essentially: I understand what this parameter does but I don't yet understand why it is important to exercise, or at what cost (e.g. I assume it probably isn't worth doubling the number of daily tests to run every single test with a default and custom state dir -- nor do we want all of one or the other).

I'd like to move toward one of:

1) We have a cheap one-off (one node install?) scenario that exercises `state_dir`.  This is probably the thing to do if we don't have known field use cases but want to check it "just in case" and the scenario doesn't really matter.
2) We incorporate `state_dir` into customer specific scenarios (e.g. upgrade from X Gravity on Y distro with Z `state_dir`).
3) `state_dir` is treated as a full fleged member of our integration test matrix, similar to OS or cluster size (this will probably have notable time and $ costs)

The current correlation without apparent cause between expand/resize and state_dir is why I removed the param, and alas neither https://github.com/gravitational/gravity.e/commit/79413bd422eba2798397634ffcfa6b8970a24a5c (state_dir added to gravity) nor https://github.com/gravitational/robotest/commit/7489002d12656ca7808f464c49e93a1442da4e85 (state_dir added to robotest) give me much to go on.

r0mant (approved):
FWIW I think that other cleanups (like removing storage driver parameter) were fine, just would keep state-dir.

build.assets/robotest_run_nightly.sh
r0mant: 1. No particular reason, just so we have a scenario that exercises custom state directory for install and making sure it's respected when joining new nodes too.
2. Yes they do. They can use whatever directory they like e.g. I know some customers mount their data in /opt/xxx and use that.

a-palchikov (approved):
Agree with Roma - the state directory test is useful even if somewhat obscure with the purpose. We could make it more explicit.

wadells:
> FWIW I think that other cleanups (like removing storage driver parameter) were fine, just would keep state-dir.

> Agree with Roma - the state directory test is useful even if somewhat obscure with the purpose. We could make it more explicit.

I'll loop back to both of these (re-removing storage_driver and making state_directory more explicit/self explanatory) in a future PR.

wadells:
The robotest failures seen in earlier are issues latent in master.  Because I don't have admin privileges (to force merge) and I know this patch doesn't affect the PR build robotest, I've retriggered the build without the robotest step to get a green run. This is likely an abuse of the letter of the `continuous-integration/jenkins/pr-head  : Required` rule, but not the spirit of it.

I'll be sure to be timely in my triage of nightly failures for the next couple days.

Issue #1640, opened 2020-06-01
System upgrades to master fail


a-palchikov:
Seeing this in agent logs when upgrading to master:
```
Jun 01 13:29:09 node-0 /usr/bin/planet[764]: WARN             "k8s collector failed after 1.828659ms: failed to query nodes: Get https://leader.telekube.local:6443/api/v1/nodes: http2: no cached connection was available\n\tGet https://leader.telekube.local:6443/api/v1/nodes: http2: no cached connection was available" collector/collector.go:115
Jun 01 13:29:09 node-0 /usr/bin/planet[764]: WARN             "k8s collector failed after 493.666µs: failed to query nodes: Get https://leader.telekube.local:6443/api/v1/nodes: http2: no cached connection was available\n\tGet https://leader.telekube.local:6443/api/v1/nodes: http2: no cached connection was available" collector/collector.go:115
```
Also, `planet version` returns:
```
{"version":"v0.0.0-master+$Format:%h$","gitCommit":"$Format:%H$","gitTreeState":"not a git tree"}
```
I initially thought this has something to do with skew (as I was testing with upgrades from `6.1.0` which is based on k8s `1.15.1`) but upgrades from the latest `6.3.18` show the same symptoms.

I'm suspecting either Go version incompatibility (although this should not be the case) between `1.12.x` and `1.13.9` or the k8s go client is behind.

Issue #1209, opened 2020-03-05
[BUG] Node degraded due to kube-apiserver after upgrade


r0mant:
**Describe the bug**

This was observed in robotest. The upgrade operation completes successfully but the status shows unhealthy due to kube-apiserver unit being in a failed state on one of the nodes:

`"nodes":[{"hostname":"robotest-660fd05b-node-1","advertise_ip":"10.142.15.236","role":"master","profile":"node","status":"healthy"},{"hostname":"robotest-660fd05b-node-2","advertise_ip":"10.142.15.238","role":"master","profile":"node","status":"degraded","failed_probes":["kube-apiserver.service (failed)"]},{"hostname":"robotest-660fd05b-node-0","advertise_ip":"10.142.15.237","role":"master","profile":"node","status":"healthy"}]`

Looking into the logs, apiserver shut down with an error on that node:

```
Mar 05 04:52:23 robotest-660fd05b-node-2 systemd[1]: kube-apiserver.service: Main process exited, code=exited, status=255/EXCEPTION
Mar 05 04:52:23 robotest-660fd05b-node-2 systemd[1]: kube-apiserver.service: Failed with result 'exit-code'.
Mar 05 04:52:23 robotest-660fd05b-node-2 systemd[1]: Stopped Kubernetes API Server.
```

On another node, kube-apiserver was running fine. Another interesting data point is that according to the logs it seemed like two apiservers were running simultaneously (which shouldn't happen) until one of them shut down with the error.

**To Reproduce**

Unclear, it happened during a 3-node robotest upgrade.

**Expected behavior**

**Logs**

[10.142.15.236-planet-journal-export.log.gz](https://github.com/gravitational/gravity/files/4294397/10.142.15.236-planet-journal-export.log.gz)
[10.142.15.238-planet-journal-export.log.gz](https://github.com/gravitational/gravity/files/4294399/10.142.15.238-planet-journal-export.log.gz)

To view logs, unzip and:

```
cat ./planet-journal-export.log | /lib/systemd/systemd-journal-remote -o ./journal/system.journal -
journalctl -D ./journal ...
```

**Environment (please complete the following information):**
 - OS [e.g. Redhat 7.4]:
 - Gravity [e.g. 5.5.4]:
 - Platform [e.g. Vmware, AWS]:

**Additional context**

wadells:
I hit this in `version/6.1.x` for 5 different upgrade scenarios, all 5.5.x versions -> 6.1.latest

[...]

PR #1652, opened 2020-06-03
Use kubectl's drain package directly during upgrades

a-palchikov:
## Description

This PR removes most of the ad-hoc (although copied from kubernetes repository) drain code in favor of using the kubectl's implementation.
I'm a bit on the fence about this vs. using `kubectl` directly but there maybe up but also down-sides to that. One upside is probably that if using the binary, the version that matches the installed kubernetes version runs the drain, although the versions should ideally be compatible functionality-wise.
As for the issue with `http2: no cached connection` - it seems to be related to k8s-client mismatch between the server, so I updated satellite (as I did planet) to match the kubernetes version.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1640.

* Requires https://github.com/gravitational/planet/pull/659.

## TODOs

- [x] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback

## Testing done

Robotests should ideally pass as it mostly affects them.

## Additional information

There are massive issues with upgrading to the correct version of k8s packages (v0.17.6, sometimes v1.17.6) due to incompatibility with the vendored helm (`2.15.2`). So I had to just bump the `net` (for client-go) and `kubectl` (for drain package) dependencies - otherwise, the k8s dependencies are `v0.15.7`.


r0mant (commented):
Weird, seems like this is rebased off my rollback branch? Is there an easy way to see only relevant changes?

a-palchikov:
It is since I need rollback for testing.

a-palchikov:
Merging as the robotest failures are all due another [issue](https://github.com/gravitational/gravity/issues/1209).

PR #1653, opened 2020-06-03
(5.5) Sergei/influxdb data migration

### LINKED ISSUE #1549: [BUG] Monitoring data is not preserved b/w system upgrades

UnderGreen:
## Description

This PR implements data migration for InfluxDB pod from `/var/lib/data/influxdb` to `/var/lib/gravity/monitoring/influxdb`.

## Type of change

* Closes #1549 for 5.5.x.

* Requires https://github.com/gravitational/rigging/pull/77

* Ports #

## Testing done

* Installed 5.5.41 cluster.
* Updated to the version from this PR and checked:
  1. Data on the node with InfluxDB pod was copied to /var/lib/gravity/monitoring/influxdb
  2. ConfigMap `influxdb-node` is created in monitoring namespace with the correct nodename for InfluxDB pod.
  3. Upgrade finished without fail.


lib/update/cluster/phases/bootstrap.go
```diff
@@ -277,6 +300,7 @@ type updatePhaseBootstrapLeader struct {
 	existingClusterConfig []byte
 	// updateManifest specifies the manifest of the update application
 	updateManifest schema.Manifest
+	kubernetesOperation
```
r0mant, line 303: As we discussed offline, just put `client *kubernetes.Clientset` here, you don't really need to embed kubernetesOperation.

lib/update/cluster/phases/bootstrap.go
```diff
@@ -548,6 +576,49 @@ func (p *updatePhaseBootstrap) addUpdateRuntimePackageLabel() error {
 	return nil
 }
 
+// migrateInfluxDBData copies all from /var/lib/data/influxdb if it exist on the node and if destination
+// directory /var/lib/gravity/monitoring/influxdb empty
+func (p *updatePhaseBootstrap) migrateInfluxDBData(ctx context.Context) error {
+	newDataDirectory := "/var/lib/gravity/monitoring/influxdb"
```
r0mant, line 582: I'd consider moving these to lib/defaults. Or at least make these local constants instead of variables.

Also wouldn't hurt to add a comment that "oldDataDirectory" is within planet.

lib/update/cluster/phases/bootstrap.go
```diff
@@ -548,6 +576,49 @@ func (p *updatePhaseBootstrap) addUpdateRuntimePackageLabel() error {
...
+	oldDataDirectory := "/var/lib/data/influxdb"
+
+	if _, err := os.Stat(oldDataDirectory); os.IsNotExist(err) {
```
r0mant, line 585: I don't think this will work as expected. "oldDataDirectory" is inside the planet, right? So this will always return "not found"

lib/update/cluster/phases/bootstrap.go
```diff
@@ -548,6 +576,49 @@ func (p *updatePhaseBootstrap) addUpdateRuntimePackageLabel() error {
...
+		return trace.Wrap(err)
+	}
+	if !empty {
+		// directory already exist and not empty, probably data already migrated
+		return nil
+	}
+
+	p.Info("Copying InfluxDB data to the new directory.")
+	_, err = utils.RunCommand(ctx, p.FieldLogger,
+		utils.PlanetCommandArgs("cp", "-a", fmt.Sprintf("%s/.", oldDataDirectory), newDataDirectory)...)
```
r0mant, line 601: Does "newDataDirectory" need to be created first? Or at least check that it exists?

lib/update/cluster/phases/bootstrap.go
```diff
@@ -548,6 +576,49 @@ func (p *updatePhaseBootstrap) addUpdateRuntimePackageLabel() error {
...
+	if err != nil {
+		trace.Wrap(err)
```
r0mant, line 603: `return`. Also, I think it wouldn't hurt to capture output of the above command (first return parameter) and add it to the error.

lib/update/cluster/phases/bootstrap.go
```diff
@@ -548,6 +576,49 @@ func (p *updatePhaseBootstrap) addUpdateRuntimePackageLabel() error {
...
+	}
+	p.Infof("Setting ownership of %v to %v:%v.", newDataDirectory, p.ServiceUser.UID, p.ServiceUser.GID)
+	_, err = utils.RunCommand(ctx, p.FieldLogger,
+		utils.PlanetCommandArgs("chown", "-R", fmt.Sprintf("%v:%v", p.ServiceUser.UID, p.ServiceUser.GID), newDataDirectory)...)
+	if err != nil {
+		trace.Wrap(err)
```
r0mant, line 609: Same as above.

lib/update/cluster/phases/kubernetes.go
```diff
@@ -378,4 +379,84 @@ func removeKubeletPermissions(client *kubeapi.Clientset) error {
 	return nil
 }
 
+func getInfluxDBNodename(client *kubeapi.Clientset, logger log.FieldLogger) (string, error) {
+	var labels map[string]string
+	deployment, err := getInfluxDBDeployment(client)
+	if deployment.Spec.Selector != nil {
+		labels = deployment.Spec.Selector.MatchLabels
+	}
+	replicaSets, err := rigging.CollectReplicaSets(deployment.Namespace, labels, logger, client, func(ref metav1.OwnerReference) bool {
```
r0mant, line 388: Can't we just query the influxdb pod by its labels in the monitoring namespace? Should be easy enough, a single API call:

```
ubuntu@g5-cluster-node-1:~$ kubectl -nmonitoring get pods/influxdb-677c446f6f-k28g4 -oyaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    kubernetes.io/psp: privileged
    seccomp.security.alpha.kubernetes.io/pod: docker/default
  creationTimestamp: "2020-05-29T22:24:42Z"
  generateName: influxdb-677c446f6f-
  labels:
    app: monitoring
    component: influxdb
    pod-template-hash: 677c446f6f
  name: influxdb-677c446f6f-k28g4
  namespace: monitoring
...
```

lib/update/cluster/phases/kubernetes.go
```diff
@@ -378,4 +379,84 @@ func removeKubeletPermissions(client *kubeapi.Clientset) error {
...
+		if trace.IsAlreadyExists(err) {
+			logger.WithField("configmap", defaults.InfluxDBNodenameConfigMap).Info("ConfigMap already exists, updating.")
+			err = updateInfluxDBConfigMap(client, nodename)
+		}
+	}
+	return trace.Wrap(err)
+}
+
+func updateInfluxDBConfigMap(client *kubeapi.Clientset, nodename string) error {
+	configMap, err := client.CoreV1().ConfigMaps(defaults.MonitoringNamespace).Get(defaults.InfluxDBNodenameConfigMap, metav1.GetOptions{})
```
r0mant, line 450: You don't really need to Get it here, you can just Update it if you got AlreadyExists on the Create call above.

We also have a function that does basically this: https://github.com/gravitational/gravity/blob/version/5.5.x/lib/ops/opsservice/monitoring.go#L214.

I would try to adopt that one instead: e.g. move it as exported function somewhere, say, to lib/utils/kubernetes.UpdateConfigMap and use here. Looking at it, it hardcodes the data map key, so you'd need to modify it a bit for example to accept the entire "data" map instead.

lib/update/cluster/phases/bootstrap.go
```diff
@@ -548,6 +576,49 @@ func (p *updatePhaseBootstrap) addUpdateRuntimePackageLabel() error {
 	return nil
 }
 
+// migrateInfluxDBData copies all from /var/lib/data/influxdb if it exist on the node and if destination
```
a-palchikov, line 579: copies all? all data?

lib/update/cluster/phases/bootstrap.go
```diff
@@ -548,6 +576,49 @@ func (p *updatePhaseBootstrap) addUpdateRuntimePackageLabel() error {
...
+		// influxdb data is on another node
+		return nil
+	}
+
+	empty, err := isEmptyDir(newDataDirectory)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	if !empty {
+		// directory already exist and not empty, probably data already migrated
```
a-palchikov, line 595: exist**s**

UnderGreen:
I am going to close this PR and work in another branch on different implementation based on review.

PR #1669, opened 2020-06-08
Add 5.0.36 notes

r0mant:


docs/7.x/changelog.md
```diff
@@ -1279,6 +1279,12 @@ to learn how to gain insight into how the cluster status changes over time.
 * Wait for `kube-system` namespace to be created during the installation.
 * Update `tele push` to treat existing applications and their dependencies gracefully.
 
+### 5.0.36 LTS (June 8th, 2020)
+
+#### Bugfixes
+
+* Update `--docker-device` flag that sets devicemapper device to support symlinks to actual block devices.
```
knisbet, line 1286: PR?

Also, I'm not sure I understand the statement. If I read this correctly, perhaps something along these lines would work:
Fixes an issue where `--docker-device` wasn't able to follow symlinks to block devices.

r0mant: I didn't put the PR because this is 5.0 which is pre-OSS and the PR is in the enterprise repo. I guess we could add it but nobody will be able to see it except for us.

Also yes, your wording works better, I'll update.

PR #1687, opened 2020-06-11
(5.5) Check cluster status after upload

r0mant:
## Description

This PR updates the "gravity update upload" command that uploads new version of cluster image to the local cluster so it checks cluster health after all uploading has been done.

This is to accommodate a scenario where immediately after upload the cluster may degrade due to increased i/o and the following upgrade command will fail.

Now the upload command will give cluster a few minutes to settle before exiting.

## Type of change

* Internal change (not necessarily a bug fix or a new feature)

## Linked tickets and other PRs

* Refs https://github.com/gravitational/gravity/issues/1659.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

### When everything's fine

The "verifying" step completes immediately.

```
ubuntu@node-1:~/upgrade$ sudo ./upload
Thu Jun 11 00:33:34 UTC	Importing cluster image telekube v5.5.48-dev.8
Thu Jun 11 00:34:18 UTC	Synchronizing application with Docker registry 192.168.99.102:5000
Thu Jun 11 00:34:26 UTC	Verifying cluster health
Thu Jun 11 00:34:26 UTC	Cluster image has been uploaded
```

### When cluster is temporarily degraded after upload

By the timestamp you can tell it spent some time in "verifying" step, it is visible in the logs also.

```
ubuntu@node-1:~/upgrade$ sudo ./upload
Thu Jun 11 00:46:28 UTC	Importing cluster image telekube v5.5.48-dev.8
Thu Jun 11 00:47:10 UTC	Synchronizing application with Docker registry 192.168.99.102:5000
Thu Jun 11 00:47:18 UTC	Verifying cluster health
Thu Jun 11 00:48:52 UTC	Cluster image has been uploaded
ubuntu@node-1:~/upgrade$ echo $?
0
```

Issue #1659, opened 2020-06-04
Intermittent etcd inavailability causes upgrade to fail to create an operation


a-palchikov:
## Current Result
If etcd is intermittently unavailable when the upgrade is initiated, the operation might fail to create. This can be problematic if the upgrade is used from an automation script.

## Expected Result
Instead, if the issue is intermittent, retry the attempt in a loop.

knisbet:
Do we have any insight into why etcd is going intermittently unavailable? While we should retry on failures, it seems if etcd is constantly flapping something else is also going wrong.

a-palchikov:
IIRC the etcd was becoming unresponsive for a short period of time including the embedded health check (i.e. the etcd logs displayed it as well) failing with unable to reach server.

This was triggered by using a script to routinely perform the same set of steps:
 * pull the installer package
 * uploading the packages to cluster store
 *  trigger the upgrade operation

Before the upload is initiated (`gravity update upload`), it checks the cluster health which means the cluster is healthy prior to the steps above. Somewhere into the operation, etcd gets a busy spike (I believe it was related to read requests) and becomes reproducibly unavailable long enough for planet agent to transition the cluster to degraded.

r0mant:
Released in [5.5.48](https://gravitational.com/gravity/docs/changelog/#5548-lts-june-15th-2020), forward-port PRs are linked above.

PR #1709, opened 2020-06-16
(6.1) Check cluster status after upload

r0mant:
## Description

Backport https://github.com/gravitational/gravity/pull/1687 to 6.1.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1659.
* Ports https://github.com/gravitational/gravity/pull/1687.

## TODOs

- [x] Self-review the change
- [x] Address review feedback

## Testing done

None, having robotest pass should be enough for this forward-port.


a-palchikov (approved):

lib/status/wait.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/gravitational/gravity/lib/defaults"
+	"github.com/gravitational/gravity/lib/ops"
+	"github.com/gravitational/gravity/lib/utils"
+
+	"github.com/gravitational/trace"
+	"github.com/sirupsen/logrus"
+)
+
+var log = logrus.WithField(trace.Component, "status")
```
a-palchikov, line 30: nit: all the refactoring so we can use the package global? It could also be used as an attribute in some configuration struct which I would prefer to a global.

a-palchikov (approved):

lib/status/wait.go
r0mant: I actually think in cases like this it is more convenient to have a package-global logger with the proper component field set that can be used from anywhere in the package (and extended with additional fields if needed).

Issue #1724, opened 2020-06-17
[7.0.x] wizard installation broken


a-palchikov:
When installing in wizard mode, the installer service fails with:
```
[ERROR]: unknown long flag '--no-selinux'
```

r0mant:
@a-palchikov Was this ported to master as well?

a-palchikov:
It is now: https://github.com/gravitational/gravity/pull/1881

PR #1725, opened 2020-06-17
[7.0.x] remove selinux configuration for wizard installer

a-palchikov:
## Description

SElinux configuration is not available for the wizard installer.

## Type of change

* Refs  https://github.com/gravitational/gravity/issues/1724.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

### Wizard-based installer

 1. Run wizard installer
```
$ ./install
```
 2. Follow the wizard steps and ensure the successful completion of the installer

### CLI-based installer

 1. Install 3-node cluster by first installing a single node
 1. Add additional nodes with `gravity join`
 1. Ensure the operations are successful

### DIFF
--- a/tool/gravity/cli/install.go
+++ b/tool/gravity/cli/install.go
@@ -793,10 +793,14 @@ var InterruptSignals = signals.WithSignals(
 func NewInstallerConnectStrategy(env *localenv.LocalEnvironment, config InstallConfig, commandArgs cli.CommandArgs) (strategy installerclient.ConnectStrategy, err error) {
 	commandArgs.FlagsToAdd = append(commandArgs.FlagsToAdd,
 		cli.NewFlag("token", config.Token),
-		cli.NewBoolFlag("selinux", config.SELinux),
 		cli.NewBoolFlag("from-service", true),
 		cli.NewArg("path", config.StateDir),
 	)
+	if config.Mode != constants.InstallModeInteractive {
+		commandArgs.FlagsToAdd = append(commandArgs.FlagsToAdd,
+			cli.NewBoolFlag("selinux", config.SELinux),
+		)
+	}
 	commandArgs.FlagsToRemove = append(commandArgs.FlagsToRemove, "token", "selinux", "path", "from-service")
 	args, err := commandArgs.Update(os.Args[1:])
 	if err != nil {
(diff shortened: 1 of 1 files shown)


tool/gravity/cli/install.go
```diff
@@ -793,10 +793,14 @@ var InterruptSignals = signals.WithSignals(
 func NewInstallerConnectStrategy(env *localenv.LocalEnvironment, config InstallConfig, commandArgs cli.CommandArgs) (strategy installerclient.ConnectStrategy, err error) {
 	commandArgs.FlagsToAdd = append(commandArgs.FlagsToAdd,
 		cli.NewFlag("token", config.Token),
-		cli.NewBoolFlag("selinux", config.SELinux),
 		cli.NewBoolFlag("from-service", true),
 		cli.NewArg("path", config.StateDir),
 	)
+	if config.Mode != constants.InstallModeInteractive {
+		commandArgs.FlagsToAdd = append(commandArgs.FlagsToAdd,
```
a-palchikov, line 800: Although I initially thought to move the decision about SELinux configuration outside this method, it really has all the context to make that decision (i.e. `config.Mode`) and would require far more refactoring otherwise.

tool/gravity/cli/install.go
```diff
@@ -817,13 +821,20 @@ func NewInstallerConnectStrategy(env *localenv.LocalEnvironment, config InstallC
 
 // newReconfiguratorConnectStrategy returns a new service connect strategy
 // for the agent executing the cluster reconfiguration operation.
-func newReconfiguratorConnectStrategy(env *localenv.LocalEnvironment, config InstallConfig, commandArgs cli.CommandArgs) (strategy installerclient.ConnectStrategy, err error) {
+func newReconfiguratorConnectStrategy(
+	env *localenv.LocalEnvironment,
+	config InstallConfig,
+	commandArgs cli.CommandArgs,
+) (strategy installerclient.ConnectStrategy, err error) {
+	commandArgs.FlagsToAdd = append(commandArgs.FlagsToAdd,
```
a-palchikov, line 829: These changes are to sync with the implementation of `NewInstallerConnectStrategy` (also below).

r0mant (approved):
I would also test [reconfigure operation](https://gravitational.com/gravity/docs/cluster/#changing-node-advertise-address) to make sure it doesn't regress since the changes affect it also.

tool/gravity/cli/install.go
bernardjkim, line 800: I'm not too familiar with the code yet, but could there be any problems if the `selinux` flag is not appended to `FlagsToAdd` but appended to `FlagsToRemove`?

bernardjkim (approved):

tool/gravity/cli/install.go
a-palchikov: The way it works is `FlagsToRemove` are handled first and they do not require the flag to actually be present on command line. After that, the code will update the command line with the flags given with `FlagsToAdd` so it should not break anything.

tool/gravity/cli/install.go
```diff
@@ -793,10 +793,14 @@ var InterruptSignals = signals.WithSignals(
 func NewInstallerConnectStrategy(env *localenv.LocalEnvironment, config InstallConfig, commandArgs cli.CommandArgs) (strategy installerclient.ConnectStrategy, err error) {
 	commandArgs.FlagsToAdd = append(commandArgs.FlagsToAdd,
 		cli.NewFlag("token", config.Token),
-		cli.NewBoolFlag("selinux", config.SELinux),
 		cli.NewBoolFlag("from-service", true),
 		cli.NewArg("path", config.StateDir),
 	)
+	if config.Mode != constants.InstallModeInteractive {
```
knisbet, line 799: I'm probably missing something, but doesn't this basically silently discard selinux configuration if trying to use the interactive installer?

a-palchikov: It is not discarding the SELinux configuration - the interactive installer is special in a way that it only installs "remotely" - it, on its own, is not supposed to change the system it is running on hence it omits this configuration.

Now, you're not restricted from running the agent on the same node where wizard is running and the agent could potentially configure SELinux if necessary (note, that due to lack of UI configuration and the off default for the SELinux flag, it will not be able to do that in UI-based case unless the agent's command is manually updated to include the SELinux flag).

Issue #1698, opened 2020-06-15
[BUG] Do not allow to launch upgrade if last upgrade plan isn't completed/rolled back


r0mant:
**Describe the bug**

We should not allow launching an upgrade if the last upgrade operation plan isn't completed or rolled back, even if the cluster state is "active".

**To Reproduce**

**Expected behavior**

**Logs**

**Environment (please complete the following information):**
 - OS [e.g. Redhat 7.4]:
 - Gravity [e.g. 5.5.4]:
 - Platform [e.g. Vmware, AWS]:

**Additional context**

r0mant:
Done in 5.5, tracking forward-ports in https://github.com/gravitational/gravity/issues/1740.

PR #1731, opened 2020-06-18
(5.5) Check last upgrade state when creating new upgrade operation

r0mant:
## Description

This PR adds more sanity checks when creating a new upgrade operation.

Previously, we only checked for "active" cluster state, which can pose a problem if somebody abandons an upgrade operation midway and manually resets the cluster state using internal `status-reset` command (which unfortunately support folks don't hesitate to use - more on that below too).

The additional checks (which can be bypassed with `--force` flag) are:

* That previous upgrade operation (if any) successfully completed, or
* That previous upgrade operation failed and its plan was fully rolled back.

Also, I updated the `gravity status-reset` command to display a large warning message to explain that the command can potentially lead to inconsistent cluster state and confirm their intent. The warning can be bypassed by providing `--confirm` flag.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1698.

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

```sh
# Start operation in manual mode.
Wed Jun 17 23:31:58 UTC	Upgrading cluster from 5.5.48-dev.8 to 5.5.49-dev.5

# Attempt to start next one right away fails.
Wed Jun 17 23:32:06 UTC	Upgrading cluster from 5.5.48-dev.8 to 5.5.49-dev.5
[ERROR]: Upgrade operation can only be triggered for active clusters. This cluster is currently updating.
Use gravity status to see the cluster status and make sure that the cluster is active and healthy before retrying.

# Force-set active state and retry - also fails since the first operation is in progress.
ubuntu@node-1:~/upgrade$ sudo ./gravity status-reset --confirm
Cluster test state has been set to active
Wed Jun 17 23:32:24 UTC	Upgrading cluster from 5.5.48-dev.8 to 5.5.49-dev.5
[ERROR]: The last operation_update operation (c4433ee8-2f72-4499-b112-79fe02765844) is still in progress.
Use gravity plan to view the operation plan, and either resume or rollback it before attempting to start another operation.
You can provide --force flag to override this check.

# Complete first operation - next one should launch fine.
ubuntu@node-1:~/upgrade$ sudo ./gravity plan complete --operation-id=c4433ee8-2f72-4499-b112-79fe02765844
Wed Jun 17 23:32:59 UTC	Upgrading cluster from 5.5.48-dev.8 to 5.5.49-dev.5

# Execute a few phases, make one of them fail
ubuntu@node-1:~/upgrade$ sudo ./gravity plan execute --phase=/init
ubuntu@node-1:~/upgrade$ sudo ./gravity plan execute --phase=/bootstrap
Wed Jun 17 23:34:48 UTC	Executing "/bootstrap/node-1" locally
Wed Jun 17 23:34:49 UTC	Executing phase "/bootstrap" finished in 1 second
[ERROR]: rename /var/lib/gravity/site/update/394376198 /var/lib/gravity/site/update/gravity: file exists

# Complete the plan - next attempt should still fail since plan isn't rolled back.
ubuntu@node-1:~/upgrade$ sudo ./gravity plan complete --operation-id=2e28ade7-b450-47a3-a61a-b12720024cef
Wed Jun 17 23:35:29 UTC	Upgrading cluster from 5.5.48-dev.8 to 5.5.49-dev.5
[ERROR]: The last operation_update operation (2e28ade7-b450-47a3-a61a-b12720024cef) is in a failed state but its operation plan isn't fully rolled back.
Use gravity plan to view the operation plan, and either resume or rollback it before attempting to start another operation.
You can provide --force flag to override this check.

# Make sure can be forced.
ubuntu@node-1:~/upgrade$ sudo ./gravity upgrade --manual --force
Thu Jun 18 01:02:48 UTC	Upgrading cluster from 5.5.48-dev.8 to 5.5.49-dev.5
```

### DIFF
--- a/lib/ops/opsservice/clusterconfig.go
+++ b/lib/ops/opsservice/clusterconfig.go
@@ -145,7 +145,7 @@ func (s *site) createUpdateConfigOperation(req ops.CreateUpdateConfigOperationRe
 			Config:     req.Config,
 		},
 	}
-	key, err := s.getOperationGroup().createSiteOperation(op, false)
+	key, err := s.getOperationGroup().createSiteOperation(op)
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
--- a/lib/ops/opsservice/clusterconfig.go
+++ b/lib/ops/opsservice/clusterconfig.go
@@ -145,7 +145,7 @@ func (s *site) createUpdateConfigOperation(req ops.CreateUpdateConfigOperationRe
 			Config:     req.Config,
 		},
 	}
-	key, err := s.getOperationGroup().createSiteOperation(op)
+	key, err := s.getOperationGroup().createSiteOperation(op, false)
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
--- a/lib/ops/opsservice/operationgroup.go
+++ b/lib/ops/opsservice/operationgroup.go
@@ -20,6 +20,7 @@ import (
 	"sync"
 
 	"github.com/gravitational/gravity/lib/defaults"
+	"github.com/gravitational/gravity/lib/fsm"
 	"github.com/gravitational/gravity/lib/ops"
 	"github.com/gravitational/gravity/lib/schema"
 	"github.com/gravitational/gravity/lib/storage"
@@ -62,11 +63,11 @@ func (s swap) Check() error {
 }
 
 // createSiteOperation creates the provided operation if the checks allow it to be created
-func (g *operationGroup) createSiteOperation(operation ops.SiteOperation) (*ops.SiteOperationKey, error) {
+func (g *operationGroup) createSiteOperation(operation ops.SiteOperation, force bool) (*ops.SiteOperationKey, error) {
 	g.Lock()
 	defer g.Unlock()
 
-	err := g.canCreateOperation(operation)
+	err := g.canCreateOperation(operation, force)
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
@@ -99,7 +100,7 @@ func (g *operationGroup) createSiteOperation(operation ops.SiteOperation) (*ops.
 //
 // In case of failed checks returns trace.CompareFailed error to indicate that
 // the cluster is not in the appropriate state.
-func (g *operationGroup) canCreateOperation(operation ops.SiteOperation) error {
+func (g *operationGroup) canCreateOperation(operation ops.SiteOperation, force bool) error {
 	cluster, err := g.operator.GetSite(g.siteKey)
 	if err != nil {
 		return trace.Wrap(err)
@@ -111,6 +112,11 @@ func (g *operationGroup) canCreateOperation(operation ops.SiteOperation) error {
 	case ops.OperationExpand:
 		// expand has to undergo some checks
 		err := g.canCreateExpandOperation(*cluster, operation)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+	case ops.OperationUpdate:
+		err := g.canCreateUpgradeOperation(*cluster, operation, force)
 		if err != nil {
 			return trace.Wrap(err)
 		}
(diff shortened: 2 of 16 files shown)


lib/ops/opsservice/operationgroup.go
```diff
@@ -133,6 +139,70 @@ func (g *operationGroup) canCreateOperation(operation ops.SiteOperation) error {
...
+			log.WithError(err).Warn("Force-creating upgrade operation.")
+			return nil
+		}
+		return trace.Wrap(err)
+	}
+	// All last operation checks passed.
+	return nil
+}
+
+func (g *operationGroup) checkLastOperation(operation ops.SiteOperation) error {
```
r0mant, line 171: Made this more or less generic on purpose, so we can potentially check for other operations too, not just last upgrade.

lib/ops/utils.go
```diff
@@ -123,18 +123,16 @@ func GetLastCompletedOperation(key SiteKey, operator Operator) (*SiteOperation,
 		key.SiteDomain)
 }
 
-// GetLastUpdateOperation returns the last update operation
-//
-// If there're no operations or the last operation is not of type 'update', returns NotFound error
-func GetLastUpdateOperation(siteKey SiteKey, operator Operator) (*SiteOperation, error) {
-	lastOperation, _, err := GetLastOperation(siteKey, operator)
+// GetLastUpgradeOperation returns the most recent upgrade operation or NotFound.
+func GetLastUpgradeOperation(key SiteKey, operator Operator) (*SiteOperation, error) {
```
r0mant, line 127: This function wasn't actually used so it's ok that its logic changed.

knisbet:
I'm curious, why this approach, instead of say, seeing that we're already upgrading and just resuming the previous upgrade attempt?

a-palchikov:
@knisbet I think it might be confusing to the user if the command previously used to create a new upgrade operation would now change the behavior to resuming the last operation in certain cases. I think explicit warning and the ability to resume (there's still `gravity plan resume`) is better.

On the other hand, if what you suggest is the only mode of operation, it might be complicated to back out of an upgrade that cannot be resumed (i.e. when one of the nodes failed permanently).

r0mant:
Yes, exactly, this is to catch scenarios when someone started an upgrade, it failed and was abandoned/forgotten about, then cluster state was reset and then they upload a new version and try to upgrade again. This warning will remind them about an incomplete upgrade and prompt to either finish the last one of rollback. The decision will be explicit.

lib/ops/opsservice/operationgroup.go
```diff
@@ -133,6 +139,70 @@ func (g *operationGroup) canCreateOperation(operation ops.SiteOperation) error {
...
+	}
+	if trace.IsNotFound(err) {
+		return nil
+	}
+	if !fsm.IsRolledBack(plan) {
+		return trace.CompareFailed(
+			`The last %v operation (%v) is in a %v state but its operation plan isn't fully rolled back.
+Use gravity plan to view the operation plan, and either resume or rollback it before attempting to start another operation.
+You can provide --force flag to override this check.`,
+			operation.Type,
```
a-palchikov, line 199: `operation.TypeString()`
```
and either resume it or roll it back ...
```

lib/ops/opsservice/clusterconfig.go
```diff
@@ -145,7 +145,7 @@ func (s *site) createUpdateConfigOperation(req ops.CreateUpdateConfigOperationRe
 			Config:     req.Config,
 		},
 	}
-	key, err := s.getOperationGroup().createSiteOperation(op)
+	key, err := s.getOperationGroup().createSiteOperation(op, false)
```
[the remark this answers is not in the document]
r0mant: Yeah, I wasn't too happy about this either. I'll see if I can make it cleaner.

lib/ops/opsservice/operationgroup.go
```diff
@@ -63,10 +64,18 @@ func (s swap) Check() error {
 
 // createSiteOperation creates the provided operation if the checks allow it to be created
 func (g *operationGroup) createSiteOperation(operation ops.SiteOperation) (*ops.SiteOperationKey, error) {
+	return g.createSiteOperationWithOptions(operation, createOperationOptions{})
+}
+
+type createOperationOptions struct {
+	force bool
```
a-palchikov, line 71: So you don't want to move it to `operationGroup`? It could be an aspect that cuts not only through the operation creation but perhaps something else - if you look at it that way.

a-palchikov (approved):

knisbet:
> Yes, exactly, this is to catch scenarios when someone started an upgrade, it failed and was abandoned/forgotten about, then cluster state was reset and then they upload a new version and try to upgrade again. This warning will remind them about an incomplete upgrade and prompt to either finish the last one of rollback. The decision will be explicit.

I didn't say this should work if someone is trying to upgrade a different version. But what we do see users do, is run upgrade, see it fail, and then run upgrade again not knowing or having awareness of what to do to resume the upgrade, inspect the plan, etc.

lib/ops/opsservice/operationgroup.go
r0mant: Not really, I'm not sure it belongs as an operation group property, because operation group is maintained per operation, to me it's more like a property of the "create" call which is why I opted to make it this way.

Issue #1689, opened 2020-06-11
[BUG] Manual resumption of upgrades can be triggered on node with incomplete state


knisbet:
**Describe the bug**

Etcd upgrades contain a critical section, where the node planned to run upgrade stages, is expected to be the same node coordinating the upgrade. If an upgrade fails however, and someone manually resumes the upgrade from a different node, the node tracking state while etcd is down is now different that the node executing steps as per the plan, allowing the plan to be inconsistent and leading to failures due to missing dependencies.

**To Reproduce**

On a multi-master cluster
1. Trigger an upgrade on node 1 that will pause before the etcd phases (fail an automatic upgrade, or use manual upgrade)
2. On node 2, resume the upgrade.
3. Observe that the upgrade will fail on /etcd/restore phase

**Expected behavior**

The upgrade to succeed

**Logs**

**Environment (please complete the following information):**
 - OS [e.g. Redhat 7.4]:
 - Gravity [e.g. 5.5.4]: 5.5.47
 - Platform [e.g. Vmware, AWS]:

**Additional context**

During etcd upgrades, only one node has the complete plan. Phases that are dependent on prior phases, have been written in a way to execute on the node that is expected to have the complete plan. However, a failed upgrade, and a resumption within a terminal, can allow the node recording state to differ from if the automatic upgrade ran to completion. When originally implemented, this was known, but the internal behaviour was to force all remote steps, which allowed the node tracking the plan and executing steps to differ. At some point this was changed, which now requires the nodes to be the same.

Resuming an upgrade, should check whether critical etcd phases are present, and if they are, indicate to the user the correct node to run the resumption from.

PR #1737, opened 2020-06-18
Kevin/master/1689 ensure resumption on upgrade leader

knisbet:
## Description

Forces resumed upgrades that contain etcd upgrade phases to be run on the node coordinating the upgrades. This prevents unsynced plan phases from blocking the upgrade due to dependency resolution failures.

## Type of change

* Closes #, refs #
https://github.com/gravitational/gravity/issues/1689

## TODOs

- [x] Self-review the change
- [ ] Write tests
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

Do a manual upgrade that contains an etcd upgrade, and then resume from a different node.

Node 1:
```
root@kevin-test1:~/build# ./upload
Thu Jun 18 20:12:20 UTC	Importing application telekube v7.1.0-alpha.1.91
Thu Jun 18 20:13:15 UTC	Synchronizing application with Docker registry 10.162.0.7:5000
Thu Jun 18 20:15:07 UTC	Synchronizing application with Docker registry 10.162.0.6:5000
Thu Jun 18 20:16:18 UTC	Application has been uploaded
root@kevin-test1:~/build# ./gravity upgrade --manual
Thu Jun 18 20:16:38 UTC	Upgrading cluster from 7.0.0 to 7.1.0-alpha.1.91
Thu Jun 18 20:16:39 UTC	Deploying agents on cluster nodes
Deployed agent on kevin-test2 (10.162.0.6)
Deployed agent on kevin-test1 (10.162.0.7)
The operation has been created in manual mode.

See https://gravitational.com/gravity/docs/cluster/#managing-an-ongoing-operation for details on working with operation plan.
```

Node 2:
```
root@kevin-test2:~/7.0.0# /var/lib/gravity/site/update/agent/gravity plan resume
[ERROR]: Plan must be resumed on node Server(AdvertiseIP=10.162.0.7, Hostname=kevin-test1, Role=node, ClusterRole=master)
```


r0mant (commented):
I'm curious what would be the downsides of just enforcing resume to launch from coordinator in all cases, not just when etcd is present. The behavior would be more consistent.

Also, this probably needs a support for a `--force` flag to override this check.

r0mant (commented):
Also, this particular fix doesn't really address the actual issue the upgrade failed in the first place, right? From the investigation it looked like the original reason upgrade failed is "etcd restore" phase was executed on "node1" while "node3" was coordinating the upgrade. But this particular fix only ensures that the manual "resume" command is run from coordinator, not that "restore" phase would be run on the correct node. Or am I missing something?

I think we need to add appropriate `ExecServer` to all etcd phases to make sure they're executed where they should be. From the plan it's visible they're missing exec servers right now.

knisbet:
> I'm curious what would be the downsides of just enforcing resume to launch from coordinator in all cases, not just when etcd is present. The behavior would be more consistent.
> 
> Also, this probably needs a support for a `--force` flag to override this check.

The reason of the narrow approach, is that it was the original design intent of the upgrade process to be resumable from any node, and the restrictions placed on the etcd upgrades when originally implemented were these violations should only occur during etcd phases. So I'm just continuing with the restrictions that were placed on me from when etcd upgrades were originally implemented.

knisbet:
> Also, this particular fix doesn't really address the actual issue the upgrade failed in the first place, right? From the investigation it looked like the original reason upgrade failed is "etcd restore" phase was executed on "node1" while "node3" was coordinating the upgrade. But this particular fix only ensures that the manual "resume" command is run from coordinator, not that "restore" phase would be run on the correct node. Or am I missing something?
> 
> I think we need to add appropriate `ExecServer` to all etcd phases to make sure they're executed where they should be. From the plan it's visible they're missing exec servers right now.

It does address the problem. The issue is, that node 1 was coordinating the upgrade, but failed on some endpoints step. The upgrade was then resumed on node 3 within a user shell, which meant node 3 was now effectively coordinating the upgrade, because the resume was allowed to run on node 3. This changes the resume to be run on the same node that started the upgrade, which would be node 1 in our failure scenario.

r0mant (approved):
Don't forget to backport to 5.5 pls.

Issue #352, opened 2019-03-27
Update release scheudule for gravity


klizhentas:
## Description

** New releases page**

We are adopting new LTS and non LTS releases for Gravity:

* 24 months LTS releases starting gravity 5.5
* 4 months for non LTS commercial releases starting Gravity 5.5

This should be reflected in our Releases page that should look like:

Every major Gravity version `x.0.0` has it's long term support release, e.g. for `3.0.0` version
LTS starts with `3.51.0` with minor backwards compatible changes added over time until the end of support cycle.

| Release       | LTS | Initial Release | Last Updated         | Supported Until      | Kubernetes Version | Teleport Version |
| --------------|-----| -------------------- | -------------------- | ------------------ |------------------|
| 5.5.3         | LTS | March 26th, 2019 | March 26th, 2019     | September 7th, 2020  | 1.13.5             | 3.0.4            |

**Most important**

I think we should explore the possibility of releases with security backports should not be downloadable from our OSS page, only through enterprise houston. Non commercial non LTS releases should not be supported on extended schedule.

kontsevoy:
Two points

## Point One

What I'm not happy about is that it's impossible to tell if a release is an LTS or not. One must locate the "release table" and look that up. The version itself doesn't say, i.e. why `3.51.0` is LTS and `3.57.1` is not? 

We can consider one of:

- All major releases are LTS releases.
- All odd/even releases are LTS releases.
- Add explicit "LTS" at the end of the version, i.e. `6.1.0-lts` 

... or the combination of all 3, i.e. anything that starts with an even number is an LTS release and there's `-lts` appended to it. This way people can refer to their production systems as "6-series" or "8-series".

## Point Two

Why "every 4 months"? If there's a vulnerability or a production-critical bug, we'll be forced to release patches sooner, and then there will be milestones that would take longer. We are already struggling with juggling several milestones in the air, often introducing "intermediate" ones (like "project $250K"), and this proposal adds yet another dimension to cut releases for.

klizhentas:
RE: numbering scheme denoting LTS releases - I think it's the best way so people will get used to it. @r0mant please suggest the scheme.

r0mant:
A couple of thoughts:

First, I think we should avoid using suffixes like `-lts` because it will break semver comparisons.

Using even/odd numbers might work but it still might create some confusion as to how to tag a certain release, i.e. we also have "stable" releases which are not LTS but have new Kubernetes.

[...]

PR #1738, opened 2020-06-18
Refactor release table.

wadells:
## Description
Drawing from the discussion at #352, this patch tries to clarify a couple
confusing things about our release table:

 1) Releases that had a release date after their "Supported Until"
    didn't make any sense. Now they have the context of a distinction
    between a branch and a release, and when the first release on
    on the branch came out (which will always be before "End of
    Support").

 2) I tried to formalize language around "branch" vs "release".
    Previously we used release to mean both an individual tagged
    release, and the line of development it was associated with.
    This was confusing as 7.0 will be "a LTS Release" but 7.0.9 is
    not "a LTS release" because it is pre-LTS-ification of 7.0.
    Honestly, this is still pretty confusing (7.0 probably should have
    been LTS all along), but at least the branch metaphor aligns with
    what is happening behind the scenes.

## Type of change

Contributes to #352.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done
`make run-docs` and captured the following screenshot:

image

### DIFF
--- a/docs/7.x/changelog.md
+++ b/docs/7.x/changelog.md
@@ -1,42 +1,42 @@
 # Releases
 
-## Supported Releases
+## Supported Branches
 
-| Release       | LTS | Release Date         | Supported Until      | Kubernetes Version | Teleport Version |
-| --------------|-----| -------------------- | -------------------- | ------------------ |------------------|
-| 7.0.9         | No  | June 16th, 2020      | 7.1 is released      | 1.17.6             | 3.2.13           |
-| 6.1.29        | Yes | June 12th, 2020      | November 10th, 2021  | 1.15.12            | 3.2.12           |
-| 5.5.48        | Yes | June 15th, 2020      | March 7th, 2021      | 1.13.11            | 3.0.6-gravity    |
+| Branch | LTS | Latest Release | Latest Release Date | Initial Release Date | End of Support          | Kubernetes Version   | Teleport Version |
+| ------ | --- | -------------- | ------------------- | -------------------- | ----------------------- | -------------------- | ---------------- |
+| 7.0    | No  | 7.0.9          | June 16, 2020       | April 3, 2020        | 7.1 is released         | 1.17.6               | 3.2.13           |
+| 6.1    | Yes | 6.1.29         | June 12, 2020       | August 2, 2019       | November 10, 2021       | 1.15.12              | 3.2.12           |
+| 5.5    | Yes | 5.5.48         | June 15, 2020       | March 8, 2019        | March 8, 2021           | 1.13.11              | 3.0.6-gravity    |
 
-Gravity offers one Long Term Support (LTS) release for every 2nd Kubernetes
+Gravity offers one Long Term Support (LTS) branch for every 2nd Kubernetes
 minor version, allowing for seamless upgrades per Kubernetes
 [supported version skew policy](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-version-skew).
-LTS Gravity versions are supported with security and bug fixes for two years
-after the x.x.0 release date.
+Gravity LTS branches are supported with security and bug fixes for two years.
 
-Non-LTS (regular) versions of Gravity have the latest features and offer
-more current versions of Kubernetes. Regular versions of Gravity are supported
-with security and bug fixes until the subsequent Gravity release is published.
+Non-LTS (regular) branches of Gravity offer the latest features and include
+more current versions of Kubernetes. Regular branches of Gravity are supported
+with security and bug fixes until first release of the subsequent Gravity
+branch is published.
 
-## Unsupported Releases
+## Unsupported Branches
 
-These releases are past their End of Life date, and no longer receive security
+These branches are past their End of Support date, and no longer receive security
 and bug fixes. [Gravity customers](https://gravitational.com/gravity/demo/) can
-extend updates past EOL through customer agreements if required.
+extend updates past End of Support through customer agreements if required.
 
-| Release       | LTS | Release Date         | Supported Until      | Kubernetes Version | Teleport Version |
-| --------------|-----| -------------------- | -------------------- | ------------------ |------------------|
-| 6.3.18        | No  | June 1st, 2020       | April 3rd, 2020      | 1.17.6             | 3.2.13           |
-| 6.2.5         | No  | December 3rd, 2019   | December 18th, 2019  | 1.16.3             | 3.2.13           |
-| 6.0.10        | No  | October 17th, 2019   | August 2nd, 2019     | 1.14.7             | 3.2.12           |
-| 5.6.8         | No  | September 18th, 2019 | July 17th, 2019      | 1.14.7             | 3.0.6-gravity    |
-| 5.4.10        | No  | March 26th, 2019     | March 8th, 2019      | 1.13.5             | 2.4.10           |
-| 5.3.9         | No  | March 7th, 2019      | December 14, 2018    | 1.12.3             | 2.4.7            |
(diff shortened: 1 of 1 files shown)


r0mant (approved):

a-palchikov (approved):

knisbet:
I'm not sure I see reframing releases in the terms of branches makes this any clearer. It seems like we're just throwing more information out there.

> 1. Releases that had a release date after their "Supported Until"
> didn't make any sense. Now they have the context of a distinction
> between a branch and a release, and when the first release on
> on the branch came out (which will always be before "End of
> Support").

While I'm sure it does come up from time to time, I don't believe it's necessarily that uncommon for companies to release patches ontop of unsupported software by customer request. We sort of say that in the comments, but I do see the knee jerk question of why is there a release if it's unsupported. I'm not sure how much true confusion this really causes though. To me, showing three dates, with an original release, doesn't really change the fact that there was a patch after the official end of support.

> 2. I tried to formalize language around "branch" vs "release".
> Previously we used release to mean both an individual tagged
> release, and the line of development it was associated with.
> This was confusing as 7.0 will be "a LTS Release" but 7.0.9 is
> not "a LTS release" because it is pre-LTS-ification of 7.0.
> Honestly, this is still pretty confusing (7.0 probably should have
> been LTS all along), but at least the branch metaphor aligns with
> what is happening behind the scenes.

I think you're right, that we should have designated 7.0 LTS from the beginning, or when we do designate the LTS release mark it 7.1. But I don't think it's really correct to say that 7.0.9 is not an LTS release if we "promote" the branch to be our LTS release. There isn't an outward facing difference between 7.0.9 LTS and 7.0.9 Non-LTS. Basically from an outside perspective, all we've done is change/extend the supported until date for a major version. While we may do an extended set of tests before we extend the supported date, or contractually do something like run conformance testing, it doesn't mean from an outside perspective there is a substantial difference in the releases, other than when they exit support.

wadells:
@knisbet  TLDR: You bring up good clarifying points.  Is there anything you would like to see changed in the .md?

> I'm not sure I see reframing releases in the terms of branches makes this any clearer.

I stand pretty firmly behind the "branch" vs "release" vocabulary shift. This my attempt to clarify that a branch/line of development is what we provide support for, not any particular build off that release E.g. we won't backport a security fix to each 7.0.x release, only the 7.0 branchline of development as a whole.  Before this patch, we didn't have vocabulary in our documentation for distinguishing between the two -- everything was a release.

> While I'm sure it does come up from time to time, I don't believe it's necessarily that uncommon for companies to release patches on top of unsupported software by customer request. We sort of say that in the comments, but I do see the knee jerk question of why is there a release if it's unsupported. I'm not sure how much true confusion this really causes though. To me, showing three dates, with an original release, doesn't really change the fact that there was a patch after the official end of support.

That's fair.  I agree for the extra date column is mostly superfluous info.  If they hurt the scan-ability of the table I'd be glad to remove them.  They have parity with Sasha's suggestion in https://github.com/gravitational/gravity/issues/352 -- and because I agreed with his suggestion due quirks like post-support releases.

> I think you're right, that we should have designated 7.0 LTS from the beginning, or when we do designate the LTS release mark it 7.1. But I don't think it's really correct to say that 7.0.9 is not an LTS release if we "promote" the branch to be our LTS release. There isn't an outward facing difference between 7.0.9 LTS and 7.0.9 Non-LTS. Basically from an outside perspective, all we've done is change/extend the supported until date for a major version. While we may do an extended set of tests before we extend the supported date, or contractually do something like run conformance testing, it doesn't mean from an outside perspective there is a substantial difference in the releases, other than when they exit support.

Understood -- I believe the docs reflect this, and any confusion is limited to the discussion in this PR. Perhaps a better example than 7.0 would be what @klizhentas discussed in https://github.com/gravitational/gravity/issues/352#issue-426209051

> Every major Gravity version x.0.0 has it's long term support release, e.g. for 3.0.0 version
LTS starts with 3.51.0 with minor backwards compatible changes added over time until the end of support cycle.

"[S]tarts with 3.51.x" connotes 3.x  releases < 3.51.x are not LTS, which is moot (3.x is dead), but also confusing. The current 7.0 situation has echos of 3.x, and [I'm not the only one who is confused](https://grv8.slack.com/archives/CEZM4JFDF/p1592452729103700) -- this is the discussion that prompted me to write this patch.

knisbet:
> @knisbet TLDR: You bring up good clarifying points. Is there anything you would like to see changed in the .md?

Yes, we need to align on our messaging, otherwise we'll just go back and fourth.

> I stand pretty firmly behind the "branch" vs "release" vocabulary shift. This my attempt to clarify that a branch/line of development is what we provide support for, not any particular build off that release E.g. we won't backport a security fix to each 7.0.x release, only the 7.0 branch / line of development as a whole. Before this patch, we didn't have vocabulary in our documentation for distinguishing between the two -- everything was a release.

I don't think using "Branch" as the nomenclature for a Major.Minor version is the best fit. It seems to uplift the VCS terminology for how the code is organized, not reflect an outward facing version policy. So it makes lots of sense to us, because we live it in source code every day, but an outside observer shouldn't be expected to know this. 

What we have traditionally indicated, is an outward version policy of SemVer, which operates in terms of Major, Minor, and Patches, where Major.Minor make up a supported version, some of which will have LTS patch cycles. 

Polling some well known software to try and see how they communicate similar information, seems to be fairly varied:
Ubuntu: https://wiki.ubuntu.com/Releases
RedHat: https://access.redhat.com/articles/3078
Openshift: https://docs.openshift.com/container-platform/4.2/release_notes/versioning-policy.html
Kubernetes: https://kubernetes.io/docs/setup/release/version-skew-policy/ (Does use release branches)
Oracle: https://en.wikipedia.org/wiki/Oracle_Database#Releases_and_versions
InfluxData: https://docs.influxdata.com/influxdb/v1.8/about_the_project/releasenotes-changelog/ (I don't see a table at cursory glance)
Docker EE: https://docs.mirantis.com/docker-enterprise/v3.0/dockeree-products/docker-ee/release-notes.html (Appears to use version)
Linux Kernel: https://www.kernel.org/category/releases.html
Rancher: https://rancher.com/docs/rancher/v2.x/en/installation/options/server-tags/
Tanzu: https://docs.pivotal.io/pks/1-6/release-notes.html

Overall, I don't think there is any generally accepted terminology here. What I see most often doing a quick survey, is to refer to the Major.Minor as a Version. So if anything, I would consider going the other direction, and from a release cycle perspective, only present the following information: 

Version: Major.Minor
LTS: Yes/No
Release Date: Date of the .0 Patch release
Supported Until: 3 years or 4 months above Release Date based on LTS designation. Or preferably we'd get to a point where we separate End of Support versus End of Security, but we don't currently distinguish that when we should as Ubuntu does.

Patches can then be individually dated as they already are, sticking with SemVer nomenclature. When we say LTS release, from a SemVer standpoint, we're talking about the Major.Minor combination. 

Example

| Version | LTS | Release Date | Supported Until |
| ------- | --- | -------------- | ---------------- |
| 7.0 | Yes | January 1st, 2020 | January 1st, 2023 |
| 6.3 | No | March 3rd, 2019 | June 3rd, 1029 |
| 6.1 | Yes | August 2nd, 2019 | November 10th, 2021 |

This doesn't address how we communicate Kubernetes/Teleport and dependency versions, although there was already a side discussion around this, because we don't communicate other dependencies (etcd as an example). For the short term I'd continue to include Kubernetes/Teleport in this table, but long term a wider variety of dependencies will be communicated in some other type of table, maybe something like Tanzu does.

wadells:
@knisbet I drafted some changes incorporating your feedback here: https://github.com/gravitational/gravity/pull/1751

We can continue the conversation there.

Issue #1740, opened 2020-06-18
Tracking back/forward ports


r0mant:
The following issues are missing back or forward ports.

---

* [#1655 https://github.com/gravitational/gravity/issues/1655](https://github.com/gravitational/gravity/issues/1655) (@a-palchikov)
  - [x] 6.1
  - [x] 7.0
  - [ ] master

---

* [#1611 Set up priority classes for system pods](https://github.com/gravitational/gravity/issues/1611) (@UnderGreen)
  - [ ] 6.1
  - [ ] 7.0
  - [ ] master

---

* [#1623 Revise RPC credentials handling](https://github.com/gravitational/gravity/issues/1623) (@a-palchikov)
  - [x] 6.1
  - [x] 7.0
  - [ ] master

---

* [#1506 Monitoring doesn't provide Compute Resource/Pod](https://github.com/gravitational/gravity/issues/1506) (@UnderGreen)
  - [ ] 6.1
  - [x] 7.0
  - [ ] master

---

* [#1836 Tele login fixes](https://github.com/gravitational/gravity/pull/1836) (@r0mant)
  - [x] 7.0
  - [ ] master

r0mant:
Completed ports:

---

* [#1209 Node degraded due to kube-apiserver after upgrade](https://github.com/gravitational/gravity/issues/1209) (@a-palchikov)
  - [x] 6.1
  - [x] 7.0
  - [x] master

---

* [#1703 CVE-2020-13379 - incorrect access control vulnerability Grafana](https://github.com/gravitational/gravity/issues/1703) (@UnderGreen)
  - [x] 6.1
  - [x] 7.0
  - [x] master

---

* [#1695 Ops Center Monitoring Console displays "Dashboard not found"](https://github.com/gravitational/gravity/issues/1695) (@UnderGreen)
  - [x] 6.1
  - [x] 7.0
  - [x] master

---

PR #1747, opened 2020-06-22
[6.1.x] backport RPC credentials rotation from 5.5.x

a-palchikov:
## Description

The PR does not port the RPC `rotate` sub-command as the taken approach is to refresh the RPC credentials on each cluster upgrade operation.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1740.

* Ports https://github.com/gravitational/gravity/pull/1629, https://github.com/gravitational/gravity/pull/1363.

## TODOs

- [x] Self-review the change
- [ ] Write tests
- [ ] Perform manual testing
- [ ] Address review feedback

### DIFF
--- a/lib/app/docker/imageservice.go
+++ b/lib/app/docker/imageservice.go
@@ -138,21 +138,26 @@ func newImageService(req RegistryConnectionRequest) (*imageService, error) {
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
+	// Strip the scheme prefix if specified in the address as we manipulate
+	// the registry address without the URL scheme when wrapping/unwrapping image references
+	registryPrefix := strings.TrimPrefix(req.RegistryAddress, "http://")
+	registryPrefix = strings.TrimPrefix(registryPrefix, "https://")
 	return &imageService{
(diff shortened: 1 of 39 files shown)


knisbet (approved):

lib/app/docker/imageservice.go
```diff
@@ -138,21 +138,26 @@ func newImageService(req RegistryConnectionRequest) (*imageService, error) {
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
+	// Strip the scheme prefix if specified in the address as we manipulate
+	// the registry address without the URL scheme when wrapping/unwrapping image references
+	registryPrefix := strings.TrimPrefix(req.RegistryAddress, "http://")
```
a-palchikov, line 143: The problem with `make test` is that it executes the tests from within a container which skips the tests tha require Docker themselves, hence I stumbled on a couple of failing tests.
But this seems to only affect 6.1.x - i.e. the unit tests are not failing for 7.0.x (and possibly for master) which makes me think I may need to look into that in more detail.

PR #1751, opened 2020-06-22
Remove patch releases from release tables.

wadells:
## Description
This PR incorporates some feedback for #1738 that came in after it merged:

  https://github.com/gravitational/gravity/pull/1738#issuecomment-647720409

Branch is changed to version, and rapidly changing info (latest release
& latest release date) are removed from the table entirely.

## Type of change

This is a continuation of #1738, incorporating feedback after the that had merged.

Contributes to #352.

## TODOs
- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done
`make run-docs` and captured the following screenshot.

image

### DIFF
--- a/docs/7.x/changelog.md
+++ b/docs/7.x/changelog.md
@@ -42,7 +42,7 @@ extend updates past End of Support through customer agreements if required.
 
 # Release Notes
 
-## 7.x Releases
+## 7.0 Releases
 
 ### 7.0.10 (June 23rd, 2020)
 
@@ -201,7 +201,7 @@ them to any Kubernetes cluster.
 * [Cluster Status History](cluster.md#cluster-status-history)
 to learn how to gain insight into how the cluster status changes over time.
 
-## 6.x Releases
+## 6.3 Releases
 
 ### 6.3.18 (June 1st, 2020)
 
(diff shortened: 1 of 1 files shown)


helgi:
I would have to read the changelog to find the latest release of each?

wadells:
> I would have to read the changelog to find the latest release of each?

@helgi Should this be merged in its present form, I'd recommend grabbing latest patch releases from https://gravitational.com/gravity/download. Even in the present state the downloads page is more current & reliable than these charts -- the changelog can lag a day or two behind publishing as changelogs frequently get batched up when there are lots of releases close together.

I'm not sure if it will go out as is: this is a draft continuing some late feedback for #1738.  I put it up for review because I find it easier to critique/get feedback when there is a concrete proposal.

knisbet:
@wadells The changelog is not usually a day or two behind. The only circumstances this generally happens, is if there are two simaltanous PRs that don't account for eachother. What happens, is while the PR is still outstanding, most release managers publish from the PR branch instead of master, so the proposed docs are publicly available, while the PR is still under review.

@helgi This is correct. The recent changes are to try and reduce confusion, although to continue the overall structure update, what we should do is organize the changelog by "Version". So in the sub-topics on the left, the changelog would take you strait to 7.1, 6.3, etc, with the most recent change at the top.

docs/7.x/changelog.md
```diff
@@ -1,42 +1,44 @@
...
 
-## Supported Branches
+## Supported Versions
 
-| Branch | LTS | Latest Release | Latest Release Date | Initial Release Date | End of Support          | Kubernetes Version   | Teleport Version |
-| ------ | --- | -------------- | ------------------- | -------------------- | ----------------------- | -------------------- | ---------------- |
-| 7.0    | No  | 7.0.9          | June 16, 2020       | April 3, 2020        | 7.1 is released         | 1.17.6               | 3.2.13           |
-| 6.1    | Yes | 6.1.29         | June 12, 2020       | August 2, 2019       | November 10, 2021       | 1.15.12              | 3.2.12           |
-| 5.5    | Yes | 5.5.48         | June 15, 2020       | March 8, 2019        | March 8, 2021           | 1.13.11              | 3.0.6-gravity    |
+Find the latest releases of supported Gravity versions at [Gravity Downloads](https://gravitational.com/gravity/download).
```
knisbet, line 5: I wonder if we should specifically indicate this as Gravity OSS.

knisbet (approved):
This fits closer to my personal preference. 

Although, not so much for this PR, but it might be worth considering helgi's comment, where we also organize the changelog by "Version", and maybe provide links in the table. So Instead of having a sub-category for 7.x, we create ones for 7.0, 7.1, 6.3, etc.

docs/7.x/changelog.md
wadells: Probably.  If we indicate this is OSS only, I'd like to give a pointer to enterprise users too.  However I haven't found the a similar resource to the OSS downloads page.  Helgi's question leads me to believe it may be this changelog page.

knisbet: No, as I mentioned in slack, the references to enterprise downloads got completely removed from the docs for some reason. As such, currently someone from professional services needs to instruct how to get the enterprise downloads.

wadells: Specified OSS in 5f84e8a.

wadells:
> Although, not so much for this PR, but it might be worth considering helgi's comment, where we also organize the changelog by "Version", and maybe provide links in the table. So Instead of having a sub-category for 7.x, we create ones for 7.0, 7.1, 6.3, etc.

Refactored to sort changelogs by version in a5ccf1e and added jumplinks in 0926cfe.  I think this brings back most of the usability that having the most current version in the table provided, though it does require an extra click.

r0mant (approved):
This is useful 👍

PR #1719, opened 2020-06-16
[5.5.x] Add since flag to gravity report

bernardjkim:
## Description

This PR adds the `since` flag to `gravity report` command. This flag is used to filter the log entries by time. The current implementation only affects the log entries collected from `kubectl logs` and `journalctl`.

Add a few more system collectors that have come up recently:
* Reboots history: `last -x`
* Kernel information: `uname -a`
* Serf cluster: `serf members`
* Swap status: `swapon -s (we have vmstat but wouldn't hurt)`
* Etcd metrics: `curl -s --insecure --tlsv1.2 --cacert /var/state/root.cert --cert /var/state/etcd.cert --key /var/state/etcd.key https://localhost:4001/metrics`

## Type of change

* New feature (non-breaking change which adds functionality)
* This change has a user-facing impact

## Linked tickets and other PRs

* Updates https://github.com/gravitational/gravity/issues/1273

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

**Verify additional collectors**
```
[gravity-report] tree report                                                                                                                   bernard/5.5.x/time-filter  ✭ ✱
report
├── cluster.json
├── node-1-debug-logs
│   ├── bridge-fdb
│   ├── cluster-status
│   ├── df
│   ├── df-inodes
│   ├── dmesg
│   ├── dmsetup
│   ├── etcd-status
│   ├── etc-logs.tar.gz
│   ├── fdisk
│   ├── free
│   ├── gravity-install.log
│   ├── gravity-packages.yaml
│   ├── gravity-system.log
│   ├── gravity-system.log.gz
│   ├── host-system-failed
│   ├── host-system-jobs
│   ├── host-system-status
│   ├── ifconfig
│   ├── ip-addr
│   ├── ip-link
│   ├── ip-maddress
│   ├── ip-neighbor
│   ├── ip-netconf
│   ├── ip-ntable
│   ├── ip-route
│   ├── ip-rule
│   ├── iptables
│   ├── ip-tcp_metrics
│   ├── lsblk
│   ├── lscpu
│   ├── lsmod
│   ├── planet-journal-export.log.gz
│   ├── planet-status
│   ├── planet-system-failed
│   ├── planet-system-jobs
│   ├── planet-system-status
│   ├── reboot-history  <---
│   ├── route
│   ├── running-processes
│   ├── serf-members  <---
│   ├── slabinfo
│   ├── slabtop
│   ├── swapon <---
│   ├── uname  <---
[...]

PR #1755, opened 2020-06-23
[5.5.x] update report collector

a-palchikov:
## Description

This PR introduces the following changes:
* propagates --since to planet journal log collector
* fixes journalctl binary location inside planet

## Type of change

* Refs https://github.com/gravitational/gravity/pull/1719

## TODOs

- [x] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback

## Testing done

Ran various combinations of `gravity report` and `gravity system report` and verified availability of expected data points (e.g. planet journal logs).

### DIFF
--- a/lib/defaults/defaults.go
+++ b/lib/defaults/defaults.go
@@ -326,7 +326,7 @@ const (
 	SerfBin = "/usr/bin/serf"
 
 	// JournalctlBin is the default location of the journalctl inside planet
-	JournalctlBin = "/usr/bin/journalctl"
+	JournalctlBin = "/bin/journalctl"
 
 	// SystemctlBin is systemctl executable inside planet
 	SystemctlBin = "/bin/systemctl"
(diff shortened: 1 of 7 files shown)


a-palchikov:
Also, I saw that https://github.com/gravitational/gravity/pull/1719 removed k8s logs collection using `kubectl cluster-info dump` - why?

r0mant (approved):
@a-palchikov "cluster-info dump" should still be collected if you don't specify the time filter. It was only supposed to be skipped (related with regular kctl logs) if you provide --since flag because it "cluster-info dump" doesn't allow to filter logs by time. Maybe it's a side-effect of the fact that --since has default value? cc @bernardjkim

bernardjkim:
@a-palchikov ah, the intention was that passing `--since 0s` to `gravity report` would collect k8s logs using `kubectl cluster-info dump`. Looks like I made a mistake.

A default `since` value was specified for both `gravity report` and `gravity system report`. So if `--since 0s` is passed to `gravity report` `gravity system report` will be called with unspecified `--since` and it will defer to the default value. 

We can fix this by removing the default value for `gravity system report`. Or we can explicitly pass `0s` to `gravity system report` when calling `gravity report`.

Issue #1748, opened 2020-06-22
Disk checker kicks in even if requirements are not specified


a-palchikov:
Reported from a 7.0.9 installation:
```
[  129.608996] cloud-init[1120]: [×] disk capacity at /proc is 0 (failed to validate storage requirements)
```

burdzwastaken:
we are also seeing this behaviour in `7.0.9`:
```
[  118.100276] cloud-init[1123]: [31m[ERROR]: The following pre-flight checks failed:
[  118.104108] cloud-init[1123]: [×] disk capacity at /proc is 0 (failed to validate storage requirements)
```

r0mant:
Released in [7.0.11](https://gravitational.com/gravity/docs/changelog/#7011-june-24th-2020), ports are tracked in https://github.com/gravitational/gravity/issues/1740.

PR #1758, opened 2020-06-23
[7.0.x] Skip disk usage check during pre-flight checks

bernardjkim:
## Description

This PR fixes a bug that was introduced during changes to the storage check due to satellite setting a default HighWatermark value to the storage checker. Storage checkers with unspecified or 0 HighWatermark value will skip the disk usage check.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1748
* Requires https://github.com/gravitational/satellite/pull/237, https://github.com/gravitational/planet/pull/692

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

**Verify successful install of single node cluster with mounted /proc volume**

* Create mount for /proc node profile in telekube manifest
```
nodeProfile
  - name: node
     ...
       volumes:
       - path: /proc
          targetPath: /hostproc
          createIfMissing: false
```

```
// Before changes install fails due to failed pre-flight check.
TASK [Install gravity application] *******************************************************************************************************************************************
task path: /home/bernard/go/src/github.com/gravitational/gravity/vagrant/ansible/install.yaml:8
fatal: [node-1]: FAILED! => {"changed": true, "cmd": "umask 0066\n cd /vagrant/installer\n  ./gravity install --debug --cloud-provider=generic --flavor=standalone --cluster=dev.test --advertise-addr=172.28.128.101 --token=token123\n ", "delta": "0:00:00.271563", "end": "2020-06-24 01:21:21.531486", "msg": "non-zero return code", "rc": 255, "start": "2020-06-24 01:21:21.259923", "stderr": "\u001b[31m[ERROR]: The following pre-flight checks failed:\n\t[×] disk capacity at /proc is 0 (failed to validate storage requirements)\n\n\u001b[0m", "stderr_lines": ["\u001b[31m[ERROR]: The following pre-flight checks failed:", "\t[×] disk capacity at /proc is 0 (failed to validate storage requirements)", "", "\u001b[0m"], "stdout": "\u001b[1mWed Jun 24 01:21:21 UTC\u001b[0m\tStarting installer", "stdout_lines": ["\u001b[1mWed Jun 24 01:21:21 UTC\u001b[0m\tStarting installer"]}
```

```
// After changes install is sucessful
TASK [Install gravity application] *******************************************************************************************************************************************
task path: /home/bernard/go/src/github.com/gravitational/gravity/vagrant/ansible/install.yaml:8
changed: [node-1] => {"changed": true, "cmd": "umask 0066\n cd /vagrant/installer\n  ./gravity install --debug --cloud-provider=generic --flavor=standalone --cluster=dev.test --advertise-addr=172.28.128.101 --token=token123\n ", "delta": "0:11:27.694958", "end": "2020-06-24 01:48:37.844384", "rc": 0, "start": "2020-06-24 01:37:10.149426", ...
```


lib/schema/checks.go
```diff
@@ -138,6 +138,7 @@ func ValidateRequirements(reqs Requirements, stateDir string) (failed []*pb.Prob
 				WillBeCreated:     utils.BoolValue(vol.CreateIfMissing),
 				Filesystems:       vol.Filesystems,
 				MinFreeBytes:      vol.Capacity.Bytes(),
+				HighWatermark:     0, // skip disk usage check
```
r0mant, line 141: Hmm, not sure how this helps? If HighWatermark is not initialized, its value will also be 0, right?

a-palchikov:
I'm not sure this is headed in the right direction.
The original premise is avoiding handling of virtual (or pseudo) file systems in the first place - not adding knobs to conditionally switch off testing. I don't see how skipping disk usage test will help.

I would look in the direction of programmatically detecting a filesystem type and excluding them from test explicitly.
One way this seems this could be achieved is via the `stat` syscall:
```Go
var stat unix.Stat_t
if err := unix.Stat(path, &stat); err != nil {
	return trace.ConvertSystemError(err)
}
if unix.Major(stat.Dev) == 0 && stat.Size == 0 {
	// exclude this filesystem from disk-specific tests
}
```
This is not 100% as I could not verify it from official documentation certain but close t
Also, this needs to be implemented taking OS environment into account - i.e. only providing a `Linux` implementation with empty stub for `Darwin`.

Kernel [device documentation](https://www.mjmwired.net/kernel/Documentation/devices.txt)
[Stat_t](https://godoc.org/golang.org/x/sys/unix#Stat_t)
[SO discussion](https://unix.stackexchange.com/questions/80324/determining-if-a-file-system-object-is-a-procfs-sysfs-etc-virtual-file-which)

knisbet:
> I'm not sure this is headed in the right direction.
> The original premise is avoiding handling of virtual (or pseudo) file systems in the first place - not adding knobs to conditionally switch off testing. I don't see how skipping disk usage test will help.
> 
> I would look in the direction of programmatically detecting a filesystem type and excluding them from test explicitly.
> One way this seems this could be achieved is via the `stat` syscall:
> 
> ```go
> var stat unix.Stat_t
> if err := unix.Stat(path, &stat); err != nil {
> 	return trace.ConvertSystemError(err)
> }
> if unix.Major(stat.Dev) == 0 && stat.Size == 0 {
> 	// exclude this filesystem from disk-specific tests
> }
> ```
> 
> This is not 100% as I could not verify it from official documentation certain but close t
> Also, this needs to be implemented taking OS environment into account - i.e. only providing a `Linux` implementation with empty stub for `Darwin`.
> 
> Kernel [device documentation](https://www.mjmwired.net/kernel/Documentation/devices.txt)
> [Stat_t](https://godoc.org/golang.org/x/sys/unix#Stat_t)
> [SO discussion](https://unix.stackexchange.com/questions/80324/determining-if-a-file-system-object-is-a-procfs-sysfs-etc-virtual-file-which)

I might be missing something, but do we know why we're even trying to check disk capacity here? I assume it's not specified in the app.yaml? My expectation would be that we're only trying to test or bubble up errors for volumes that specifically indicate a capacity within the app.yaml, and wouldn't require the effort in trying to filter by various types of volumes, as presumably that's a more complicated issue than just detecting virtual filesystems such as /proc.

r0mant:
@knisbet @a-palchikov Yes, this is a simple regression that we try to check disk usage where we shouldn't (since it's not specified in the manifest but satellite sets the default and checks it anyway). I discussed it with @bernardjkim yesterday, we're gonna remove setting defaults in satellite.

r0mant (approved):
@bernardjkim Please fill out testing done before merging to show the scenario that was failing before and that now's passing.

PR #1760, opened 2020-06-23
Update robotest upgrade targets.

wadells:
## Description
This patch drops validation of 6.1 to 7.1.0-alpha upgrades.
These are replaced with validation from 7.0. Furthermore 6.2 and 6.3
are included per our <=2 minor kubernetes version skew policy and some
recent discussion with @knisbet about whether we support non-LTS to
non-LTS upgrades.

## Type of change

## TODOs

- [x] Self-review the change
- [x] Perform testing
- [x] Address review feedback

## Testing done
The PR build is sufficient testing.

### DIFF
--- a/build.assets/robotest_run_nightly.sh
+++ b/build.assets/robotest_run_nightly.sh
@@ -9,15 +9,17 @@ DOCKER_STORAGE_DRIVERS="overlay2"
 declare -A UPGRADE_MAP
 
 # latest patch release on compatible LTS, keep this up to date
-UPGRADE_MAP[6.1.18]="centos:7 debian:9 ubuntu:18"
+UPGRADE_MAP[7.0.10]="centos:7 debian:9 ubuntu:18"
 
-# latest patch release on supported non-LTS version, keep this up to date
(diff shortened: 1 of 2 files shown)


a-palchikov (approved):

knisbet (approved):
Overall, I think this reflects our model and the underlying kubernetes version skew, so good stuff.

The only thing we may want to consider, as an exception to the underlying model, is 6.2 has been out of support for more than 6 months at this point, and 6.3 hasn't had much adoption, with enterprise customers going to 7.0, so it might be safe to try and save a few CPU cycles, dropping 6.2 and maybe 6.3 for the time being.

wadells:
> The only thing we may want to consider, as an exception to the underlying model, is 6.2 has been out of support for more than 6 months at this point, and 6.3 hasn't had much adoption, with enterprise customers going to 7.0, so it might be safe to try and save a few CPU cycles, dropping 6.2 and maybe 6.3 for the time being.

I yanked 6.2 and 6.3 from PR builds in 5269615, but I left them in nightly builds.  This is probably an appropriate level of testing for "technically supported but unlikely to ever be used"

Issue #1691, opened 2020-06-11
[BUG] 6.1.25 to 6.1.28 upgrade fails: "[ERROR]: client: etcd cluster is unavailable or misconfigured"


wadells:
**Describe the bug**
Upgrades from 6.1.25 to 6.1.28 (and potentially other versions -- TBD) fail when they cannot connect to etcd after the upgrade.

**To Reproduce**
Install 6.1.25
Upgrade to 6.1.28
Monitor the install operation from the master and shortly you'll see:
```
$ sudo ./gravity status --operation-id=c993ed62-8608-4b2a-a909-16a196457c88
[ERROR]: client: etcd cluster is unavailable or misconfigured; error #0: dial tcp 127.0.0.1:2379: connect: connection refused

$
````

Alternatively, the following robotest scenario will recreate the issue:
```
upgrade={"flavor":"three","nodes":3,"role":"node","service_uid":997,"service_gid":994,"os":"redhat:7","installer_url":"s3://hub.gravitational.io/gravity/oss/app/telekube/6.1.28/linux/x86_64/telekube-6.1.28-linux-x86_64.tar","from":"s3://hub.gravitational.io/gravity/oss/app/telekube/6.1.25/linux/x86_64/telekube-6.1.25-linux-x86_64.tar"}
```

**Expected behavior**
I expected the upgrade to complete successfully.

**Logs**
[robotest logs](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&minLogLevel=0&timestamp=2020-06-11T04:00:06.964000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%2250483a87-ff13-48fb-8aa7-0051e9dfc506%22%0Alabels.__suite__%3D%22fcedae8b-0888-4699-8d71-2fe9ceba26b1%22&dateRangeEnd=2020-06-11T04:00:04.047Z&interval=PT1H&scrollTimestamp=2020-06-11T02:03:53.620347971Z&dateRangeUnbound=backwardInTime)

**Environment (please complete the following information):**
 - OS [e.g. Redhat 7.4]: SUSE 12 sp5
 - Gravity [e.g. 5.5.4]: 6.1.25/6.1.28
 - Platform [e.g. Vmware, AWS]: GCP

wadells:
Not sure why merging 1694 didn't close this automatically. Closing now.

Issue #1199, opened 2020-03-04
Find a way to keep upgrade versions in sync with release versions more easily


wadells:
As suggested by @a-palchikov in https://github.com/gravitational/gravity/pull/1195#discussion_r387532679 we should find a way to keep our "upgrade from" (and "upgrade to" if those are in the code base by the time this lands) versions current.  The versions found here:

https://github.com/gravitational/gravity/blob/7007ed17a583c0b3c6668770d7affe62f3337eb6/build.assets/robotest_run_suite.sh#L8-L20

https://github.com/gravitational/gravity/blob/7007ed17a583c0b3c6668770d7affe62f3337eb6/build.assets/robotest_run_nightly.sh#L10-L22

These currently need to be manually updated & committed by employees each time there is a new release, which is quite tedious and low ROI work.

wadells:
One suggestion is moving these upgrade versions closer to `version.sh` or the code it is replaced with in #1196.  In the Makefile as @a-palchikov suggested:

> One thing that popped up a while ago is keeping these versions in sync. I think if we moved them to the makefile closer to where the versions are defined / queried, it would be much easier to handle.
> Something along the lines of:
> 
> ```
> VERSIONS_LTS_TO_LATEST = 5.5.37 5.2.15
> VERSIONS_TO_LATEST = 5.2.12 5.5.19 5.5.20 5.5.28 5.5.36
> ```

Another approach would be to have a tool that could dynamically query releases (unfortunately involving 1+ network round trips) to always come up with an active list.  This would utilize a hybrid of info in the wild (e.g. latest LTS patch release before & after the build under test) as well as some statically defined info (important releases in the wild).

We can do a bit more design when we find priority for this bug.

PR #1761, opened 2020-06-23
Add logic to test upgrade from previous release.

wadells:
## Description
This changeset seeks to catch issues like #1691 by ensuring we always test upgrade from the most current previous release.  This is definitely a halfway measure as it only covers intra-branch upgrade issues, and not inter-branch ones. Still, intra would have caught #1691.

 Also bump latest releases on other branches.

## Type of change

* Contributes to #1199.
* Intended to catch issues like #1691.

## TODOs
- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done
I put an exit in the script before it ran and gathered the following output:

```
walt@work:~/git/gravity$ make robotest-run-suite GET_GRAVITATIONAL_IO_APIKEY=a | tr " " "\n"                                                                                                  
./build.assets/robotest_run_suite.sh
/home/walt/git/gravity/upgrade_from
resize={"to":3,"flavor":"one","nodes":1,"role":"node","state_dir":"/var/lib/telekube","os":"ubuntu:18","storage_driver":"overlay2"}
shrink={"nodes":3,"flavor":"three","role":"node","os":"redhat:7"}
upgrade={"flavor":"three","nodes":3,"role":"node","service_uid":997,"service_gid":994,"os":"ubuntu:18","storage_driver":"overlay2","from":"/telekube_6.1.29.tar"}
upgrade={"flavor":"three","nodes":3,"role":"node","service_uid":997,"service_gid":994,"os":"ubuntu:18","storage_driver":"overlay2","from":"/telekube_7.0.9.tar"}
upgrade={"flavor":"three","nodes":3,"role":"node","service_uid":997,"service_gid":994,"os":"ubuntu:16","storage_driver":"overlay2","from":"/telekube_6.1.0.tar"}
upgrade={"flavor":"three","nodes":3,"role":"node","service_uid":997,"service_gid":994,"os":"ubuntu:16","storage_driver":"overlay2","from":"/telekube_6.2.5.tar"}
upgrade={"flavor":"three","nodes":3,"role":"node","service_uid":997,"service_gid":994,"os":"ubuntu:18","storage_driver":"overlay2","from":"/telekube_6.3.18.tar"}
install={"flavor":"three","nodes":3,"role":"node","os":"redhat:7","storage_driver":"overlay2"}
install={"installer_url":"/installer/opscenter.tar","nodes":1,"flavor":"standalone","role":"node","os":"ubuntu:18","ops_advertise_addr":"example.com:443"}
walt@work:~/git/gravity$ ./build.assets/robotest_run_suite.sh 
./build.assets/robotest_run_suite.sh: line 15: GRAVITY_VERSION: Set GRAVITY_VERSION to the current SemVer
```
The PR build should be everything else this needs.

## Implementation Details

I chose to put this 'calculate previous version' logic in our shell scripts because:
* the other versioning is handed in shell (though we'd like to move away from that -- doing so would be huge scope creep for this pretty small change)
* It keeps the impact localized, this doesn't bleed into the makefile at all (other than depending on the `GRAVITY_VERSION` in addtion to `GRAVITY_BUILDDIR`)

I'm not super happy with how it turned out.  However this will probably get entirely replaced with a proper solution to #1199 hits.  This is just a low hanging fruit to catch a known escape without investing more time in a proper refactoring.

## Additional info
I've got all the ports of this ready to go, but I wanted to incorporate any PR feedback before posting them, so I don't need to make the same change 4 times.

### DIFF
--- a/build.assets/robotest_run_suite.sh
+++ b/build.assets/robotest_run_suite.sh
@@ -3,17 +3,32 @@ set -eu -o pipefail
 
 readonly UPGRADE_FROM_DIR=${1:-$(pwd)/../upgrade_from}
 
+# decrement_patch returns x.y.(z-1) given valid x.y.z semver.
+decrement_patch() {
+    major=$(echo $1 | cut -d'.' -f1)
+    minor=$(echo $1 | cut -d'.' -f2)
+    patch=$(echo $1 | cut -d'.' -f3)
+    patch=$((patch - 1))
+    echo "${major}.${minor}.${patch}"
+}
+
+readonly GRAVITY_VERSION=${GRAVITY_VERSION:?Set GRAVITY_VERSION to the current SemVer}
+readonly THIS_RELEASE=$(echo $GRAVITY_VERSION | cut -f1 -d'+' | cut -f1 -d'-')
+readonly PREVIOUS_RELEASE=$(decrement_patch $THIS_RELEASE)
+unset -f decrement_patch THIS_RELEASE
+
 # UPGRADE_MAP maps gravity version -> list of OS releases to upgrade from
 declare -A UPGRADE_MAP
 
-# latest patch release on this branch, keep this up to date
-UPGRADE_MAP[7.0.5]="ubuntu:18"
+# latest patch release on this branch
+UPGRADE_MAP[$PREVIOUS_RELEASE]="ubuntu:18"
 
 # latest patch release on compatible LTS, keep this up to date
-UPGRADE_MAP[6.1.24]="ubuntu:18"
+UPGRADE_MAP[6.1.29]="ubuntu:18"
 
-# latest patch release on supported non-LTS version, keep this up to date
-UPGRADE_MAP[6.3.13]="ubuntu:18"
+# latest patch release on compatible non-LTS versions
+UPGRADE_MAP[6.3.18]="ubuntu:18"
+UPGRADE_MAP[6.2.5]="ubuntu:16"
 
 # important versions in the field, these are static
 UPGRADE_MAP[6.1.0]="ubuntu:16"
--- a/build.assets/robotest_run_nightly.sh
+++ b/build.assets/robotest_run_nightly.sh
@@ -5,21 +5,35 @@ readonly UPGRADE_FROM_DIR=${1:-$(pwd)/../upgrade_from}
 
 DOCKER_STORAGE_DRIVERS="overlay2"
 
+# decrement_patch returns x.y.(z-1) given valid x.y.z semver.
+decrement_patch() {
+    major=$(echo $1 | cut -d'.' -f1)
+    minor=$(echo $1 | cut -d'.' -f2)
+    patch=$(echo $1 | cut -d'.' -f3)
+    patch=$((patch - 1))
+    echo "${major}.${minor}.${patch}"
+}
+
+readonly GRAVITY_VERSION=${GRAVITY_VERSION:?Set GRAVITY_VERSION to the current SemVer}
+readonly THIS_RELEASE=$(echo $GRAVITY_VERSION | cut -f1 -d'+' | cut -f1 -d'-')
+readonly PREVIOUS_RELEASE=$(decrement_patch $THIS_RELEASE)
+unset -f decrement_patch THIS_RELEASE
+
 # UPGRADE_MAP maps gravity version -> list of OS releases to upgrade from
 declare -A UPGRADE_MAP
 
-# latest patch release on this branch, keep this up to date
-UPGRADE_MAP[7.0.3]="ubuntu:18"
+# latest patch release on this branch
+UPGRADE_MAP[$PREVIOUS_RELEASE]="centos:7 debian:9 ubuntu:18"
 
 # latest patch release on compatible LTS, keep this up to date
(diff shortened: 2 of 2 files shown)


build.assets/robotest_run_suite.sh
```diff
@@ -3,17 +3,32 @@ set -eu -o pipefail
 
 readonly UPGRADE_FROM_DIR=${1:-$(pwd)/../upgrade_from}
 
+# decrement_patch returns x.y.(z-1) given valid x.y.z semver.
+decrement_patch() {
```
a-palchikov, line 7: nit: I'm pretty sure you can source this from a separate file to avoid duplication.

a-palchikov (approved):

build.assets/robotest_run_suite.sh
```diff
@@ -3,17 +3,32 @@ set -eu -o pipefail
...
+    major=$(echo $1 | cut -d'.' -f1)
+    minor=$(echo $1 | cut -d'.' -f2)
+    patch=$(echo $1 | cut -d'.' -f3)
+    patch=$((patch - 1))
```
r0mant, line 11: This doesn't take pre-releases into account and also won't work for .0 patch releases (e.g. when we release 7.1.0-alpha.1, 7.1.0, etc.)

wadells: This is a tricky one.

When I drafted this work, my plan was this bash implementation would never hit master because finding the previous testable release is challenging on master is tricky without making one or more network calls to query release metadata.

See https://github.com/gravitational/gravity/pull/1760 for my comparable master PR.

*If* this never hits master, we don't need to worry about pre-release tags or .0 tags, and can wait for a more full fledged native golang solution to #1199. However until that proper solution is available, we will have to worry about porting the change to each release branch after .0 is cut... this porting is human dependent and likely to lag behind the actual X.X.0 release -- a real risk.

Digging into this further, consider the current master state:

```bash
walt@work:~/git/gravity$ git describe --abbrev=0 master
7.1.0-alpha.1-112-g0bdbdecd
```
7.1.0-alpha.1 is tagged:

https://github.com/gravitational/gravity/releases/tag/7.1.0-alpha.1

But not released in a way that it is consumable by robotest:

```bash
walt@work:~/git/gravity$ ./e/build/current/tele pull telekube:7.1.0-alpha.1                                                                                          
Fri Jun 26 20:59:06 UTC Using Hub https://get.gravitational.io:443
Fri Jun 26 20:59:06 UTC Requesting cluster image from https://get.gravitational.io:443
Fri Jun 26 20:59:06 UTC Download finished in now
[ERROR]: package(gravitational.io/telekube:7.1.0-alpha.1) not found

```

Beyond tags without test artifacts, lets consider testing releases that aren't a traditional patch+1 bump: published X.X.X-(alpha|beta|rc), your .0 scenario, or 6.1.24 to the non-existant https://github.com/gravitational/gravity/releases/tag/6.1.23.

Looking at the previous patch release will always fail in this case, and we need to query the prior semver release.  That can't be done without knowing our release history - unlike regular x.y.z releases, the prior pre-release is impossible to predict. It could be `rc.X`, `alpha.X` or `beta.X` for any value of X.

This can be found by looking at the git tag history:

```
walt@work:~/git/gravity$ git describe master
7.1.0-alpha.1-112-g0bdbdecd
walt@work:~/git/gravity$ git describe --abbrev=0 master
7.1.0-alpha.1
```
We check to see if we're on a tag:

https://github.com/gravitational/gravity/blob/0bdbdecd9afec9938f291bf3deab1733f851ba56/version.sh#L29

And if so, looking at the git tag history instead of a naive semver patch decrement:

```bash
walt@work:~/git/gravity$ git describe --abbrev=0 7.1.0-alpha.1^
7.0.0-rc.5
```

^ This is what we'd be testing upgrade from currently on master.

This actually feels pretty reasonable and robust -- save the tagging inconsistencies in the repo.  I need to audity parity between tags & test artifacts to be confident.  I know when I checked that changelogs were consistent & sequential in https://github.com/gravitational/gravity/pull/1751 there were consistency issues:

```
missing changelogs for 6.1.23
missing changelogs for 5.5.39
missing changelogs for 5.5.30
missing changelogs for 5.5.29
missing changelogs for 5.5.25
missing changelogs for 5.5.16
missing changelogs for 5.4.8
missing changelogs for 5.0.34
missing changelogs for 5.0.31
missing changelogs for 5.0.19
missing changelogs for 5.0.5
// snip 4.x and earlier, 
````

/ramble

I'll rework this to the git history based implementation. We'll need to fix the test artifact situation for 7.1.0-alpha.1 and revisit tagging procedures in our next team meeting.

Enforcement/automation of release consistency will take work... but with the above proposal, this script will be sure to fail our PR build every time we make a release/tagging parity mistake until then.

wadells:
After playing around with the git tag based implementation discussed at https://github.com/gravitational/gravity/pull/1761#discussion_r446410967, I've decided to discard this PR entirely and go with a different approach.

build.assets/robotest_run_suite.sh
wadells: I won't be returning to this PR, but I will include this guidance in the next edition, as appropriate.

PR #1762, opened 2020-06-24
[5.5.x] Explicitly pass since value to gravity system report

bernardjkim:
## Description

Pass --since value to `gravity system report` even if the value is 0. The reason for this is so that k8s logs can be collected with `kubectl cluster-info dump` when `--since 0s` is specified.

## Type of change

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

**Verfiy `gravity report --since 0s` collects `kubectl cluster-info dump`**
```
report
├── cluster.json
├── node-1-debug-logs.tar.gz 
├── node-1-etcd.tar.gz
├── node-1-k8s-logs
│   ├── k8s-cluster-info-dump.tgz  <---
│   ├── k8s-daemonsets
│   ├── k8s-deployments
│   ├── k8s-describe-daemonsets
│   ├── k8s-describe-deployments
│   ├── k8s-describe-endpoints
│   ├── k8s-describe-events
│   ├── k8s-describe-jobs
│   ├── k8s-describe-nodes
│   ├── k8s-describe-pods
│   ├── k8s-describe-replicasets
│   ├── k8s-describe-services
│   ├── k8s-endpoints
│   ├── k8s-events
│   ├── k8s-jobs
│   ├── k8s-logs-kube-system-site-app-post-install-4fe702-xwz6f-post-install-hook-prev
│   ├── k8s-logs-monitoring-influxdb-6b55955dfc-cx2mk-watcher-prev
│   ├── k8s-nodes
│   ├── k8s-pods
│   ├── k8s-replicasets
│   ├── k8s-replicationcontrollers
│   └── k8s-services
├── node-1-k8s-logs.tar.gz
├── node-1-status.tar.gz
└── operation_install.4932c557-68c4-4b90-979f-564aa3b62da4
```

### DIFF
--- a/lib/ops/opsservice/report.go
+++ b/lib/ops/opsservice/report.go
@@ -210,13 +210,11 @@ func (s *site) collectDebugInfo(reportWriter report.FileWriter, runner *serverRu
 	}
 	defer w.Close()
 
-	args := []string{"system", "report", "--filter", report.FilterSystem, "--compressed"}
-	if since != 0 {
-		args = append(args, "--since", since.String())
-	}
-
 	var stderr bytes.Buffer
-	err = runner.RunStream(w, &stderr, s.gravityCommand(args...)...)
+	err = runner.RunStream(w, &stderr, s.gravityCommand("system", "report",
+		"--filter", report.FilterSystem,
+		"--compressed",
+		"--since", since.String())...)
 	if err != nil {
 		return trace.Wrap(err, "failed to collect diagnostics: %s", stderr.String())
 	}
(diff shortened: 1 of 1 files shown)


a-palchikov:
Not sure I understand what this is fixing.
My understanding of the problem is as follows:
 1. `gravity report` changed to generate `cluster-info dump` report when no time window is specified
 1. time window is always specified (by virtue of using kingpin's default API)
 1. `cluster-info dump` can only be generated with an explicit `--since=0` which is counter-intuitive

How I expect `gravity report` to work:

 * `gravity report` without `--since` generates `cluster-info dump` (as does `gravity system report`)
 * `gravity report --since=xxx` does not generate `cluster-info dump` (as does not `gravity system report`)

Let me know if my understanding is flawed.
To achieve this, the code might need to abandon using the kingpin's default to be able to differentiate between set/unset and propagate the value accordingly.

r0mant:
Your `how I expect gravity report to work` makes sense. I think that time filter should apply to k8s logs as well. I don't think cluster-info dump really collects anything that we don't already collect otherwise (via kctl describe, get yaml, etc), it just dumps it in a more convenient format, so we don't really lose anything when it's skipped.

r0mant (commented):
I'm curious if we need to care about including `cluster-info dump` too much at all. As mentioned above, it doesn't really collect anything we can't (and don't) already collect ourselves - it's only a bit more convenient as it's a single API call - and it turned out to be inflexible. To me, having a sane time filter default has more "UX weight" than collecting cluster-info dump. I would consider just getting rid of it completely to simplify things.

a-palchikov:
I would like to keep it. Let's leave the explicit `--since=0s` as a requirement to capture it.
The benefits are is that it's a standard tool that captures all of the information when we might be selective about it (especially for reasons of not having thought about capturing specific resources).

a-palchikov (approved):

r0mant (approved):

PR #1797, opened 2020-06-25
(5.5) Bump election timeout

r0mant:
## Description

Just had a failure on join due to 

```
failed to enable election for 10.128.0.20: [ERROR]: client: etcd cluster is unavailable or misconfigured; error #0: client: endpoint https://127.0.0.1:2379 exceeded header timeout
```

Looking at the logs, etcd was having some temporary issues on the new node, and the current leader election timeout is pretty short - 1m - esp. considering it uses exponential backoff, so there may be not a lot of attempts.

This PR increases the timeout to 5m and also makes sure that max retry interval is 10s so the phase gets plenty of attempts to account for temp. etcd issues.

## Type of change

## TODOs

- [x] Self-review the change
- [x] Address review feedback

## Testing done

None, really. Just a simple constant bump.

### DIFF
--- a/lib/defaults/defaults.go
+++ b/lib/defaults/defaults.go
@@ -1006,7 +1006,9 @@ const (
 
 	// ElectionWaitTimeout specifies the maximum amount of time to wait to resume elections
 	// on a master node
-	ElectionWaitTimeout = 1 * time.Minute
+	ElectionWaitTimeout = 5 * time.Minute
+	// ElectionRetryInterval is the interval between resuming election attempts
+	ElectionRetryInterval = 10 * time.Second
 
 	// ImageRegistryVar is a local cluster registry variable that gets
 	// substituted in Helm templates.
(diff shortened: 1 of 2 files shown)


a-palchikov (approved):

lib/defaults/defaults.go
```diff
@@ -1006,7 +1006,9 @@ const (
 
 	// ElectionWaitTimeout specifies the maximum amount of time to wait to resume elections
 	// on a master node
-	ElectionWaitTimeout = 1 * time.Minute
+	ElectionWaitTimeout = 5 * time.Minute
+	// ElectionRetryInterval is the interval between resuming election attempts
+	ElectionRetryInterval = 10 * time.Second
```
knisbet, line 1011: ElectionRetryMaxInterval because we're technically using exponential backoffs?

knisbet (approved):

PR #1854, opened 2020-07-09
Revise license test plan.

wadells:
## Description
No major changes, just a couple ease of life tweaks to avoid an
interactive wizard during cert generation and suppress/preempt errors.

## Type of change

* Docs

## Linked tickets and other PRs
Developed & tested while executing #1828

## TODOs
- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done
Not much to be done.  Here is proof the openssl command works without prompting:

```
$ openssl req -newkey rsa:2048 -nodes -keyout domain.key -x509 -days 365 -out domain.crt -subj "/C=US/ST=The Internet/L=127.0.0.1/O=Gravitational/OU=Hastily Generated Values Division/CN=Test"   
Generating a 2048 bit RSA private key                                                          
................................................+++++
........................................................................+++++
writing new private key to 'domain.key'                                                        
----- 
```

Everything else is self explanatory and has output captured in #1852 or #1853.

PR #1851, opened 2020-07-09
Revise garbage collection test plan.

wadells:
## Description
I found the former plan challenging to interpret as a beginner. I've
attempted to add detail & run-able shell commands such that the next
employee getting started with this test will boostrap a bit more quickly.

I expect some of these shell commands aren't quite right -- or the most 
efficient way to gather the necessary info. Per Cunningham's Law
I want to provide something concrete to critique. :smile: 

## Type of change
* Documentation

## Linked tickets and other PRs
Developed while executing #1823.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done
I developed/followed these instructions while working on #1823, leading me to catch #1850.  Some of the shell command output can be seen there.  An entire transcript would be rather lengthy, but here are some excerpts:

<details><summary><code>du</code> changing over time</summary>

Before upgrade:

(lost before my 10k line scrollback buffer)

After upgrade:
```
$ sudo du -sh /var/lib/gravity/{local/packages,site/packages,planet/registry}                                                                                     
6.7G    /var/lib/gravity/local/packages                                                                                                                                                       
6.3G    /var/lib/gravity/site/packages                                                                                                                                                        
2.1G    /var/lib/gravity/planet/registry  
```

After GC:
```
$ sudo du -sh /var/lib/gravity/{local/packages,site/packages,planet/registry}                                                                                     
2.7G    /var/lib/gravity/local/packages                                                                                                                                                       
6.3G    /var/lib/gravity/site/packages                                                                                                                                                        
1.8G    /var/lib/gravity/planet/registry      
```
</details>

<details><summary><code>sudo gravity system gc package --dry-run</code></summary>

```
$ sudo gravity system gc package --dry-run                                                                                                                        
Thu Jul  9 00:09:27 UTC [dry-run] Mark package gravitational.io/gravity:7.0.11 as required.                                                                                                   
Thu Jul  9 00:09:27 UTC [dry-run] Mark package gravitational.io/web-assets:7.0.11 as required.                                                                                                
Thu Jul  9 00:09:27 UTC [dry-run] Mark package gravitational.io/teleport:3.2.14 as required.                                                                                                  
Thu Jul  9 00:09:27 UTC [dry-run] Mark package gravitational.io/planet:7.0.35-11706 as required.                                                                                              
Thu Jul  9 00:09:27 UTC [dry-run] Mark package gravitational.io/fio:3.15.0 as required.                                                                                                       
[...]

PR #1857, opened 2020-07-10
Add 6.1.30 and 7.0.12 changelogs

r0mant:
## Description

Chagelog.

### DIFF
--- a/docs/7.x/changelog.md
+++ b/docs/7.x/changelog.md
@@ -4,10 +4,10 @@ Find the latest Open Source Gravity releases at [Gravity Downloads](https://grav
 
 ## Supported Versions
 
-| Version             | Latest Patch | LTS | Release Date         | End of Support          | Kubernetes Version   | Teleport Version |
+| Version             | Latest Patch | LTS | Release Date         | End of Support *        | Kubernetes Version   | Teleport Version |
 | ------------------- | ------------ | --- | -------------------- | ----------------------- | -------------------- | ---------------- |
-| [7.0](#70-releases) | 7.0.11       | No  | April 3, 2020        | 7.1 is released         | 1.17.6               | 3.2.13           |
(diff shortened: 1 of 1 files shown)


a-palchikov (approved):

docs/7.x/changelog.md
```diff
@@ -20,6 +20,8 @@ more current versions of Kubernetes. Regular branches of Gravity are supported
 with security and bug fixes until first release of the subsequent Gravity
 branch is published.
 
+_* Unless extended through customer agreement._
```
wadells, line 23: Why this addendum instead of adjusting the earlier line to 3 years?

> Gravity LTS versions are supported with security and bug fixes for two years.

Is there a difference between LTS support for one customer and across-the-board LTS support?  Either way it seems like we need to do the same proactive security backports.

It seems like the edge case would be in year 3 where an OSS customer (or paying customer with without an agreement) hits a an LTS bug that the paying sponsor doesn't care about, and we don't backport it?

wadells (approved):
When you merge this, would you please also merge the docs PRs https://github.com/gravitational/gravity/pull/1854 and https://github.com/gravitational/gravity/pull/1851 at the same time?

They're both cleared, and I don't want to juggle rebasing & PR builds to merge them as a non-admin.

docs/7.x/changelog.md
r0mant: > Is there a difference between LTS support for one customer and across-the-board LTS support?

Yes, it might make other customers upgrade faster :) Increasing this to 3 years across the board will significantly increase our support burden.

r0mant:
@wadells Sure, I'll merge those once CI passes (not technically required for them but don't want to interrupt something in the middle and leave resources hanging).

wadells:
>  don't want to interrupt something in the middle and leave resources hanging

From the Jenkins side, it is safe to merge PRs even when they have active jobs running.  The branch source plugin we currently use lets the run complete and then cleans up the job afterwards.

PR #1847, opened 2020-07-08
(7.0) Refactor etcd disk check to be more tolerant

r0mant:
## Description

The current fio-based etcd disk performance check is too strict and often fails in the environment where etcd performance may not be of a big concern, without any ability to skip/override it. It results in poor experience for users, to the point where this check actually does more harm than good.

This PR splits the limits the checker verifies into soft and hard thresholds, where a soft threshold only produces a warning and hard threshold leads to a critical failure. So the behavior is as follows:

* If any of the soft limits are hit, the installer prints a warning in the output and proceeds as normal.
* If any of the hard limits are hit, it fails pre-checks and installation like now.
* As an added bonus, all warnings/failures are printed in the installer process right away now.

The old hard limits (50ms latency and 50 IOPS) are now soft limits and produce warnings. The hard limits are 150ms latency and 10 IOPS. The hard limits were chosen by gathering information from users who experienced these issues.

In addition, I added ability to override any of the soft/hard limits by setting respective environment variables. These are manual knobs that are convenient to use for debugging/troubleshooting.

## Type of change

* Internal change (not necessarily a bug fix or a new feature)

## Linked tickets and other PRs

* Closes https://github.com/gravitational/gravity/issues/1834.

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [ ] Write documentation
- [x] Address review feedback

## Implementation

I had to change the signature of opsservice's `ValidateServers` method to return failed probes as well so I could process these results in the client (installer) and print warnings/failures properly. Before, it only returns a pre-formatted error. It should not cause any incompatibility issues b/c it is only used during the installation (not upgrade).

### Override soft limits to trigger the warnings

```
ubuntu@node-1:~/installer$ export GRAVITY_ETCD_MAX_LATENCY_SOFT=1
ubuntu@node-1:~/installer$ export GRAVITY_ETCD_MIN_IOPS_SOFT=5000
ubuntu@node-1:~/installer$ sudo -E ./gravity install --advertise-addr=192.168.99.102 --cluster=test
Wed Jul  8 20:53:22 UTC	Starting enterprise installer
...
Wed Jul  8 20:53:38 UTC	Executing "/checks" locally
Wed Jul  8 20:53:38 UTC	Running pre-flight checks
Wed Jul  8 20:53:39 UTC	Execute preflight checks
Wed Jul  8 20:53:48 UTC		Still running pre-flight checks (10 seconds elapsed)
[...]

PR #1862, opened 2020-07-11
(7.0) Pull only required packages during join

r0mant:
## Description

This pull request updates join's `/pull` phase to only pull necessary packages from the cluster (gravity, teleport, planet) instead of the entire application. This should help some customers who are experiencing issues when pulling large packages from the cluster via internal load balancer on AWS (something that retries won't fix), and will also make join a little faster and more efficient as we don't really need those packages to be laying around on all nodes.

This is a much smaller and less risky approach than the original one I took with excluding disabled packages from built cluster images (the more I worked on it the deeper it got - something I'd like to avoid making in LTS release).

Also, I have just realized that when working on etcd disk check warning in https://github.com/gravitational/gravity/pull/1847, I only updated the install operation but didn't touch join. So I have also updated join's `/checks` phase with the same behavior: to print warning probes and fail on critical probes.

## Type of change

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

Join master and regular node and verify only required packages were pulled:

```
ubuntu@node-2:~$ sudo gravity package list

[gravitational.io]
------------------

* gravitational.io/gravity:7.0.12 111MB installed:installed
* gravitational.io/planet:7.0.35-11706 515MB purpose:runtime,installed:installed
* gravitational.io/teleport:3.2.14 33MB installed:installed
```

Also, run the same tests as in https://github.com/gravitational/gravity/pull/1847 to make sure the disk check warnings are working properly.

### DIFF
--- a/lib/expand/phases/checks.go
+++ b/lib/expand/phases/checks.go
@@ -24,9 +24,9 @@ import (
 	"github.com/gravitational/gravity/lib/fsm"
 	"github.com/gravitational/gravity/lib/ops"
 	"github.com/gravitational/gravity/lib/rpc"
-	"github.com/gravitational/satellite/agent/proto/agentpb"
 
 	"github.com/fatih/color"
+	"github.com/gravitational/satellite/agent/proto/agentpb"
 	"github.com/gravitational/trace"
 	"github.com/sirupsen/logrus"
 )
--- a/lib/expand/phases/checks.go
+++ b/lib/expand/phases/checks.go
@@ -24,7 +24,9 @@ import (
 	"github.com/gravitational/gravity/lib/fsm"
 	"github.com/gravitational/gravity/lib/ops"
 	"github.com/gravitational/gravity/lib/rpc"
+	"github.com/gravitational/satellite/agent/proto/agentpb"
 
+	"github.com/fatih/color"
 	"github.com/gravitational/trace"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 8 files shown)


lib/install/phases/pull.go
```diff
@@ -107,11 +107,21 @@ type pullExecutor struct {
 
 // Execute executes the pull phase
 func (p *pullExecutor) Execute(ctx context.Context) error {
-	err := p.pullUserApplication()
-	if err != nil {
-		return trace.Wrap(err)
+	// If the list of packages to pull was explicitly provided, pull only those
+	// (e.g. during join), otherwise pull the entire user application (e.g.
+	// during initial installation).
```
a-palchikov, line 112: When is the application package required during the installation? Can you elaborate here?
Also the `pullUserApplication` hints at the possibility of skipping the app pull on regular nodes - is it something this PR can also do?
And if the pulling ends up requiring the application package only on a single node - how about splitting the pull step into pull app and pull packages which will make the decision more emphasized as now it is pretty much buried inside and not explicit about when and whether the application package is pulled.

lib/expand/phases/checks.go
```diff
@@ -24,7 +24,9 @@ import (
 	"github.com/gravitational/gravity/lib/fsm"
 	"github.com/gravitational/gravity/lib/ops"
 	"github.com/gravitational/gravity/lib/rpc"
+	"github.com/gravitational/satellite/agent/proto/agentpb"
```
bernardjkim, line 27: nit: move to external packages

a-palchikov (approved):

PR #1865, opened 2020-07-13
Revert "[5.5.x] Update DNS application hook"

UnderGreen:
Reverts gravitational/gravity#1429

Reverting it because upgrade from 5.2.x version should delete kube-dns and all that resources.


r0mant (approved):

a-palchikov (approved):
Would still be helpful to know why it is reverted?

PR #1531, opened 2020-05-08
Hook 'system cluster-info' command

a-palchikov:
## Description
This PR adds the hook up for the `system cluster-info` command added in https://github.com/gravitational/gravity/pull/1361. Without this, the debug reports will contain a figment of kubernetes resources (i.e. will not have container logs).
For some reason, the forward port did not have the command hook up code.

## Type of change

* New feature (non-breaking change which adds functionality)
* This change has a user-facing impact

## Linked tickets and other PRs

* Ports https://github.com/gravitational/gravity/pull/1081

## TODOs

- [x] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback

### Debug logs (focused on kubernetes) captured before the change:
```
$ gravity system cluster-info --help
[ERROR]: expected command but got "cluster-info"
$ gravity system report --filter=kubernetes >report.tgz
$ tar tvf report.tgz
-rw-r--r-- root/root      2375 2020-05-08 20:49 k8s-daemonsets
-rw-r--r-- root/root      6142 2020-05-08 20:49 k8s-deployments
-rw-r--r-- root/root     18996 2020-05-08 20:49 k8s-describe-daemonsets
-rw-r--r-- root/root     40553 2020-05-08 20:49 k8s-describe-deployments
-rw-r--r-- root/root     17206 2020-05-08 20:49 k8s-describe-endpoints
-rw-r--r-- root/root     41833 2020-05-08 20:49 k8s-describe-jobs
-rw-r--r-- root/root     31004 2020-05-08 20:49 k8s-describe-nodes
-rw-r--r-- root/root    204307 2020-05-08 20:49 k8s-describe-pods
-rw-r--r-- root/root     37824 2020-05-08 20:49 k8s-describe-replicasets
-rw-r--r-- root/root     11683 2020-05-08 20:49 k8s-describe-services
-rw-r--r-- root/root      3796 2020-05-08 20:49 k8s-endpoints
-rw-r--r-- root/root      2501 2020-05-08 20:49 k8s-jobs
-rw-r--r-- root/root      5454 2020-05-08 20:49 k8s-logs-kube-system-coredns-98n5q-coredns-prev
... <more previous container logs skipped>
-rw-r--r-- root/root       614 2020-05-08 20:49 k8s-nodes
-rw-r--r-- root/root      6393 2020-05-08 20:49 k8s-pods
-rw-r--r-- root/root      6704 2020-05-08 20:49 k8s-replicasets
[...]

PR #1361, opened 2020-04-13
[7.0.x] expand improvements

a-palchikov:
Forward port changes from https://github.com/gravitational/gravity/pull/1348

PR #1882, opened 2020-07-15
Cumulative 7.0 port

a-palchikov:
## Description

Cumulative port of PRs from the 7.0.x branch.

## Type of change

* Requires https://github.com/gravitational/gravity.e/pull/4317

* Ports #1361, #1424, #1530, #1531, #1542, #1603, #1648

## TODOs

- [x] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback


assets/site-app/resources/images/site/start.sh
```diff
@@ -1,12 +1,15 @@
 #!/bin/sh
```
r0mant, line 1: Hmm, I'm just noticing this but why do we have the same `start.sh` script twice - in site-app/images/site (above) and here? Do we need both?

lib/checks/checks.go
```diff
@@ -772,28 +772,58 @@ func checkRAM(info ServerInfo, ram schema.RAM) error {
...
-		osToNodes[os] = append(osToNodes[os], fmt.Sprintf("%v (%v)",
-			server.ServerInfo.GetHostname(), server.AdvertiseAddr))
+		info := server.GetOS()
+		distros[info.ID] = append(distros[info.ID], info.Version)
 	}
+	if len(distros) != 1 {
+		return trace.BadParameter("servers have different OS distributions: %v", formatKeysAsList(distros))
+	}
+	// Version verification is purposely simple and will compare the prefixes
+	// up to to either the first '.' or end of line
```
r0mant, line 789: "to to"

r0mant (approved):

lib/blob/cluster/cluster_test.go
```diff
@@ -164,12 +164,11 @@ func (s *ClusterMultiPeers) SetUpTest(c *C) {
...
 			ID:               fmt.Sprintf("%v", i),
 			AdvertiseAddr:    "https://localhost",
-			TestMode:         true,
 			GracePeriod:      gracePeriod,
 		})
 		c.Assert(err, IsNil)
-		objects[i] = obj.(*cluster)
-		c.Assert(objects[i].heartbeat(), IsNil)
+		objects[i] = obj
+		objects[i].heartbeat()
```
bernardjkim, line 171: Is there a reason for not asserting heartbeat returns nil anymore?

lib/ops/opsservice/report.go
```diff
@@ -145,17 +145,18 @@ func (s *site) getReport(runner remoteRunner, servers []remoteServer, masters []
...
 		serverRunner := &serverRunner{server: server, runner: runner}
 		reportWriter := getReportWriterForServer(dir, server)
+		logger := log.WithField("server", server.Address())
 		if err := s.collectKubernetesInfo(reportWriter, serverRunner, since); err != nil {
-			log.WithError(err).Error("Failed to collect Kubernetes info.")
+			logger.WithError(err).Error("Failed to collect Kubernetes info.")
 		}
 		if err := s.collectEtcdInfoFromMasters(dir, masters, runner); err != nil {
-			log.WithError(err).Error("Failed to collect etcd info.")
+			logger.WithError(err).Error("Failed to collect etcd info.")
```
bernardjkim, line 153: Use `log` instead of `logger` since we are collecting etcd info from all masters instead of the just first selected server.

lib/ops/opsservice/report.go
```diff
@@ -253,17 +256,18 @@ func (s *site) collectEtcdInfoFromMasters(dir string, masters []remoteServer, ru
 	return nil
 }
 
-// collectEtcdInfo collects etcd backup and metrics.
+// collectEtcdInfo collects etcd metrics and captures a snapshot of the data.
 func (s *site) collectEtcdInfo(reportWriter report.FileWriter, runner *serverRunner) error {
-	w, err := reportWriter.NewWriter("etcd.tar.gz")
+	w, err := reportWriter.NewWriter("etcd.json.tar.gz")
```
bernardjkim, line 261: I updated to `etcd.tar.gz` when I added the metrics snapshot. `etcd-backup.json` and `etcd-metrics` are zipped together in `etcd.tar.gz`

a-palchikov: Thanks for pointing out - I feared I'd break a few things.

lib/ops/opsservice/report.go
a-palchikov: Will revert this change.

lib/blob/cluster/cluster_test.go
a-palchikov: That's a good question - not that I see any. I'll revert this line.

assets/site-app/resources/images/site/start.sh
a-palchikov: I think the one in `resources/images/site` not in-use - will remove.

lib/ops/opsservice/report.go
a-palchikov: The thing is is that it remained (or became?) `etcd.json.tar.gz` in `version/7.0.x` [PR](https://github.com/gravitational/gravity/pull/1806/files#diff-9fa9aa39035553b520c7dff230be31a9R262) and I think that's what confused me.

bernardjkim: Oh, that's my bad. Looks like I didn't update it correctly in 7.0. It was intended to be `etcd.tar.gz`

Issue #1799, opened 2020-06-25
Provide ability to view/collect Gravity CLI commands history


aelkugia:
## Problem

Inside Gravity clusters there is no visibility into what kind of actions users/support perform on the nodes using `gravity` CLI tooling. Having this information available would aid in troubleshooting to see the commands that were executed on the nodes, what their results were, etc.

## Solution

Each invocation of a `gravity` CLI command should be logged in the systemd journal, along with a journald tag which the journal can be filtered by.

## Success

- [x] Each `gravity` CLI command and its arguments is recorded in the journal, with the following properties:
  - [x] With `gravity-cli` journald tag to allow filtering (e.g. using `journalctl -t gravity-cli`).
  - [x] The command's start/finish should both be logged.
  - [x] If the command finishes with the error, the error should be logged alongside the failed command.
  - [ ] ~~Each command invocation is also assigned a UUID4 that will let track start/finish of a particular command (e.g. to distinguish one `gravity status` call from another).~~
  - [x] Potentially sensitive information should be excluded from the logs (tokens, passwords that can be passed on CLI, etc.).
- [x] In addition, `gravity` CLI commands history is included in the debug report.

## Scope

This feature should be ported to 5.5, 6.1, 7.0 and master.

PR #1889, opened 2020-07-16
Sanitize gravity cli cmd logs

bernardjkim:
## Description

Gravity cli commands will now be logged twice. Once when the command starts execution and another after it completes. Potentially sensitive data will also be redacted from the logs.

The log entries already include the process ID so a UUID is not included with the log entries. Use `journalctl _PID=[pid]` to track start/finish of a particular comand.
## Type of change

* Updates #https://github.com/gravitational/gravity/issues/1799

* Requires #https://github.com/gravitational/gravity.e/pull/4316

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

**Verify log entry of a successfully completed gravity command**
```
Jul 11 12:00:22 node-1 gravity-cli[13177]: [RUNNING]: gravity status history
Jul 11 12:00:22 node-1 gravity-cli[13177]: [SUCCESS]: gravity status history
```
**Verify log entry of a failed gravity command. Also verify sensitive data is redacted**
```
Jul 11 12:00:36 node-1 gravity-cli[13238]: [RUNNING]: gravity install --token *****
Jul 11 12:00:36 node-1 gravity-cli[13238]: [FAILURE]: gravity install --token *****: [ERROR]: install token is too short, min length is 6
```

## Limitations?

One thing that seems inconsistent was when `gravity status` reported a degraded cluster. The command runs and displays the status, but an error is still returned. Is this okay as is?
```
Jul 11 06:29:31 node-1 gravity-cli[6226]: [RUNNING] gravity status
Jul 11 06:29:31 node-1 gravity-cli[6226]: [FAILURE] gravity status: [ERROR]: degraded
```

There is still one inconsistent case I ran into. `gravity status` has a boolean `--token` flag that is used to request just the cluster token. Output:
```
Jul 11 14:23:45 node-1 gravity-cli[8759]: [RUNNING]: /usr/bin/gravity status cluster --token "*****"
```

### DIFF
--- a/lib/utils/cli/cli.go
+++ b/lib/utils/cli/cli.go
@@ -63,6 +63,13 @@ func (r *CommandArgs) Update(command []string) (args []string, err error) {
 func (r *CommandArgs) addRemoveFlags(ctx *kingpin.ParseContext) (flags []Flag) {
 	var args []Flag
 	seen := make(map[string]struct{})
+
+	// replace is an index to easily look up flags that need to be replaced
+	replace := make(map[string]Flag)
+	for _, flag := range r.FlagsToReplace {
+		replace[flag.Name()] = flag
+	}
+
 	for _, el := range ctx.Elements {
 		switch c := el.Clause.(type) {
 		case *kingpin.ArgClause:
@@ -78,10 +85,12 @@ func (r *CommandArgs) addRemoveFlags(ctx *kingpin.ParseContext) (flags []Flag) {
 				continue
 			}
 			seen[c.Model().Name] = struct{}{}
-			if flagInFlags(r.FlagsToReplace, c.Model().Name) {
-				// Skip the flag after it's been marked as seen
+
+			if flag, exists := replace[c.Model().Name]; exists {
+				flags = append(flags, flag)
 				continue
 			}
+
 			if _, ok := c.Model().Value.(boolCmdlineFlag); ok {
 				flags = append(flags, newBoolFlag(c.Model().Name, *el.Value))
 			} else {
@@ -100,22 +109,8 @@ func (r *CommandArgs) addRemoveFlags(ctx *kingpin.ParseContext) (flags []Flag) {
 			flags = append(flags, flag)
 		}
 	}
-	for _, flag := range r.FlagsToReplace {
-		if _, exists := seen[flag.Name()]; exists {
-			flags = append(flags, flag)
-		}
-	}
 	// Return the new command line with positional arguments following non-positional flags
 	return append(flags, args...)
-}
-
-func flagInFlags(flags []Flag, flag string) bool {
-	for _, f := range flags {
-		if f.Name() == flag {
-			return true
-		}
-	}
-	return false
 }
 
 // NewArg creates a new positional argument
--- a/lib/utils/cli/cli.go
+++ b/lib/utils/cli/cli.go
@@ -32,6 +32,9 @@ type CommandArgs struct {
 	Parser ArgsParser
 	// FlagsToAdd is a list of additional flags to add to resulting command.
 	FlagsToAdd []Flag
+	// FlagsToReplace is a list of flags to replace in the command.
+	// A flag is only replaced if it has been provided.
+	FlagsToReplace []Flag
 	// FlagsToRemove is a list of flags to omit from the resulting command.
 	FlagsToRemove []string
 }
@@ -75,6 +78,10 @@ func (r *CommandArgs) addRemoveFlags(ctx *kingpin.ParseContext) (flags []Flag) {
 				continue
 			}
 			seen[c.Model().Name] = struct{}{}
+			if flagInFlags(r.FlagsToReplace, c.Model().Name) {
+				// Skip the flag after it's been marked as seen
+				continue
+			}
 			if _, ok := c.Model().Value.(boolCmdlineFlag); ok {
 				flags = append(flags, newBoolFlag(c.Model().Name, *el.Value))
 			} else {
@@ -93,8 +100,22 @@ func (r *CommandArgs) addRemoveFlags(ctx *kingpin.ParseContext) (flags []Flag) {
 			flags = append(flags, flag)
 		}
 	}
+	for _, flag := range r.FlagsToReplace {
+		if _, exists := seen[flag.Name()]; exists {
+			flags = append(flags, flag)
+		}
+	}
 	// Return the new command line with positional arguments following non-positional flags
 	return append(flags, args...)
+}
+
+func flagInFlags(flags []Flag, flag string) bool {
+	for _, f := range flags {
+		if f.Name() == flag {
+			return true
+		}
+	}
+	return false
 }
 
 // NewArg creates a new positional argument
--- a/lib/utils/cli/cli_test.go
+++ b/lib/utils/cli/cli_test.go
@@ -85,11 +85,35 @@ func (*S) TestUpdatesCommandLine(c *check.C) {
 			removeFlags: []string{"selinux"},
 		},
 		{
-			comment:    "Redact flags",
+			comment:    "Redact install token",
 			inputArgs:  []string{"install", `--token=token`, "--debug"},
 			outputArgs: []string{"install", "--token", fmt.Sprintf(`"%s"`, constants.Redacted), "--debug"},
 			replaceFlags: []Flag{
 				NewFlag("token", constants.Redacted),
+			},
+		},
+		{
+			comment:   "Redact user create password",
+			inputArgs: []string{"user", "create", `--email=email`, `--password=password`},
+			outputArgs: []string{"user create",
+				"--email", `"email"`,
+				"--password", fmt.Sprintf(`"%s"`, constants.Redacted),
+			},
+			replaceFlags: []Flag{
+				NewFlag("password", constants.Redacted),
+			},
+		},
+		{
+			comment:   "Redact multiple flags",
+			inputArgs: []string{"test", `--secret1`, `secret1`, `--secret2`, `secret2`, `test`},
+			outputArgs: []string{"test",
+				"--secret1", fmt.Sprintf(`"%s"`, constants.Redacted),
+				"--secret2", fmt.Sprintf(`"%s"`, constants.Redacted),
+				`"test"`,
+			},
+			replaceFlags: []Flag{
+				NewFlag("secret1", constants.Redacted),
+				NewFlag("secret2", constants.Redacted),
 			},
 		},
 	}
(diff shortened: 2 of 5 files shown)


tool/gravity/cli/logging.go
```diff
@@ -0,0 +1,109 @@
...
+			cmdString = Redact(cmdString, *g.UserCreateCmd.Password)
+		}
+	}
+
+	return cmdString
+}
+
+// Redact replaces any instances of value found in cmd with the redacted string.
+func Redact(cmd, value string) string {
+	return strings.ReplaceAll(cmd, value, constants.Redacted)
```
a-palchikov, line 108: Although unlikely, such context-free application of text replacement can replace unrelated portions if two flags happen to have the same value.
Also, we've already solved a similar problem when moving the install/expand functionality to a server running as a systemd service where we had to manipulate the existing command line and add/remove/modify parameters on the fly - this is implemented as [CommandArgs](https://pkg.go.dev/github.com/gravitational/gravity@v0.0.0-20200603182942-2a15a2722622/lib/utils/cli?tab=doc#CommandArgs):

For this to work, we'll need to tweak `CommandArgs` to support flag replacement but it is trivial:

Add the corresponding attribute (`FlagsToReplace`):
```go
// CommandArgs manipulates command-line arguments.
type CommandArgs struct {
	// Parser is used to parse provided command-line.
	Parser ArgsParser
	// FlagsToAdd is a list of additional flags to add to resulting command.
	FlagsToAdd []Flag
	// FlagsToReplace is a list of flags to replace in the command.
	// A flag is only replaced if it has been provided
	FlagsToReplace []Flag
	// FlagsToRemove is a list of flags to omit from the resulting command.
	FlagsToRemove []string
}
```
update `addRemoveFlags` to take that into account:
```go
func (r *CommandArgs) addRemoveFlags(ctx *kingpin.ParseContext) (flags []Flag) {
	var args []Flag
	seen := make(map[string]struct{})
	for _, el := range ctx.Elements {
		switch c := el.Clause.(type) {
		case *kingpin.ArgClause:
			if utils.StringInSlice(r.FlagsToRemove, c.Model().Name) {
				// Remove the positional argument
				continue
			}
			seen[c.Model().Name] = struct{}{}
			args = append(args, NewArg(c.Model().Name, *el.Value))
		case *kingpin.FlagClause:
			if utils.StringInSlice(r.FlagsToRemove, c.Model().Name) {
				// Remove the flag
				continue
			}
			seen[c.Model().Name] = struct{}{}
			if flagInFlags(r.FlagsToReplace, c.Model().Name) {
				// Skip the flag after it's been marked as seen
				continue
			}
			if _, ok := c.Model().Value.(boolCmdlineFlag); ok {
				flags = append(flags, newBoolFlag(c.Model().Name, *el.Value))
			} else {
				flags = append(flags, NewFlag(c.Model().Name, *el.Value))
			}
		}
	}
	for _, flag := range r.FlagsToAdd {
		if arg, ok := flag.(arg); ok {
			if _, exists := seen[arg.name]; !exists {
				args = append(args, arg)
			}
			continue
		}
		if _, exists := seen[flag.Name()]; !exists {
			flags = append(flags, flag)
		}
	}
	for _, flag := range r.FlagsToReplace {
		if _, exists := seen[flag.Name()]; exists {
			flags = append(flags, flag)
		}
	}
	// Return the new command line with positional arguments following non-positional flags
	return append(flags, args...)
}

// ...

func flagInFlags(flags []Flag, flag string) bool {
	for _, f := range flags {
		if f.Name() == flag {
			return true
		}
	}
	return false
}
```

and then just use it as following:

```go
// This is just for illustration - the function already exists in tool/gravity/cli
func parseArgs(args []string) (*kingpin.ParseContext, error) {
	app := kingpin.New("gravity", "")
	app.Terminate(func(int) {})
	app.Writer(ioutil.Discard)
	return RegisterCommands(app).ParseContext(args)
}

func redactCmd(args ...string) (cmd []string, err error) {
	commandArgs := cli.CommandArgs{
		Parser: cli.ArgsParserFunc(parseArgs),
		FlagsToReplace: []Flag{
			cli.NewFlag("token", defaults.Redacted),
			cli.NewFlag("registry-password", defaults.Redacted),
			cli.NewFlag("password", defaults.Redacted),
		},
	}
	args, err = commandArgs.Update(args)
	if err != nil {
		return nil, trace.Wrap(err)
	}
	return append([]string{utils.Exe.Path}, args...), nil
}
```

tool/gravity/cli/logging.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/gravitational/gravity/lib/constants"
+	"github.com/gravitational/gravity/lib/utils"
+
+	"github.com/gravitational/trace"
+)
+
+// LogCLIRunning writes the running cmd as a log entry into the system journal
+// with the gravity-cli tag.
+func LogCLIRunning(cmd string) {
```
r0mant, line 32: Do all these need to be exported?

tool/gravity/cli/run.go
```diff
@@ -81,7 +76,15 @@ func Run(g *Application) error {
...
 		return trace.Wrap(err)
 	}
-	return Execute(g, cmd, extraArgs)
+
+	cmdString := fmt.Sprintf("gravity %s", strings.Join(os.Args[1:], " "))
+	cmdString = SanitizeCmd(g, cmd, cmdString)
+
+	LogCLIRunning(cmdString)
+	err = Execute(g, cmd, extraArgs)
+	LogCLICompleted(cmdString, err)
```
r0mant, line 85: I would defer this and extended to catch/log panics as well (they sometimes do happen).

```go

logCLIRunning(cmdString)
defer logCLICompleted(cmdString, err)
err = Execute(g, cmd, extraArgs)

```

tool/gravity/cli/run.go
```diff
@@ -81,7 +76,15 @@ func Run(g *Application) error {
...
+
+	return err
```
r0mant, line 87: trace.Wrap?

tool/gravity/cli/run.go
```diff
@@ -81,7 +76,15 @@ func Run(g *Application) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	return Execute(g, cmd, extraArgs)
+
+	cmdString := fmt.Sprintf("gravity %s", strings.Join(os.Args[1:], " "))
```
r0mant, line 80: Use real path instead of just "gravity".

tool/gravity/cli/logging.go
```diff
@@ -0,0 +1,109 @@
...
+	entry := fmt.Sprintf("[RUNNING]: %s", cmd)
```
r0mant, line 33: Also, what about my suggestion to assign a unique uuid4 tag to each command run and log it as well? So we can correlate start/finish of the commands?

tool/gravity/cli/logging.go
bernardjkim: @a-palchikov Thanks for pointing this out. This will help simplify the implementation quite a bit.

tool/gravity/cli/logging.go
bernardjkim: These functions will be imported in gravity.e.

tool/gravity/cli/logging.go
bernardjkim: The journal entries already include the process ID which can be used to track start/finish. Is there any benefit to adding a UUID4 tag in this case?

We can track the command with the `_PID` flag. Example:
```
[vagrant@node-1 ~]$ sudo journalctl -t gravity-cli _PID=29778
-- Logs begin at Fri 2020-07-10 20:26:59 UTC, end at Sat 2020-07-11 14:36:30 UTC. --
Jul 11 14:05:37 node-1 gravity-cli[29778]: [RUNNING]: /usr/bin/gravity install --debug --token "*****"
Jul 11 14:05:37 node-1 gravity-cli[29778]: [FAILURE]: /usr/bin/gravity install --debug --token "*****": [ERROR]: install token is too short, min length is 6
```

r0mant: Good point. No need for additional ids then.

tool/gravity/cli/logging.go
```diff
@@ -0,0 +1,96 @@
...
+	sanitizedCmd, err := r.redactCmd()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	cmdString := fmt.Sprintf("%s -- %s", strings.Join(sanitizedCmd, " "), strings.Join(r.ExtraArgs, " "))
+
+	logEntry(fmt.Sprintf("[RUNNING]: %s", cmdString))
+	defer func() {
+		if r := recover(); r != nil {
+			logEntry(fmt.Sprintf("[FAILURE]: %s: [PANIC]: %v", cmdString, r))
```
a-palchikov, line 58: I don't think we should be silencing a panic. It might be worthwhile to capture to report the failing command but I'd still want to see the stack so either re-panic or log `string(debug.Stack())` to get stack (since recover runs in the context of the failing function).

lib/utils/cli/cli.go
```diff
@@ -93,10 +100,24 @@ func (r *CommandArgs) addRemoveFlags(ctx *kingpin.ParseContext) (flags []Flag) {
 			flags = append(flags, flag)
 		}
 	}
+	for _, flag := range r.FlagsToReplace {
```
a-palchikov, line 103: Maybe also add a test for the new attribute?

bernardjkim: Added a test case for flag replacement. I noticed that the replacement could alter the order of the flags provided. I've updated the implementation a bit in 4320279.

tool/gravity/cli/logging.go
bernardjkim: Ah, did not intend to silence the panic. Updated to re-panic after logging.

lib/utils/cli/cli_test.go
```diff
@@ -80,14 +84,23 @@ func (*S) TestUpdatesCommandLine(c *check.C) {
 			},
 			removeFlags: []string{"selinux"},
 		},
+		{
+			comment:    "Redact flags",
+			inputArgs:  []string{"install", `--token=token`, "--debug"},
```
a-palchikov, line 89: I'd like to test with a couple of commands since this change has an interesting twist of supporting multiple commands with a single set of replacement flags.

tool/gravity/cli/logging.go
a-palchikov: nit: `[PANIC]: ...` will repeat what the std panic handling also dumps to the console.

a-palchikov (approved):

tool/gravity/cli/logging.go
```diff
@@ -0,0 +1,96 @@
...
+			cli.NewFlag("token", constants.Redacted),
+			cli.NewFlag("registry-password", constants.Redacted),
+			cli.NewFlag("password", constants.Redacted),
+			cli.NewFlag("license", constants.Redacted),
+			cli.NewFlag("ops-token", constants.Redacted),
+			cli.NewFlag("ops-tunnel-token", constants.Redacted),
+			cli.NewFlag("encryption-key", constants.Redacted),
+		},
+	}
+	r.Args, err = commandArgs.Update(r.Args)
```
knisbet, line 83: Just a nitpick: I'm not sure I understand why this is mutating the CmdExecer internals, and although I don't think this causes any problems, I'm not sure this is an expected side effect of calling Execute/redactCommand, is that the Args/Config in the CmdExecer structure will change by side effect of the execution. I could easily be missing something, but my view would be that this function should be something like getRedactedCmd and only return the []string to be used for logging.

knisbet (approved):

tool/gravity/cli/logging.go
bernardjkim: I don't quite understand the comment. Could you clarify?

tool/gravity/cli/logging.go
bernardjkim: Yeah that makes sense. There is no need to mutate the CmdExecer Args field here. Will update.

tool/gravity/cli/logging.go
a-palchikov: Say you have an index out of bounds somewhere. The code as written will output the following:
```sh
[FAILURE]: [/path/tobinary args...] [PANIC]: runtime error: index out of range [0] with length 0.
panic: runtime error: index out of range [0] with length 0 [recovered]
	panic: runtime error: index out of range [0] with length 0
...
```
The same panic message is output 3 times in this case.

bernardjkim: Oh, I see. I think this okay. The first line `[FAILURE] ...] will only be written to the system journal. The panic message will still be output twice, but this seems like the standard way of re-panicking.

Issue #1807, opened 2020-06-26
Gravity plan resume resiliency improvements


r0mant:
## Problem

Currently `gravity plan resume` command (and its various counterparts such as `gravity upgrade --resume`) resumes plan execution in foreground which means it will interrupt upon disconnect.

Users often hit this when resuming an upgrade from the control panel terminal - the connection to the control panel drops during etcd upgrade or gravity-site restart and the phase remains in progress.

## Solution

`gravity plan resume` should have 2 modes of operation:

* Blocking, which works like now by executing all plan phases in foreground.
* Non-blocking which launches a systemd service to execute the phases.

In non-blocking mode, the plan progress can be monitored using a (new) `gravity plan watch` command that will print the phase changes to the terminal.

## Success

- [ ] `gravity plan resume` launches a background job to execute phases by default.
- [ ] `gravity plan resume --block` resumes the plan in foreground, same behavior as now.
- [ ] `gravity plan watch` command is available to track the plan progress.

r0mant:
Implemented in 5.5; [6.1](https://github.com/gravitational/gravity/pull/1935), [7.0](https://github.com/gravitational/gravity/pull/1936) and [master](https://github.com/gravitational/gravity/pull/1934) PRs are out.

Issue #841, opened 2019-10-24
gravity plan --tail


klizhentas:
## Description

Ops teams constantly execute `gravity plan` to verify the current stage of the execution.

`gravity plan --follow` or `gravity plan --tail` should show the current operation progress with auto refresh.

jeff-1amstudios:
The usage case behind this is to have the ability to wait for the upgrade to complete, so our customers can just execute a single command. In this mode, `plan` should exit with a non-zero status code if one of the steps fails

r0mant:
Closing as a part of https://github.com/gravitational/gravity/issues/1807.

PR #1899, opened 2020-07-22
(5.5) Background plan resume / plan tail

r0mant:
## Description

This pull requests implements a couple of resiliency improvements for the `plan resume` command. The main issue was that users would often launch it and lose SSH connection (esp. relevant if launching from our web UI's terminal) which would basically terminate the resume since it executes in foreground.

* Update `gravity plan resume` to launch resume from a one-shot systemd service by default.
  * It also accepts `--block` flag for the old behavior to run in foreground.
* Add `gravity plan --tail` to be able to monitor plan progress.
  * This is useful not just for resume but in general for watching the plan.
  * Tail will exit with 0 if plan completes successfully, and non-0 if a phase completes with an error.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1807, refs https://github.com/gravitational/gravity/issues/841.

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [x] Address review feedback

### Resume in background (now default)

```
ubuntu@node-1:~/upgrade$ sudo ./upload
Wed Jul 22 01:10:36 UTC	Importing cluster image telekube v5.5.50-dev.8
Wed Jul 22 01:11:15 UTC	Synchronizing application with Docker registry 192.168.99.102:5000
Wed Jul 22 01:11:35 UTC	Verifying cluster health
Wed Jul 22 01:11:35 UTC	Cluster image has been uploaded
Wed Jul 22 01:12:04 UTC	Upgrading cluster from 5.5.40 to 5.5.50-dev.8
Wed Jul 22 01:12:04 UTC	Deploying agents on cluster nodes
Wed Jul 22 01:12:08 UTC	Deployed agent on node-1 (192.168.99.102)
The operation has been created in manual mode.

See https://gravitational.com/gravity/docs/cluster/#managing-an-ongoing-operation for details on working with operation plan.
ubuntu@node-1:~/upgrade$ sudo ./gravity plan resume --help
usage: gravity plan resume [<flags>]

Resume last aborted operation

Flags:
      --log-file="/var/log/gravity-install.log"
      --operation-id=OPERATION-ID
                             ID of the active operation. If not specified, the last operation will be used
      --block                Launch plan resume in foreground instead of a systemd unit

ubuntu@node-1:~/upgrade$ sudo ./gravity plan resume
Wed Jul 22 01:12:23 UTC	Starting gravity-resume.service service
Wed Jul 22 01:12:23 UTC	Service gravity-resume.service has been launched.

To monitor the operation progress:

  sudo gravity plan --operation-id=d6d43cae-edcd-49e7-b056-cb1a9f50ffed --tail

To monitor the service logs:

  sudo journalctl -u gravity-resume.service -f

ubuntu@node-1:~/upgrade$ sudo ./gravity plan --operation-id=d6d43cae-edcd-49e7-b056-cb1a9f50ffed --tail
Wed Jul 22 01:12:25 UTC	[  1/ 26] Phase /init/node-1 is completed
Wed Jul 22 01:12:26 UTC	[  2/ 26] Phase /checks is completed
Wed Jul 22 01:12:27 UTC	[  3/ 26] Phase /pre-update is in_progress
Wed Jul 22 01:12:33 UTC	[  3/ 26] Phase /pre-update is completed
Wed Jul 22 01:12:33 UTC	[  4/ 26] Phase /bootstrap/node-1 is in_progress
...
Wed Jul 22 01:18:07 UTC	[ 26/ 26] Phase /gc/node-1 is in_progress
Wed Jul 22 01:18:08 UTC	[ 26/ 26] Phase /gc/node-1 is completed
Wed Jul 22 01:18:09 UTC	Operation plan is completed
```

### Resume blocking (old behavior)

```
ubuntu@node-1:~/upgrade$ sudo ./upload
Wed Jul 22 01:31:33 UTC	Importing cluster image telekube v5.5.50-dev.8
Wed Jul 22 01:32:18 UTC	Synchronizing application with Docker registry 192.168.99.102:5000
Wed Jul 22 01:32:36 UTC	Verifying cluster health
Wed Jul 22 01:32:36 UTC	Cluster image has been uploaded
Wed Jul 22 01:32:41 UTC	Upgrading cluster from 5.5.40 to 5.5.50-dev.8
Wed Jul 22 01:32:41 UTC	Deploying agents on cluster nodes
Wed Jul 22 01:32:45 UTC	Deployed agent on node-1 (192.168.99.102)
The operation has been created in manual mode.

See https://gravitational.com/gravity/docs/cluster/#managing-an-ongoing-operation for details on working with operation plan.
ubuntu@node-1:~/upgrade$ sudo ./gravity plan resume --block
Wed Jul 22 01:32:52 UTC	Executing "/init/node-1" locally
Wed Jul 22 01:32:53 UTC	initializing the operation
Wed Jul 22 01:32:53 UTC	Executing "/checks" locally
Wed Jul 22 01:32:55 UTC	Executing "/pre-update" locally
Wed Jul 22 01:33:01 UTC	Executing "/bootstrap/node-1" locally
	Still executing "/bootstrap/node-1" locally (10 seconds elapsed)
...
Wed Jul 22 01:37:57 UTC	Executing "/migration/labels" locally
Wed Jul 22 01:37:58 UTC	Executing "/app/telekube" locally
Wed Jul 22 01:37:59 UTC	Executing "/gc/node-1" locally
Wed Jul 22 01:38:00 UTC	operation(update(b95dcc7c-29c6-48f8-8f4e-a47a14abd018), cluster=test, created=2020-07-22 01:32) finished in 5 minutes
```

### DIFF
--- a/lib/fsm/follow.go
+++ b/lib/fsm/follow.go
@@ -31,32 +31,37 @@ type GetPlanFunc func() (*storage.OperationPlan, error)
 
 // PlanEvent represents an operation plan event.
 type PlanEvent interface {
-	// IsPlanEvent returns true for events representing plan changes.
-	IsPlanEvent() bool
+	// isTerminalEvent returns true for terminal plan events, such as if
+	// a plan was completed or fully rolled back.
+	isTerminalEvent() bool
 }
 
-// PlanChangeEvent is sent when plan phases change states.
-type PlanChangeEvent struct {
+// PlanChangedEvent is sent when plan phases change states.
+type PlanChangedEvent struct {
 	// Change is a plan phase change.
 	Change storage.PlanChange
-	// Plan is the resolved operation plan.
-	Plan storage.OperationPlan
 }
 
-// IsPlanEvent is for satisfying PlanEvent interface.
-func (e *PlanChangeEvent) IsPlanEvent() bool { return true }
+// isTerminalEvent returns false for the plan change event.
+func (e *PlanChangedEvent) isTerminalEvent() bool { return false }
 
-// PlanFinishEvent is sent when the plan is fully completed or rolled back.
-type PlanFinishEvent struct {
-	// Plan is the resolved operation plan.
-	Plan storage.OperationPlan
-}
+// PlanCompletedEvent is sent when the plan is fully completed.
+type PlanCompletedEvent struct{}
 
-// IsPlanEvent is for satisfying PlanEvent interface.
-func (e *PlanFinishEvent) IsPlanEvent() bool { return true }
+// isTerminalEvent returns true for the completed plan event.
+func (e *PlanCompletedEvent) isTerminalEvent() bool { return true }
 
-// FollowOperationPlan returns a channel that receives phase updates for the
-// specified plan.
+// PlanRolledBackEvent is sent when the plan is fully rolled back.
+type PlanRolledBackEvent struct{}
+
+// isTerminalEvent returns true for the rolled back plan event.
+func (e *PlanRolledBackEvent) isTerminalEvent() bool { return true }
+
+// FollowOperationPlan returns a channel that will be receiving phase updates
+// for the specified plan.
+//
+// The watch will stop upon entering one of the terminal operation states, for
+// example if the obtained plan is completed or fully rolled back.
 func FollowOperationPlan(ctx context.Context, getPlan GetPlanFunc) <-chan PlanEvent {
 	ch := make(chan PlanEvent, 100)
 	// Send an initial batch of events from the initial state of the plan.
@@ -65,22 +70,19 @@ func FollowOperationPlan(ctx context.Context, getPlan GetPlanFunc) <-chan PlanEv
 		logrus.WithError(err).Error("Failed to load plan.")
 	}
 	if plan != nil {
-		for _, change := range GetPlanProgress(*plan) {
-			ch <- &PlanChangeEvent{Change: change, Plan: *plan}
-		}
-		if IsCompleted(plan) || IsRolledBack(plan) {
-			ch <- &PlanFinishEvent{Plan: *plan}
-			return ch
+		for _, event := range getPlanEvents(GetPlanProgress(*plan), *plan) {
+			ch <- event
+			if event.isTerminalEvent() {
+				return ch
+			}
 		}
 	}
 	// Then launch a goroutine that will be monitoring the progress.
 	go func() {
 		tickerBackoff := getFollowBackoffPolicy()
 		ticker := backoff.NewTicker(tickerBackoff)
-		defer func() {
-			ticker.Stop()
-			logrus.Info("Operation plan watcher done.")
-		}()
+		defer ticker.Stop()
+		defer logrus.Info("Operation plan watcher done.")
 		for {
 			select {
 			case <-ticker.C:
@@ -94,12 +96,11 @@ func FollowOperationPlan(ctx context.Context, getPlan GetPlanFunc) <-chan PlanEv
 					logrus.WithError(err).Error("Failed to diff plans.")
 					continue
 				}
-				for _, change := range changes {
-					ch <- &PlanChangeEvent{Change: change, Plan: *nextPlan}
-				}
-				if IsCompleted(nextPlan) || IsRolledBack(nextPlan) {
-					ch <- &PlanFinishEvent{Plan: *nextPlan}
-					return
+				for _, event := range getPlanEvents(changes, *nextPlan) {
+					ch <- event
+					if event.isTerminalEvent() {
+						return
+					}
 				}
 				// Update the current plan for comparison on the next cycle and
 				// reset the backoff so the ticker keeps ticking every second
@@ -112,6 +113,20 @@ func FollowOperationPlan(ctx context.Context, getPlan GetPlanFunc) <-chan PlanEv
 		}
 	}()
 	return ch
+}
+
+// getPlanEvents returns a list of plan events from the provided list of
+// changes and the current state of the plan.
+func getPlanEvents(changes []storage.PlanChange, plan storage.OperationPlan) (events []PlanEvent) {
+	for _, change := range changes {
+		events = append(events, &PlanChangedEvent{Change: change})
+	}
+	if IsCompleted(&plan) {
+		events = append(events, &PlanCompletedEvent{})
+	} else if IsRolledBack(&plan) {
+		events = append(events, &PlanRolledBackEvent{})
+	}
+	return events
 }
 
 // getFollowBackoffPolicy returns retry/backoff policy for the plan follower.
--- a/lib/fsm/follow.go
+++ b/lib/fsm/follow.go
@@ -1,0 +1,95 @@
+/*
+Copyright 2020 Gravitational, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fsm
+
+import (
+	"context"
+	"time"
+
+	"github.com/gravitational/gravity/lib/storage"
+
+	"github.com/gravitational/trace"
+	"github.com/sirupsen/logrus"
+)
+
+// GetPlanFunc is a function that returns an operation plan.
+type GetPlanFunc func() (*storage.OperationPlan, error)
+
+// PlanEvent represents an operation plan event.
+type PlanEvent interface{}
+
+// PlanChanged is sent when plan phases change states.
+type PlanChanged struct {
+	// Change is a plan phase change.
+	Change storage.PlanChange
+	// Plan is the resolved operation plan.
+	Plan storage.OperationPlan
+}
+
+// PlanFinished is sent when the plan is fully completed or rolled back.
+type PlanFinished struct {
+	// Plan is the resolved operation plan.
+	Plan storage.OperationPlan
+}
+
+// FollowOperationPlan returns a channel that receives phase updates for the
+// specified plan.
+func FollowOperationPlan(ctx context.Context, getPlan GetPlanFunc) (<-chan PlanEvent, error) {
+	previousPlan, err := getPlan()
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	ch := make(chan PlanEvent, 100)
+	go func() {
+		// Send initial events to the channel, excluding unstarted phases.
+		for _, change := range GetPlanProgress(*previousPlan) {
+			ch <- &PlanChanged{Plan: *previousPlan, Change: change}
+		}
+		if IsCompleted(previousPlan) || IsRolledBack(previousPlan) {
+			ch <- &PlanFinished{Plan: *previousPlan}
+			return
+		}
+		ticker := time.NewTicker(time.Second)
+		defer ticker.Stop()
+		for {
+			select {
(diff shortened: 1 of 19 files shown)


tool/gravity/cli/clusterupdate.go
```diff
@@ -183,8 +185,86 @@ func executeUpdatePhaseForOperation(env *localenv.LocalEnvironment, environ Loca
...
+	// Before launching the service, perform a few prechecks, for example to
+	// make sure that the operation is being resumed from the correct node.
+	//
+	// TODO(r0mant): Also, make sure agents are running on the cluster nodes:
+	// https://github.com/gravitational/gravity/issues/1667
+	if err := updater.Check(params.toFSM()); err != nil {
+		return trace.Wrap(err)
+	}
+	// Make sure to launch the unit command with the --block flag.
+	args := append(os.Args[1:], "--debug", "--block")
```
a-palchikov, line 216: I think this is susceptible to the same issue as fixed [here](https://github.com/gravitational/gravity/pull/1255) - something to watch for and optionally rewrite with changes from https://github.com/gravitational/gravity/pull/1895.

tool/gravity/cli/clusterupdate.go
```diff
@@ -183,8 +185,86 @@ func executeUpdatePhaseForOperation(env *localenv.LocalEnvironment, environ Loca
...
+To monitor the service logs:
+
+  sudo journalctl -u %[1]v -f
+`, gravityResumeServiceName, operation.ID)
+	return nil
+}
+
+// launchOneshotService launches the specified command as a one-shot systemd
+// service with the specified name.
+func launchOneshotService(name string, args []string) error {
```
a-palchikov, line 236: I wonder how that differs from https://github.com/gravitational/gravity/blob/version/5.5.x/tool/gravity/cli/system.go#L313

tool/gravity/cli/plan.go
```diff
@@ -87,101 +91,170 @@ func displayOperationPlan(localEnv *localenv.LocalEnvironment, environ LocalEnvi
...
-	return trace.Wrap(outputPlan(*plan, format))
 }
 
 // displayExpandOperationPlan shows plan of the join operation from the local join backend
-func displayExpandOperationPlan(environ LocalEnvironmentFactory, opKey ops.SiteOperationKey, format constants.Format) error {
-	joinEnv, err := environ.NewJoinEnv()
-	if err != nil {
-		return trace.Wrap(err)
+func displayExpandOperationPlan(localEnv *localenv.LocalEnvironment, environ LocalEnvironmentFactory, opKey ops.SiteOperationKey, opts displayPlanOptions) error {
+	return outputOrFollowPlan(localEnv, getExpandOperationPlanFunc(environ, opKey), opts)
```
a-palchikov, line 194: I guess you could move getXXXPlanFunc inside of the respective displayXXXPlan variant since they're no used elsewhere.

lib/fsm/fsm.go
```diff
@@ -620,6 +606,8 @@ type StateChange struct {
 	State string
 	// Error is the error that happened during phase execution
 	Error trace.Error
+	// created overrides the time for the state change, only used in tests.
+	created time.Time
```
a-palchikov, line 610: So just by looking at the way this is used, I don't think you need this in the main code if that's only for tests. There's nothing special about [testEngine](https://github.com/gravitational/gravity/blob/roman/5.5/resume/lib/fsm/testhelpers.go#L33). More over, I think it's confusing that the test engine is implementing an additional functionality (as opposed to existing main code counterparts) just to relay a test attribute.
I think for your purpose, you could implement an additional test-only API:
```go
// ChangePhaseState records the provided phase state change in the test engine.
func (t *testEngine) ChangePhaseState(ctx context.Context, ch StateChange) error {
	t.changelog = append(t.changelog, storage.PlanChange{
		PhaseID:  ch.Phase,
		NewState: ch.State,
		Created:  t.clock.Now().Add(time.Duration(len(t.changelog)) * time.Minute),
	})
	return nil
}
...
func (t *testEngine) changePhaseStateWithTimestamp(ctx context.Context, ch StateChange, created time.Time) error {
	t.changelog = append(t.changelog, storage.PlanChange{
		PhaseID:  ch.Phase,
		NewState: ch.State,
		Created:  created,
	})
	return nil
}
```
and use `changePhaseStateWithTimestamp` where necessary.

lib/fsm/follow.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/gravitational/trace"
+	"github.com/sirupsen/logrus"
+)
+
+// GetPlanFunc is a function that returns an operation plan.
+type GetPlanFunc func() (*storage.OperationPlan, error)
+
+// PlanEvent represents an operation plan event.
+type PlanEvent interface{}
```
a-palchikov, line 33: I'd also suggest to add an opaque unexported API to the interface so it could not be satisfied by an unrelated type:
```go
type PlanEvent interface {
	isPlanEvent()
}
```

lib/fsm/follow.go
```diff
@@ -0,0 +1,95 @@
...
+			select {
+			case <-ticker.C:
+				newPlan, err := getPlan()
+				if err != nil {
+					logrus.WithError(err).Error("Failed to reload plan.")
+					continue
+				}
+				diff, err := DiffPlan(*previousPlan, *newPlan)
+				if err != nil {
+					logrus.WithError(err).Error("Failed to diff plans.")
```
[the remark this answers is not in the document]
r0mant: This is a little tricky, the diff should always succeed, yes, as long as `getPlan()` always returns the same plan (which it should). Otherwise this would likely mean a programming error on our part, so not much a user can do.

lib/fsm/utils.go
```diff
@@ -139,6 +140,59 @@ func ResolvePlan(plan storage.OperationPlan, changelog storage.PlanChangelog) *s
...
+// DiffPlan returns the difference between the next and previous plans in the
+// form of a changelog.
+func DiffPlan(prevPlan, nextPlan storage.OperationPlan) (diff []storage.PlanChange, err error) {
+	// Quick sanity check that this is the same plan.
+	if prevPlan.OperationID != nextPlan.OperationID {
+		return nil, trace.BadParameter("can't diff different plans: %v %v", prevPlan, nextPlan)
+	}
+	// Since this is the same plan, should be safe to assume they have the
+	// same phases with different states.
+	prevPhases := prevPlan.GetLeafPhases()
```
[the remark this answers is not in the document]
r0mant: I'm not sure about this. At best this logic will just move elsewhere and will have to be handled there. But this method should still check for plans similarity, otherwise something might panic (e.g. if it blindly assumes the plans are the same) / not work as expected.

lib/fsm/utils.go
```diff
@@ -254,3 +308,21 @@ func addPhases(phase *storage.OperationPhase, result *[]*storage.OperationPhase)
...
+// coordinator node specified in the plan.
+//
+// This is mainly important for making sure the plan is executed on the lead
+// node for a particular plan - for example, for etcd upgrades, where state
+// can only be kept in sync on the lead master node itself.
+func CheckPlanCoordinator(p *storage.OperationPlan) error {
+	if p.OfflineCoordinator == nil {
+		return nil
+	}
+	err := systeminfo.HasInterface(p.OfflineCoordinator.AdvertiseIP)
```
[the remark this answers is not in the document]
r0mant: Yeah, I just moved this piece to a different location, so this is out of scope.

tool/gravity/cli/clusterupdate.go
r0mant: I don't think it's susceptible to the same issue. In the cases you're referring to we had to either replace the flags (remove/add), or update their values. In this case we just need to append a `--block` flag. Worst case scenario, it'll end up with two `--debug` flags which is fine. So unless I'm missing anything, I'd like to keep this simple.

tool/gravity/cli/clusterupdate.go
r0mant: Yeah, we have a couple of similar methods which I originally tried to see if they will fit this use-case, but they're different in a couple of ways. This method, for example, makes sure there's no another service running to prevent resume from being launched multiple times simultaneously.

tool/gravity/cli/plan.go
r0mant: That's how I did it originally, but it was a bit messy. I also think having those method in global scope might be a bit more convenient if something/someone ever needs to use them.

lib/fsm/utils.go
```diff
@@ -139,6 +140,64 @@ func ResolvePlan(plan storage.OperationPlan, changelog storage.PlanChangelog) *s
 	return &plan
 }
 
+// DiffPlan returns the difference between the previous and the next plans in the
+// form of a changelog.
+func DiffPlan(prevPlan *storage.OperationPlan, nextPlan storage.OperationPlan) (diff []storage.PlanChange, err error) {
```
bernardjkim, line 145: nit: pass both plans as pointers. Or is there a reason for passing nextPlan as a struct?

r0mant: Yeah, it's actually on purpose. To indicate that the `prevPlan` is optional (nil can be passed), while the `nextPlan` should always be passed in.

lib/fsm/follow.go
```diff
@@ -22,74 +22,107 @@ import (
...
-			ch <- &PlanChanged{Plan: *previousPlan, Change: change}
+	if plan != nil {
+		for _, change := range GetPlanProgress(*plan) {
+			ch <- &PlanChangeEvent{Change: change, Plan: *plan}
 		}
-		if IsCompleted(previousPlan) || IsRolledBack(previousPlan) {
-			ch <- &PlanFinished{Plan: *previousPlan}
-			return
+		if IsCompleted(plan) || IsRolledBack(plan) {
+			ch <- &PlanFinishEvent{Plan: *plan}
```
a-palchikov, line 72: Also this hints that two final event types would be better: `CompletedPlan` and `RolledbackPlan` - then you would not need to relay the plan itself.

lib/fsm/follow.go
```diff
@@ -22,74 +22,107 @@ import (
...
-		return nil, trace.Wrap(err)
+		logrus.WithError(err).Error("Failed to load plan.")
 	}
-	ch := make(chan PlanEvent, 100)
-	go func() {
-		// Send initial events to the channel, excluding unstarted phases.
-		for _, change := range GetPlanProgress(*previousPlan) {
-			ch <- &PlanChanged{Plan: *previousPlan, Change: change}
+	if plan != nil {
+		for _, change := range GetPlanProgress(*plan) {
```
a-palchikov, line 68: this could potentially be encapsulated in a function and reused.
Also - why not pass the changes as a slice into the channel instead of one after another - it would make sense if changes were available separately but they are available all at once.

lib/fsm/follow.go
```diff
@@ -22,74 +22,107 @@ import (
...
+}
 
-// PlanChanged is sent when plan phases change states.
-type PlanChanged struct {
+// PlanChangeEvent is sent when plan phases change states.
+type PlanChangeEvent struct {
 	// Change is a plan phase change.
 	Change storage.PlanChange
 	// Plan is the resolved operation plan.
 	Plan storage.OperationPlan
```
a-palchikov, line 43: You don't seem to have a use for the plan itself other than the number of phases (aka its `Length`) - do you need to pack the whole plan with every change? Even the length is a static artifact - i.e. it does not change as you follow the plan.

tool/gravity/cli/plan.go
```diff
@@ -202,6 +202,10 @@ func outputOrFollowPlan(localEnv *localenv.LocalEnvironment, getPlan fsm.GetPlan
 }
 
 func followPlan(localEnv *localenv.LocalEnvironment, getPlan fsm.GetPlanFunc) error {
+	plan, err := getPlan()
```
a-palchikov, line 205: nit: this will now make the follow command half-resilient - before it could tolerate initial transient failure and now will fail immediately forcing the user to retry. Not sure which is better though.

a-palchikov (approved):

Issue #1667, opened 2020-06-08
Upgrade agents resiliency improvements


r0mant:
There are a few issues with the upgrade agents.

## 1. Agents do not restart after crash/reboot

Right now our upgrade (or any other operation) agents are installed as one-shot systemd units and will not restart in case of failure, node reboot, etc. which leads to a bad upgrade experience because users in general don't know how to redeploy the agents.

We should make them into regular services so they're more resilient and automatically restart upon failures, reboots and do not require manual intervention during upgrades.

Success:

- [x] Upgrade (and others) agents automatically restart after reboots/crashes and upgrade operation can continue without needing to redeploy the agents manually.

A couple of things to watch out for:

* What happens if the "lead" upgrade agent (the one that runs automatic upgrade) restarts automatically? Is it going to attempt to resume the automatic upgrade operation?
* Agents should be made to be shut down and disabled properly after each operation to make sure they're not restarting automatically when they're not needed.

## 2. Make sure `gravity upgrade --resume` (or `plan resume`) check that all agents are running

When resuming an operation (e.g. after a node reboot), the resume command should check if all required agents are running and alert the user to redeploy any of them if some aren't.

Success:

- [x] The resume command bails out with a clear error message if any of the upgrade agents are not running.

## 3. It is possible to deploy agents of incorrect version

IIRC the current behavior is that deployed upgrade agents are of the same version as the "gravity" binary that performs the deployment. So if a user executes "gravity agent deploy" with gravity 5.0.0, the same gravity 5.0.0 will be used to run agents. This may be causing incompatibility issues because during upgrade the agents should be running the new gravity version, but users sometimes manage to deploy the old ones.

Success:

- [ ] Make sure upgrade operation always deploys agents of correct version (the version we're upgrading to). We may need to make sure here that "gravity upgrade" can only be launched using the "new" binary (we should have this check in place, but double-check).
- [ ] Make sure "gravity agent deploy" launched separately deploys agents of correct version if there's an ongoing upgrade operation, regardless of which "gravity" binary was used.
- [ ] Make sure versions of deployed agents are printed in the console, just for better visibility.

## 4. Agents deployment CLI improvements

These are more of a nice-to-haves, but it would be helpful to update our "gravity agent" CLI with following:

- [x] Have an ability to deploy an agent <strike>on a particular node and</strike> of a particular version. For example, `gravity agent deploy node-2 --version=7.0.1`.
- [x] Have a command to check the deployed agents and their statuses. For example:

```
[...]

PR #1906, opened 2020-07-22
Upgrade agent changes

bernardjkim:
## Description

This PR makes a few improvements to the gravity upgrade agents. 
- The gravity-agents will now run as regular services instead of oneshot services so that they restart on failure and node reboot. 
- A `gravity agent status` command is now available to view the status of the deployed gravity agents. Will look something like:
```
// Successfully received status response from all agents
Tue Jul 14 07:11:19 UTC Collecting RPC agent status
Hostname     Address            Status       Version
--------     -------            ------       -------
node-2       172.28.128.102     Deployed     7.1.0-alpha.1.141
node-3       172.28.128.103     Deployed     7.1.0-alpha.1.141
node-1       172.28.128.101     Deployed     7.1.0-alpha.1.141

// Failed to receive status response from all agents
Tue Jul 14 07:12:30 UTC Collecting RPC agent status
Hostname     Address            Status      Version
--------     -------            ------      -------
node-2       172.28.128.102     Offline
node-3       172.28.128.103     Offline
node-1       172.28.128.101     Offline
```

- The `gravity agent deploy` command now supports a `--version` flag to specify a particular agent version to deploy.
- The `gravity agent deploy` command also accepts a node hostname as an argument to deploy an agent on the specified node.
## Type of change

* Refs https://github.com/gravitational/gravity/issues/1667
* Requires https://github.com/gravitational/gravity.e/pull/4321

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

**Verify gravity-agent is running after reboot**

**Verify lead update agent does not automatically attempt to resume upgrade**

**Verify upgrade can be resumed after node reboot with `gravity upgrade --resume`**

**Verify all gravity-agents are shutdown after upgrade**

**Verify gravity-agents are disabled and do not restart on next reboot**

**Verify upgrade can be resumed after non lead agent restarts**

**Verify upgrade after rollback**

**Deploy particular agent version with --version flag**
```
[vagrant@node-1 installer]$ sudo gravity agent deploy --version=7.1.0-alpha.1.153
Wed Jul 29 19:13:08 UTC Deploying agents on the cluster nodes
Deployed agent on node-3 (172.28.128.103)
Deployed agent on node-1 (172.28.128.101)
Deployed agent on node-2 (172.28.128.102)

[vagrant@node-1 installer]$ sudo gravity agent status
Wed Jul 29 19:13:31 UTC Collecting RPC agent status
Hostname     Address            Status       Version
--------     -------            ------       -------
node-3       172.28.128.103     Deployed     7.1.0-alpha.1.153
node-1       172.28.128.101     Deployed     7.1.0-alpha.1.153
node-2       172.28.128.102     Deployed     7.1.0-alpha.1.153
```

**Deploy onto a particular node**
```
[vagrant@node-1 installer]$ sudo gravity agent deploy node-1
Wed Jul 29 19:00:36 UTC Deploying agents on the cluster nodes
Deployed agent on node-1 (172.28.128.101)

[vagrant@node-1 installer]$ sudo gravity agent status
Wed Jul 29 19:00:43 UTC Collecting RPC agent status
Hostname     Address            Status       Version
--------     -------            ------       -------
node-1       172.28.128.101     Deployed     7.1.0-alpha.1.154
node-3       172.28.128.103     Offline
node-2       172.28.128.102     Offline
```

**Verify agents show deployed even if older server version**
```
[vagrant@node-1 ~]$ sudo gravity agent status
Wed Jul 29 21:22:44 UTC Collecting RPC agent status
Hostname     Address            Status       Version
--------     -------            ------       -------
node-2       172.28.128.102     Deployed     N/A
node-3       172.28.128.103     Deployed     N/A
node-1       172.28.128.101     Deployed     N/A
```

**Verify deploy without args works as usual**
```
[vagrant@node-1 ~]$ sudo gravity agent deploy
Thu Jul 30 21:21:04 UTC Deploying agents on the cluster nodes
Deployed agent on node-2 (172.28.128.102)
Deployed agent on node-3 (172.28.128.103)
Deployed agent on node-1 (172.28.128.101)

[vagrant@node-1 ~]$ sudo gravity agent status
Thu Jul 30 21:21:12 UTC Collecting RPC agent status
Hostname     Address            Status       Version
--------     -------            ------       -------
node-2       172.28.128.102     Deployed     7.1.0-alpha.1.162
node-3       172.28.128.103     Deployed     7.1.0-alpha.1.162
node-1       172.28.128.101     Deployed     7.1.0-alpha.1.162
```


tool/gravity/cli/rpcagent.go
```diff
@@ -366,6 +367,30 @@ func rpcAgentShutdown(env *localenv.LocalEnvironment) error {
 	return trace.Wrap(err)
 }
 
+// rpcAgentStatus requests and writes the gravity agent statuses to stdout.
+func rpcAgentStatus(env *localenv.LocalEnvironment) error {
```
r0mant, line 371: How does this method behave / what does it output if there are no deployed agents?

lib/update/cluster/automatic.go
```diff
@@ -132,3 +132,14 @@ func ShutdownClusterAgents(ctx context.Context, remote rpc.AgentRepository) erro
 	err = rpc.ShutdownAgents(ctx, nodes, log.StandardLogger(), remote)
 	return trace.Wrap(err)
 }
+
+// AgentStatus requests the statuses of all agents
+func AgentStatus(ctx context.Context, remote rpc.AgentRepository) ([]rpc.GravityAgentStatus, error) {
+	nodes, err := kubectl.GetNodesAddr(ctx)
```
r0mant, line 138: I think we should retrieve the nodes from the cluster state, not k8s. If you pass an operator here, you can use it:

```go
// in cli/rpcagent.go
func rpcAgentStatus(env *localenv.LocalEnvironment) error {
  operator, err := env.ClusterOperator()
  ...
  cluster.AgentStatus(ctx, operator, remote)
}

// here
func AgentStatus(ctx context.Context, operator ops.Operator, remote ...) () {
  cluster, err := operator.GetLocalSite()
  ...
  cluster.ClusterState.Servers // <-- cluster nodes
}
```

lib/update/cluster/automatic.go
```diff
@@ -132,3 +132,14 @@ func ShutdownClusterAgents(ctx context.Context, remote rpc.AgentRepository) erro
 	err = rpc.ShutdownAgents(ctx, nodes, log.StandardLogger(), remote)
 	return trace.Wrap(err)
 }
+
+// AgentStatus requests the statuses of all agents
+func AgentStatus(ctx context.Context, remote rpc.AgentRepository) ([]rpc.GravityAgentStatus, error) {
```
r0mant, line 137: I think this better belongs in `lib/rpc`?

tool/gravity/cli/rpcagent.go
r0mant: Also, what does it output if there's a failure contacting one of the agents / retrieving info from it?

lib/rpc/status.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/gravitational/gravity/lib/modules"
+	"github.com/gravitational/gravity/lib/storage"
+	"github.com/gravitational/gravity/lib/utils"
+
+	"github.com/gravitational/trace"
+	log "github.com/sirupsen/logrus"
+)
+
+// GravityAgentStatus contains a gravity agent's status information.
+type GravityAgentStatus struct {
```
r0mant, line 33: Nit: Just `AgentStatus`, or `RPCAgentStatus`.

lib/rpc/status.go
```diff
@@ -0,0 +1,118 @@
...
+	Node string
+	// IP specifies the IP address of the node running the agent.
+	IP string
+	// Status indiciates the current status of the agent. An agent is `Deployed`
+	// if the gravity-agent service is active. The agent is `Offline` if it
+	// fails to respond to the status request.
+	Status string
+	// Version describes gravity agent version.
+	Version string
+}
```
r0mant, line 44: I would add another field here with `Error error` just so it can be displayed in case there's a failure contacting the agent.

lib/rpc/status.go
```diff
@@ -0,0 +1,118 @@
...
+
+func AgentStatus(ctx context.Context, servers []string, logger log.FieldLogger, rpc AgentRepository) (
+	[]GravityAgentStatus, error) {
+	errs := make(chan error, len(servers))
+	statusCh := make(chan GravityAgentStatus, len(servers))
+	for _, srv := range servers {
+		go func(host string) {
+			systemInfo, err := getSystemInfo(ctx, host, rpc)
```
r0mant, line 52: I would encapsulate these two (getSystemInfo and getVersion) into a separate function (getAgentStatus) just so the error handling is a little easier, for example:

```go
func getAgentStatus(ctx context.Context, address string, rpc AgentRepository) AgentStatus {
  systemInfo, err := getSystemInfo(ctx, host, rpc)
  // return status with Error set
  version, err := getVersion(ctx, host, rpc)
  // return status with Error set
  return AgentStatus{
    Address:  address,
    Status:   constants.GravityAgentDeployed,
    Hostname: systemInfo.GetHostname(),
    Version:  version.Version,    
  }, nil
}

// Then the goroutine becomes trivial:

go func(address string) {
  statusCh <- getAgentStatus(ctx, address, rpc)
}
```

lib/rpc/status.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/gravitational/trace"
+	log "github.com/sirupsen/logrus"
+)
+
+// GravityAgentStatus contains a gravity agent's status information.
+type GravityAgentStatus struct {
+	// Node specifies the name of the node running the agent.
+	Node string
+	// IP specifies the IP address of the node running the agent.
+	IP string
```
r0mant, line 37: Nit: I'd rename `IP -> Address` and `Node -> Hostname` to be consistent with naming in other places.

r0mant (commented):
Preliminary feedback for your draft.

lib/rpc/status.go
bernardjkim: Realized we don't even need to request systemInfo since we are only collecting hostname from it. We can pull hostname from the cluster state.

tool/gravity/cli/rpcagent.go
bernardjkim: Added additional details and added a 5 sec timeout on the status request.
If an agent fail to return a status response, the agent will be displayed as `Offline` and the version column will be empty.

lib/constants/constants.go
```diff
@@ -230,6 +230,12 @@ const (
 	// OneshotService is a service that executes one time
 	OneshotService = "oneshot"
 
+	// SimpleService is a simple service that is recommended for long running processes
```
a-palchikov, line 233: I think a link to the systemd docs as a reference to service types would be useful.
Also, I think these could be more appropriate in [lib/system/service](https://github.com/gravitational/gravity/tree/bernard/master/upgrade-agent/lib/system/service) but it's debatable.

lib/rpc/proto/discovery.proto
```diff
@@ -16,6 +16,15 @@ service Discovery {
...
     rpc GetCurrentTime(google.protobuf.Empty) returns (google.protobuf.Timestamp);
+
+    // GetVersion returns the agent's version information.
+    rpc GetVersion(google.protobuf.Empty) returns (Version);
+}
+
+// Version describes an agent's version information
+message Version {
+    // Payload is the version information payload.
+    bytes payload = 1;
```
a-palchikov, line 27: I would rather have the protobuf to be the source of truth for data types.

lib/rpc/status.go
```diff
@@ -0,0 +1,96 @@
...
+	Hostname string
+	// Address specifies the IP address of the node running the agent.
+	Address string
+	// Status indiciates the current status of the agent. An agent is `Deployed`
+	// if the gravity-agent service is active. The agent is `Offline` if it
+	// fails to respond to the status request.
+	Status string
+	// Version describes gravity agent version.
+	Version string
+	// Error contains errors while collected while requesting agent status.
```
a-palchikov, line 42: nit: maybe rephrase this to avoid duplicate `while`: `Error contains an error that might have occurred when requesting agent status`

lib/rpc/status.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/gravitational/trace"
+)
+
+// AgentStatus contains a gravity agent's status information.
+type AgentStatus struct {
+	// Hostname specifies the hostname of the node running the agent.
+	Hostname string
+	// Address specifies the IP address of the node running the agent.
+	Address string
+	// Status indiciates the current status of the agent. An agent is `Deployed`
```
a-palchikov, line 36: indicates

lib/rpc/status.go
```diff
@@ -0,0 +1,96 @@
...
+		return agentStatus
+	}
+
+	agentStatus.Version = version.Version
+	agentStatus.Status = constants.GravityAgentDeployed
+	return agentStatus
+}
+
+func getVersion(ctx context.Context, addr string, rpc AgentRepository) (version modules.Version, err error) {
+	ctx, cancel := context.WithTimeout(ctx, defaults.DialTimeout)
```
a-palchikov, line 82: This sounds like a bug - `DialTimeout` should only affect the connection step and not the actual API call.

tool/gravity/cli/system.go
```diff
@@ -280,6 +282,33 @@ func reinstallOneshotService(env *localenv.LocalEnvironment, serviceName string,
 	return trace.Wrap(err)
 }
 
+// reinstallService stops and reinstalls the service specified by serviceName
+// as a simple service.
+func reinstallService(env *localenv.LocalEnvironment, serviceName string, cmd []string) error {
```
a-palchikov, line 287: I'd really like that we moved general purpose functions to specific packages - i.e. `lib/system/service` if there's no such API yet.

lib/defaults/defaults.go
```diff
@@ -219,6 +219,9 @@ const (
 	// AgentStopTimeout is amount of time agent gets to gracefully shut down
 	AgentStopTimeout = 10 * time.Second
 
+	// AgentStatusTimeout specifies the timeout for agent status query
+	AgentStatusTimeout = 5 * time.Second
```
a-palchikov, line 223: I'm sort of on the fence with timeouts - I feel these should be configurable via envars at least to avoid any issues with slower networks and arbitrary cluster sizes.

bernardjkim: Gravity will now check for `GRAVITY_AGENT_STATUS_TIMEOUT` env var. Will default to 5sec if env var is not specified.

lib/rpc/proto/discovery.proto
bernardjkim: Moved Version definition from modules to proto pkg. a583350

lib/update/cluster/automatic.go
```diff
@@ -43,6 +43,21 @@ func AutomaticUpgrade(ctx context.Context, localEnv, updateEnv *localenv.LocalEn
...
+
+	plan, err := fsm.GetOperationPlan(updateEnv.Backend, siteOp.Key())
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	// Check if the upgrade is already in progress, which would indicate an
+	// agent is already running. Return without creating a new agent.
+	for _, phase := range plan.Phases {
+		if phase.IsInProgress() {
```
r0mant, line 56: I think we should be checking for "not unstarted" phases. Otherwise, this will not catch scenarios where a phase is "failed" or "completed". So we should basically look for phases in any state other than the initial "unstarted" one.

lib/update/cluster/automatic.go
```diff
@@ -43,6 +43,21 @@ func AutomaticUpgrade(ctx context.Context, localEnv, updateEnv *localenv.LocalEn
...
+			return nil
```
r0mant, line 57: We should definitely log something in this case.

lib/utils/env.go
```diff
@@ -131,6 +132,19 @@ func GetenvInt(name string) (int, error) {
 	return valueI, nil
 }
 
+//GetenvDuration returns the specified environment variable value parsed as a duration.
```
r0mant, line 135: Nit: Space after //.

tool/gravity/cli/rpcagent.go
```diff
@@ -315,16 +331,44 @@ func deployAgents(ctx context.Context, env *localenv.LocalEnvironment, req deplo
...
+		verifiedServers, err := verifyCluster(ctx, req)
+		if err != nil {
+			return nil, trace.Wrap(err)
+		}
+		servers = append(servers, verifiedServers...)
+	}
+
+	// If version is not specified in the request, use the current build version
+	if req.version == "" {
+		req.version = strings.Split(version.Get().Version, "+")[0]
```
r0mant, line 358: What's "split by +" for?

tool/gravity/cli/rpcagent.go
```diff
@@ -388,6 +443,51 @@ func rpcAgentShutdown(env *localenv.LocalEnvironment) error {
...
+
+	t := goterm.NewTable(0, 10, 5, ' ', 0)
+	common.PrintTableHeader(t, []string{"Hostname", "Address", "Status", "Version"})
+	for _, status := range statusList {
+		fmt.Fprintf(t, "%s\t%s\t%s\t%s\n", status.Hostname, status.Address, status.Status, status.Version)
+		if status.Error != nil {
+			log.WithError(err).Debugf("Failed to collect query agent version on %s.", status.Address)
+		}
+	}
+	fmt.Println(t.String())
```
r0mant, line 486: We usually use `env.Println` - it considers some flags potentially passed on the CLI, like "quiet" or "verbose" that might affect the output.

tool/gravity/cli/rpcagent.go
```diff
@@ -388,6 +443,51 @@ func rpcAgentShutdown(env *localenv.LocalEnvironment) error {
...
+	defer cancel()
+
+	statusList := rpc.CollectAgentStatus(ctx, cluster.ClusterState.Servers, fsm.NewAgentRunner(creds))
+
+	t := goterm.NewTable(0, 10, 5, ' ', 0)
+	common.PrintTableHeader(t, []string{"Hostname", "Address", "Status", "Version"})
+	for _, status := range statusList {
+		fmt.Fprintf(t, "%s\t%s\t%s\t%s\n", status.Hostname, status.Address, status.Status, status.Version)
+		if status.Error != nil {
+			log.WithError(err).Debugf("Failed to collect query agent version on %s.", status.Address)
```
r0mant, line 483: Shouldn't this be `log.WithError(status.Error)`?

tool/gravity/cli/rpcagent.go
```diff
@@ -388,6 +443,51 @@ func rpcAgentShutdown(env *localenv.LocalEnvironment) error {
...
+		}
+	}
+	fmt.Println(t.String())
+
+	return trace.Wrap(err)
```
r0mant, line 488: It's a bit weird - what is this error that's returned here? The duration parsing error from above?

tool/gravity/cli/rpcagent.go
```diff
@@ -345,6 +389,17 @@ func newDeployAgentsRequest(ctx context.Context, env *localenv.LocalEnvironment,
 	}, nil
 }
 
+// findServerByHostname returns the server with the matching hostname. Returns
+// a NotFound error if a server with a matching hostname does not exist.
+func findServerByHostname(clusterState storage.ClusterState, hostname string) (server storage.Server, err error) {
```
r0mant, line 394: I'm pretty sure we already have this function in storage.go.

https://github.com/gravitational/gravity/blob/master/lib/storage/storage.go#L696

lib/rpc/status.go
```diff
@@ -0,0 +1,100 @@
...
+
+func getAgentStatus(ctx context.Context, server storage.Server, rpc AgentRepository) AgentStatus {
+	agentStatus := AgentStatus{
+		Hostname: server.Hostname,
+		Address:  server.AdvertiseIP,
+		Status:   constants.GravityAgentOffline,
+	}
+
+	version, err := getVersion(ctx, server.AdvertiseIP, rpc)
+	if err != nil {
```
r0mant, line 71: I think this should also consider a "not implemented" error, in case it contacts the agent that doesn't have GetVersion method yet. In this case the agent is basically online, we just don't know its version (could display it as `N/A` for example).

tool/gravity/cli/rpcagent.go
bernardjkim: https://github.com/gravitational/gravity/blob/46b9ee065b121ed99ac593cb98084ef079ac020b/tool/gravity/cli/rpcagent.go#L416
Copied over from `getGravityPackage` func. It doesn't seem like we need to split version by '+'. I guess we had a different version format in the past?

lib/rpc/status.go
```diff
@@ -0,0 +1,105 @@
...
+		go func(server storage.Server) {
+			statusCh <- getAgentStatus(ctx, server, rpc)
+		}(srv)
+	}
+
+	var statusList []AgentStatus
+	for range servers {
+		statusList = append(statusList, <-statusCh)
+	}
+
```
a-palchikov, line 59: nit: could also close the channel at this point to help gc.

lib/update/cluster/automatic.go
```diff
@@ -43,6 +43,22 @@ func AutomaticUpgrade(ctx context.Context, localEnv, updateEnv *localenv.LocalEn
...
+	siteOp := (*ops.SiteOperation)(operation)
+
+	plan, err := fsm.GetOperationPlan(updateEnv.Backend, siteOp.Key())
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	// Check if the upgrade is already in progress, which would indicate an
+	// agent is already running. Return without creating a new agent.
+	for _, phase := range plan.Phases {
```
a-palchikov, line 55: This should be better encapsulated as `IsOperationInProgress` or `IsPlanInProgress` - you would not need the comment then. I'd check whether we might already have an API like that.

tool/gravity/cli/checks.go
```diff
@@ -134,7 +134,7 @@ func checkUpgrade(ctx context.Context, env *localenv.LocalEnvironment, config pr
 		return trace.Wrap(err)
 	}
 	// Deploy RPC agents that will be used for running checks on the nodes.
-	credentials, err := rpcAgentDeployHelper(ctx, env, "", "")
+	credentials, err := rpcAgentDeployHelper(ctx, env, "", "", "", "")
```
a-palchikov, line 137: Pretty hard to tell what the values are and why empty. In this case I'd suggest an options struct which would be provided to the function as empty.

tool/gravity/cli/rpcagent.go
```diff
@@ -388,6 +432,51 @@ func rpcAgentShutdown(env *localenv.LocalEnvironment) error {
...
+	defer cancel()
+
+	statusList := rpc.CollectAgentStatus(ctx, cluster.ClusterState.Servers, fsm.NewAgentRunner(creds))
+
+	t := goterm.NewTable(0, 10, 5, ' ', 0)
+	common.PrintTableHeader(t, []string{"Hostname", "Address", "Status", "Version"})
+	for _, status := range statusList {
+		fmt.Fprintf(t, "%s\t%s\t%s\t%s\n", status.Hostname, status.Address, status.Status, status.Version)
+		if status.Error != nil {
+			log.WithError(status.Error).Debugf("Failed to collect agent status on %s.", status.Address)
```
a-palchikov, line 472: I wonder if we need to indicate with that command whether (any) agents are offline via a non-0 exit code?
Also, if used with `--quiet` the command will not output anything - not sure if this is intentional/desired.

tool/gravity/cli/rpcagent.go
```diff
@@ -315,16 +331,44 @@ func deployAgents(ctx context.Context, env *localenv.LocalEnvironment, req deplo
 
 // newDeployAgentsRequest creates a new request to deploy agents on the local cluster
 func newDeployAgentsRequest(ctx context.Context, env *localenv.LocalEnvironment, req deployAgentsRequest) (*rpc.DeployAgentsRequest, error) {
-	servers, err := verifyCluster(ctx, req)
-	if err != nil {
-		return nil, trace.Wrap(err)
+	// If hostname is specified in the request, deploy agent only on specified node
+	var servers []rpc.DeployServer
+	if req.hostname != "" {
```
a-palchikov, line 336: Not sure about the need to leak hostname into deployAgentRequest - if it's desired that only an agent on a specific node is started, add an optional server (or a server slice - a single element means only on a single node) to the request.

tool/gravity/cli/rpcagent.go
bernardjkim: It could be useful to have the command return a non-0 exit code if any agents are offline. Updated in cd7ee63.

I think it was preferred to use env.Println. Not entirely sure when someone would want to run the agent status command without any output, but since it now returns an error if any agents are offline, it might be useful somewhere.

lib/system/service/simple.go
```diff
@@ -0,0 +1,55 @@
...
+// specified serviceName.
+func ReinstallSimpleService(serviceName string, cmd []string) error {
+	services, err := systemservice.New()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	err = services.StopService(serviceName)
+	if err != nil {
+		logrus.Warnf("Error stopping service %v: %v.", serviceName, err)
```
r0mant, line 40: `logrus.WithError(err).Warnf()`

r0mant (approved):
Code looks good, one additional suggestion for "testing done" is in addition to `Verify all gravity-agents are shutdown after upgrade` scenario, let's also make sure that the shutdown agents are not automatically restarted after reboot (i.e. complete the upgrade, reboot the node(-s) and make sure agents aren't running).

tool/gravity/cli/rpcagent.go
r0mant: @a-palchikov Do you recall by any chance why we had this "split by +"? I remember it was needed for something but can't remember why.

a-palchikov (approved):

tool/gravity/cli/rpcagent.go
a-palchikov: It might have had something to do with semver parsing but I need to go through the history to know more/remember.

Issue #1732, opened 2020-06-18
Failed remote commands are ignored during upgrade


a-palchikov:
During upgrade, the controller might receives OK from remote command invocation whereas the commands themselves have clearly failed:
```
2020-06-17T11:45:56Z INFO [FSM:REMOT] "Executing phase \"/masters/robotest-e39637e1-node-1/system-upgrade\" finished in 6 seconds \n" CMD:/var/lib/gravity/site/update/agent/gravity#1 gravity:[plan execute --phase /masters/robotest-e39637e1-node-1/system-upgrade --operation-id 9fea8b14-86ed-4b6d-b017-5a7c093cf356] server:robotest-e39637e1-node-1/10.162.0.33 utils/logging.go:103
2020-06-17T11:45:56Z WARN [FSM:REMOT] "\x1b[31m[ERROR]: failed uninstalling gravitational.io/planet:7.0.26-11704 service: error stopping service \"gravity__gravitational.io__planet__7.0.26-11704.service\": failed\n\n\x1b[0m" CMD:/var/lib/gravitysite/update/agent/gravity#1 gravity:[plan execute --phase /masters/robotest-e39637e1-node-1/system-upgrade --operation-id 9fea8b14-86ed-4b6d-b017-5a7c093cf356] server:robotest-e39637e1-node-1/10.162.0.33 utils/logging.go:103
2020-06-17T11:45:56Z DEBU [RPC]       Completed. exit:255 gravity:[plan execute --phase /masters/robotest-e39637e1-node-1/system-upgrade --operation-id 9fea8b14-86ed-4b6d-b017-5a7c093cf356] seq:1 server:robotest-e39637e1-node-1/10.162.0.33 utils/logging.go:103
2020-06-17T11:45:56Z DEBU [FSM:UPDAT] Apply. change:StateChange(Phase=/masters/robotest-e39637e1-node-1/system-upgrade, State=completed) utils/logging.go:103
```
followed by
```
[...]

PR #1914, opened 2020-07-23
[7.0.x] minor fixes

a-palchikov:
## Description

This PR fixes the following issues:
  * annotate errors for each etcd step (i.e invocation of `planet etcd`)  with command output for better visibility.
  * when executing a command for the remote caller, the command execution has precedence over streaming gRPC events. So when a command fails, exec should relay this error consistently.
  * when describing hook jobs in logs, avoid use of naked `interface{}` values to catch type violations.

These are some things I stumbled upon when working on intermediate upgrades port and I wanted to PR separately.

## Type of change

Updates https://github.com/gravitational/gravity/issues/1732

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

The changes were tested as part of a bigger changeset and involved upgrading `5.5.49` to a working intermediate upgrades branch via `6.1.31` as intermediary on 3 controller plane node cluster and a 3 node cluster with a single controller and 2 worker nodes.

### DIFF
--- a/lib/app/hooks/diff.go
+++ b/lib/app/hooks/diff.go
@@ -21,7 +21,8 @@ import (
 	"fmt"
 
 	batchv1 "k8s.io/api/batch/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 // diffPodSets returns a difference in Pods between existing and new.
@@ -126,14 +127,15 @@ func (p *podDiff) String() string {
 	return out.String()
 }
 
-func describe(v interface{}) string {
-	switch val := v.(type) {
+func describe(obj runtime.Object) string {
+	switch obj := obj.(type) {
 	case *v1.Pod:
-		return fmt.Sprintf("Pod %q in namespace %q", val.Name, val.Namespace)
+		return fmt.Sprintf("Pod %q in namespace %q", obj.Name, obj.Namespace)
 	case *batchv1.Job:
-		return fmt.Sprintf("Job %q in namespace %q", val.Name, val.Namespace)
+		return fmt.Sprintf("Job %q in namespace %q", obj.Name, obj.Namespace)
+	default:
+		return fmt.Sprintf("<unknown>")
 	}
-	return "<unknown>"
 }
 
 type phaseDiff struct {
--- a/lib/app/hooks/diff.go
+++ b/lib/app/hooks/diff.go
@@ -133,9 +133,8 @@ func describe(obj runtime.Object) string {
 		return fmt.Sprintf("Pod %q in namespace %q", obj.Name, obj.Namespace)
 	case *batchv1.Job:
 		return fmt.Sprintf("Job %q in namespace %q", obj.Name, obj.Namespace)
-	default:
-		return fmt.Sprintf("<unknown>")
 	}
+	return "<unknown>"
 }
 
 type phaseDiff struct {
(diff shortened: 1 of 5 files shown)


lib/app/hooks/diff.go
```diff
@@ -126,14 +127,15 @@ func (p *podDiff) String() string {
 	return out.String()
 }
 
-func describe(v interface{}) string {
-	switch val := v.(type) {
+func describe(obj runtime.Object) string {
```
a-palchikov, line 130: This ensures that only qualified types can be accepted as parameter.

lib/defaults/defaults.go
```diff
@@ -355,7 +355,7 @@ const (
 	SerfBin = "/usr/bin/serf"
 
 	// JournalctlBin is the default location of the journalctl inside planet
-	JournalctlBin = "/usr/bin/journalctl"
+	JournalctlBin = "/bin/journalctl"
```
a-palchikov, line 358: This path changed in planet at some point. I did not find a ticket for this but will probably need to create and verify that it's been consistently fixed in active branches.

lib/rpc/server/callable.go
```diff
@@ -71,11 +80,14 @@ func (c *osCommand) exec(ctx context.Context, stream pb.OutgoingMessageStream, a
 		}
 	}
 
-	return stream.Send(&pb.Message{Element: &pb.Message_ExecCompleted{ExecCompleted: &pb.ExecCompleted{
```
a-palchikov, line 74: Do not suppress the result of executing the command.

lib/rpc/server/callable.go
```diff
@@ -48,20 +49,28 @@ func (c *osCommand) exec(ctx context.Context, stream pb.OutgoingMessageStream, a
...
 	}}})
 	if err != nil {
-		// Do not wrap gRPC-specific error
-		return err
+		log.WithError(err).Warn("Failed to notify stream of command start.")
 	}
 	err = cmd.Wait()
 	if err == nil {
-		return stream.Send(&pb.Message{Element: &pb.Message_ExecCompleted{ExecCompleted: &pb.ExecCompleted{Seq: seq}}})
+		if err := stream.Send(&pb.Message{
```
r0mant, line 64: Nit: I would wrap stream.Send and error logging into a "void" function, would make the 3 places where you now just log the error instead of returning it a bit cleaner.

```go
func sendMessage(stream XXX, message *pb.Message) {
  err := stream.Send(message)
  if err != nil {
    log.WithError(err).Warn("Failed to send message %T.", message)
  }
}
```

r0mant (approved):

lib/app/hooks/diff.go
```diff
@@ -126,14 +127,15 @@ func (p *podDiff) String() string {
...
+	switch obj := obj.(type) {
 	case *v1.Pod:
-		return fmt.Sprintf("Pod %q in namespace %q", val.Name, val.Namespace)
+		return fmt.Sprintf("Pod %q in namespace %q", obj.Name, obj.Namespace)
 	case *batchv1.Job:
-		return fmt.Sprintf("Job %q in namespace %q", val.Name, val.Namespace)
+		return fmt.Sprintf("Job %q in namespace %q", obj.Name, obj.Namespace)
+	default:
+		return fmt.Sprintf("<unknown>")
```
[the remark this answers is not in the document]
a-palchikov: No reason - I was rearranging the implementation when debugging - will revert.

Issue #1905, opened 2020-07-22
Check tiller health before upgrade


r0mant:
### Feature Request

Before triggering the upgrade we should verify that tiller pod is running and healthy to make sure that users' helm charts can be installed/upgraded.

Success:

- [x] Upgrade executes a pre-check that verifies tiller health.
  - [x] The check should not run if "catalog" feature is disabled in manifest.

r0mant:
Done in 5.5, PRs to [6.1](https://github.com/gravitational/gravity/pull/1929), [7.0](https://github.com/gravitational/gravity/pull/1928) and [master](https://github.com/gravitational/gravity/pull/1927) are created.

PR #1916, opened 2020-07-23
(5.5) Add tiller server pre-check

r0mant:
## Description

Add a pre-upgrade check that makes sure that Tiller server is up and running and alerts the user if it's not, otherwise they might run into issues later when their application upgrade tries to do helm upgrade.

The pre-check can be bypassed by providing a `--force` flag, as other warnings (which may be useful if, for example, Tiller is down for some reason by the user doesn't care since the application doesn't need it). In this case the warning will still be printed to the terminal.

## Type of change

* Closes https://github.com/gravitational/gravity/issues/1905.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

### Scale tiller to 0 and try to launch the upgrade:

```
ubuntu@node-1:~/upgrade$ kubectl -nkube-system scale deploy tiller-deploy --replicas=0
Thu Jul 23 22:47:19 UTC	Upgrading cluster from 5.5.40 to 5.5.50-dev.9
[ERROR]: Tiller server health check failed with the following error:

    "could not find tiller"

This means that "helm upgrade" and other Helm commands may not work correctly. If
the application upgrade requires Helm, make sure that Tiller pod is up, running
and reachable before retrying the upgrade.

This warning can be bypassed by providing a --force flag to the upgrade command.
```

### Try to launch with --force and see a warning

```
ubuntu@node-1:~/upgrade$ sudo ./gravity upgrade --manual --force
Thu Jul 23 22:48:24 UTC	Upgrading cluster from 5.5.40 to 5.5.50-dev.9
Thu Jul 23 22:48:24 UTC	Tiller server health check failed, "helm upgrade" may not work!
Thu Jul 23 22:48:25 UTC	Deploying agents on cluster nodes
Thu Jul 23 22:48:28 UTC	Deployed agent on node-1 (192.168.99.102)
```

### Scale it back and verify upgrade launches fine

```
ubuntu@node-1:~/upgrade$ kubectl -nkube-system scale deploy tiller-deploy --replicas=1
Thu Jul 23 22:50:18 UTC	Upgrading cluster from 5.5.40 to 5.5.50-dev.9
Thu Jul 23 22:50:18 UTC	Deploying agents on cluster nodes
Thu Jul 23 22:50:21 UTC	Deployed agent on node-1 (192.168.99.102)
```

### Disable catalog and verify tiller is not checked

```yaml
extensions:
  catalog:
    disabled: true
```

```
ubuntu@node-1:~/upgrade$ kubectl -nkube-system scale deploy tiller-deploy --replicas=0
ubuntu@node-1:~/upgrade$ sudo ./gravity --debug upgrade --manual
Thu Jul 23 22:56:00 UTC	Upgrading cluster from 5.5.40 to 5.5.50-dev.9
Thu Jul 23 22:56:00 UTC	Deploying agents on cluster nodes
Thu Jul 23 22:56:02 UTC	Deployed agent on node-1 (192.168.99.102)
```

### DIFF
--- a/lib/helm/client.go
+++ b/lib/helm/client.go
@@ -35,6 +35,8 @@ import (
 
 	"github.com/gravitational/trace"
 	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 )
 
 // Client is the Helm client.
@@ -230,7 +232,18 @@ func (c *Client) Revisions(name string) ([]Release, error) {
 
 // Ping pings the Tiller pod and ensures it's up and running.
 func (c *Client) Ping() error {
-	return c.client.PingTiller()
+	err := c.client.PingTiller()
+	if err != nil {
+		// Not all Helm versions implement the ping endpoint, so fall back
+		// to getting the server version.
+		if grpc.Code(err) != codes.Unimplemented {
+			return trace.Wrap(err)
+		}
+		if _, err := c.client.GetVersion(); err != nil {
+			return trace.Wrap(err)
+		}
+	}
+	return nil
 }
 
 // Close closes the Helm client.
--- a/lib/helm/client.go
+++ b/lib/helm/client.go
@@ -258,6 +258,7 @@ func Ping(dnsAddress string) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
+	defer client.Close()
 	return client.Ping()
 }
 
--- a/lib/helm/client.go
+++ b/lib/helm/client.go
@@ -228,10 +228,24 @@ func (c *Client) Revisions(name string) ([]Release, error) {
 	return releases, nil
 }
 
+// Ping pings the Tiller pod and ensures it's up and running.
+func (c *Client) Ping() error {
+	return c.client.PingTiller()
+}
+
 // Close closes the Helm client.
 func (c *Client) Close() error {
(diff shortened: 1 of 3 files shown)


lib/helm/client.go
```diff
@@ -228,12 +230,37 @@ func (c *Client) Revisions(name string) ([]Release, error) {
...
+
 // Close closes the Helm client.
 func (c *Client) Close() error {
 	c.tunnel.Close()
 	return nil
 }
 
+// Ping pings the cluster's Tiller pod and ensures it's up and running.
+func Ping(dnsAddress string) error {
+	client, err := NewClient(ClientConfig{DNSAddress: dnsAddress})
```
bernardjkim, line 257: Close client after ping.

a-palchikov, line 257: nit: never was a fan of spreading low-level details such as DNS address instead of the high-level info such as the kubernetes client - I would replace the DNS address with that.

a-palchikov (approved):

lib/helm/client.go
r0mant: Good catch!

r0mant: This code is soon-to-be-refactored anyway (when we finally migrate to Helm 3), don't want to bother now.

PR #1917, opened 2020-07-24
[experimental] mage/magnet build

knisbet:
## Description

Initial implementation of using magnet to speed up our build process. As discussed at the last team meeting, this is experimental, but probably in a state where the team can try and use it and fix problems as we go.

## Type of change

- OSS users trying to build gravity themselves.

* Requires #
- https://github.com/gravitational/gravity.e/pull/4319

## TODOs

- [ ] Self-review the change
- [x] Perform manual testing
- [x] Write documentation
- [ ] Address review feedback

## Implementation

See mage/README.md for most of the notes about trying/using.

## Testing done

Tried to run various build scenarios, but obviously not able to cover everything and what other team members are doing. So expect plenty of glitches and issues.


knisbet:
Hmmm... think we probably need to update docker on the build server.

a-palchikov:
I would not worry about this working on CI - like discussed, we'd focus on using this internally for development.

knisbet:
> I would not worry about this working on CI - like discussed, we'd focus on using this internally for development.

@a-palchikov The problem is I changed the container image builds to be buildkit based, even for the Makefiles, which is why CI/CD failed. While I suppose I could rollback that one small thing, it's also one small thing that helps out the existing build process.

a-palchikov:
Ah, I see. And I suppose it would be difficult (if not impossible) to move the changes to environment (or other configuration) in this case. Maybe we can live with some duplicates (eg dockerfiles) and move dockerfile references in the makefile to configuration (environment) so they can be swapped where necessary.

knisbet:
Yea, I think there are sort of 3 options here:
1. Upgrade docker on the build server. Downside is the additional caching may make our disk space issues worse.
2. Don't buildkit the images in the short term, the main benefit is being able to use a shared apt-cache and isn't used in too many places, and overall the changes are small.
3. Use alternative dockerfiles for Magnet that use buildkit, and use the originals for make, requiring devs to keep them in sync.

UnderGreen:
I don't believe we change dockerfiles ofter in gravity project. Maybe until next time, one needs to change something we already migrate to another build system.

wadells (commented):
Looks great!

My major concern is:

We'll probably see significant drift between the two build versions as long as they exist in parallel and there isn't a single source of truth (e.g. for dependency versions).  If/when we take this, I think we *should* add the new build into CI (in addition to the old) to make sure it remains topically working.  Otherwise folks will make changes that break this, and then discovering/fixing those changes or inconsistencies is pushed onto the users of the new system instead of the dev who made the change.

Going further, until we use this for integration tests and publishing, we run the risk of generated artifacts being subtly different, and thus not trusted.  With that in mind: I'd recommend we aim to make this the sanctioned build system before 7.1 (but not in this PR).  I'd further council we avoid back-porting it as the sanctioned build system to LTS releases (this introduces risk for no end user benefit).  I recognize this leaves us in the unenviable position of working with both systems until our current crop of LTS branches die -- a couple years.

P.S.  For testing done, I encourage you to "Show, don't tell".  I know you gave a demo in the  team meeting, but can you add some shell output, e.g. for a clean build and a cached build?  A link to a gist is good if it is too big for the description.

wadells:
> Yea, I think there are sort of 3 options here:
> 
>     1. Upgrade docker on the build server. Downside is the additional caching may make our disk space issues worse.

`/var/lib/docker` is on its own 1TB device that we've never filled during my tenure -- at least that I'm aware of.  All the space issues I'm tracking of come from `/` or `/var/lib/jenkins/workspace`.  The main thing here is scheduling downtime and figuring out if we want to cowboy an unmanaged docker upgrade or get ansible ship shape first.

>     2. Don't buildkit the images in the short term, the main benefit is being able to use a shared apt-cache and isn't used in too many places, and overall the changes are small.

:+1: until we can get to #1

>     3. Use alternative dockerfiles for Magnet that use buildkit, and use the originals for make, requiring devs to keep them in sync.

This seems like a maintenance headache.

build.assets/Dockerfile
```diff
@@ -33,13 +42,6 @@ ENV LANGUAGE="en_US.UTF-8" \
      GOPATH="/gopath" \
      PATH="$PATH:/opt/protoc/bin:/opt/go/bin:/gopath/bin"
 
-RUN (mkdir -p /gopath/src/github.com/gravitational && \
-     cd /gopath/src/github.com/gravitational && \
-     git clone https://github.com/gravitational/version.git && \
-     cd /gopath/src/github.com/gravitational/version && \
-     git checkout ${VERSION_TAG} && \
-     go install github.com/gravitational/version/cmd/linkflags)
-
```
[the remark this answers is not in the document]
knisbet: Doesn't appear to be used. I don't see any references grepping the code base.

go.mod
```diff
@@ -208,18 +210,19 @@ replace (
 	github.com/hashicorp/hcl => github.com/hashicorp/hcl v1.0.0
 	github.com/hashicorp/memberlist => github.com/hashicorp/memberlist v0.1.4
 	github.com/imdario/mergo => github.com/imdario/mergo v0.0.0-20160517064435-50d4dbd4eb0e
+	github.com/jaguilar/vt100 => github.com/tonistiigi/vt100 v0.0.0-20190402012908-ad4c4a574305
 	github.com/json-iterator/go => github.com/json-iterator/go v1.1.5
 	github.com/julienschmidt/httprouter => github.com/julienschmidt/httprouter v1.1.0
 	github.com/kr/pty => github.com/kr/pty v1.0.0
 	github.com/kylelemons/godebug => github.com/kylelemons/godebug v0.0.0-20160406211939-eadb3ce320cb
+	github.com/magefile/mage => github.com/knisbet/mage v1.9.1-0.20200719045837-eabe8cda6d46
```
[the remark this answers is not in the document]
knisbet: I need to reach out to the devs, and see about adding a feature to mage to run a pre-exit hook, so that we can shutdown the UI library. For now this is just hacked into a fork, so it's not something that can be upstreamed directly.

mage/README.md
```diff
@@ -0,0 +1,53 @@
...
+Mage doesn't support build arguments, so all customization of the build process is done via environment variables. For
+a complete listing of environment variables, run the help:envs target. This will give a listing of all env variables
+and any current settings.
+
+`go run mage.go help:envs`
+
+## Notes and Tips
+
+1. There isn't really a good way to rebuild specific packages bypassing caching, besides changing the version that
+would be assigned. So for example to force rebuild RBAC would require changing the version by env variable.
```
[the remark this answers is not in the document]
knisbet: This is in line with caching behaviour today, but a little more exposed. Even today we try and be clever with some caching, which is effectively controlled by the version number. So if rbac-app:1.0.0 has been built before and uploaded to AWS, if we're able to download rbac-app:1.0.0 it will use that instead of rebuilding it, even if the contents have changed. 

The difference here is I don't try and download from AWS at all, but still cache locally by the version number. So if you build rbac-app:1.0.0-wadells.test.1, make some changes, and build again, the resulting installer may have the cached/old version. 

The way I'd imagine this working in the short term would by to do something like:
`FORCE_APP_REBUILD=true go run mage.go package:rbac && go run mage.go cluster:gravity` 

This would force rebuild some app being worked on, allow all other apps to be served from cache, but haven't gotten around to implementing it. For now it means changing the tag every build to avoid caches, so rbac-app:1.0.0-wadells.test.2, test.3, etc. 

Ideally we'd be able to use BuildKit as an Inputs cache, but as I explained on the last team call one of the problems we have is that the gravity binary is an input to the content cache, meaning each change to the gravity binary invalidates the caches. We don't change the format very frequently, so we might be able to get to a point where we can do something here, or maybe it would be better to just add content caching directly to our tools. 

Changing the tag to build though on the other hand is a good practice. Gravity internally uses the version number to decide to upgrade, so it has happened more than once where contents have been changes, but the version number has been missed.

mage/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+For Example: `RBAC_APP_TAG=7.0.0-rbac.1 go run mage.go cluster:gravity`
+
+2. The automatic version assignment this uses is a bit different than the version.sh script currently being used. I'm 
+not really sure what's better, but version.sh will generate something like `7.1.0-alpha.1.149` and this will use
+`7.1.0-alpha.1-149-g86435b2c-dirty`. The downside is this can generate more packages to rebuild when getting marked as
+dirty.
```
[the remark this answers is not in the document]
knisbet: This tag is really the result of what `git describe --tags --dirty` will produce, so yes it's not something meant to produce metadata as defined in semver. 

```
https://github.com/gravitational/magnet/blob/master/util.go#L12
```

mage/enterprise.go
```diff
@@ -0,0 +1,30 @@
...
+	"os"
+
+	"github.com/gravitational/trace"
+	"github.com/magefile/mage/mg"
+)
+
+type Enterprise mg.Namespace
+
+func hasE() error {
+	if _, err := os.Stat("e/.github"); !os.IsNotExist(err) {
```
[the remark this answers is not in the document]
knisbet: No reason really, just wanted a simple and quick heuristic that would be unlikely to change in the future. .github seemed reasonable as it's unlikely we'll remove CODEOWNERS in the foreseeable future.

mage/misc.go
```diff
@@ -0,0 +1,28 @@
...
+
+package mage
+
+import (
+	"os"
+
+	"github.com/gravitational/trace"
+)
+
+// Clean cleans up the build directory and caches.
```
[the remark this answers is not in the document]
knisbet: Fair enough. Following the XDG structure was a relatively recent addition.

mage/packages.go
```diff
@@ -0,0 +1,982 @@
...
+		p.DefaultCachePath(),
+	)
+	return trace.Wrap(err)
+}
+
+func (p gravityPackage) localAppImport(m *magnet.Magnet, stateDir string) error {
+	_, err := m.Exec().Run(
+		context.TODO(),
+		consistentGravityBin(),
+		"--state-dir", stateDir,
```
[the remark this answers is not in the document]
knisbet: No.

The idea is each app builds independently in its own state directory, which allows the build to be parallel, working on an independent state directory. Only steps that interact with the same state directory are protected by a mutex at of an abundance of caution incase parallel interactions with the same state directory are unsound.

mage/vars.go
```diff
@@ -0,0 +1,267 @@
...
+
+	return fmt.Sprintf("%d%02d%02d", version.Major, version.Minor, version.Patch)
+}
+
+func buildFlags() []string {
+	return []string{
+		fmt.Sprint(`-X github.com/gravitational/version.gitCommit=`, magnet.DefaultHash()),
+		fmt.Sprint(`-X github.com/gravitational/version.version=`, buildVersion),
+		fmt.Sprint(`-X github.com/gravitational/gravity/lib/defaults.WormholeImg=`, wormholeImage),
+		fmt.Sprint(`-X github.com/gravitational/gravity/lib/defaults.TeleportVersionString=`, teleportTag),
```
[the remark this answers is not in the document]
knisbet: https://github.com/gravitational/gravity/blob/master/Makefile#L81-L84

mage/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+3. To better interact with gopls, dependent apps are no longer extracted to the build/ directory, and instead generate
+temporary directory for extraction and build.
+
+4. For now I've set caching to default to the build directory. So build/cache will be used for golang's caches, and 
+build/apps get's used for internal applications. The cache directory can be overriden by `XDG_CACHE_HOME`.
```
r0mant, line 46: Nit: get's -> gets

mage/build.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	err = m.GolangBuild().
+		SetGOOS("linux").
+		SetGOARCH("amd64").
+		SetEnv("GO111MODULE", "on").
+		SetMod("vendor").
+		AddTag("selinux").
+		AddTag("selinux_embed").
+		SetBuildContainer(fmt.Sprint("gravity-build:", buildVersion)).
```
r0mant, line 59: Use `buildBoxName()` function here instead?

mage/build.go
```diff
@@ -0,0 +1,120 @@
...
+	m := root.Target("build:go")
+	defer func() { m.Complete(err) }()
+
+	packages := []string{"github.com/gravitational/gravity/tool/gravity", "github.com/gravitational/gravity/tool/tele"}
+	if enterprise != "" {
+		if err = hasE(); err != nil {
+			return trace.Wrap(err)
+		}
+
+		packages = []string{"github.com/gravitational/gravity/e/tool/gravity", "github.com/gravitational/gravity/e/tool/tele"}
```
r0mant, line 49: Move these paths (and the ones above) into constants? I see they're used in multiple places.

mage/build.go
```diff
@@ -0,0 +1,120 @@
...
+		SetBuildArg("UID", fmt.Sprint(os.Getuid())).
+		SetBuildArg("GID", fmt.Sprint(os.Getgid())).
+		SetDockerfile("build.assets/Dockerfile").
+		Build(context.TODO(), "./build.assets")
+
+	return trace.Wrap(err)
+}
+
+// Selinux builds internal selinux code
+func (Build) Selinux() (err error) {
```
r0mant, line 110: Nit: Selinux -> SELinux

mage/cluster.go
```diff
@@ -0,0 +1,66 @@
...
+	_, err = m.Exec().SetEnv("GRAVITY_K8S_VERSION", k8sVersion).Run(context.TODO(),
+		filepath.Join(consistentBinDir(), "tele"),
+		"--debug",
+		"build",
+		"assets/telekube/resources/app.yaml",
+		"-f",
+		"--version", buildVersion,
+		"--state-dir", consistentStateDir(),
+		"--skip-version-check",
+		"-o", filepath.Join(consistentBuildDir(), "gravity.tar"),
```
r0mant, line 42: I think a bunch of publishing depends on this tarball being called `telekube.tar`.

mage/build.go
```diff
@@ -0,0 +1,120 @@
...
+	"context"
+	"fmt"
+	"os"
+	"runtime"
+
+	"github.com/gravitational/trace"
+	"github.com/magefile/mage/mg"
+)
+
+type Build mg.Namespace
```
r0mant, line 26: Godoc

mage/build.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (Build) All() {
```
r0mant, line 28: Godoc

mage/cluster.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"context"
+	"path/filepath"
+
+	"github.com/gravitational/trace"
+	"github.com/magefile/mage/mg"
+)
+
+type Cluster mg.Namespace
```
r0mant, line 24: Godoc

mage/enterprise.go
```diff
@@ -0,0 +1,30 @@
...
+package mage
+
+import (
+	"os"
+
+	"github.com/gravitational/trace"
+	"github.com/magefile/mage/mg"
+)
+
+type Enterprise mg.Namespace
```
r0mant, line 23: Godoc

mage/enterprise.go
```diff
@@ -0,0 +1,30 @@
...
+
+func hasE() error {
+	if _, err := os.Stat("e/.github"); !os.IsNotExist(err) {
+		return nil
+	}
+	return trace.BadParameter("Enterprise sources are not available.")
```
r0mant, line 29: Add a hint on how to make them available? `git submodule init e && git submodule update`. Also, `trace.NotFound` is prob. more appropriate here.

mage/misc.go
```diff
@@ -0,0 +1,39 @@
...
+
+// Clean cleans up the build directory.
+func Clean() (err error) {
+	m := root.Target("build:buildContainer")
+	defer func() { m.Complete(err) }()
+
+	return trace.Wrap(os.RemoveAll("build"))
+}
+
+// Env
```
r0mant, line 32: Unfinished godoc

mage/packages.go
```diff
@@ -0,0 +1,982 @@
...
+	"strings"
+	"sync"
+
+	"github.com/gravitational/magnet"
+	"github.com/gravitational/magnet/pkg/cp"
+	"github.com/gravitational/trace"
+	"github.com/magefile/mage/mg"
+)
+
+type Package mg.Namespace
```
r0mant, line 31: Godoc

mage/packages.go
```diff
@@ -0,0 +1,982 @@
...
+type gravityPackage struct {
+	repository string
+	name       string
+	version    string
+
+	// env are any env variables that need to be passed to the underlying build commands
+	env map[string]string
+
+	// apps subcommands
+	// apps from local repository
```
r0mant, line 42: Seems like either misplaced or unfinished comment.

mage/packages.go
```diff
@@ -0,0 +1,982 @@
...
+	pkgGravity = gravityPackage{
+		repository: "gravitational.io",
+		name:       "gravity",
+		version:    buildVersion,
+		force:      true,
+	}
+
+	pkgTeleport = gravityPackage{repository: "gravitational.io", name: "teleport", version: teleportTag}
+	pkgFio      = gravityPackage{repository: "gravitational.io", name: "fio", version: fioPkgTag}
+	pkgSelinux  = gravityPackage{repository: "gravitational.io", name: "selinux", version: selinuxVersion}
```
r0mant, line 72: Nit: SELinux

mage/packages.go
```diff
@@ -0,0 +1,982 @@
...
+			pkgStorageApp.Locator(),
+			pkgBandwagonApp.Locator(),
+			pkgTillerApp.Locator(),
+			pkgSiteApp.Locator(),
+		},
+		force: true,
+	}
+)
+
+func (Package) Telekube() (err error) {
```
r0mant, line 206: Godocs on exported methods here and below are missing

mage/packages.go
```diff
@@ -0,0 +1,982 @@
...
+	m := root.Target("package:k8s")
+	defer func() { m.Complete(err) }()
+
+	return trace.Wrap(pkgKubernetes.localAppImport(m, consistentStateDir()))
+}
+
+func (Package) GravityPackage() (err error) {
+	mg.Deps(Build.Go)
+
+	m := root.Target("package:gravity-package")
```
r0mant, line 229: Hmm, so we have packages (gravity, planet) and apps (site, monitoring, etc) so I wonder if using the same naming in targets would make sense. I.e. this target name would be `package:gravity`. Targets above would be `app:telekube`, `app:k8s`, `app:bandwagon`, etc.

mage/packages.go
```diff
@@ -0,0 +1,982 @@
...
+		filepath.Join(buildDir, "/rootfs/usr/share/teleport"),
+		tmpDir,
+	} {
+		err = os.MkdirAll(dir, 0755)
+		if err != nil {
+			return trace.ConvertSystemError(err)
+		}
+	}
+
+	_, err = m.Exec().SetWD(tmpDir).Run(context.TODO(),
```
r0mant, line 302: Looks like this always builds Teleport from source? In the current Makefile we first try to download a pre-built asset from S3. Same for monitoring, logging and other apps. Or is that logic elsewhere?

mage/packages.go
```diff
@@ -0,0 +1,982 @@
...
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	defer os.Remove("build/fio")
+
+	return trace.Wrap(pkgFio.ImportPackage(m, "build/fio"))
+}
+
+func (Package) Selinux() (err error) {
```
r0mant, line 388: Nit: SELinux

mage/packages.go
```diff
@@ -0,0 +1,982 @@
...
+
+func (Package) Web() (err error) {
+	mg.Deps(Build.Go)
+
+	m := root.Target("package:web-assets")
+	defer func() { m.Complete(err) }()
+
+	webImage := fmt.Sprint("telekube-oss-web:", pkgWebAssets.version)
+	contextPath := "web/"
+	if enterprise != "" {
```
r0mant, line 487: I'm curious how hard it would be (or even worth the effort) to try and move enterprise-specific logic to e/ submodule so open-source version doesn't have enterprise-specific code.

mage/tests.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/gravitational/magnet"
+	"github.com/gravitational/trace"
+	"github.com/magefile/mage/mg"
+)
+
+type Test mg.Namespace
+
+func (Test) All() {
+	//mg.SerialDeps(Build.Go, Build.BuildContainer)
```
r0mant, line 16: Let's either add a todo item describing what needs to happen here or remove the commented code.

mage/vars.go
```diff
@@ -0,0 +1,281 @@
...
+
+	// planetTag is <planet version>-<encoded kubernetes version> as would be tagged in the planet repo
+	// TODO: We should consider a way to import planet directly from a docker image for OSS users customizing planet
+	// or add support for building off forks of the repo
+	//planetTag = fmt.Sprintf("7.1.4-%v", k8sVersionToPlanetFormat(k8sVersion))
+	planetTag = ""
+
+	planetBranch = magnet.E(magnet.EnvVar{
+		Key: "PLANET_BRANCH",
+		//Default: planetTag,
```
r0mant, line 96: Remove this?

mage/vars.go
```diff
@@ -0,0 +1,281 @@
...
+	// k8sVersion is the version of kubernetes we're shipping
+	k8sVersion = magnet.E(magnet.EnvVar{
+		Key:   "K8S_VER",
+		Short: "The k8s version to use (and locate the planet tag)",
+	})
+
+	// planetTag is <planet version>-<encoded kubernetes version> as would be tagged in the planet repo
+	// TODO: We should consider a way to import planet directly from a docker image for OSS users customizing planet
+	// or add support for building off forks of the repo
+	//planetTag = fmt.Sprintf("7.1.4-%v", k8sVersionToPlanetFormat(k8sVersion))
```
r0mant, line 91: Is this supposed to be commented out?

r0mant (commented):
Just a first pass, haven't tried using this yet...

mage/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+### OSS
+
+`go run mage.go cluster:gravity`
+
+This is roughly equivelant to `make production telekube` within the OSS source directory. 
+
+### Enterprise
+
+`ENTERPRISE=true go run mage.go cluster:gravity`
```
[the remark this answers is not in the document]
knisbet: No. I've been taking a bit of a different approach from the Makefile, it's not exactly a secret we have an e submodule and separate code for enterprise. But to me it didn't make as much sense to duplicate build logic in the subdir, so instead I just made the build tooling capable of understanding when e is/isn't present, and can use the same targets for building either version.

build.assets/Dockerfile
```diff
@@ -33,13 +42,6 @@ ENV LANGUAGE="en_US.UTF-8" \
      GOPATH="/gopath" \
      PATH="$PATH:/opt/protoc/bin:/opt/go/bin:/gopath/bin"
 
-RUN (mkdir -p /gopath/src/github.com/gravitational && \
-     cd /gopath/src/github.com/gravitational && \
-     git clone https://github.com/gravitational/version.git && \
-     cd /gopath/src/github.com/gravitational/version && \
-     git checkout ${VERSION_TAG} && \
-     go install github.com/gravitational/version/cmd/linkflags)
```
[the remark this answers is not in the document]
knisbet: So I wasn't able to find anything that actually called linkflags, which is why I removed it from the container.

mage/README.md
```diff
@@ -0,0 +1,53 @@
...
+## Configuration
+Mage doesn't support build arguments, so all customization of the build process is done via environment variables. For
+a complete listing of environment variables, run the help:envs target. This will give a listing of all env variables
+and any current settings.
+
+`go run mage.go help:envs`
+
+## Notes and Tips
+
+1. There isn't really a good way to rebuild specific packages bypassing caching, besides changing the version that
```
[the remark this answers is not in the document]
knisbet: Yea, I have a sort of clean target. I think eventually I'd like to support something like:
`NO_CACHE=true go run mage.go app:monitoring-app && go run mage.go cluster:gravity`. This way, just the monitoring-app is rebuilt ignoring the cache, but the rest of the cluster image can be rebuilt using the cache as opposed to deleting everything. 

So this comment was more of a we don't support the model I'd like here.

mage/packages.go
knisbet: IIRC for stuff that's in golang, the idea is to have a working GOCACHE. So building teleport should be relatively quick off the gocache versus the added complexity of trying to download a built binary from somewhere else.

So atleast as it stands today, magnet depends more on local caching, and doesn't do the remote caching stuff. Means a slower build every once in awhile, but on average it's much faster without the try to download if possible logic.

mage/packages.go
knisbet: I'm not sure what benefit that gets us?

wadells:
Superseded by https://github.com/gravitational/gravity/pull/2485

PR #1936, opened 2020-07-29
(7.0) Background plan resume / plan tail

r0mant:
## Description

https://github.com/gravitational/gravity/pull/1899 port.

### DIFF
--- a/tool/gravity/cli/clusterupdate.go
+++ b/tool/gravity/cli/clusterupdate.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 	"text/tabwriter"
 
 	"github.com/gravitational/gravity/lib/app"
@@ -38,6 +39,7 @@ import (
 	statusapi "github.com/gravitational/gravity/lib/status"
 	"github.com/gravitational/gravity/lib/storage"
 	"github.com/gravitational/gravity/lib/system/selinux"
+	"github.com/gravitational/gravity/lib/systemservice"
 	"github.com/gravitational/gravity/lib/update"
 	clusterupdate "github.com/gravitational/gravity/lib/update/cluster"
 	"github.com/gravitational/gravity/lib/utils/helm"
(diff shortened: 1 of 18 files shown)


tool/gravity/cli/clusterupdate.go
```diff
@@ -222,8 +224,86 @@ func executeUpdatePhaseForOperation(env *localenv.LocalEnvironment, environ Loca
...
+	status, err := systemd.StatusService(name)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	// Since we're using a one-shot service, it will be in "activating" state
+	// for the duration of the command execution. In fact, one-shot services
+	// never reach "active" state, but we're checking it too just in case.
+	switch status {
+	case systemservice.ServiceStatusActivating, systemservice.ServiceStatusActive:
+		return trace.AlreadyExists("service %v is already running", name)
```
a-palchikov, line 290: Hmm, active means the service has already run?
Activating also does not mean the same service - could be one from previous attempts with a different command line.

a-palchikov (approved):

tool/gravity/cli/clusterupdate.go
r0mant: > active means the service has already run

One-shot services never reach the "active" state. They go straight from "activating" to "exited" (or whatever the terminal state is called).

> Activating also does not mean the same service

Which is fine, since we don't want to launch another resume while another one is going, even if it's a different one.

r0mant: To clarify a bit more on the one-shot services and "active" state - they never reach it unless `RemainAfterExit=true` is set (which it is not the for the "gravity-resume" service). From systemd docs on one-shot services:

> Note that if this option is used without RemainAfterExit= the service will never enter "active" unit state, but directly transition from "activating" to "deactivating" or "dead" since no process is configured that shall run continuously. In particular this means that after a service of this type ran (and which has RemainAfterExit= not set) it will not show up as started afterwards, but as dead.

For "gravity-agent" service we do set `RemainAfterExit=true` for some reason - I don't know what was the reasoning behind it, so the agent service indeed shows as "active" after successful upgrade, which is weird.

PR #1950, opened 2020-07-30
[7.0.x] Create directories for monitoring data

UnderGreen:
## Description

1. Create directories for storing Prometheus/Alertmanager data. 
2. Bump rigging version with gravitational/prometheus-operator data to serialize Prometheus/Alertmanager CRD resources with the support of hostPath as storage.

## Type of change

* Requires https://github.com/gravitational/rigging/pull/80, https://github.com/gravitational/monitoring-app/pull/185

## TODOs

- [ ] Self-review the change
- [ ] Write tests
- [ ] Perform manual testing
- [ ] Write documentation
- [ ] Address review feedback

## Implementation

`prometheus-db` and `alertmanager-db` are hardcoded subPath for volume mounts in prometheus-operator.

## Testing done

- [x] 3-node installation. After installation, there are 2 pods of prometheus and 2 pods of alertmanager. All graphs are showing data.
- [x] 1+2 nodes installation. One controller and 2 workers.
- [ ] Upgrade 3 nodes.
- [ ] Upgrade 1+2 nodes.


lib/install/phases/bootstrap.go
```diff
@@ -220,6 +220,8 @@ func (p *bootstrapExecutor) configureSystemDirectories(ctx context.Context) erro
 		filepath.Join(stateDir, "secrets"),
 		filepath.Join(stateDir, "backup"),
 		filepath.Join(stateDir, "logrange"),
+		filepath.Join(stateDir, "monitoring", "prometheus-db"),
```
a-palchikov, line 223: nit: but why not simply `prometheus` and `alertmanager` - are we planning to have more directories under monitoring prefixed with `prometheus`/`alertmanager`?

a-palchikov (approved):

lib/install/phases/bootstrap.go
r0mant: Yeah, I would just name "prometheus" and "alertmanager".

UnderGreen:
```
Implementation

prometheus-db and alertmanager-db are hardcoded subPath for volume mounts in prometheus-operator.```

UnderGreen: I would like to create only `monitoring` path and leave it like that. But prometheus operator makes volume mounts with hardcoded paths `alertmanager/prometheus-db` and if directory does not exist Kubernetes creates it with the permissions of kubelet, which is root/root. In that case I can't write in them because those containers are running with service-user. This is why I explicitly create `alertmanager-db/prometheus-db`.

a-palchikov: Ok - lgtm then.

r0mant (approved):

Issue #1940, opened 2020-07-29
[7.0.x] custom planet container override broken


a-palchikov:
### Description

**What happened**:
If a cluster image overrides the system container globally, the resulting installer will fail to install as it's missing the container image package:

```yaml
systemOptions:
  baseImage: repo/planet:<version>
```

**What you expected to happen**:
The container image package is properly generated and installation works as expected.

**How to reproduce it (as minimally and precisely as possible)**:
Build an installer with a custom system container image - installation fails.

digger18:
It worked for me with Gravity version 7.0.15 and a global custom planet image.

My app.yaml:
```
baseImage: "gravity:7.0.15"

systemOptions:
  baseImage: gcr.io/[REDACTED]/planet:7.0.41-11709
```

```
root@ip-10-83-2-215:~/gravity-base-k8s-7.0.15-2# gravity status
Cluster name:		cluster.local
Cluster status:		active
Cluster image:		gravity-base-k8s, version 7.0.15-2
Gravity version:	7.0.15 (client) / 7.0.15 (server)
Join token:		2079da4ebfdfe360e133558acd9445db
Last completed operation:
    * 1-node install
      ID:		ba94cac7-1877-4299-9ee6-376f61c5d393
      Started:		Thu Aug 20 08:48 UTC (6 minutes ago)
      Completed:	Thu Aug 20 08:49 UTC (5 minutes ago)
Cluster endpoints:
    * Authentication gateway:
        - 10.83.2.215:32009
    * Cluster management URL:
        - https://10.83.2.215:32009
Cluster nodes:
    Masters:
        * ip-10-83-2-215 / 10.83.2.215 / master
            Status:		healthy
            Remote access:	online
root@ip-10-83-2-215:~/gravity-base-k8s-7.0.15-2# gravity package list

[cluster.local]
---------------

* cluster.local/cert-authority:0.0.1 12kB operation-id:ba94cac7-1877-4299-9ee6-376f61c5d393,purpose:ca
* cluster.local/planet-10.83.2.215-secrets:7.0.41-11709 71kB advertise-ip:10.83.2.215,installed:installed,operation-id:ba94cac7-1877-4299-9ee6-376f61c5d393,purpose:planet-secrets
[...]

PR #1962, opened 2020-08-04
[6.1.x] custom planet containers

a-palchikov:
## Description

This PR fixes the issue with customizing the planet (system) container.

Previously, the installer tarball built with custom docker image replacing the default image still included the redundant default image. The installer built with a custom docker image for any of the node profile correctly appended the custom container package.

Now, the resulting tarball will not include the default package if it has been customized globally (using top-level `systemOptions`).

## Type of change

* Updates https://github.com/gravitational/gravity/issues/1940

* Requires https://github.com/gravitational/gravity.e/pull/4326, https://github.com/gravitational/planet/pull/717.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

#### Prerequisites
1. Custom planet container replacing the default package:

```yaml
systemOptions:
  baseImage: quay.io/gravitational/planet:<custom-build>
```

2. Custom planet container for a (subset of) node profile

```yaml
nodeProfiles:
  - name: profile
    systemOptions:
      baseImage: quay.io/gravitational/planet:<custom-build>
```

#### Clusters

- [x] Installation, 3-node cluster, all control plane nodes
- [x] Installation, 3-node cluster, one control plane, 2 workers
- [x] Installation, 1-node cluster,  then expand to 3-nodes (both controller plane and workers)
- [x] Upgrade from a previous version, 3-node cluster
- [x] Upgrade from a previous version, run gc and expand (both controller plane and workers)

## Implementation
I decided to adopt the package/app pull/sync API changes from 7.0.x intermediate upgrades since they included the necessary changes for this PR as well.
Note, that I also did not include the planet changes directly as they are used indirectly - during `tele build` when the planet docker image is translated into a gravity package.


lib/update/cluster/builder.go
```diff
@@ -638,43 +637,38 @@ func supportsTaints(gravityPackage loc.Locator) (supports bool, err error) {
 	return defaults.BaseTaintsVersion.Compare(*ver) <= 0, nil
 }
 
-func shouldUpdateEtcd(p planConfig) (updateEtcd bool, installedEtcdVersion string, updateEtcdVersion string, err error) {
+func shouldUpdateEtcd(installedRuntimePackage, updateRuntimePackage loc.Locator, packages pack.PackageService) (etcd *etcdVersion, err error) {
```
a-palchikov, line 640: To determine whether to update etcd, we need to look at the planet packages of the top-level application and not the runtime application as before. This is because the top-level application can override the planet package, in which case we might be looking for the wrong package since the default planet package is not packaged in the resulting tarball if the system container has been replaced in global `systemOptions`.

tool/gravity/cli/operation.go
```diff
@@ -82,13 +82,8 @@ type SetPhaseParams struct {
...
 // resumeOperation resumes the operation specified with params
 func resumeOperation(localEnv *localenv.LocalEnvironment, environ LocalEnvironmentFactory, params PhaseParams) error {
-	err := executePhase(localEnv, environ, PhaseParams{
-		PhaseID:          fsm.RootPhase,
-		Force:            params.Force,
-		Timeout:          params.Timeout,
-		SkipVersionCheck: params.SkipVersionCheck,
-		OperationID:      params.OperationID,
-	})
+	params.PhaseID = fsm.RootPhase
```
a-palchikov, line 85: This fixes the issue with `gravity plan resume` introduced [here](https://github.com/gravitational/gravity/pull/1935) not relaying the `block` flag to the `executePhase` which decides how to execute - blocking or from service.

r0mant (approved):

Issue #1923, opened 2020-07-28
Allow a configurable robotest image


wadells:
### Feature Request
Expand robotest testing to include on/off testing for various elements of the Gravity application manifest that are baked into the robotest image (e.g. OpenEBS, Ingress)

### Motivation
This was requested by @a-palchikov here:

https://github.com/gravitational/gravity/issues/1908#issuecomment-663440933

> Having a single image will still be insufficient to test on/off scenarios. Something like the ability to override(all) values from the manifest at install/upgrade is necessary.

### Who's it for?
Gravity Developers

wadells:
@a-palchikov: Can you give me an example of a specific regression issue or two where a customized image would have helped catch the issue -- or we need to test both the on and off case?

I want to weigh the cost-benifit of thi -- it could dramatically increase the dimensional of our automated testing.

a-palchikov:
I meant this as a way to test additional build/install paths when the system container image is completely replaced. See my other [comment](https://github.com/gravitational/gravity/issues/1925#issuecomment-665542539)

wadells:
> I meant this as a way to test additional build/install paths when the system container image is completely replaced. See my other [comment](https://github.com/gravitational/gravity/issues/1925#issuecomment-665542539)

Is this something you think we should do as an automated test:
* on every PR
* on nightly builds on master?
* as one-off testing not integrated with CI (e.g. using robotest to aid manual testing or execution of a test plan)

I see this as a potentially very high cost ask given number of dimensions it would add to any CI based test matrix (PR or Nightly) which means it looks low ROI.  If it is for one-off testing with robotest, that isn't a use case the tool presently targets -- though it is certainly a laudable goal.

Please let me know if I'm misinterpreting the request.

a-palchikov:
We could start with a one-off. Ideally this would be tested automatically but I'm not sure how often and whether on each PR.

Issue #1915, opened 2020-07-23
Implement upgrade testing with the new cluster image


wadells:
### Problem
Currently our build system builds and persists telekube.tar artifacts that are consumed days/weeks later by robotest upgrade testing:

https://github.com/gravitational/gravity/blob/12ea7dbb1462fd93b040202399aefac20a240851/build.assets/robotest/run.sh#L45-L52

In order to do upgrade testing with a new image (#1908) we need a way to generate or retrieve appropriate tarballs for all versions upgraded to/from.

### Motivation
Upgrades tend to be our biggest pain point in the field.  A significant portion of the value from #1908 comes from testing upgrades. However including this work in the initial design delays the value provided by testing an application in our install, expand, and shrink test cases -- so I've split it out into
a next step task.

Issue #1909, opened 2020-07-23
Implement support for NFS provisioner in robotest cluster image


r0mant:
### Problem

Customers have complained in the past that NFS provisioning failed when upgrading between the versions due to a regression with missing package from planet: https://github.com/gravitational/planet/pull/687.

### Solution

We should enhance our Robotest test cluster image to deploy a NFS provisioner (so everything is contained within the cluster), and update the application to mount an NFS share.

NFS provisioner Helm chart: https://github.com/helm/charts/tree/master/stable/nfs-server-provisioner
https://github.com/kubernetes-incubator/external-storage/tree/master/nfs
https://medium.com/@myte/kubernetes-nfs-and-dynamic-nfs-provisioning-97e2afb8b4a9

### Success

- [ ] `robotest` cluster image deploys an NFS provisioner Helm chart alongside other charts.
  - [ ] Optionally, it can be made optional by using Helm values so it's not installed in every test.
- [ ] One of `robotest` cluster image applications (charts) requests an NFS persistent volume.
- [ ] Robotest verifies that the pod was created and NFS share was mounted correctly.

### Timeline

Best case: 7 days
Worst case: 14 days

tosih:
This should work well since we can test both NFS and StorageClass (we can mount an ebs volume) with this chart. We would need to add this:
```apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: standard
provisioner: kubernetes.io/aws-ebs
parameters:
  type: gp2
reclaimPolicy: Retain
allowVolumeExpansion: true
volumeBindingMode: Immediate```

wadells:
During testing: make sure we try this against a broken gravity version (7.0.9) and it fails.

PR #1967, opened 2020-08-05
Custom Robotest Image

### LINKED ISSUE #1908: Implement a new test cluster image for Robotest
### Problem

Currently, Robotest uses our base [telekube](https://github.com/gravitational/gravity/tree/master/assets/telekube) image to exercise tests. The problem with this approach is that we cannot extend this base image with additional scenarios, to keep it "vanilla".

### Solution

Instead of testing with base image, we should build another application that will be used specifically by Robotest to exercise all sorts of test scenarios and which we will be constantly iterating on and extending with new features and use-cases.

Let's call this new test cluster image `robotest` and place it under [assets](https://github.com/gravitational/gravity/tree/master/assets) in this repo, alongside other images there. We can base it off of `telekube` and iterate on it.

### Success

We will consider our first step on the path to a feature-rich Robotest application a success when the following is completed:

- [x] There's a cluster image called `robotest` under the [assets](https://github.com/gravitational/gravity/tree/master/assets) directory.
- [x] The `robotest` cluster image has the following properties (see [Image Manifest](https://gravitational.com/gravity/docs/pack/#image-manifest) for details on these):
  - [x] It defines multiple node profiles and different flavors, can base it off of telekube app.
    - [x] Similar to telekube app, there should be an "auto" profile (that auto-selects the role), "master" and "node".
    - [x] There should be a single-node flavor, an HA flavor (say, with 3 and 6 masters), and a flavor that contains regular nodes (say, 1 master + 1 node and 3 masters + 3 nodes).
  - [x] It contains a simple Helm chart. For the first iteration, something simple like alpine/nginx will work.
  - [x] It defines **all** hooks that are on the "execution path" of current Robotest scenarios: install, postInstall, preUpdate, update, postUpdate, status, preNodeAdd, postNodeAdd, preNodeRemove, postNodeRemove, etc. (check what else).
  - [x] Node profiles specify volume mounts requirements with custom uid/gid.
  - [x] It enables additional features like OpenEBS and nginx ingress controller.
- [x] Robotest uses this `robotest` cluster image to exercise all test suites except upgrade (upgrade tracked in #1915)

### Time estimate

Best case scenario: 5 days.
Worst case scenario: 10 days.

wadells:
## Description
This PR introduces a new cluster image `robotest.tar` for our automated integration testing.  The contents of the image are minimal, this largely the build & infra work to support an the new artifact.

Why? Gravity, like the kubernetes it wraps has a plethora of customization options.  Running it without any of these customization options set and without an application running inside the cluster is not realistic for our customers. We want to move towards more realistic integration scenarios.

See https://github.com/gravitational/gravity/issues/1908 for more context.

## Type of change

* Closes https://github.com/gravitational/gravity/issues/1908
* Related to https://github.com/gravitational/robotest/issues/160 and https://github.com/gravitational/robotest/issues/197

## TODOs
- [x] Self-review the change
- [x] Perform manual testing
- [ ] Write documentation
- [x] Address review feedback

## Implementation
I've kept the implementation minimal for the initial version, preferring an incremental approach to trying for a more substantial image, as a more waterfall-y design first strategy wasn't providing timely results.

Follow up work that is explicitly out of scope *for this PR* includes:
* Integrating with upgrade testing https://github.com/gravitational/gravity/issues/1915
* Implementing a NFS provisioner https://github.com/gravitational/gravity/issues/1909
* Allowing the contents to be tweaked in an ad-hoc manner https://github.com/gravitational/gravity/issues/1923
* 

## Testing done
TBD


Makefile
```diff
@@ -522,15 +520,35 @@ scan-artifacts: telekube
...
 	GRAVITY_K8S_VERSION=$(K8S_VER) $(GRAVITY_BUILDDIR)/tele build \
 		$(ASSETSDIR)/telekube/resources/app.yaml -f \
 		--version=$(TELEKUBE_APP_TAG) \
 		--state-dir=$(PACKAGES_DIR) \
 		--skip-version-check \
-		-o $(GRAVITY_BUILDDIR)/telekube.tar
+		-o $(TELEKUBE_OUT)
+
+$(GRAVITY_OUT):
+	$(MAKE) -C build.assets build
```
r0mant, line 534: Nit: This will build all binaries, not just `gravity`.

Makefile
```diff
@@ -652,63 +670,17 @@ wizard-gen:
...
-robotest-installer: EXPORT_DIR := $(LOCAL_STATE_DIR)/export
-robotest-installer: EXPORT_APP_TARBALL := $(EXPORT_DIR)/app.tar.gz
-robotest-installer: ROBOTEST_APP_PACKAGE ?= $(TELEKUBE_APP_PKG)
-robotest-installer: ROBOTEST_APP_PACKAGE_SRCDIR ?= $(TOP)/assets/telekube
-robotest-installer: ROBO_BUCKET_URL = s3://builds.gravitational.io/robotest
-robotest-installer: ROBO_GRAVITY_BUCKET := $(ROBO_BUCKET_URL)/gravity/$(GRAVITY_VERSION)
-robotest-installer: INSTALLER_FILE := $(subst /,-,$(subst :,-,$(ROBOTEST_APP_PACKAGE)))-$(GRAVITY_VERSION)-installer.tar.gz
-robotest-installer: INSTALLER_URL := $(ROBO_BUCKET_URL)/$(INSTALLER_FILE)
-robotest-installer: robotest-publish-gravity
-robotest-installer:
```
r0mant, line 671: Is this no longer used? How will robotest use the new application for install scenarios? IIRC currently it builds it locally?

assets/robotest/resources/app.yaml
```diff
@@ -0,0 +1,163 @@
+apiVersion: bundle.gravitational.io/v2
```
r0mant, line 1: "bundle" is the legacy version/kind, let's use new header:

```yaml
apiVersion: cluster.gravitational.io/v2
kind: Cluster
```

assets/robotest/resources/app.yaml
```diff
@@ -0,0 +1,163 @@
...
+        - path: /var/lib/gravity
+          capacity: "10GB"
+          filesystems: ["xfs", "ext4"]
+hooks:
+  install:
+    job: file://install.yaml
+  status:
+    job: file://status.yaml
+  postInstall:
+    job: file://status.yaml
```
r0mant, line 142: Is "status.yaml" used as a placeholder for other hooks for now?

assets/robotest/resources/app.yaml
```diff
@@ -0,0 +1,163 @@
...
+  postUpdate:
+    job: file://status.yaml
+# ingress:
+#   nginx:
+#     enabled: true
+storage:
+  openebs:
+    enabled: true
+systemOptions:
+  runtime:
```
r0mant, line 160: systemOptions.runtime is also an obsolete field, I'd just remove it.

And systemOptions.docker as well, overlay2 is the default driver.

So I'd remove the entire systemOptions section for now.

assets/robotest/resources/app.yaml
```diff
@@ -0,0 +1,163 @@
...
+          versions: ["8", "9"]
+        - name: suse
+          versions: ["12"]
+        - name: sles  # Suse Linux Enterprise Server
+          versions: ["12"]
+      volumes:
+        - path: /var/lib/gravity
+          capacity: "10GB"
+          filesystems: ["xfs", "ext4"]
+      volumes:
```
r0mant, line 97: I would add the same custom volume to other node profiles as well.

assets/robotest/resources/install.yaml
```diff
@@ -0,0 +1,22 @@
...
+   template:
+     metadata:
+       name: echoserver-install
+     spec:
+       restartPolicy: Never
+       containers:
+         - name: echoserver-install
+           image: quay.io/gravitational/debian-tall:buster
+           command:
+             - /usr/local/bin/helm
```
r0mant, line 16: I would also set a release name (`--name xxx`) so it can be used in "helm upgrade" command in the upgrade hook.

Makefile
wadells: This is unused for many years.  More context in the commit message:

https://github.com/gravitational/gravity/pull/1967/commits/dd357c87cab1303eb537ea4c175546df2995cf56

The `robotest-image` target produces the new installer tarball which will be used for all scenarios.  I don't have a strong opinion on the name.

Makefile
wadells: What is the correct way to build just `GRAVITY_OUT`?  I am having a deal of trouble figuring that out, because the binary targets are lumped together under a parallel dockerized make in `build-in-container` -- which I didn't really want to take as a pretty slow (2m22s on my system) `.PHONY` dependency.

This `$(GRAVITY_OUT)` target allows me to cheat and make sure a binary exists on the FS without taking the costly `.PHONY` dependency.  It is still prone to failure because it in turn doesn't depend on any of the upstreams that feed into `$(GRAVITY_OUT), and thus won't get rebuilt if a source changes -- but that is rampant in our makefiles and not a problem I want to tackle in this PR.

r0mant: Hmm, IIRC there's a `BINARIES` variable that sets the binaries to build so you could try to use something like

```
BINARIES=gravity $(MAKE) -C build.assets build
```

assets/robotest/resources/app.yaml
wadells: Removed in 79d539d.

assets/robotest/resources/app.yaml
wadells: Added in 82fef03.

Makefile
wadells: Fixed in beeb7c5.  This may lead to non-parallel rebuilds in some cases.

assets/robotest/resources/app.yaml
wadells: Fixed in 9bc4135.

Jenkinsfile
```diff
@@ -84,6 +84,7 @@ timestamps {
         robotest : {
           if (env.RUN_ROBOTEST == 'run') {
             withCredentials([
+                [$class: 'StringBinding', credentialsId: 'GET_GRAVITATIONAL_IO_APIKEY', variable: 'GET_GRAVITATIONAL_IO_APIKEY'],
```
r0mant, line 87: Still need this with https://github.com/gravitational/gravity/pull/2023?

r0mant (approved):

Jenkinsfile
wadells: This is no longer needed because of #1997. #2023 isn't relevant to this PR because it targets 7.0.x and upgrade-via which hasn't hit master yet (as far as I know).

There will be a bit of merge headache between https://github.com/gravitational/gravity/issues/1915 and the upgrade via work, but nothing out of the ordinary given our port strategy.

PR #1941, opened 2020-07-29
Remove `~jenkins/.gravity/config` shared state from robotest

wadells:
## Description
This PR replaces a `tele logout` with a custom `--state-directory` to achieve the same effect.

## Context
In this Gravity 6.1.32  Jenkins publish job, we saw publishing fail with the following error:

```
----> Publishing Gravity to https://get.gravitational.io...
/tmp/tmp.vfPmgoL1D0/gravity/e//build/6.1.32/gravity package delete --ops-url=https://get.gravitational.io  --force gravitational.io/gravity_linux_x86_64:6.1.32 && /tmp/tmp.vfPmgoL1D0/gravity/e//build/6.1.32/gravity package import --ops-url=https://get.gravitational.io  /tmp/tmp.vfPmgoL1D0/gravity/e//build/6.1.32/gravity gravitational.io/gravity_linux_x86_64:6.1.32
\u001b[31m[ERROR]: access denied to perform action 'delete' on repository
```

This is because several different parts of the build/release process rely on `~jenkins/.gravity/config`. Some want to be logged out (robotest), and some logged in (publishing artifacts). Exacerbating the issue, different builds on the same executor all share this single file leading to highly racy behavior.  This has been a theoretical risk since ce93812c3edcd4095afbf03373afad3841cd1750, but was moved from theoretical to practical by https://github.com/gravitational/gravity/pull/1890 and ports.

## Type of change
* Regression fix (non-breaking change which fixes a regression)
* Internal change (not necessarily a bug fix or a new feature)

## Linked tickets and other PRs
Fixes a release automation regression introduced in #1890 and ports.

## TODOs
- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback
- [ ] Delete uploaded test artifacts

## Testing done
From a previous version of the PR:
<details><summary><code>make publish-telekube</code></summary>

```
walt@work:~/git/gravity$ make publish-telekube   
make -C build.assets publish-telekube
make[1]: Entering directory '/home/walt/git/gravity/build.assets'
// snip build stuff ...
if [ -z "redacted" ] || [ -z "https://get.gravitational.io" ]; then \
   echo "TELE_KEY or DISTRIBUTION_OPSCENTER are not set"; exit 1; \
fi;
tele login --ops https://get.gravitational.io --key=redacted --state-dir=/home/walt/git/gravity//build/7.1.0-alpha.1.147/.publish
Ops Center:     get.gravitational.io
Username:       meta@gravitational.io
[...]

PR #1982, opened 2020-08-07
[7.0.x] custom planet containers

a-palchikov:
## Description

Forward port of https://github.com/gravitational/gravity/pull/1962.
Also fixes an issue with the upgrade unable to determine existing etcd server version if the planet container did not expose this information.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1940.

* Requires https://github.com/gravitational/planet/pull/723, https://github.com/gravitational/gravity.e/pull/4328.

* Ports https://github.com/gravitational/gravity/pull/1962.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

#### Prerequisites
1. Custom planet container replacing the default package:

```yaml
systemOptions:
  baseImage: quay.io/gravitational/planet:<custom-build>
```

2. Custom planet container for a (subset of) node profile

```yaml
nodeProfiles:
  - name: profile
    systemOptions:
      baseImage: quay.io/gravitational/planet:<custom-build>
```

#### Clusters

- [x] Installation, 3-node cluster, all control plane nodes
- [x] Installation, 3-node cluster, one control plane, 2 workers
- [x] Upgrade from a previous version (`6.1.33`), 3-node cluster
- [ ] Upgrade from a previous version, run gc and expand (both controller plane and workers)
- [ ] Installation, 1-node cluster,  then expand to 3-nodes (both controller plane and workers)


lib/install/phases/postsystem.go
```diff
@@ -164,7 +164,7 @@ type healthExecutor struct {
 func (p *healthExecutor) Execute(ctx context.Context) error {
 	p.Progress.NextStep("Waiting for the planet to start")
 	p.Info("Waiting for the planet to start.")
-	err := utils.Retry(defaults.RetryInterval, defaults.RetryAttempts,
+	err := utils.RetryWithInterval(ctx, defaults.ExponentialBackOff(),
```
a-palchikov, line 167: Found this issue that prevents timely abort of the installer during the planet status wait phase.

lib/update/cluster/builder_test.go
```diff
@@ -391,113 +391,50 @@ func (s *PlanSuite) TestPlanWithIntermediateRuntimeUpdate(c *check.C) {
 	})
 }
 
-func (s *PlanSuite) TestUpdatesEtcdFromManifestWithoutLabels(c *check.C) {
```
a-palchikov, line 394: Removed the test about assumed etcd version since it is expected that we have both installed and upgrade versions available.

build.assets/robotest/run.sh
```diff
@@ -27,6 +27,7 @@ export GCE_VM=${GCE_VM:-custom-4-8192}
 # Parallelism & retry, tuned for GCE
 export PARALLEL_TESTS=${PARALLEL_TESTS:-4}
 export REPEAT_TESTS=${REPEAT_TESTS:-1}
+GRAVITY_PULL_OPTIONS=${GRAVITY_PULL_OPTIONS:-""}
```
wadells, line 30: This doesn't appear to change anything for CI robotest runs -- I don't see a reference to this variable anywhere else in our Makefiles & toolchain. Is this for local testing?

Jenkinsfile
```diff
@@ -84,6 +84,7 @@ timestamps {
                 file(credentialsId:'ROBOTEST_LOG_GOOGLE_APPLICATION_CREDENTIALS', variable: 'GOOGLE_APPLICATION_CREDENTIALS'),
                 file(credentialsId:'OPS_SSH_KEY', variable: 'SSH_KEY'),
                 file(credentialsId:'OPS_SSH_PUB', variable: 'SSH_PUB'),
+                string(credentialsId:'GET_GRAVITATIONAL_IO_APIKEY', variable: 'BUILD_APIKEY'),
```
wadells, line 87: We shouldn't need this to fetch images, as far as I know. `tele pull` will download the correct artifacts when it isn't logged in to get.gravitational.io -- so long as it isn't logged in elsewhere either.

We should test both authenticated & unauthenticated fetches, but if we only do one, unauthenticated are more important to customers in my opinion.  Especially for the new user experience.

a-palchikov: Do we always logout? Since if there is login state (from an unrelated job), this will still use it (this is what's been happening here).

build.assets/robotest/run.sh
a-palchikov: It is coming from [outside](https://github.com/gravitational/gravity.e/pull/4328/files#diff-b67911656ef5d18c4ae36cb6741b7965R30)

lib/builder/builder.go
```diff
@@ -270,13 +246,9 @@ func (b *Builder) SyncPackageCache(ctx context.Context, runtimeVersion semver.Ve
...
 		return trace.Wrap(err)
 	}
-	syncer, err := b.NewSyncer(b)
-	if err != nil {
-		return trace.Wrap(err)
-	}
 	for _, runtimeVersion := range append([]semver.Version{runtimeVersion}, intermediateVersions...) {
 		b.NextStep("Syncing packages for %v", runtimeVersion)
-		if err := b.syncPackageCache(ctx, runtimeVersion, syncer, apps); err != nil {
+		if err := b.syncPackageCache(ctx, runtimeVersion, b.Syncer, apps); err != nil {
```
bernardjkim, line 251: nit: can get `b.Syncer` from inside `syncPackageCache`

lib/builder/syncer.go
```diff
@@ -41,122 +39,86 @@ type Syncer interface {
...
-// NewSyncer returns a new syncer instance for the provided builder
-//
-// Satisfies NewSyncerFunc type.
-func NewSyncer(b *Builder) (Syncer, error) {
-	return newS3Syncer()
-}
-
 // s3Syncer synchronizes local package cache with S3 bucket
-type s3Syncer struct {
+type S3Syncer struct {
```
bernardjkim, line 43: nit: Update godoc `s3Syncer` -> `S3Syncer`

Jenkinsfile
wadells: We did briefly, but that caused issues for publishing.

I think login is the way to go for now -- it is consistent with the old behavior.

I'm working on a patch that containerizes more of the robotest workflow such that `~jenkins` state cannot affect fetching robotest artifacts (and vice versa), but that is probably a day or two out -- it is a more substantial refactoring

a-palchikov: I have nothing against using login but it seems that publishing is a very focused task which is not necessary for PR jobs - I think explicit token configuration is as good as logging in in a per-task environment, no?

wadells: > I have nothing against using login but it seems that publishing is a very focused task which is not necessary for PR jobs

Publishing & PR shouldn't be related, but are due to sharing login state in `~jenkins`.  See https://jenkins.gravitational.io/view/Publishing/job/Telekube-Publish-5.2/301/console for an example failure

> I think explicit token configuration is as good as logging in in a per-task environment, no?

My goal for test purposes would be to fetch artifacts while not logged in.  I've yet to find a good way to use unauthenticated pulls in a way that is non

On a more holistic note, I want to isolate because sharing state in `~/` has proven to be prone to races and a large pain point.  This stolon/pithos CI breaking robotest topic is the 3rd notable disruption stemming from "spooky action at a distance" when unrelated parts of our CI/release infra breaking each other due to sharing state in `~/`.  See https://github.com/gravitational/gravity/pull/1941 and https://github.com/gravitational/gravity/pull/177 for the first two.

While I *think* I understand enough to not make mistakes regarding this shared state after having been burned by it twice, will the next employee who touches our release infra understand?  Why leave this edge around for them to trip on?

a-palchikov: I think we're not following each other. I'm not saying we should use shared state and using hub/token does not create shared (login) state - it is a way to use the login information at that particular point where it is applied - isn't this what we're after?

wadells: > I'm not saying we should use shared state and using hub/token does not create shared (login) state - it is a way to use the login information at that particular point where it is applied - isn't this what we're after?

A couple issues with the hub/token approach:

1) How do we apply this advice to 6.1 and 5.5 where the hub/token flags do not exist?  Those branches are subject to the same shared state modes of failure. 

2) How do we test unauthenticated fetch with the hub and token flags?  Related: How could open source users run any of these targets?

To be clear: I'm not asking you to change anything here in this PR.  We need an immediate mitigation and your PR provides this in the best way!  This discussion is about a long term plan.

wadells: The answer to all the above could be "product fixes & backports to all active releases".  But per our earlier discussion about "do unnecessary feature backports" I'm inclined to look at non-product fixes to our build architecture.  Lets fix the product in master, but use non-product solutions (could be sandboxing pushes/pulls in a container, could be something else) to fix the issue in older branches.  This reduces surface area for product regressions, especially since this wouldn't be work a customer cares about -- it is purely internal work.

knisbet (approved):

r0mant (commented):
I think that this massive builder refactoring makes this a pretty risky change esp. for an LTS release. My worry is we'll break one of our subtle tele build use-cases as we did in the past.

I would either see if it's possible to fix the custom planet problem with less of unnecessary refactoring, or extensively test all tele build use-cases (for both enterprise and open-source where applicable) to make sure nothing regresses:

- [ ] That when logged out, it pulls stuff from default hub.
- [ ] That when logged into a hub (both via "tele login" and "tsh login"), it uses that hub.
- [ ] That when --hub/--token/--repository provided, it uses the indicated hub.
- [ ] Behavior of --state-dir flag, should be source for both packages and credentials (when logged in via tele login --state-dir), and when used in conjunction with --hub/--token/--repository flags.

Maybe I'm forgetting something else...

a-palchikov:
Checked the following for enterprise builds:

- [x] That when logged out, it pulls stuff from default hub.
- [x] That when logged into a hub (both via "tele login" and "tsh login"), it uses that hub.
- [x] [!] That when --hub/--token/--repository provided, it uses the indicated hub. **This did not work when using --token/--repository combination - seems like --hub is always required**.
- [x] Behavior of --state-dir flag, should be source for both packages and credentials (when logged in via tele login --state-dir), and when used in conjunction with --hub/--token/--repository flags.

dorsany:
Hi @a-palchikov, do you think that this pr will be a part of the next release version of `7.0.x` ?

a-palchikov:
@dorsany I have a couple more things to take care as part of this PR that were discovered in the neighbor branch, but I hope it will.

lib/pack/localpack/packageserver.go
```diff
@@ -327,22 +327,14 @@ func (p *PackageServer) UpsertRepository(repository string, expires time.Time) e
...
-	if !trace.IsNotFound(err) {
-		return trace.Wrap(err)
-	}
 	repo := storage.NewRepository(repository)
 	if !expires.IsZero() {
 		repo.SetExpiry(expires)
 	}
-	_, err = p.backend.CreateRepository(repo)
-	if err != nil {
+	_, err := p.backend.CreateRepository(repo)
```
r0mant, line 334: Bit weird that the function is called "UpsertRepository" yet it only creates it. Prob. not a big deal since it's been like this forever, only just noticing.

lib/update/cluster/plan.go
```diff
@@ -425,26 +425,18 @@ func shouldUpdateCoreDNS(client *kubernetes.Clientset) (bool, error) {
...
 	// TODO: should somehow maintain etcd version invariant across runtime packages
-	runtimePackage, err := schema.GetDefaultRuntimePackage(installedRuntimeApp.Manifest)
-	if err != nil {
-		return nil, trace.Wrap(err, "error fetching runtime package for %v", installedRuntimeApp.Package)
-	}
 	var updateEtcd bool
-	installedVersion, err := getEtcdVersion(*runtimePackage, packages)
+	installedVersion, err := getEtcdVersion(installedRuntimePackage, packages)
 	if err != nil {
 		if !trace.IsNotFound(err) {
```
r0mant, line 433: Shouldn't "not found" also be a terminal scenario in this case? Since the etcd migrate operation is now basically copying old directory to new directory, the old one must always be present. Otherwise "installedVersion" below will be empty and we'll run into the same scenario as the customer S reported to us yesterday. Am I missing something?

lib/update/cluster/steps.go
```diff
@@ -71,7 +71,15 @@ func (r *phaseBuilder) initSteps(ctx context.Context) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	etcd, err := shouldUpdateEtcd(installedRuntimeApp, r.updateRuntimeApp, r.packages)
+	installedRuntimePackage, err := schema.GetDefaultRuntimePackage(r.installedApp.Manifest)
```
r0mant, line 74: Does this properly handle custom planets? It's just the name "Get**Default**RuntimePackage" gives the impression that it always uses the default planet.

r0mant (approved):
Left a couple of question, otherwise lgtm.

lib/update/cluster/plan.go
a-palchikov: It should be the other way around - the etcd data directory migration path needs to support the case of the empty source version and fallback to a version file as it used to be.

lib/update/cluster/steps.go
a-palchikov: Yeah, it is confusing. I think I'll drop the default part.

lib/pack/localpack/packageserver.go
a-palchikov: This should propagate to the storage interface which should offer `UpsertRepository` in the same vein. Not going to change that though in this PR.

lib/update/cluster/plan.go
```diff
@@ -475,6 +483,35 @@ func getEtcdVersion(locator loc.Locator, packageService pack.PackageService) (*s
...
 
+// getCurrentEtcdVersion attempts to read the version of etcd from the etcd version
+// file. This assumes that all cluster (master) nodes run the same version of etcd
+// although this might be invalidated with the use of custom planet image.
+// This does not make sense to have this descripancy though and it should be flagged
+// elsewhere - i.e. at 'tele build' time
+func getCurrentEtcdVersion(stateDir string) (version *semver.Version, err error) {
+	// https://github.com/gravitational/planet/blob/version/7.0.x/tool/planet/constants.go#L87
+	const etcdVersionKey = "PLANET_ETCD_VERSION"
+	f, err := os.Open(filepath.Join(state.EtcdDir(stateDir), "etcd-version.txt"))
```
r0mant, line 494: I'm curious, instead of relying on this file being present, why not just do `etcd --version` (or a respective API call)?

a-palchikov: This an excellent point - I've addressed it in https://github.com/gravitational/gravity/pull/1982/commits/7c19daac69f80530b0eaaf888083aa267ba14593

Issue #1931, opened 2020-07-28
Provide a way to make gravity-site not a load balancer


r0mant:
### Feature Request

The use-case is that a customer would like to prevent gravity-site service from creating a load balancer when deploying with cloud integration enabled.

In another scenario a customer wants to have more control over the service annotations e.g. to enable/disable internal load balancer.

### Proposed spec

Extend the existing [ClusterConfiguration](https://gravitational.com/gravity/docs/config/#general-cluster-configuration) resource with the gravityController section that allows to control certain parts of gravity-site service:

```yaml
kind: ClusterConfiguration
version: v1
spec:
  # global and kubelet are existing fields
  global:
  kubelet:
  # new field
  gravityControllerService:
    type: NodePort
    annotations:
      "cloud.google.com/load-balancer-type": "Internal"
```

* The "type" field should allow only `NodePort` or `LoadBalancer` (`ClusterIP` will break internal gravity stuff).
* Default values should be taken from the current gravity-site service:
  * "type" should be `LoadBalancer`.
  * "annotations" should contain two AWS-specific annotations (see service definition linked below).

### Behavior

- [ ] When this resource (with gravityControllerService section) is provided during install (via `--config` flag), gravity-site service should be created with those settings applied.
- [ ] If this resource (with gravityControllerService) is created/updated when the cluster is deployed, it should update gravity-site service accordingly (same as if it was updated using kubectl, no operation needed).
- [ ] During upgrade, these settings should also be considered and applied to the updated gravity-site service.

### Implementation details

The gravity-site service is currently statically defined among site-app resources here:

https://github.com/gravitational/gravity/blob/version/7.0.x/assets/site-app/resources/site.yaml#L276-L293

The change proposed in this ticket means that the service will now have to be created dynamically by install procedure and updated during an upgrade.

PR #1995, opened 2020-08-11
[WIP] Enable gravity-site service configuration through ClusterConfiguration resource

bernardjkim:
## Description

This PR enables the configuration of gravity-site service through the ClusterConfiguration resource. 

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1931

## TODOs

- [ ] Self-review the change
- [ ] Write tests
- [ ] Perform manual testing
- [ ] Write documentation
- [ ] Address review feedback

## Testing done

Default gravity-site config:
```
[vagrant@node-1 ~]$ sudo kubectl describe svc gravity-site -n kube-system
Name:                     gravity-site
Namespace:                kube-system
Labels:                   app=gravity-site
Annotations:              service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: 3600
                          service.beta.kubernetes.op/aws-load-balancer-internal: 0.0.0.0/0
Selector:                 app=gravity-site
Type:                     LoadBalancer
IP:                       100.100.42.132
Port:                     web  3009/TCP
TargetPort:               3009/TCP
NodePort:                 web  32009/TCP
Endpoints:                172.28.128.101:3009
Session Affinity:         None
External Traffic Policy:  Cluster
Events:                   <none>
```
cluster-config.yaml
```
kind: ClusterConfiguration
version: v1
spec:
  global:
    featureGates:
      AllAlpha: true
  gravityControllerService:
    type: NodePort
    annotations:
      "cloud.google.com/load-balancer-type": "Internal"
      "service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout": "4000"
```

**Verify gravity-site service configuration is updated when updating cluster configuration**
**Verify gravity-site service configuration is updated on install**
**Verify gravity-site service configuration is updated on upgrade**

### DIFF
--- a/lib/ops/opsservice/site.go
+++ b/lib/ops/opsservice/site.go
@@ -44,7 +44,6 @@ import (
 	"github.com/gravitational/trace"
 	"github.com/mailgun/timetools"
 	log "github.com/sirupsen/logrus"
-	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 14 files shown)


lib/storage/clusterconfig/clusterconfig.go
```diff
@@ -227,6 +227,8 @@ type Spec struct {
 type ComponentConfigs struct {
 	// Kubelet defines kubelet configuration
 	Kubelet *Kubelet `json:"kubelet,omitempty"`
+	// GravityControllerService defines gravity-site service configuration
+	GravityControllerService `json:"gravityControllerService,omitempty"`
```
a-palchikov, line 231: You will not be able to take advantage of `omitempty` unless this is a primitive with a known zero-value (incl. a pointer).

lib/storage/clusterconfig/clusterconfig.go
```diff
@@ -430,7 +446,23 @@ const specSchemaTemplate = `{
             },
             "extraArgs": {"type": "array", "items": {"type": "string"}}
           }
-        }
+		},
+		"gravityControllerService": {
```
a-palchikov, line 450: nit: this looks a bit off indentation-wise. I believe the rest of the YAML uses `ts=2 sts=2 sw=2 expandtab` (in vim terms) and this is using tabs instead of whitespace.

lib/ops/opsservice/site.go
```diff
@@ -604,6 +605,49 @@ func (s *site) getClusterEnvironmentVariables() (env storage.EnvironmentVariable
...
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	services := client.CoreV1().Services(defaults.KubeSystemNamespace)
+
+	svc, err := services.Get("gravity-site", metav1.GetOptions{})
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	svc.Spec.Type = v1.ServiceType(controllerConfig.Type)
```
r0mant, line 631: I think this should check whether the service already has correct type and annotations and do nothing in this case.

lib/ops/opsservice/site.go
```diff
@@ -604,6 +605,49 @@ func (s *site) getClusterEnvironmentVariables() (env storage.EnvironmentVariable
 	return env, nil
 }
 
+// updateServiceConfiguration updates the gravity-site service configuration if
+// the gravityControllerService configuration has been modified.
+func (s *site) updateServiceConfiguration() error {
```
r0mant, line 610: Seems like this function will always try to update the service as long as gravityController is present in the resource (even if it hasn't changed)? I think it should only be called if there really was a change in the gravity-site settings when a user updates the clusterconfig resource.

r0mant:
Closing this as it is no longer a priority.

Issue #1996, opened 2020-08-11
SELinux installation fails on centOS/redhat 7


a-palchikov:
### Description

**What happened**:
Gravity 7.0.x fails to install nodes with SELinux. The installer will get stuck at `/wait` step, and the container logs will have:

```
[RUNNING]: /usr/bin/gravity package command "start" "gravitational.io/planet:7.0.40-11709" "dev.test/planet-config-10128018devtest:7.0.40-11709"
Aug 11 16:44:51 node-0 gravity[15876]: standard_init_linux.go:211: exec user process caused "permission denied"
Aug 11 16:44:51 node-0 gravity[15876]: [ERROR]: failed to initialize container factory
Aug 11 16:44:51 node-0 gravity[15876]: standard_init_linux.go:211: exec user process caused "permission denied"
```

**What you expected to happen**:
Installation succeeds.

### Environment

 - Gravity version [e.g. 7.0.11]: 7.0.14
 - OS [e.g. Redhat 7.4]: CentOS Linux release 7.8.2003 (Core)
 - Platform [e.g. Vmware, AWS]: onprem

PR #2012, opened 2020-08-14
[7.0.x] SELinux update

a-palchikov:
## Description

This PR includes the fix for SELinux to make installs work again on rhel/CentOS 7. SELinux support is currently broken on `rhel 8` - this will be addressed in a future PR.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1996

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Write documentation
- [ ] Address review feedback

## Testing done

I've only been able to test installs/expands with the branch so upgrades from an older version where SELinux still works should be recommended.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ BANDWAGON_TAG ?= 6.0.1
 RBAC_APP_TAG := $(GRAVITY_TAG)
 TILLER_VERSION = 2.15.2
 TILLER_APP_TAG = 7.0.1
-SELINUX_VERSION ?= 6.0.0
+SELINUX_VERSION ?= 6.0.1
 # URI of Wormhole container for default install
 WORMHOLE_IMG ?= quay.io/gravitational/wormhole:0.3.3
 # set this to true if you want to use locally built planet packages
--- a/Makefile
+++ b/Makefile
@@ -381,7 +381,7 @@ ci:
 .PHONY: packages
 packages: planet-packages binary-packages teleport-package gravity-packages dns-packages\
 	rbac-app-package bandwagon-package tiller-package monitoring-package \
-	storage-package log-package k8s-packages telekube-packages selinux-policy-package
+	storage-package log-package k8s-packages telekube-packages
 
 .PHONY: teleport-package
 teleport-package:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -382,7 +381,7 @@ ci:
 .PHONY: packages
 packages: planet-packages binary-packages teleport-package gravity-packages dns-packages\
 	rbac-app-package bandwagon-package tiller-package monitoring-package \
-	storage-package log-package k8s-packages telekube-packages selinux-policy-package
+	storage-package log-package k8s-packages telekube-packages
```
a-palchikov, line 384: I removed the policy package since this was never implemented. Will re-add once I've settled on the best way to deal with it.

lib/report/timeline.go
```diff
@@ -38,6 +39,8 @@ func (r TimelineCollector) Collect(ctx context.Context, reportWriter FileWriter,
 		return trace.Wrap(err)
 	}
 	defer w.Close()
+	ctx, cancel := context.WithTimeout(ctx, defaults.StatusCollectionTimeout)
```
a-palchikov, line 42: During testing, I've been hitting the issues with abort when the planet container has not yet initialized properly (and hence the agent was not running). It does not really make sense to wait for the agent indefinitely (or for the lifetime of the parent context which defaults to 10min).

r0mant (approved):

build.assets/Makefile
```diff
@@ -191,9 +191,9 @@ endif
 # generate gRPC code
 #
 .PHONY: grpc
-grpc: buildbox $(GRPC_PROTO_OUTPUTS)
+grpc: $(GRPC_PROTO_OUTPUTS)
 
-$(GRPC_PROTO_OUTPUTS): $(GRPC_PROTOS)
+$(GRPC_PROTO_OUTPUTS): buildbox $(GRPC_PROTOS)
```
a-palchikov, line 196: Moved the `buildbox` dependency to the list of dependencies for the proto output files - if left on the `grpc` target, the two dependencies might be built in parallel (`buildbox` and `$GRPC_PROTO_OUTPUTS`) while the latter depends on the first which leads to inadvertent modifications to the generated protobuf sources.

PR #2017, opened 2020-08-17
Verify agents are active before resuming upgrade

bernardjkim:
## Description

This change will make sure that all upgrade agents are online before resuming an upgrade or before rolling back an upgrade. Resume/Rollback operations will attempt to re-deploy agents if any agents are offline.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1667

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

**Verify upgrade resume re-deploys agents if some are offline**
- Start upgrade.
- Shutdown agents with `./gravity agent shutdown`.
- Resume upgrade with `./gravity upgrade --resume`.
- Verify agents are re-deployed and upgrade resumes.

**Verify rollback re-deploys agents if some are offline**
- Start upgrade.
- Shutdown agents with `./gravity agent shutdown`.
- Rollback upgrade with `/./gravity rollback`.
- Verify agents are re-deployed and rolls back upgrade.

### DIFF
--- a/tool/gravity/cli/clusterupdate.go
+++ b/tool/gravity/cli/clusterupdate.go
@@ -235,8 +235,7 @@ func executeUpdatePhaseForOperation(env *localenv.LocalEnvironment, environ Loca
 // executeOrForkPhase either directly executes the specified operation phase,
 // or launches a one-shot systemd service that executes it in the background.
 func executeOrForkPhase(env *localenv.LocalEnvironment, updater updater, params PhaseParams, operation ops.SiteOperation) error {
-	// "/" PhaseID indicates a resume operation. Verify all agents are active before resuming.
-	if params.PhaseID == "/" {
+	if params.isResume() {
 		if err := verifyAgentsActive(env); err != nil {
 			return trace.Wrap(err)
 		}
--- a/tool/gravity/cli/clusterupdate.go
+++ b/tool/gravity/cli/clusterupdate.go
@@ -214,18 +214,21 @@ func executeUpdatePhase(env *localenv.LocalEnvironment, environ LocalEnvironment
 	if operation.Type != ops.OperationUpdate {
 		return trace.NotFound("no active update operation found")
 	}
+
+	statusList, err := collectAgentStatus(env)
+	if err != nil {
+		return trace.Wrap(err, "failed to collect agent status")
+	}
+
+	if !statusList.AgentsActive() {
+		env.Println(statusList.String())
+		return trace.BadParameter("some agents are offline; ensure all agents are deployed with `gravity agent deploy`")
+	}
+
 	return executeUpdatePhaseForOperation(env, environ, params, operation.SiteOperation)
 }
 
 func executeUpdatePhaseForOperation(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, params PhaseParams, operation ops.SiteOperation) error {
-	allActive, err := verifyActiveAgents(env)
-	if err != nil {
-		return trace.Wrap(err, "failed to verify agent status")
-	}
-	if !allActive {
-		return trace.BadParameter("some agents are offline; ensure all agents are deployed with `gravity agent deploy`")
-	}
-
 	updateEnv, err := environ.NewUpdateEnv()
 	if err != nil {
 		return trace.Wrap(err)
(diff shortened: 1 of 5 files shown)


tool/gravity/cli/clusterupdate.go
```diff
@@ -217,6 +217,14 @@ func executeUpdatePhase(env *localenv.LocalEnvironment, environ LocalEnvironment
 }
 
 func executeUpdatePhaseForOperation(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, params PhaseParams, operation ops.SiteOperation) error {
+	allActive, err := verifyActiveAgents(env)
+	if err != nil {
+		return trace.Wrap(err, "failed to verify agent status")
+	}
+	if !allActive {
+		return trace.BadParameter("some agents are offline; ensure all agents are deployed with `gravity agent deploy`")
```
r0mant, line 225: Any way to just show the same table as "agent status" command shows to the user doesn't have to guess which are offline?

tool/gravity/cli/clusterupdate.go
```diff
@@ -217,6 +217,14 @@ func executeUpdatePhase(env *localenv.LocalEnvironment, environ LocalEnvironment
 }
 
 func executeUpdatePhaseForOperation(env *localenv.LocalEnvironment, environ LocalEnvironmentFactory, params PhaseParams, operation ops.SiteOperation) error {
+	allActive, err := verifyActiveAgents(env)
```
r0mant, line 220: Also, I don't think this check should be placed here. This function is invoked for every manual phase invocation e.g. using "gravity plan execute" command. I think this check should only be executed when a user does "gravity resume" or "gravity rollback".

bernardjkim: 54103b0 Does this look right?

`gravity plan resume` can still be executed if an agent is offline. `gravity upgrade resume` and now `gravity rollback` will fail if an agent is offline.

bernardjkim:
Now that we are verifying the agent status before resuming upgrades and rollbacks, I'm wondering if users might run into issues due to the status requests timing out. I've updated the status query timeout from 5 sec to 10 sec and it is still configurable by setting an env variable. Is this okay as is? Should we change the implementation so that the status is updated on a loop like how satellite handles status?

tool/gravity/cli/clusterupdate.go
```diff
@@ -213,6 +213,17 @@ func executeUpdatePhase(env *localenv.LocalEnvironment, environ LocalEnvironment
 	if operation.Type != ops.OperationUpdate {
 		return trace.NotFound("no active update operation found")
 	}
+
+	statusList, err := collectAgentStatus(env)
```
r0mant, line 217: So this will also be executed not just for resume, but for any individual phase execution as well (i.e. when you do `gravity plan execute --phase=xxx`), which is probably not what we want. If you look at how resume is invoked, it sets the `PhaseID` to `/` which indicates it is resume being requested: https://github.com/gravitational/gravity/blob/version/5.5.x/tool/gravity/cli/run.go#L430. I would make sure that this agents check is only executed for `gravity plan resume` and not for `gravity plan execute --phase=`.

tool/gravity/cli/operation.go
```diff
@@ -182,6 +182,17 @@ func rollbackPlan(localEnv *localenv.LocalEnvironment, environ LocalEnvironmentF
 	default:
 		return trace.BadParameter(unsupportedRollbackWarning, op.TypeString())
 	}
+
+	statusList, err := collectAgentStatus(localEnv)
```
r0mant, line 186: This code block is basically identical to the one in the "resume", can we factor it out into a helper method?

tool/gravity/cli/operation.go
```diff
@@ -182,6 +182,17 @@ func rollbackPlan(localEnv *localenv.LocalEnvironment, environ LocalEnvironmentF
...
+	if err != nil {
+		return trace.Wrap(err, "failed to collect agent status")
+	}
+
+	if !statusList.AgentsActive() {
+		localEnv.Println(statusList.String())
+		return trace.BadParameter("some agents are offline; ensure all agents are deployed with `gravity agent deploy`")
```
r0mant, line 193: I would update the message to say `./gravity agent deploy` to indicate that it should be run with the binary from the upgrade tarball.

r0mant (approved):

Issue #2031, opened 2020-08-19
Upgrades of custom planet containers fails: [ERROR]: REPLACE_ETCD_LATEST_VERSION is not in dotted-tri format


knisbet:
### Description

**What happened**:

When performing upgrade, error appears:
```
Encountered error, will shutdown agents
Shutting down the agents
[ERROR]: REPLACE_ETCD_LATEST_VERSION is not in dotted-tri format
```

**What you expected to happen**:

The upgrade to execute

**How to reproduce it (as minimally and precisely as possible)**:

Unknown. I've tried to replicate the issue reported but failed for an unknown reason.

### Environment

 - Gravity version [e.g. 7.0.11]: 6.1.35
 - OS [e.g. Redhat 7.4]: N/A
 - Platform [e.g. Vmware, AWS]: N/A

**Relevant Debug Logs If Applicable**
```
2020-08-18T22:40:27Z WARN             Operation failed. error:[
ERROR REPORT:
Original Error: *errors.errorString REPLACE_ETCD_LATEST_VERSION is not in dotted-tri format
Stack Trace:
        /gopath/src/github.com/gravitational/gravity/lib/update/cluster/builder.go:684 github.com/gravitational/gravity/lib/update/cluster.getEtcdVersion
        /gopath/src/github.com/gravitational/gravity/lib/update/cluster/builder.go:643 github.com/gravitational/gravity/lib/update/cluster.shouldUpdateEtcd
        /gopath/src/github.com/gravitational/gravity/lib/update/cluster/plan.go:360 github.com/gravitational/gravity/lib/update/cluster.newOperationPlan
        /gopath/src/github.com/gravitational/gravity/lib/update/cluster/plan.go:196 github.com/gravitational/gravity/lib/update/cluster.NewOperationPlan
        /gopath/src/github.com/gravitational/gravity/lib/update/cluster/plan.go:91 github.com/gravitational/gravity/lib/update/cluster.InitOperationPlan
        /gopath/src/github.com/gravitational/gravity/tool/gravity/cli/clusterupdate.go:426 github.com/gravitational/gravity/tool/gravity/cli.clusterInitializer.newOperationPlan
        /gopath/src/github.com/gravitational/gravity/tool/gravity/cli/update.go:106 github.com/gravitational/gravity/tool/gravity/cli.newUpdater
[...]

PR #2032, opened 2020-08-19
[6.1] fallback on corrupted planet metadata

knisbet:
## Description

This patch uses the previous method of determining whether to upgrade etcd as a fallback method when reading corrupted metadata from planet. The metadata stored within /etc/planet/orbit.manifest.json didn't have the sed applied to it, leading to many versions of planet to return the string REPLACE_ETCD_LATEST_VERSION instead of the actual version. In this case, we fallback to trying to determine the version from the default runtime package, which would be the original shipped planet image which has the correct metadata.

## Type of change

Updates https://github.com/gravitational/gravity/issues/2031

## TODOs

- [x] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback

## Implementation

I basically used https://github.com/gravitational/gravity/pull/1962 as the reference, to use the previous method of determining whether to upgrade etcd as a fallback.

## Testing done

I haven't tried to test this yet, as yesterday when I was trying I wasn't able to reproduce the problem, and I'm not exactly sure why. I'll make sure this passes robotest, but we might need to release it for the customer experiencing this error to verify.

### DIFF
--- a/lib/update/cluster/builder.go
+++ b/lib/update/cluster/builder.go
@@ -637,16 +637,22 @@ func supportsTaints(gravityPackage loc.Locator) (supports bool, err error) {
 	return defaults.BaseTaintsVersion.Compare(*ver) <= 0, nil
 }
 
-func shouldUpdateEtcd(installedRuntimePackage, updateRuntimePackage loc.Locator, packages pack.PackageService) (etcd *etcdVersion, err error) {
+func shouldUpdateEtcd(installedRuntimePackage, updateRuntimePackage, altInstalledRuntimePackage loc.Locator, packages pack.PackageService) (etcd *etcdVersion, err error) {
 	// TODO: should somehow maintain etcd version invariant across runtime packages
 	var updateEtcd bool
 	installedVersion, err := getEtcdVersion("version-etcd", installedRuntimePackage, packages)
 	if err != nil {
-		if !trace.IsNotFound(err) {
+		// Workaround for https://github.com/gravitational/gravity/issues/2031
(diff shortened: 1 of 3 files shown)


r0mant (approved):

lib/update/cluster/builder.go
```diff
@@ -637,16 +637,22 @@ func supportsTaints(gravityPackage loc.Locator) (supports bool, err error) {
...
 	// TODO: should somehow maintain etcd version invariant across runtime packages
 	var updateEtcd bool
 	installedVersion, err := getEtcdVersion("version-etcd", installedRuntimePackage, packages)
 	if err != nil {
-		if !trace.IsNotFound(err) {
+		// Workaround for https://github.com/gravitational/gravity/issues/2031
+		// If we loaded a corrupted version of the planet version information, fallback to the default runtime package
+		// on versions where this was included.
+		// TODO(knisbet): Obsolete when no longer supporting upgrades from 7.0
+		if strings.Contains(err.Error(), "REPLACE_ETCD_LATEST_VERSION") && !altInstalledRuntimePackage.IsEmpty() {
```
r0mant, line 649: Wouldn't it be more reliable to just always fall back to the default runtime package?

knisbet: Probably, my thought is that would potential hide unknown forms of corruption. So the narrow approach allows us to target the bug we know about, and avoid masking related problems we might not know about (WDKWWDK).

knisbet:
Tested briefly on customer environment using unmatched version. It appears to progress.

PR #2049, opened 2020-08-22
(5.5) Add check for http(-s) proxy format

### LINKED ISSUE #2021: Validate that HTTP(S) proxy format is valid
### Feature Request

When a user sets an `HTTP(S)_PROXY` via a RuntimeEnvironment resource, Gravity should validate that the proxy format is correct i.e. that the `http(s)://` scheme is present as some services require it and don't work if scheme is not present.

### Motivation

Customer reported misconfigured proxy led to some upgrade problems. In the specific case there was an issue due to malformed configuration. 

### Who's it for?
OSS User, Enterprise

r0mant:
## Description

This pull request makes sure that RuntimeEnvironment resource is validated, specifically that `HTTP_PROXY` and `HTTPS_PROXY` environment variables, when set, are valid URLs and have corresponding scheme.

There are really two parts to the fix:

* Validate `HTTP(-S)_PROXY` variables as mentioned above (case-insensitive) when user creates/updates RuntimeEnvironment resource.
* Add a pre-upgrade check to perform the same validation on the existing resource, to catch potentially invalid configurations before upgrade and let the user fix it (we've run into such problems before).

## Type of change

* Refs https://github.com/gravitational/gravity/issues/2021.

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

Install 5.5.40 and create a RuntimeEnvironment resource with invalid proxies format:

```yaml
kind: RuntimeEnvironment
version: v1
spec:
  data:
    HTTP_PROXY: abcdge
    HTTPS_PROXY: qweasd
    NO_PROXY: 0.0.0.0/0,*
```

```
ubuntu@node-1:~/resources$ sudo gravity resource get runtimeenvironment
Environment
-----------
HTTPS_PROXY=qweasd
HTTP_PROXY=abcdge
NO_PROXY=0.0.0.0/0,*
```

Try to upgrade to this branch and see a warning:

```
Fri Aug 21 23:36:07 UTC	Upgrading cluster from 5.5.40 to 5.5.52-dev.4
[ERROR]: There was an issue detected with runtime environment variables:

    "HTTP_PROXY URL must include http:// scheme, HTTPS_PROXY URL must include https:// scheme"

This may cause problems during the upgrade. Please review configured environment
variables using "gravity resource get runtimeenvironment" command and update it
appropriately before proceeding with the upgrade:

kind: runtimeenvironment
metadata:
  name: runtimeenvironment
spec:
  data:
    HTTP_PROXY: abcdge
    HTTPS_PROXY: qweasd
    NO_PROXY: 0.0.0.0/0,*
version: v1

See https://gravitational.com/gravity/docs/config/#runtime-environment-variables
for more information on managing runtime environment variables.

This warning can be bypassed by providing a --force flag to the upgrade command.
```

Fix the resource, upgrade and try to update to the incorrect resource again:

```
ubuntu@node-1:~/resources$ sudo gravity resource create runtimeenv.yaml
[ERROR]: HTTP_PROXY URL must include http:// scheme: "123s123", HTTPS_PROXY URL must include https:// scheme: "http://qweasd:20"
```

## Additional information

Needs forward-ports.

### DIFF
--- a/lib/storage/environment.go
+++ b/lib/storage/environment.go
@@ -19,12 +19,12 @@ package storage
 import (
 	"encoding/json"
 	"fmt"
-	"net/url"
 	"strings"
 	"time"
 
 	"github.com/gravitational/gravity/lib/constants"
 	"github.com/gravitational/gravity/lib/defaults"
+	"github.com/gravitational/gravity/lib/utils"
 
 	teleservices "github.com/gravitational/teleport/lib/services"
 	teleutils "github.com/gravitational/teleport/lib/utils"
@@ -119,34 +119,26 @@ func (r *EnvironmentV1) CheckAndSetDefaults() error {
 	return trace.NewAggregate(errors...)
 }
 
-// checkHTTPProxy verifies HTTP_PROXY (if present) is valid a URL and has http:// scheme.
+// checkHTTPProxy verifies HTTP_PROXY (if present) is a valid proxy URL.
 func (r *EnvironmentV1) checkHTTPProxy() error {
 	httpProxy := r.getVariable(constants.HTTPProxyEnvVar)
 	if httpProxy == "" {
 		return nil
 	}
-	parsed, err := url.Parse(httpProxy)
-	if err != nil {
-		return trace.Wrap(err, "HTTP_PROXY must be a valid URL: %q", httpProxy)
-	}
-	if parsed.Scheme != "http" {
-		return trace.BadParameter("HTTP_PROXY URL must include http:// scheme: %q", httpProxy)
+	if _, err := utils.ParseProxy(httpProxy); err != nil {
+		return trace.Wrap(err, "failed to parse HTTP_PROXY")
 	}
 	return nil
 }
 
-// checkHTTPSProxy verifies HTTPS_PROXY (if present) is valid a URL and has https:// scheme.
+// checkHTTPSProxy verifies HTTPS_PROXY (if present) is a valid proxy URL.
 func (r *EnvironmentV1) checkHTTPSProxy() error {
 	httpsProxy := r.getVariable(constants.HTTPSProxyEnvVar)
 	if httpsProxy == "" {
 		return nil
 	}
-	parsed, err := url.Parse(httpsProxy)
-	if err != nil {
-		return trace.Wrap(err, "HTTPS_PROXY must be a valid URL: %q", httpsProxy)
-	}
-	if parsed.Scheme != "https" {
-		return trace.BadParameter("HTTPS_PROXY URL must include https:// scheme: %q", httpsProxy)
+	if _, err := utils.ParseProxy(httpsProxy); err != nil {
+		return trace.Wrap(err, "failed to parse HTTPS_PROXY")
 	}
 	return nil
 }
--- a/lib/storage/environment.go
+++ b/lib/storage/environment.go
@@ -19,6 +19,8 @@ package storage
 import (
 	"encoding/json"
 	"fmt"
+	"net/url"
+	"strings"
 	"time"
 
 	"github.com/gravitational/gravity/lib/constants"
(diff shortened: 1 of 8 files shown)


lib/storage/environment.go
```diff
@@ -107,9 +109,60 @@ func (r *EnvironmentV1) GetKeyValues() map[string]string {
...
+	if err := r.checkHTTPProxy(); err != nil {
+		errors = append(errors, err)
+	}
+	if err := r.checkHTTPSProxy(); err != nil {
+		errors = append(errors, err)
+	}
+	return trace.NewAggregate(errors...)
+}
+
+// checkHTTPProxy verifies HTTP_PROXY (if present) is valid a URL and has http:// scheme.
```
bernardjkim, line 122: Typo: is **a** valid URL ~a~. Same for `checkHTTPSProxy` below.

lib/storage/environment.go
```diff
@@ -107,9 +109,60 @@ func (r *EnvironmentV1) GetKeyValues() map[string]string {
...
+func (r *EnvironmentV1) checkHTTPProxy() error {
```
knisbet, line 123: These validation rules don't appear to match atleast golang validation logic: https://github.com/golang/go/blob/master/src/vendor/golang.org/x/net/http/httpproxy/proxy.go#L149-L170

1. it appears the scheme isn't required, and defaults to http.
2. socks5 isn't considered a valid scheme.
3. it's valid to use an http proxy for https traffic. And an https proxy for http traffic.

Also I'm actually confused about the originally stated problem now, since it looks like the golang validation logic just drops any parsing errors. I must be missing something.
https://github.com/golang/go/blob/master/src/vendor/golang.org/x/net/http/httpproxy/proxy.go#L211-L216

knisbet (changes requested):

lib/storage/environment.go
r0mant: @knisbet I thought that the original problem was that some software didn't like the fact that proxy URL is missing the scheme?

r0mant: @knisbet I've relaxed the requirements to only check for the presence of a valid scheme (http/https/socks5) in the provided proxy address to behave more like the Go parse function. Unlike the Go parser though, I have decided to make the scheme mandatory rather than assuming "http" - I think it may cause more trouble than help, as after some experimentation is looks like that `url.Parse()` will parse almost anything without an error, so if we try to prepend "http" everywhere, it will happily let some bogus address through as well. So, I think that having a bit stricter requirements to have scheme explicitly will help us filter out potentially bad addresses better.

lib/storage/environment.go
```diff
@@ -107,9 +109,52 @@ func (r *EnvironmentV1) GetKeyValues() map[string]string {
...
+		return nil
+	}
+	if _, err := utils.ParseProxy(httpProxy); err != nil {
+		return trace.Wrap(err, "failed to parse HTTP_PROXY")
+	}
 	return nil
 }
 
+// checkHTTPSProxy verifies HTTPS_PROXY (if present) is a valid proxy URL.
+func (r *EnvironmentV1) checkHTTPSProxy() error {
```
knisbet, line 135: Nit: I think checkHTTPProxy / checkHTTPSProxy are largely the same function now. They're not long, but usually I get called out for having anything that looks like duplicate code.

lib/storage/environment.go
knisbet: > @knisbet I thought that the original problem was that some software didn't like the fact that proxy URL is missing the scheme?

I thought so as well, just I don't see how now. I suppose unless it's failing somewhere later down the line, not in parsing the proxy, but when it actually tries to use it, throws an unexpected error. In which case would a valid proxy that isn't working possibly cause problems for docker?

> Unlike the Go parser though, I have decided to make the scheme mandatory rather than assuming "http" - I think it may cause more trouble than help, as after some experimentation is looks like that url.Parse() will parse almost anything without an error, so if we try to prepend "http" everywhere, it will happily let some bogus address through as well.

I dunno. I think our objective on this one is not to crash on invalid configuration. But introducing requirements beyond what is normally accepted does create some of it's own risks. Does mandating the scheme really make the url.Parse method more strict, or will it just require a bogus address to contain a scheme?

r0mant: Yeah it's a bit tough without actually seeing the logs of the alleged Docker crash due to the missing scheme.

You're right that mandating the scheme doesn't really make url.Parse() more strict (if someone wants to break something, they will), but I do think that it reduces the chance of accidental error in user's proxy configuration.

Just looking online, pretty much every single example of http_proxy includes an explicit scheme - the fact that Go/curl default to http seems more of an "assumption" (and not a great one if you ask me) and requiring the scheme would also protect us in case some software does actually require it.

lib/storage/environment.go
```diff
@@ -107,9 +109,41 @@ func (r *EnvironmentV1) GetKeyValues() map[string]string {
 
 // CheckAndSetDefaults validates this resource and sets defaults
 func (r *EnvironmentV1) CheckAndSetDefaults() error {
+	var errors []error
+	if err := r.checkProxy(constants.HTTPProxyEnvVar); err != nil {
```
knisbet, line 113: One other Nitpick I just thought of (sorry). The golang code and I think other implementations accept both lower and upper case versions of the environment variable. 

https://github.com/golang/go/blob/master/src/vendor/golang.org/x/net/http/httpproxy/proxy.go#L93-L94

knisbet (approved):

lib/storage/environment_test.go
```diff
@@ -30,6 +32,72 @@ type EnvS struct{}
...
+		{
+			vars:    `{}`,
+			error:   false,
+			comment: "empty resource is ok",
+		},
+	}
+	for _, test := range testCases {
+		data := fmt.Sprintf(`{"kind": "runtimeenvironment", "metadata": {"name": "foo"}, "version": "v1", "spec": {"data": %v}}`, test.vars)
+		env, err := UnmarshalEnvironmentVariables([]byte(data))
+		c.Assert(err, IsNil)
```
a-palchikov, line 90: nit: c.Assert(err, IsNil, Commentf(test.comment))  // include comment

PR #2066, opened 2020-08-27
Do not set priorityClassName for hooks

UnderGreen:
----

## Description

Reverting previous changes for hooks. cluster-critical priority could
be set only for kube-system resources and this could break existing
pre-update hooks for applications.

## Type of change

## TODOs

- [x] Perform manual testing
- [ ] Address review feedback

### DIFF
--- a/lib/app/hooks/configure.go
+++ b/lib/app/hooks/configure.go
@@ -129,11 +129,6 @@ func configureMetadata(job *batchv1.Job, p Params) error {
 		job.Spec.Template.Spec.SecurityContext = defaults.HookSecurityContext()
 	}
 
-	// if priorityClassName is not specified, set the default one
-	if job.Spec.Template.Spec.PriorityClassName == "" {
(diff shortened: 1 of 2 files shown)


r0mant (commented):
Can you explain how this breaks stuff?

> cluster-critical priority could be set only for kube-system resources

What happens if it's set on resources in other namespaces? Cause we also set them on monitoring components IIRC.

UnderGreen:
For monitoring components, we set their own priorities declared in monitoring app resources. 
> Can you explain how this breaks stuff?
Any hooks in non-kube-system namespaces will fail and they must be rewritten to use kube-system namespace, which is an unexpected change from a user perspective in the middle of LTS release cycle.
We could leave it light that for master branch, but for the LTS branches I would like to remove this change for now.

Also setting our own priority class will not work for pre-update hook which I explored as another solution for the problem.

r0mant (approved):
Okay, that makes sense. Yeah, I would remove this assumption from everywhere (incl. master) as I think that hook don't necessarily have to run in kube-system.

PR #2070, opened 2020-08-27
Do not set priorityClassName for hooks

UnderGreen:
----

## Description

Reverting previous changes for hooks. cluster-critical priority could
be set only for kube-system resources and this could break existing
pre-update hooks for applications.

## Type of change

## TODOs

- [x] Perform manual testing


r0mant (approved):

knisbet:
Should it be removed entirely, or should we just have a different priority class that we're creating on clusters that is still relatively high priority to allow for preemption but below system critical?

UnderGreen:
@knisbet as I mentioned in the previous Pr for 7.0.x it is impossible to set our own priority class because it will break pre-update hook when class does not exist yet in old version. Technically it is possible if we somehow skip usage of custom priority just for that hook.

a-palchikov (approved):

knisbet:
Or it could be coded to check for the priority class, or to try and create it and continue if it already exists.

PR #2071, opened 2020-08-27
[6.1.x] Verify agents are active before upgrade resume or rollback

bernardjkim:
## Description

This change will make sure that all upgrade agents are online before resuming an upgrade or before rolling back an upgrade.

## Type of change

* Ports https://github.com/gravitational/gravity/pull/2017

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

**Verify upgrade and rollback will re-deploy agents if some are offline**

### DIFF
--- a/tool/gravity/cli/clusterupdate.go
+++ b/tool/gravity/cli/clusterupdate.go
@@ -173,6 +173,7 @@ func executeUpdatePhase(env *localenv.LocalEnvironment, environ LocalEnvironment
 	if operation.Type != ops.OperationUpdate {
 		return trace.NotFound("no active update operation found")
 	}
+
 	return executeUpdatePhaseForOperation(env, environ, params, operation.SiteOperation)
 }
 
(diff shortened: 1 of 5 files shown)


tool/gravity/cli/clusterupdate.go
```diff
@@ -196,6 +197,12 @@ const gravityResumeServiceName = "gravity-resume.service"
 // executeOrForkPhase either directly executes the specified operation phase,
 // or launches a one-shot systemd service that executes it in the background.
 func executeOrForkPhase(env *localenv.LocalEnvironment, updater updater, params PhaseParams, operation ops.SiteOperation) error {
+	if params.isResume() {
+		if err := verifyAgentsActive(env); err != nil {
```
a-palchikov, line 201: I would try to attempt to start agents automatically and only report errors if this is not possible - less friction for a user.

a-palchikov (commented):
I think this should simply start the agents if possible to avoid having to burden the user with actions that can be performed automatically. If there's a problem deploying the agents then fail, show error message and let the user check the environment.

tool/gravity/cli/clusterupdate.go
bernardjkim: Good suggestion :+1:. Updated in c1b4e6b

r0mant (approved):

PR #2083, opened 2020-09-01
upgraded to new docs theme

alexwolfe:
## Description
I updated the docs templates to match with the upgraded Teleport docs template. This will also fix many of our SEO issues including js/css that was not minified. 

## Type of change
This is mostly a cosmetic change but will have a impact on SEO


docs/theme/toc.html
```diff
@@ -17,7 +17,7 @@ <h3 class="side-nav-section-header">
                   <a class="{% if nav_item.active%}is-active{%endif%}" href="{{ nav_item.url|url }}">{{ nav_item.title }}</a>
                 </li>
               </ul>
-              {% if nav_item.active%}
+              <!-- {% if nav_item.active%}
```
a-palchikov, line 20: if this is not needed, I'd rather remove it than comment it.

a-palchikov (approved):
hard to validate the change - if things are visual, would be helpful to see the differences but otherwise - lgtm

docs/theme/toc.html
alexwolfe: I'm keeping this around as a reference when we expand the navigation again. I'll remove all old stuff when we upgrade to the new design.

r0mant (approved):

benarent (approved):
Tested locally and everything looks good.

PR #2095, opened 2020-09-02
adjust pod subnet defaults

knisbet:
## Description

- Updates default pod subnet to be 1023 /24's (flannel skips the first subnet). 
- Updates minimum subnet size to be a /22 (3 nodes)
  - This request has come up several times. The proper check should probably be a combination of app.yaml requirements for number of nodes + a satellite check at join time. 

## Type of change

  - Support for 1k node clusters
  - Default pod CIDR is now the range 100.96.0.0/14 - 100.99.255.0/14
  - Users can now specify smaller pod subnets.

* Closes #, refs #

* Requires #

* Ports #

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
  - Tested large subnet assignments
- [ ] Address review feedback

### DIFF
--- a/lib/defaults/defaults.go
+++ b/lib/defaults/defaults.go
@@ -698,7 +698,7 @@ const (
 	// this seems like a sensible default. The setting can also be overridden at installation time.
 	ServiceSubnet = "100.100.0.0/16"
 	// PodSubnet is a subnet dedicated to the pods in the cluster
-	PodSubnet = "100.96.0.0/16"
+	PodSubnet = "100.96.0.0/14"
 
 	// MaxRouterIdleConnsPerHost defines tha maximum number of idle connections for "opsroute" transport
 	MaxRouterIdleConnsPerHost = 5
--- a/lib/validate/net.go
+++ b/lib/validate/net.go
@@ -36,11 +36,11 @@ func KubernetesSubnets(podCIDR, serviceCIDR string) error {
 				"invalid pod subnet: %q", podCIDR)
 		}
 
-		// the pod network should be /16 minimum so k8s can allocate /24 to each node
+		// the pod network should be /22 minimum so k8s can allocate /24 to each node (minimum 3 nodes)
 		ones, _ := podNet.Mask.Size()
-		if ones > 16 {
+		if ones > 22 {
 			return trace.BadParameter(
-				"pod subnet should be a minimum of /16: %q", podCIDR)
+				"pod subnet should be a minimum of /22: %q", podCIDR)
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


a-palchikov (approved):

lib/defaults/defaults.go
```diff
@@ -697,7 +697,7 @@ const (
 	// this seems like a sensible default. The setting can also be overridden at installation time.
 	ServiceSubnet = "100.100.0.0/16"
 	// PodSubnet is a subnet dedicated to the pods in the cluster
-	PodSubnet = "100.96.0.0/16"
+	PodSubnet = "100.96.0.0/14"
```
r0mant, line 700: Any downsides to making the default subnet even larger, say /12 or /11 to potentially allow to scale up to max K8s supported size and make it more future-proof?

lib/validate/net.go
```diff
@@ -36,11 +36,11 @@ func KubernetesSubnets(podCIDR, serviceCIDR string) error {
 				"invalid pod subnet: %q", podCIDR)
 		}
 
-		// the pod network should be /16 minimum so k8s can allocate /24 to each node
+		// the pod network should be /22 minimum so k8s can allocate /24 to each node (minimum 3 nodes)
 		ones, _ := podNet.Mask.Size()
-		if ones > 16 {
+		if ones > 22 {
```
r0mant, line 41: Just curious where this change is coming from. Has anyone tried/wanted to allocate smaller subnets for their clusters?

r0mant (approved):
Just a couple of questions.

lib/defaults/defaults.go
knisbet: Not really, we'd have to move the start range which isn't a big deal because we service cidr starts at 100.100.0.0...

It's also not hard to change this default at any time, or pass to the installer, so whatever the team prefers.

lib/validate/net.go
knisbet: It comes up with customer M from time to time. 

The problem is the pod cidr does need to be coordinated with the network to some degree, if pods within the cluster need to be able to connect to customer services. The way this tends to manifest, is the customer than allocates subnets for the pod/service cidr from their own subnet tracking, where a /16 is a quite large allocation. It does help that if they have multiple clusters they can re-use the /16 where some customers interpreted it as a /16 for each cluster(dev/staging/prod each get their own subnet), and also why we moved to the cgnat private IP range by default which is outside the RFC 1918 address space in use by most networks.

wadells:
Housekeeping: This doesn't seem to be headed anywhere at the moment.

Please re-open this PR if/when it is needed.

Issue #1983, opened 2020-08-07
Cluster gets stuck in shrinking w/o any recovery when removing a node


aelkugia:
### Description

During removing a master node the cluster gets stuck in shrinking without any recovery

### Environment

 - Gravity version: 6.1.31

**Relevant Debug Logs If Applicable**

```
User Message: failed to start shrink agent: [ERROR]: error enabling the service
failed to enable shrink-9278f72f-b710-4064-99fe-10c81c087729.service: Failed to execute operation: Connection timed out
```

More information in internal zendesk ticket: [#1807](https://gravitational.zendesk.com/agent/tickets/1807)

PR #2096, opened 2020-09-02
[6.1.x] shrink improvements

a-palchikov:
## Description

This PR makes the following changes:
  * reuses the default service name for the shrink agent (i.e. `gravity-agent`). The reasoning behind it is that the service setup path already handles clean of the previous service should it be active on the node since the shrink operation is exclusive to any other. We have had problems with agent service running in parallel with the shrink in which case the latter would fail to bind to the agent port. Also, using the `shrink-<operation-id>` naming generates a lot of litter which is never cleaned up.
  * rewrites the etcd member removal using the v2 API. The removal is wrapped in a retry loop that retries on transient errors

## Type of change

* Updates https://github.com/gravitational/gravity/issues/1983

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

### Shrinking a 3-node cluster to a single node

Verify that shrinks are successful in a healthy environment.

### Shrinking a 3-node cluster with simulated network failures to test timeouts:

```
# on the nodes which are left in the clustrer:
$ iptables -A OUTPUT -p tcp --destination-port 3012 -j DROP

$ gravity remove node-2 --confirm
launched operation "bed7ea74-9dec-495e-8093-6242252271e2", use 'gravity status' to poll its progress
18:41:59root@node-0:/home/centos/installer# gravity status --tail
Wed Sep  2 18:41:59 UTC [INFO] initializing the operation
Wed Sep  2 18:41:59 UTC [INFO] starting "node-2" removal
Wed Sep  2 18:43:01 UTC [ERROR] failed to wait for shrink agent
[ERROR]: failed to wait for shrink agent
```
and cluster transitions to `active` as expected.

### DIFF
--- a/lib/ops/opsservice/shrink.go
+++ b/lib/ops/opsservice/shrink.go
@@ -511,7 +511,6 @@ func (s *site) launchAgent(ctx *operationContext, server storage.Server) (*serve
 	}
 
 	serverAddr := s.service.cfg.Agents.ServerAddr()
-	serviceName := fmt.Sprintf("shrink-%v.service", ctx.operation.ID)
 	command := []string{
 		"ops", "agent", s.packages().PortalURL(),
 		"--advertise-addr", server.AdvertiseIP,
@@ -520,7 +519,7 @@ func (s *site) launchAgent(ctx *operationContext, server storage.Server) (*serve
 		"--vars", fmt.Sprintf("%v:%v", ops.AgentMode, ops.AgentModeShrink),
 		"--service-uid", s.uid(),
 		"--service-gid", s.gid(),
-		"--service-name", serviceName,
+		"--service-name", defaults.GravityRPCAgentServiceName,
 		"--cloud-provider", s.provider,
 	}
 	out, err := teleportRunner.Run(s.gravityCommand(command...)...)
--- a/lib/ops/opsservice/shrink.go
+++ b/lib/ops/opsservice/shrink.go
@@ -456,6 +456,9 @@ func (s *site) waitForServerToDisappear(hostname string) error {
 }
 
 func (s *site) removeFromEtcd(ctx context.Context, opCtx *operationContext, server storage.Server, masters []storage.Server) error {
+	peerURL := server.EtcdPeerURL()
+	logger := opCtx.WithField("peer", peerURL)
+	logger.Info("Remove peer from etcd cluster.")
 	b := utils.NewExponentialBackOff(defaults.EtcdRemoveMemberTimeout)
 	return utils.RetryTransient(ctx, b, func() error {
 		client, err := clients.DefaultEtcdMembers()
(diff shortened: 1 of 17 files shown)


lib/ops/opsservice/shrink.go
```diff
@@ -454,62 +455,67 @@ func (s *site) waitForServerToDisappear(hostname string) error {
...
-	}
-
-	return nil
+	})
 }
 
 func (s *site) uninstallSystem(ctx *operationContext, runner *serverRunner) error {
 	commands := [][]string{
-		s.gravityCommand("system", "uninstall", "--confirm"),
+		s.gravityCommand("system", "uninstall", "--confirm", "--no-uninstall-service"),
```
a-palchikov, line 486: Hidden flag to avoid uninstalling the agent service as long as the shrink operation is not complete as `system uninstall` automatically uninstalls and removes the agent services.

lib/rpc/proto/utils.go
```diff
@@ -57,10 +57,9 @@ type OutgoingMessageStream interface {
 func EncodeError(err error) *Error {
 	var errWithTraces *trace.TraceErr
 	var ok bool
-	if errWithTraces, ok = err.(*trace.TraceErr); ok {
+	if errWithTraces, ok = err.(*trace.TraceErr); !ok {
```
a-palchikov, line 60: Seems like a very old bug that triggered when I used the default service name.

lib/rpc/server/agent.go
```diff
@@ -156,15 +156,6 @@ func (srv *agentServer) Abort(ctx context.Context, req *types.Empty) (resp *type
 }
 
 func (srv *agentServer) command(req pb.CommandArgs, stream pb.Agent_CommandServer, log *log.Entry) (err error) {
-	defer func() {
```
a-palchikov, line 159: When the bug above triggered panic, it was impossible to see where it failed. In general, I don't think it's useful to consume the panics as here it is a definite programming error.

lib/ops/opsservice/shrink.go
r0mant: Is the service really still needed after "system uninstall" completes? Why not just let it uninstall itself (as a very last step of system uninstall)?

lib/ops/opsservice/shrink.go
```diff
@@ -454,62 +455,67 @@ func (s *site) waitForServerToDisappear(hostname string) error {
...
-
-	provisionedServer := ProvisionedServer{Server: server}
-	memberID, err := utils.FindETCDMemberID(
-		string(out), provisionedServer.EtcdMemberName(s.domainName))
-	if err != nil {
+func (s *site) removeFromEtcd(ctx context.Context, opCtx *operationContext, server storage.Server) error {
+	peerURL := server.EtcdPeerURL()
+	logger := opCtx.WithField("peer", peerURL)
+	logger.Info("Remove peer from etcd cluster.")
+	b := utils.NewExponentialBackOff(defaults.EtcdRemoveMemberTimeout)
```
r0mant, line 462: Does this exponential backoff has reasonable max retry interval? To make sure we get enough retry attempts.

r0mant (approved):

knisbet (approved):

lib/ops/opsservice/shrink.go
a-palchikov: I'll fix the max interval at 10s

lib/ops/opsservice/shrink.go
a-palchikov: Not needed. I've revised the uninstall path and made the agent cleanup the last step. Also made the uninstall step fail if a package service fails to uninstall since removing state at this point will render further attempts to uninstall the services impossible (i.e. if the planet service fails for some reason - it would be impossible to clean up manually if we proceeded and removed the rootfs directory).

PR #2104, opened 2020-09-04
[master] change coredns on workers to deployment

knisbet:
## Description

- Move coredns on worker nodes to be a deployment, creating less instances on the cluster
- Introduce proportional autoscaler, to scale up the worker coredns deployment in relation to current cluster size

## Type of change

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Implementation

Overall, the intention is this should help reduce our over provisioning of coredns which is currently a daemonset in our clusters. Looking at the reference Kubernetes addon, the recommendation is one instance per 16 nodes or one instance per 256 cores. 

I ended up generating ladder configuration for the proportional autoscaler because the linear configuration doesn't allow a minimum of 0 nodes (in our case when there are 0 workers in the cluster).

I'm also on the fence on whether we should convert the masters to a deployment as well. We use the master coredns instances as a fallback for query failures on workers, and we don't usually have enough masters to really make a difference here.

One last note, a problem we run into sometimes with a failed node, is that even when drained the node will continue to be included in load balancing for DNS queries. This change won't entirely address this, as we currently tolerate all taints, because we don't know what taints a customer might include in their configuration, but moving off a Daemonset is a step towards addressing this problem. This means we also ignore all kubernetes node health related taints for coredns even as a deployment. Eventually we may need to consider the infrastructure, to tolerate any customer provided taints, but not tolerate any health related taints. 

## Testing done

- Install Cluster
- Expand/Shrink cluster, observe deployment size, including removal of all workers
- Upgrade Cluster
- Rollback of dns-app phase


a-palchikov:
Hmm, seems like all our hooks are implicitly assigned the `coredns` psp and the robotest image ends up with the `echoserver` running with the new `cluster-proportional-autoscaler` psp and failing.

assets/dns-app/hooks/entrypoint.sh
```diff
@@ -6,6 +6,9 @@ if [ $1 = "update" ]; then
     echo "Updating resources"
     rig upsert -f /var/lib/gravity/resources/dns.yaml
 
+    echo "Deleting coredns daemonset that has been replaced by a deployment"
+    rig delete ds/coredns-worker --resource-namespace=kube-system --force
```
r0mant, line 10: When this executes during an upgrade, workers will be falling back to coredns pods running on masters before the new deployment comes up, correct?

assets/dns-app/resources/dns.yaml
```diff
@@ -289,3 +305,685 @@ spec:
...
+        [ 4080, 256 ]
+      ]
+    }
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: autoscaler-coredns-worker
+  namespace: kube-system
+  labels:
```
r0mant, line 954: I would also mark it with `gravitational.io/critical-pod` label to make sure gravity status shows if there's a problem with it, e.g. like https://github.com/gravitational/gravity/blob/master/assets/site-app/resources/site.yaml#L117.

assets/dns-app/resources/dns.yaml
```diff
@@ -289,3 +305,685 @@ spec:
...
+    k8s-app: autoscaler-coredns-worker
+spec:
```
r0mant, line 956: Should we also include a priority class?

assets/dns-app/resources/dns.yaml
```diff
@@ -289,3 +305,685 @@ spec:
...
+      tolerations:
+        - operator: "Exists"
+      containers:
+        - image: k8s.gcr.io/cpa/cluster-proportional-autoscaler-amd64:1.8.3
+          name: autoscaler
+          command:
+            - /cluster-proportional-autoscaler
+            - --namespace=kube-system
+            - --configmap=autoscaler-coredns-worker
+            - --target=deployment/coredns-worker
```
r0mant, line 976: So I'm curious, do you know what the usual deployment model with this autoscaler is? I.e. if we want to use it for some other component (e.g. prometheus in monitoring-app), would we need to deploy another instance of autoscaler specifically for monitoring, or reuse this one (like, add another config or something).

If it's supposed to be reused, I'm wondering if it would make sense instead of bundling it with dns make it a separate system level application ("autoscaler-app") and have specific apps that need its functionality just create their configurations.

assets/dns-app/resources/dns.yaml
```diff
@@ -289,3 +305,685 @@ spec:
   - name: metrics
     port: 9153
     protocol: TCP
+#
+# Cluster Proportional Autoscaler
+# Scale the coredns deployment proportionally to the cluster size
+# https://github.com/kubernetes-sigs/cluster-proportional-autoscaler
+# https://github.com/kubernetes-sigs/cluster-proportional-autoscaler/tree/master/examples
```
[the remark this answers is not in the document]
knisbet: I don't believe that's relevant in this case, which is basically a link to the project homepage, and an examples subdirectory, not a specific view of code.

assets/dns-app/hooks/entrypoint.sh
knisbet: It might race a bit. The deployment is created first on line 6, but will only wait for the autoscaler to be scheduled, as we don't know here if the cluster has workers. Once the autosacler starts, it'll scale up the deployment appropriately. 

Whether the delete executes before the deployment pods get created I'm not sure, which would mean timeouts/errors to the worker service, and then yes fallback to the master service until the pods come up.

assets/dns-app/resources/dns.yaml
knisbet: AFAIK we would need to deploy another instance... that targets the specific deployment. So it makes sense to bundle the autoscaler with the configuration its scaling. IE. If I wanted to scale both workers and masters, I would need to deploy 2x auto scalers targeting each deployment individually.

This is why I included it within coredns, and means we should be careful to bump in multiple repos... although it's not the end of the world if we get into a situation where we have multiple versions.

assets/dns-app/resources/dns.yaml
knisbet: Yea I'm on the fence about that one. The autoscaler isn't really a critical pod, so I'm not sure it should be treated as critical, and I don't think we've really planned out a non-critical system for our stuff and customers deployments to interact with.

a-palchikov (approved):

PR #2107, opened 2020-09-04
Specify security context for robotest workload

### LINKED ISSUE #2105: Applications need explicit security context
### Description

**What happened**:
If new `PSP`s are introduced into the cluster (as is the case with [PR](https://github.com/gravitational/gravity/pull/2104)), the controller manager might select an invalid `PSP` for the pod (as it does in [here](https://console.cloud.google.com/logs/viewer?authuser=0&expandAll=false&project=kubeadm-167321&angularJsUrl=%2Flogs%2Fviewer%3FadvancedFilter%3Dresource.type%253D%2522project%2522%250Alabels.__uuid__%253D%2522bc424943-3a1f-458d-9c4f-3951c11c2613%2522%250Alabels.__suite__%253D%2522f8a79518-761c-43bb-aa96-021b3a419e2b%2522%250Aseverity%253E%253DINFO%26authuser%3D1%26expandAll%3Dfalse%26project%3Dkubeadm-167321&minLogLevel=0&timestamp=2020-09-04T09:16:53.748000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=resource.type%3D%22project%22%0Alabels.__uuid__%3D%22bc424943-3a1f-458d-9c4f-3951c11c2613%22%0Alabels.__suite__%3D%22f8a79518-761c-43bb-aa96-021b3a419e2b%22%0A&interval=NO_LIMIT&scrollTimestamp=2020-09-04T07:29:24.934379658Z&dateRangeUnbound=backwardInTime)) unless it is explicit about its security context.

The application resources are explicit about security context and are able to be scheduled regardless of additional `PSP`s in the cluster.

### Environment

 - Gravity version [e.g. 7.0.11]: 7.1.0-alpha.1.185 (8f870ecc9e840d71ff796168110a7e207c3b4b9c)
 - OS [e.g. Redhat 7.4]: redhat 7

a-palchikov:
## Description

Explicitly opt-out of running as non-root for echoserver test workload.
Remove spec redundancy in RBAC.

## Type of change

* Closes https://github.com/gravitational/gravity/issues/2105

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

### Before the change
Cluster fails to install at `postInstall` hook step for `robotest` application. Both `echoserver` replicas are in `CreateContainerConfigError` state. As a test for this step, this [PR](https://github.com/gravitational/gravity/pull/2104) was used.

### After the change
Cluster installs successfully.

### DIFF
--- a/assets/rbac-app/resources/resources.yaml
+++ b/assets/rbac-app/resources/resources.yaml
@@ -182,6 +182,7 @@ rules:
   - podsecuritypolicies
   resourceNames:
   - privileged
+  - restricted
   verbs:
   - use
 ---
@@ -225,25 +226,6 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: privileged-psp-user
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
-metadata:
-  name: telekube:controller-restricted-psp
-subjects:
-- kind: ServiceAccount
-  name: daemon-set-controller
-  namespace: kube-system
-- kind: ServiceAccount
-  name: job-controller
-  namespace: kube-system
-- kind: ServiceAccount
-  name: generic-garbage-collector
-  namespace: kube-system
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: restricted-psp-user
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
--- a/assets/robotest/resources/charts/echoserver/templates/echoserver-deployment.yaml
+++ b/assets/robotest/resources/charts/echoserver/templates/echoserver-deployment.yaml
@@ -19,6 +19,8 @@ spec:
       containers:
       - name: echoserver
         image: {{.Values.echoserverRegistry}}/echoserver:1.10
+        securityContext:
+          runAsNonRoot: false
 ---
 apiVersion: v1
 kind: Service


knisbet (approved):

r0mant (approved):

wadells:
@a-palchikov @knisbet 

> What happened:
> If new PSPs are introduced into the cluster (as is the case with PR), the controller manager might select an invalid PSP for the pod (as it does in here) unless it is explicit about its security context.

It sounds like I wrote a manifest which implicitly relied on undefined behavior. This was working for a while -- by sheer luck?  Something internal, unrelated & gravity specific changed, and broke the formerly working manifest.

This sounds like a product regression.  Is there anything that prevents our customers from making the same configuration mistake that I did and subsequently hitting the same error?

> What you expected to happen:
>  The application resources are explicit about security context and are able to be scheduled regardless of additional PSPs in the cluster.

Building on this, I'd expect that `tele build` would fail if application resources have an invalid configuration (e.g. are not explicit about security context).  Should this be an error during `tele build` instead of one during `gravity install`?

As an alternative, I'd consider some guildelines in the examples/docs & a release note about the behavior change.  However our examples & docs don't have anything about security context being suggested/required, and neither this PR nor #2104 add any.

```
walt@work:~/git/gravity$ rg -A3 securityContext docs/7.x examples 
docs/7.x/selinux.md
107:Kubernetes supports SELinux with [seLinuxOptions](https://v1-17.docs.kubernetes.io/docs/tasks/configure-pod-container/security-context/#assign-selinux-labels-to-a-container) inside the [securityContext](https://v1-17.docs.kubernetes.io/docs/tasks/configure-pod-container/security-context/). The options can be configured either for the whole Pod or per-container:
108-
109-```yaml
110-...
111:securityContext:
112-  seLinuxOptions:
113-    type: "my_type_t"
114-    level: "s0:c123,c456"

docs/7.x/pack.md
986:  securityContext:
987-    runAsUser: -1   # to use for all containers in a Pod
988-  containers:
989-  - name: nginx
--
993:    securityContext:
994-      runAsUser: -1   # to use for a single container
995-```
996-

examples/robot-shop-app/resources/charts/robot-shop/templates/mysql-deployment.yaml
25:        securityContext:
26-          capabilities:
27-            add: ["NET_ADMIN"]
28-      {{ end }}
```

a-palchikov:
I'm not certain that controller manager is capable of automatically selecting the correct PSP in any given case just like I'm not sure we can detect that a specific resource might require security configuration it currently lack just by examining the definition. In this case you had to take the container environment requirements into consideration (i.e. whether it is supposed to have additional permissions) and it's hard to do without actually running a container.
I've been thinking whether it would make sense to pin our resources (e.g. hooks) to a specific security policy since currently they're also running in an arbitrary suitable PSP (`coredns` to be more specific). This could be as simple as creating a dedicated service account or as complex as a separate PSP.

Issue #2100, opened 2020-09-03
Serf cluster brain split


r0mant:
### Description

**What happened**:

If one of serf nodes is partitioned off of the rest of the cluster for too long (>24h), serf will eventually kick it out of the cluster and it will result in the cluster brain split situation when the node comes back.

For example, here's the situation observed on one of the clusters:

```bash
# on a controller
> serf members
192_168_0_1.c01  192.168.0.1:7496  alive  role=master,publicip=192.168.0.1

# on a worker
> serf members
192_168_0_2.w01  192.168.0.2:7496  alive  role=node,publicip=192.168.0.2
192_168_0_3.w02  192.168.0.3:7496  alive  role=node,publicip=192.168.0.3
```

Another issue is that in such a case `gravity status` shows "active" status yet the nodes are offline:

```
Cluster status: active
...
Cluster nodes:
  Masters:
    * c01 (192.168.0.1, controller_node)
      Status: healthy
  Nodes:
    * w02 (192.168.0.2)
      Status: offline
    * w01 (192.168.0.3)
      Status: offline
```

**What you expected to happen**:

- [ ] `gravity status` should display "degraded" in such situation.
- [x] Gravity should attempt to re-join missing serf member to the cluster automatically.

**How to reproduce**:

1. Install a 3-node cluster.
2. On one of the nodes execute `serf force-leave -prune <another-node>`.
3. Check `serf members` and `gravity status`.

### Environment

 - Gravity version [e.g. 7.0.11]: 5.5 and later

helgi:
Time to ditch serf? ;)

PR #2111, opened 2020-09-08
Degrade system status in case of offline nodes

bernardjkim:
## Description

This PR updates the gravity status to report degraded status if any nodes are offline. Prior to this change, planet status could report an active cluster status while failing to collect the status of some nodes. This has been observed in https://github.com/gravitational/gravity/issues/2100 when a serf node is partition off the cluster for too long and is eventually kicked out of the serf cluster.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/2100

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

**Verify cluster status is degraded after a member leaves the serf cluster**
- Install 3-node cluster.
- On one of the nodes execute `serf force-leave -prune <another-node>`.
- Verify `serf members`.
```
[vagrant@node-1 ~]$ sudo gravity exec serf members
172_28_128_101.dev.test  172.28.128.101:7496  alive  role=master,publicip=172.28.128.101
172_28_128_102.dev.test  172.28.128.102:7496  alive  role=master,publicip=172.28.128.102
```
- Verify `gravity status` is degraded.
```
[vagrant@node-1 ~]$ sudo gravity status
Cluster name:           dev.test
Cluster status:         degraded
...
Cluster nodes:
    Masters:
        * node-1 / 172.28.128.101 / node
            Status:             healthy
            Remote access:      online
        * node-2 / 172.28.128.102 / node
            Status:             healthy
            Remote access:      online
    Nodes:
        * node-3 / 172.28.128.103
            Status:             offline
            Remote access:      online
[ERROR]: degraded
```

### DIFF
--- a/lib/status/status.go
+++ b/lib/status/status.go
@@ -522,6 +522,15 @@ func fromPlanetAgent(ctx context.Context, local bool, servers []storage.Server) 
 		nodes = fromSystemStatus(*status)
 	}
 
+	// Collected system status may exclude the status of some nodes due to
+	// network partition. If offline nodes are detected, degrade system status.
+	for _, node := range nodes {
+		if node.Status != NodeHealthy {
+			status.Status = pb.SystemStatus_Degraded
+			break
+		}
+	}
+
 	return &Agent{
 		SystemStatus: SystemStatus(status.Status),
 		Nodes:        nodes,


a-palchikov (approved):

lib/status/status.go
```diff
@@ -522,6 +522,15 @@ func fromPlanetAgent(ctx context.Context, local bool, servers []storage.Server)
 		nodes = fromSystemStatus(*status)
 	}
 
+	// Collected system status may exclude the status of some nodes due to
+	// network partition. If offline nodes are detected, degrade system status.
+	for _, node := range nodes {
```
r0mant, line 527: Hmm, so is the node going to be excluded completely from the list of nodes in such case (as the comment above suggests), or be present but unhealthy?

bernardjkim: Currently planet relies on serf to identify the members of the gravity cluster. planet will only be able to construct a SystemStatus that contains the NodeStatus for members of the serf cluster. So the SystemStatus that is queried by gravity may exclude the NodeStatus for some members of the gravity cluster.

gravity maintains its own view of the cluster and does not rely on serf. After querying SystemStatus from planet, if a NodeStatus was not collected for a member, an empty NodeStatus will be initialized for the member. So after this point, a NodeStatus should be present for all members of the gravity cluster.

This added code will detect this case when gravity initializes an empty NodeStatus, and degrade the SystemStatus.

r0mant (approved):

PR #2114, opened 2020-09-09
[7.0.x] shrink improvements

a-palchikov:
## Description

Backport of https://github.com/gravitational/gravity/pull/2096 to 7.0.x.

## Type of change

* Ports https://github.com/gravitational/gravity/pull/2096.

## TODOs

- [x] Self-review the change
- [ ] Address review feedback

### DIFF
--- a/lib/ops/opsservice/agents.go
+++ b/lib/ops/opsservice/agents.go
@@ -90,15 +90,12 @@ func (s *site) agentReport(ctx context.Context, opCtx *operationContext) (*ops.A
 }
 
 func (s *site) waitForAgents(ctx context.Context, opCtx *operationContext) (*ops.AgentReport, error) {
-	localCtx, cancel := defaults.WithTimeout(ctx)
-	defer cancel()
-
-	err := s.agentService().Wait(localCtx, opCtx.key(), opCtx.getNumServers())
+	err := s.agentService().Wait(ctx, opCtx.key(), opCtx.getNumServers())
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
 
-	report, err := s.agentReport(localCtx, opCtx)
+	report, err := s.agentReport(ctx, opCtx)
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
@@ -298,13 +295,12 @@ func (r *AgentService) Wait(ctx context.Context, key ops.SiteOperationKey, numAg
 
 // AbortAgents shuts down remote agents and cleans up state
 func (r *AgentService) AbortAgents(ctx context.Context, key ops.SiteOperationKey) error {
-	group, err := r.peerStore.getGroup(key)
+	group, err := r.peerStore.removeGroup(ctx, key)
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	err = group.Abort(ctx)
-	r.peerStore.removeGroup(ctx, key)
-	return trace.Wrap(err)
+	defer group.Close(ctx)
+	return trace.Wrap(group.Abort(ctx))
 }
 
 // StopAgents shuts down remote agents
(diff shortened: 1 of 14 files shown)


lib/ops/opsservice/agents.go
```diff
@@ -298,13 +295,12 @@ func (r *AgentService) Wait(ctx context.Context, key ops.SiteOperationKey, numAg
...
-	group, err := r.peerStore.getGroup(key)
+	group, err := r.peerStore.removeGroup(ctx, key)
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	err = group.Abort(ctx)
-	r.peerStore.removeGroup(ctx, key)
-	return trace.Wrap(err)
+	defer group.Close(ctx)
+	return trace.Wrap(group.Abort(ctx))
```
r0mant, line 303: Is the order change on purpose here? Before it was `group.Abort()` followed by `removeGroup()` and not it's the opposite.

lib/ops/opsservice/shrink.go
```diff
@@ -496,31 +501,33 @@ func (s *site) uninstallSystem(ctx *operationContext, runner *serverRunner) erro
 	for _, command := range commands {
 		out, err := runner.Run(command...)
 		if err != nil {
-			ctx.Warningf("failed to run %v: %v (%s)", command, trace.DebugReport(err), out)
+			ctx.WithError(err).WithFields(log.Fields{
+				"command": command,
+				"output":  string(out),
+			}).Warn("Failed to run.")
```
r0mant, line 507: Nit: Make this an Error maybe?

lib/system/environ/uninstall.go
```diff
@@ -141,6 +157,12 @@ func DisableAgentServices(logger log.FieldLogger) error {
 	return trace.NewAggregate(errors...)
 }
 
+func uninstallAgentServices(svm systemservice.ServiceManager) error {
+	return uninstallServices(svm,
+		defaults.GravityRPCInstallerServiceName,
+		defaults.GravityRPCAgentServiceName)
```
r0mant, line 163: There's also [gravity-resume.service](https://github.com/gravitational/gravity/blob/7.0.16/tool/gravity/cli/clusterupdate.go#L246) which we might want to move to defaults and add here as well...

tool/gravity/cli/system.go
```diff
@@ -672,11 +672,21 @@ func systemUninstall(env *localenv.LocalEnvironment, confirmed bool) error {
...
+		logger.WithError(err).Warn("Failed to uninstall package services.")
+		// Do not attempt to remove state if a package service failed to uninstall since
+		// this might result in inability to re-attempt uninstallation after fixing the issue.
+		// Instead bail out early and output the relevant error message
+		return trace.Wrap(err)
 	}
+	logger.Info("Uninstall system.")
 	if err := environ.UninstallSystem(context.TODO(), env, logger); err != nil {
-		log.WithError(err).Warn("Failed to uninstall system.")
+		logger.WithError(err).Warn("Failed to uninstall system.")
```
r0mant, line 685: I wonder if this and "uninstall agent services" below should also return an error if they fail, like you did above. Otherwise our uninstall is not reliable as users may rely on its successful exit code as indicator that everything was cleaned up while it in fact may not be true. Can't remember why it was done like this in the first place...

r0mant (approved):

lib/ops/opsservice/agents.go
a-palchikov: Yes, I changed order on purpose. I noticed that there's no need for a separate `getGroup` as all queries are done in the context of removing the group eventually as part of the operation so I combined these into a single removeGroup that does that - removes the group from the internal state but returns a reference to the value so it can be acted upon. There should be no change in semantics w.r.t functionality

PR #2124, opened 2020-09-13
gofmt entire project

alrs:
This cleans up some stray non-`gofmt` files in the repo all at once, so the fixes don't trickle in as the files are touched. If that's the better way for this project to handle non-`gofmt`-compliant sources, then this PR should be closed.


r0mant (approved):

r0mant:
Thanks for the contribution!

PR #2127, opened 2020-09-15
Changelog for 6.1.39 release

a-palchikov:
## Description

6.1.39 release changelog.


docs/7.x/changelog.md
```diff
@@ -570,6 +570,12 @@ to learn how to gain insight into how the cluster status changes over time.
 
 ## 6.1 Releases
 
+### 6.1.39 LTS (September 15th, 2020)
+
+#### Bugfixes
+
+* Fix regression in 'tele build' ([#2126](https://github.com/gravitational/gravity/pull/2126)).
```
r0mant, line 577: Use backticks (`) instead of apostrophes so it's rendered as code in the markdown. Also, would be helpful to specify what kind of regression it was exactly so people can see if it applies to them.

r0mant (approved):
lgtm with the comment applied

PR #2023, opened 2020-08-17
(7.0) Compare hostnames when searching for default hub credentials

r0mant:
## Description

Fixes an issue with tele not finding credentials for default hub if token isn't provided:

```
➜  e git:(version/7.0.x) GRAVITY_K8S_VERSION=1.17.9 /code/go/src/github.com/gravitational/gravity/e//build/7.0.14-dev.24/tele build \
        /code/go/src/github.com/gravitational/gravity//assets/telekube/resources/app.yaml -f \
        --version=7.0.14-dev.24 \
        --state-dir=/code/go/src/github.com/gravitational/gravity/e//build/7.0.14-dev.24/packages \
        --skip-version-check \
        --hub=https://get.gravitational.io  \
        -o /code/go/src/github.com/gravitational/gravity/e//build/7.0.14-dev.24/telekube.tar
Mon Aug 17 22:39:23 UTC	Building cluster image telekube 7.0.14-dev.24
Mon Aug 17 22:39:23 UTC	Selecting base image version
	Will use base image version 7.0.14-dev.24
Mon Aug 17 22:39:23 UTC	Build finished in now
[ERROR]: no credentials for https://get.gravitational.io:443
```

## Type of change

* Bug fix (non-breaking change which fixes an issue)
* Regression fix (non-breaking change which fixes a regression)

## Linked tickets and other PRs

* Requires https://github.com/gravitational/gravity.e/pull/4330.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

`make telekube` works fine now:

```
GRAVITY_K8S_VERSION=1.17.9 /code/go/src/github.com/gravitational/gravity/e//build/7.0.14-dev.24/tele build \
	/code/go/src/github.com/gravitational/gravity//assets/telekube/resources/app.yaml -f \
	--version=7.0.14-dev.24 \
	--state-dir=/code/go/src/github.com/gravitational/gravity/e//build/7.0.14-dev.24/packages \
	--skip-version-check \
	--hub https://get.gravitational.io \
	-o /code/go/src/github.com/gravitational/gravity/e//build/7.0.14-dev.24/telekube.tar
Mon Aug 17 23:43:06 UTC	Building cluster image telekube 7.0.14-dev.24
Mon Aug 17 23:43:06 UTC	Selecting base image version
	Will use base image version 7.0.14-dev.24
Mon Aug 17 23:43:06 UTC	Syncing packages for 7.0.14-dev.24
[...]

PR #2133, opened 2020-09-15
Exercise upgrades with custom robotest images

### LINKED ISSUE #1915: Implement upgrade testing with the new cluster image

### LINKED ISSUE #2103: Provide a mechanism for Gravity CI to specify the uid & gid of robtest state & logs
We're seeing ongoing trouble caused by robotest files owned by root:root on our Gravity CI workers.

Robotest should offer guidance & tooling to help consumers run it with all files properly owned buy the triggering user.

wadells:
## Description
This PR introduces two notable changes, as well as many smaller ones:

1) Robotest now produces custom images for a variety of gravity versions, and uses these during upgrade testing.
2) Docker permissions issues (#2103) are fixed, allowing Jenkins to properly clean up gravity build directories for the first time in over a year.

## Type of change

* Closes #1915
* Closes #2103
* Ports #2023
* Depends on https://github.com/gravitational/gravity.e/pull/4336

## TODOs

- [x] Enterprise PR
- [x] Self-review the change
- [x] Write documentation
- [x] Green PR build
- [x] Green Nightly build
- [x] Gather before/after artifact size stats, runtime stats
- [x] Address review feedback

## Testing done
Nightly run: [logs](https://console.cloud.google.com/logs/query;query=severity%3E%3DINFO%0Alabels.__suite__%3D%221501e656-9ca9-45c7-a085-b0a5704a2c99%22;timeRange=2020-09-19T06:09:00.000Z%2F2020-09-19T07:27:00.000Z?authuser=0&project=kubeadm-167321)

This PR's jenkins job shows the tests run to completion on PRs.

Examples of using the Makefile documented in the README.md.

### New Configurations

<details><summary>pr configs</summary>

```
walt@gcp:~/git/gravity$ make -C e/assets/robotest get-config ROBOTEST_CONFIG=pr     
make: Entering directory '/home/walt/git/gravity/e/assets/robotest'
install={"installer_url":"/images/telekube-7.1.0-alpha.1.245.tar","nodes":3,"flavor":"three","role":"node","os":"ubuntu:18"}
install={"installer_url":"/images/opscenter-7.1.0-alpha.1.245.tar","nodes":1,"flavor":"standalone","role":"node","os":"ubuntu:18","ops_advertise_addr":"example.com:443"}
install={"installer_url":"/images/robotest-7.1.0-alpha.1.245.tar",""flavor":"three","nodes":3,"role":"node","os":"centos:7","storage_driver":"overlay2"}
resize={"installer_url":"/images/robotest-7.1.0-alpha.1.245.tar","nodes":1,"to":3,"flavor":"one","role":"node","state_dir":"/var/lib/telekube","os":"ubuntu:18","storage_driver":"overlay2"}
shrink={"installer_url":"/images/robotest-7.1.0-alpha.1.245.tar","nodes":3,"flavor":"three","role":"node","os":"redhat:7"}
upgrade={"from":"/images/robotest-7.0.7.tar","installer_url":"/images/robotest-7.1.0-alpha.1.245.tar","flavor":"three","nodes":3,"role":"node","os":"ubuntu:16","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-7.0.13.tar","installer_url":"/images/robotest-7.1.0-alpha.1.245.tar","flavor":"three","nodes":3,"role":"node","os":"centos:7","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-7.0.12.tar","installer_url":"/images/robotest-7.1.0-alpha.1.245.tar","flavor":"three","nodes":3,"role":"node","os":"ubuntu:18","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-7.0.16.tar","installer_url":"/images/robotest-7.1.0-alpha.1.245.tar","flavor":"three","nodes":3,"role":"node","os":"centos:7","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
make: Leaving directory '/home/walt/git/gravity/e/assets/robotest'
```
</details>

<details><summary>nightly configs</summary>

```
walt@gcp:~/git/gravity$ make -C assets/robotest ROBOTEST_CONFIG=nightly get-config
make: Entering directory '/home/walt/git/gravity/assets/robotest'
install={"installer_url":"/images/telekube-7.1.0-alpha.1.255.tar","nodes":3,"flavor":"three","role":"node","os":"ubuntu:18"}
install={"installer_url":"/images/opscenter-7.1.0-alpha.1.255.tar","nodes":3,"flavor":"ha","role":"node","os":"ubuntu:18","ops_advertise_addr":"example.com:443"}
install={"installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"redhat:7","storage_driver":"overlay2"}
install={"installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"six","nodes":6,"role":"node","os":"redhat:7","storage_driver":"overlay2"}
install={"installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"centos:7","storage_driver":"overlay2"}
install={"installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"six","nodes":6,"role":"node","os":"centos:7","storage_driver":"overlay2"}
install={"installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"debian:9","storage_driver":"overlay2"}
install={"installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"six","nodes":6,"role":"node","os":"debian:9","storage_driver":"overlay2"}
install={"installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"ubuntu:16","storage_driver":"overlay2"}
install={"installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"six","nodes":6,"role":"node","os":"ubuntu:16","storage_driver":"overlay2"}
install={"installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"ubuntu:18","storage_driver":"overlay2"}
install={"installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"six","nodes":6,"role":"node","os":"ubuntu:18","storage_driver":"overlay2"}
resize={"installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","to":3,"flavor":"one","nodes":1,"role":"node","state_dir":"/var/lib/telekube","os":"ubuntu:18","storage_driver":"overlay2"}
resize={"installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","to":6,"flavor":"three","nodes":3,"role":"node","state_dir":"/var/lib/telekube","os":"ubuntu:18","storage_driver":"overlay2"}
shrink={"installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","nodes":3,"flavor":"three","role":"node","os":"redhat:7"}
upgrade={"from":"/images/robotest-7.0.16.tar","installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"centos:7","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-7.0.16.tar","installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"six","nodes":6,"role":"node","os":"centos:7","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-7.0.16.tar","installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"one","nodes":1,"role":"node","os":"centos:7","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-7.0.0.tar","installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"ubuntu:16","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-6.2.0.tar","installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"ubuntu:16","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-6.2.5.tar","installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"centos:7","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-6.3.18.tar","installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"centos:7","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-7.0.13.tar","installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"centos:7","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-7.0.12.tar","installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"ubuntu:18","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-7.0.16.tar","installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"centos:7","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-7.0.16.tar","installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"redhat:7","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-7.0.16.tar","installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"debian:9","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
upgrade={"from":"/images/robotest-7.0.16.tar","installer_url":"/images/robotest-7.1.0-alpha.1.255.tar","flavor":"three","nodes":3,"role":"node","os":"ubuntu:18","service_uid":997,"service_gid":994,"storage_driver":"overlay2"}
make: Leaving directory '/home/walt/git/gravity/assets/robotest'
```
</details>

### Impact on the Build
This adds 20+GB of build artifacts to each PR build. Stats after a `clean` and `cache-clean` followed by `images`.

```
walt@gcp:~/git/gravity$ du -sh ~/.cache/robotest/ent e/build/current/robotest 
5.7G    /home/walt/.cache/robotest/ent
16G     e/build/current/robotest
```

Caches may eventually be shared between builds/branches, reducing that impact, but as of now they're kept in the build directory.

Robotest image builds adds ~7 minutes to master Gravity PR runtime.  With the new Jenkins pipeline, this can be easily seen as robotest image builds are split out into their own step.

Both of these numbers will adjust proportionally upwards for branches that have more upgrade base versions.  E.g. 5.5 has 10 upgrade base versions to master's 4.

### Jenkinsfile work

The refactored Jenkinsfile & Make can run both Nightly and PR configurations by tweaking the `ROBOTEST_CONFIG` variable.  After this PR merges, I will move nightlies to use the regular `Jenkinsfile`, and then retire `Jenkinsfile.nightly`.

To see what the new Jenkins pipleline looks like, see [the PR build for this PR](https://jenkins.gravitational.io/job/Gravity-PR-5.2/job/PR-2133/), or r[ecent try builds where I've been iterating on Jenkins changes](https://jenkins.gravitational.io/view/Robotest/job/gravity-try-build/).


assets/robotest/upgrade-base/charts/echoserver/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+name: echoserver
+version: 0.0.1
```
wadells, line 2: This should differ between `upgrade-base` and `current`.  Probably 1.0, 1.1.

build.assets/robotest/Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG BASE
```
wadells, line 1: A comment & link to the relevant bug wouldn't go awry in this file.

build.assets/robotest/Makefile
```diff
@@ -0,0 +1,326 @@
...
+#    a race between builds could result in a half copied artifact being used.
+#  - CACHES should not be assumed to be on the same filesystem as the build directories.
+#  - Enterprise tele and OSS tele cannot share caches, nor should their outputs
+#    be confused.
+#  - If any content in ROBOTEST_IMG_SRC changes, old images in ROBOTEST_UPGRADE_IMAGE_CACHE
+#    must be ruled out. These shouldn't necessarily be removed if the cache is
+#    shared between branches, as other branches may use the older images. In this
+#    scenario, ROBOTEST_TELE_CACHE and ROBOTEST_GRAVITY_PKG_CACHE remain valid.
+#  - ROBOTEST_IMAGEDIR can be heavy (up to 30GB for 7.0.x). This can
+#    aggravate CI workers that have many active builds or don't clean up.
```
wadells, line 47: Move this content to the README.

build.assets/robotest/Makefile
```diff
@@ -0,0 +1,326 @@
...
+	docker build \
+		--build-arg BASE=$(ROBOTEST_DOCKER_UPSTREAM_IMAGE) \
+		--build-arg UID=$(shell id -u) \
+		--build-arg GID=$(shell id -g) \
+		--iidfile $(ROBOTEST_DOCKER_IMAGE_IIDFILE) \
+		--tag $(ROBOTEST_DOCKER_IMAGE) .
+
+.PHONY: images
+images: ## Create all images necessary to run robotest CI tests.
+images: image-robotest images-upgrade image-opscenter
```
wadells, line 227: Consider doing this as a `$(MAKE) -j targets` to parallelize the main image and opscenter image work with the already parallel upgrade images -- alternatively, this may may the logs very difficult to follow.  If kept this way, leave a comment as to why.

build.assets/robotest/Makefile
```diff
@@ -0,0 +1,326 @@
...
+.PHONY: populate-caches
+populate-caches: ## Populate caches with tele binaries, gravity packages, and robotest upgrade images.
+	$(MAKE) -j $(ROBOTEST_TELE_BINARIES)
+	$(MAKE) -j $(ROBOTEST_UPGRADE_CACHE_IMAGES)
+
+.PHONY: get-upgrade-from-versions
+get-upgrade-from-versions: ## Show versions used in upgrade from testing.
+get-upgrade-from-versions:
+	@echo $(ROBOTEST_UPGRADE_VERSIONS)
+export
```
wadells, line 313: export looks out of place

build.assets/robotest/Makefile
```diff
@@ -0,0 +1,326 @@
...
+
+.PHONY: populate-caches
+populate-caches: ## Populate caches with tele binaries, gravity packages, and robotest upgrade images.
+	$(MAKE) -j $(ROBOTEST_TELE_BINARIES)
+	$(MAKE) -j $(ROBOTEST_UPGRADE_CACHE_IMAGES)
+
+.PHONY: get-upgrade-from-versions
+get-upgrade-from-versions: ## Show versions used in upgrade from testing.
+get-upgrade-from-versions:
+	@echo $(ROBOTEST_UPGRADE_VERSIONS)
```
wadells, line 312: This should probably be `get config` and print out the rendered config.  I'm not sure there is much reason to care about the upgrade versions alone (aside from those hacking on this file).

build.assets/robotest/Makefile
```diff
@@ -0,0 +1,326 @@
...
+export
+
+.PHONY: run
+run: ## Run robotest.
+run: export GRAVITY_URL = $(ROBOTEST_GRAVITY_BINARY)
```
wadells, line 317: should probably be `?=` just for consistency.

build.assets/robotest/nightly_config.sh
```diff
@@ -21,15 +21,21 @@ UPGRADE_MAP[6.2.0]="ubuntu:16"
 
 UPGRADE_VERSIONS=${!UPGRADE_MAP[@]}
 
+# The following breaks a dependency loop. We need upgrade versions to generate tarballs,
+# but we need tarball names to generate the full test config.
+if [[ ${1} == "upgradeexit" ]]  ; then
+    return
+fi
```
wadells, line 28: This still feels like an ugly hack.  How could the code be factored to handle this dependency better?

build.assets/robotest/pr_config.sh
```diff
@@ -67,10 +72,11 @@ EOF
   echo -n $suite
 }
 
-SUITE=$(build_install_suite)
+SUITE=""
+SUITE="$SUITE $(build_install_suite)"
```
wadells, line 76: This change should also occur in nightly.  (for easier commenting of these lines during development.

build.assets/robotest/pr_config.sh
```diff
@@ -67,10 +72,11 @@ EOF
...
 SUITE="$SUITE $(build_resize_suite)"
 SUITE="$SUITE $(build_upgrade_suite)"
 
 echo "$SUITE" | tr ' ' '\n'
 
-export SUITE UPGRADE_VERSIONS
+export SUITE
```
wadells, line 82: Nightly needs this change too.

build.assets/robotest/run.sh
```diff
@@ -3,55 +3,114 @@ set -o errexit
...
+REPEAT_TESTS=${REPEAT_TESTS:-1}
+RETRIES=${RETRIES:-3}
+FAIL_FAST=${FAIL_FAST:-false}
+ALWAYS_COLLECT_LOGS=${ALWAYS_COLLECT_LOGS:-true}
+
+# what should happen with provisioned VMs on individual test success or failure
+DESTROY_ON_SUCCESS=${DESTROY_ON_SUCCESS:-true}
+DESTROY_ON_FAILURE=${DESTROY_ON_FAILURE:-true}
+
+# set SUITE and ROBOTEST_IMAGE_DIR_MOUNTPOINT
```
wadells, line 36: `ROBOTEST_IMAGE_DIR_MONTPOINT` should come from this higher level instead of vice versa.

Also `IMAGES_DIR` instead of `IMAGE_DIR` for consistence with `ROBOTEST_IMAGES_DIR`

build.assets/robotest/run.sh
```diff
@@ -3,55 +3,114 @@ set -o errexit
...
+
+# ROBOTEST_IMAGE_DIR_MOUNTPOINT defined by the config
+EXTRA_VOLUME_MOUNTS="-v $ROBOTEST_IMAGES_DIR:$ROBOTEST_IMAGE_DIR_MOUNTPOINT"
+
+
+# GRAVTIY_FILE/GRAVITY_URL specify the location of the up-to-date gravity binary
+if [ -d $(dirname ${GRAVITY_URL}) ]; then
+  GRAVITY_FILE='/robotest-bin/'$(basename ${GRAVITY_URL})
+  EXTRA_VOLUME_MOUNTS=${EXTRA_VOLUME_MOUNTS:-}" -v "$(dirname ${GRAVITY_URL}):$(dirname ${GRAVITY_FILE})
+fi
```
wadells, line 48: File an enhancement bug to remove this dependency.  It should be quite unnecessary in this age.

build.assets/robotest/run.sh
```diff
@@ -3,55 +3,114 @@ set -o errexit
...
+		fi
+	done
+
+	if [ ! -z $ABORT ] ; then
+		exit 1 ;
+	fi
 }
 
-export EXTRA_VOLUME_MOUNTS=$(build_volume_mounts)
+if [ $DEPLOY_TO == "gce" ] ; then
```
wadells, line 64: DEPLOY_TO can't be anything else at this point. Consider refactoring everything related to this variable.

build.assets/robotest/run.sh
```diff
@@ -3,55 +3,114 @@ set -o errexit
...
+
+if [ -n "${GCL_PROJECT_ID:-}" ] ; then
+	check_files ${GOOGLE_APPLICATION_CREDENTIALS}
+fi
+
+CLOUD_CONFIG="
+installer_url: ${INSTALLER_FILE:-${INSTALLER_URL}}
+gravity_url: ${GRAVITY_FILE:-${GRAVITY_URL}}
+script_path: /robotest/terraform/${DEPLOY_TO}
+state_dir: /robotest/state
```
wadells, line 91: Turn the state_dir mount value into a variable.

wadells (commented):
First round of self review.

assets/robotest/current/status.yaml
```diff
@@ -0,0 +1,18 @@
...
+  template:
+    metadata:
+      name: echoserver-status
+    spec:
+      restartPolicy: OnFailure
+      containers:
+        - name: echoserver-status
+          image: quay.io/gravitational/debian-tall:buster
+          command:
+            - sh
```
r0mant, line 16: Nit: I would use absolute path (`/bin/sh`) to avoid issues if the container doesn't have proper PATH set for some reason. Same for `wget`.

assets/robotest/current/status.yaml
```diff
@@ -0,0 +1,18 @@
...
+            - -c
+            - wget --quiet --output-document=- echoserver:$ECHOSERVER_SERVICE_PORT
```
r0mant, line 18: Is this still going to print an error if the query fails even with the quiet flag set?

assets/robotest/upgrade-base/charts/echoserver/values.yaml
```diff
@@ -0,0 +1,2 @@
+# echoserverRegistry is redefined to "leader.telekube.local:5000" in gravity hooks
```
r0mant, line 1: `registry.local:5000`

assets/robotest/upgrade-base/app.yaml
```diff
@@ -0,0 +1,158 @@
+apiVersion: cluster.gravitational.io/v2
+kind: Cluster
+metadata:
+  name: robotest
```
r0mant, line 4: What do we need two robotest apps for, "upgrade-base" and "current"? To be able to test application-only upgrades when runtime doesn't change?

build.assets/robotest/nightly_config.sh
```diff
@@ -21,15 +21,21 @@ UPGRADE_MAP[6.2.0]="ubuntu:16"
 
 UPGRADE_VERSIONS=${!UPGRADE_MAP[@]}
 
+# The following breaks a dependency loop. We need upgrade versions to generate tarballs,
+# but we need tarball names to generate the full test config.
+if [[ ${1} == "upgradeexit" ]]  ; then
```
r0mant, line 26: I'm not sure what this is for. Could you explain in more detail?

build.assets/robotest/Makefile
```diff
@@ -0,0 +1,326 @@
...
+
+
+# The following target is a pattern rule for populating the tele binary cache.
+# Fulfills the $(ROBOTEST_TELE_BINARIES) targets.
+# .PRECIOUS because it is a cache shared across branches/builds.
+.PRECIOUS: $(ROBOTEST_TELE_CACHE)/tele-%
+$(ROBOTEST_TELE_CACHE)/tele-%: | $(ROBOTEST_CACHE_TMP) $(ROBOTEST_TELE_CACHE)
+	$(eval TMP := $(shell mktemp -d -p $(ROBOTEST_CACHE_TMP)))
+ifdef ENTERPRISE
+	wget --quiet https://get.gravitational.io/telekube/bin/$*/linux/x86_64/tele --output-document $(TMP)/tele
```
r0mant, line 290: Can we find a way to avoid placing enterprise-specific URLs in the OSS code? For example, instead of setting `ENTERPRISE` variable in the enterprise code, set `TELE_URL` or something like this and use it here.

build.assets/robotest/Makefile
```diff
@@ -0,0 +1,326 @@
...
+	cp $< $(ATOMIC)
+	mv $(ATOMIC) $@
+	rm -rf $(TMP)
+
+# The following target is a pattern rule for populating the upgrade image cache
+# Fulfills the $(ROBOTEST_UPGRADE_CACHE_IMAGES) targets.
+# .PRECIOUS because it is a cache shared across versions.
+.PRECIOUS: $(ROBOTEST_UPGRADE_IMAGE_CACHE)/robotest-%.tar
+$(ROBOTEST_UPGRADE_IMAGE_CACHE)/robotest-%.tar: $(ROBOTEST_TELE_CACHE)/tele-% $(ROBOTEST_UPGRADE_BASE_IMAGE_SRC) | $(ROBOTEST_UPGRADE_IMAGE_CACHE) $(ROBOTEST_CACHE_TMP)
+	$(eval TMP := $(shell mktemp -d -p $(ROBOTEST_CACHE_TMP)))
```
r0mant, line 272: I wonder if we should consider trapping `rm -rf $(TMP)` to make sure that the directory is removed even if, say, tele build below fails. Same for other targets that create temp. dirs.

assets/robotest/current/install.yaml
```diff
@@ -0,0 +1,24 @@
...
+kind: Job
+metadata:
+  name: echoserver-install
+  namespace: default
+spec:
+  template:
+    metadata:
+      name: echoserver-install
+    spec:
+      restartPolicy: Never
```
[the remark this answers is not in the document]
wadells: If the helm install fails, we have no reason to believe it would succeed if run again.

Makefile
```diff
@@ -530,28 +522,13 @@ $(TELEKUBE_OUT): packages
 		--skip-version-check \
 		-o $(TELEKUBE_OUT)
 
-$(GRAVITY_OUT):
```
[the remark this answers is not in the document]
wadells: I removed the code I added in my previous custom robotest image PR:

https://github.com/gravitational/gravity/pull/1967/files#diff-b67911656ef5d18c4ae36cb6741b7965R533-R537

Is your use of `make gravity tele` more recent than that?  I can definitely add these back in.

I added this code because robotest targets depend on $(GRAVITY_OUT) and $(TELE_OUT).  They used to explicitly do so, but this was unpleasant because these tele and gravity depend on phony targets... meaning they constantly trigger a full rebuild.  For the new `robotest-run-suite` and `robotest-run-nightly`, in this PR I implicitly declare the dependencies in a comment to avoid this.  Maybe these should be an "ordering dependency"  Its not great... but I also don't want to fix our entire dependency tree in this PR -- doing so would be a lot of work and orthogonal to the changes here.

assets/robotest/upgrade-base/app.yaml
wadells: We need two (or more) separate manifests for a couple reasons:

1) Old versions of gravity do not support all the manifest directives that current versions do.  For master (this PR) we test upgrades from:
```
6.2.0 6.2.5 6.3.18 7.0.0 7.0.12 7.0.13 7.0.16 (this last one floats)
```
In particular, 6.2.x doesn't support `storage/openebs`.  We can make the entirely reasonable argument that we don't need to test 6.2 for this branch.  However the mechanism is important because this code will be backported and we do need to test upgrades from 5.5.x in 7.0.x.  And there are manifest differences between 5.5..x and 7.1.x.  We need them to test contemporary features.  This is why I alluded to upgrade testing helping with backports yesterday.

2) In order to exercise the upgrade hook, we *want* the upgrade base application to be different than the current application. Ideally the hooks for the new version will exercise both gravity and application functionality that wasn't previously available. We're not there yet, but that is where I'm aiming.

To flip your question: Why don't we have more upgrade from versions?

Considering the list from earlier:
```
6.2.0 6.2.5 6.3.18 7.0.0 7.0.12 7.0.13 7.0.16 (this last one floats)
```

We could very well have one base image for 6.2.x and a different one for 6.3 and 7.0.x. Or a different version of the application for each upgrade from version, supporting exactly what that gravity version supports.

There are a couple difficulties here:

1) Maintaining a separate application for each upgrade base version is tedious. It means each manifest change is multiplied by 5-10 edits.  There is potential to put together some code to only abstract & maintain the differences.  But this seemed complex for a v1.  I'm not looking to build out yaml/manifest templating.

2) There is no good way to keep a manifest current for the floating version. It'd require that each time you cut a release, you (or some automation) cut a PR to each branch that could be upgraded to/from.  This is not worth it at this point.

2) Makefiles are insufficient to elegantly handle this.  The pattern rules I've written are already byzantine and difficult to understand, without having many-to-one or one-to-one manifest-to-upgrade-images.  This would be considerably easier using go & @knisbet's work in #1917 but I can't afford to tie my work to his -- it'd  delay availability by weeks-to-months and make ports... challenging.

build.assets/robotest/Makefile
wadells: The issue here is that OSS gravity doesn't provide a direct link to download tele that I've found - we need to download it as part of a tarball, unpack and then copy just the `tele` binary.  Whereas for enterprise we need to download just the binary, or authenticate to get.gravitational.io to get a release tarball.  I haven't been able to find a common download path.

If we wanted to properly abstract this difference in behavior, we'd need to either:
* Find a common download path.  Two unauthenticated release tarballs or two unauthenticated raw binaries.
* Declare & call a function pointer, if such a thing is even possible in a Makefile. This seems like an poor choice from a maintenance perceptive.
* Do some sort of `sed`-esque replacement of a `TELE_URL` variable to plug in the variety of upgrade-from versions.

Let me know what you think.  The value of keeping enterprise stuff out of OSS stuff is clear to me, but the path to doing so is not.

build.assets/robotest/Makefile
wadells: Makefiles have awful signal handling.  We'd need to shell out to bash.  This is possible, but it is adding yet another layer of complexity.  I'd considered this, but was hoping that we could wait until mage/go builds, where signal handling and defer/recover is first class.  Especially with the newly enlarged jenkins workspace giving us buffer for heavy PRs

As of this PR both `$(ROBOTEST_CACHE_TMP)` and `$(ROBOTEST_BUILD_TMP)` are underneath the build root.  The caches, along with any straggling temp files will be nuked by a `make clean`. If the user puts `$(ROBOTEST_CACHE_ROOT)` outside the build dir, leaking tmp directories is a concern -- but that isn't the case for jenkins (yet).  It is more of a concern for local development.  I'd like to defer this until either:

* We move `$(ROBOTEST_CACHE_TMP)` outside `$(BUILDDIR)` or
* We migrate to mage/go for our build files

assets/robotest/current/install.yaml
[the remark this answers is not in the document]
wadells: While I was developing this, when `helm install` failed I found no reason to believe it would succeed when run a second, third, or fourth time.

`OnFailure` resulted in failures on an exponential backoff basis, resulting in longer timeouts before failure and a slower development cycle.  Perhaps if I knew more about the failure modes here I could make a more educated choice about whether to abort or retry.

Makefile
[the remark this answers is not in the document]
wadells: This shouldn't affect `make gravity tele`, that uses the `$(BINARIES)` target.  `$(GRAVITY_OUT)` and `$(TELE_OUT)` are the full paths to the binaries, not the .PHONY targets you're calling.

build.assets/robotest/Makefile
a-palchikov: I can imagine having some e-specific make functions stored in e/ which could be conditionally included with:
```
-include e/makefile.conf
```
Then we can factor the tele pull process into a function that accepts the version and output path:
```
define pull_tele
	wget --quiet https/path/to/telekube/bin/$1/linux/x86_64/tele --output-document $2
	chmod u+x $2
endef
```

wadells: Thanks @a-palchikov.  I'll run with your suggestion.

build.assets/robotest/Makefile
```diff
@@ -0,0 +1,326 @@
...
+	$(eval TGT := $(TMP)/robotest-$*.tar)
+	$< build \
```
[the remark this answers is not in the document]
wadells: Should be fixed by 5c47ccf. :crossed_fingers:

build.assets/robotest/Makefile
```diff
@@ -0,0 +1,326 @@
...
+images: image-robotest images-upgrade image-opscenter
+
+.PHONY: image-robotest
+image-robotest: ## Create robotest cluster image using the current tele.
+image-robotest: $(ROBOTEST_IMAGE)
+
+$(ROBOTEST_IMAGE): $(TELE_OUT) $(ROBOTEST_IMAGE_SRC) | $(ROBOTEST_IMAGEDIR) $(ROBOTEST_BUILD_TMP)
+	$(eval TMP := $(shell mktemp -d -p $(ROBOTEST_BUILD_TMP)))
+	$(eval TGT := $(TMP)/robotest-$(GRAVITY_VERSION).tar)
+	$(TELE_OUT) build \
```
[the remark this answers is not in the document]
wadells: Should be fixed by 5c47ccf. crossed_fingers

build.assets/robotest/tele_build.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+#
+# This file provides the following features to the tele build process:
+#
+#  1) Atomicity.  The TARGET is either complete or absent. (So long as long as
```
a-palchikov, line 5: Either `So long` or `as long as`

build.assets/robotest/tele_build.sh
```diff
@@ -0,0 +1,31 @@
...
+#     renames). This means no half finished files for Make to trip on.
+#
+# The following environment variables must be specified by the caller:
+#
+# BUILD_TMP - Partially constructed images are built here. Must be on the same filesystem as TARGET for atomic move operations.
+# TARGET - Where the cluster image should end up
+# TELE  - The tele to build the image, should typically be equal to VERSION
+# VERSION - The application version to use in the cluster image
+# APP_YAML - The image manifest to be built.
+# STATE_DIR - The gravity state dir where packages are cached/drawn from. May be a shared with parallel invocations.
```
a-palchikov, line 16: ~a~ shared?

build.assets/robotest/tele_build.sh
```diff
@@ -0,0 +1,31 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+TMP=$(mktemp -d -p $BUILD_TMP)
+trap "rm -rf $TMP" exit
+TGT=$TMP/$(basename $TARGET)
```
a-palchikov, line 25: can this have whitespace?

build.assets/robotest/tele_build.sh
```diff
@@ -0,0 +1,31 @@
...
+
+$TELE build \
```
a-palchikov, line 27: I'd also take into account that paths can contain whitespace.

build.assets/robotest/Makefile
wadells: I've refactored out all the tmp stuff into shell scripts with traps at of c91653e. Although I think ithe code is a bit more difficult to undestand, it aligned nicely with containerizing the `tele build` which was needed to handle `~/.tsh` pollution across a large range of `tele`s.

build.assets/robotest/Makefile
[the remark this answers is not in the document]
wadells: Unfortunately 5c47ccf failed because not all upgrade from `tele` binaries have the fix from #2023, leading to `[ERROR]: no credentials for https://get.gravitational.io:443` if `--hub` is blanket applied to old tele builds.

I'd rather not branch the upgrade base tele build logic at this point, so I've opted to containerize all tele builds to isolate them from ~jenkins state.  We'll see how this goes shortly.

build.assets/robotest/tele_build.sh
wadells: I added a bunch of quoting in 6deb147.  I think I caught all the possible filepaths.

build.assets/robotest/tele_build.sh
wadells: I added a bunch of quoting in 6deb147.  I think I caught all the possible filepaths.

assets/robotest/upgrade-base/charts/echoserver/values.yaml
wadells: Changed in 900be40.

assets/robotest/current/postupdate.yaml
```diff
@@ -0,0 +1,22 @@
...
+    spec:
+      restartPolicy: OnFailure
+      containers:
+        - name: upgrade
+          image: quay.io/gravitational/debian-tall:buster
+          command:
+            - /usr/local/bin/helm
+            - upgrade
+            - --set
+            - echoserverRegistry=leader.telekube.local:5000
```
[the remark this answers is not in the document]
wadells: Changed in 900be40.

assets/robotest/current/install.yaml
```diff
@@ -0,0 +1,24 @@
...
+        - name: echoserver-install
+          image: quay.io/gravitational/debian-tall:buster
+          command:
+            - /usr/local/bin/helm
+            - install
+            - /var/lib/gravity/resources/charts/echoserver
+            - --name
+            - echoserver
+            - --set
+            - echoserverRegistry=leader.telekube.local:5000
```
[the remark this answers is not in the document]
wadells: Changed in 900be40.

assets/robotest/current/status.yaml
wadells: Fixed in 2a82ca9.

build.assets/robotest/Makefile
wadells: I've got a working version of this change in place as of c91653e.

I use a shell script instead of a included Makefile to accommodate tmpdir cleanup per https://github.com/gravitational/gravity/pull/2133#discussion_r489093866.

I want to make one more structural change in this area before calling it done.  That is:

Migrate all content from `build.assets/robotest` to `assets/robotest`.  This will give a nice parity between the enterprise `assets/robotest` and the oss `assets/robotest`.

assets/robotest/current/status.yaml
wadells: Good point.  I chose quiet because of `--output-document=-`  Having wget's verbose (default) output interlaced was unsightly.  I changed this to `--no-verbose` in 2a82ca9.

build.assets/robotest/Makefile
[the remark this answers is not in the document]
wadells: Refactored to sandbox tele builds within a docker container in c91653e.  This prevents any possible contamination from ~jenkins.  It took a couple more patches to get the new dockerized builds working, but now that they are (f86dc73) we don't see this issue any more.

build.assets/robotest/run.sh
wadells: https://github.com/gravitational/robotest/issues/232

build.assets/robotest/Makefile
wadells: Parallelism won't work with a shared cache, as discussed in ecc6916.

build.assets/robotest/tele_build.sh
wadells: Fixed in cbb8da7.

build.assets/robotest/tele_build.sh
wadells: Fixed in cbb8da7.

build.assets/robotest/Dockerfile
wadells: This entire file has been replaced by a simpler `--user=$(id -u):$(id -g)` in 
2e7bfe1

build.assets/robotest/Makefile
wadells: Moved & integrated in f5a5aa3.

.gitignore
```diff
@@ -18,3 +18,5 @@ gravity.log
 terraform.tfstate*
 # Ansible retry files
 *.retry
+# Robotest credentials
+robotest-gcp-credentials.json
```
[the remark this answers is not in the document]
wadells: Moved into the robotest directory in dfe9106.

build.assets/robotest/pr_config.sh
wadells: Refactored nightly to include a newer version of this in 43c08e9.

build.assets/robotest/run.sh
wadells: Renamed in 40600dc49b149ad223bd7671aa3696da5a9abff4.

build.assets/robotest/nightly_config.sh
wadells: I added extra documentation on these config files & how they're used in ccff414d9431d0e51400df43114baa851c31d8ee.  Check out `assets/robotest/config/README.md` and see if it sufficiently clarifies the implementation.

build.assets/robotest/Makefile
wadells: Take a look at `assets/robotest/download_tele.sh` in the two repos and see if the new implementation suits your fancy.

assets/robotest/upgrade-base/charts/echoserver/Chart.yaml
wadells: Changed in 3355ed7.

Jenkinsfile
```diff
@@ -33,69 +33,85 @@ def setRobotestParameters() {
...
   }
   throttle(['robotest']) {
-    node {
-      stage('build-and-test') {
-        parallel (
-        build : {
+    stage('run robotest') {
+      if (runRobotest) {
+        withCredentials([
+					file(credentialsId:'ROBOTEST_LOG_GOOGLE_APPLICATION_CREDENTIALS', variable: 'GOOGLE_APPLICATION_CREDENTIALS'),
```
wadells, line 69: Looks like use of spaces vs tabs is inconsistent in this file. Probably need to install a Groovy indent plugin.

assets/robotest/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+```bash
+# ssh keys must not have a password. Generating unique dedicated keys is recommended.
+export ROBOTEST_SSH_KEY=~/.ssh/robotest
+export ROBOTEST_SSH_PUB=~/.ssh/robotest.pub
+# See https://cloud.google.com/docs/authentication/production
+export ROBOTEST_GOOGLE_APPLICATION_CREDENTIALS=~/.config/gcloud/legacy_credentials/robotest.json
+```
+
+Once these are set, robotest with:
```
wadells, line 47: *run* robotest with

Jenkinsfile
wadells: Fixed in f3a02f8.

assets/robotest/README.md
wadells: Fixed in 593ede2.

assets/robotest/config/pr.sh
```diff
@@ -59,18 +64,25 @@ function build_install_suite {
...
 }
 
-SUITE=$(build_install_suite)
-SUITE="$SUITE $(build_resize_suite)"
-SUITE="$SUITE $(build_upgrade_suite)"
-
-echo "$SUITE" | tr ' ' '\n'
-
-export SUITE UPGRADE_VERSIONS
+if [[ ${1} == "upgradeversions" ]] ; then
```
r0mant, line 74: Maybe not in this PR but I would seriously consider moving all of this logic for generating test suite configurations into a Go program so the only shell/makefile code we have is a simple harness for building tarballs and launching robotest. Just don't want to end up in a situation with a mountain of non-trivial shell scripts to maintain.

How hard do you think it would be to do?

assets/robotest/current/install.yaml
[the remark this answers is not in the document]
r0mant: > when helm install failed I found no reason to believe it would succeed when run a second, third, or fourth time

Technically that's not entirely true, helm install can fail for example due to temporary etcd issues and succeed on the retry.

assets/robotest/config/pr.sh
wadells: I couldn't agree more.  These old shell scripts (that I've evolved considerably) are difficult to follow & debug. I don't think it would be hard.  I was actually tempted to rewrite in go it for this PR, but it felt like unnecessary scope creep.

r0mant (approved):

assets/robotest/current/install.yaml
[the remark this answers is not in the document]
wadells: Changed to `OnFailure` in 43c65ee.

assets/robotest/tele_build.sh
```diff
@@ -0,0 +1,53 @@
...
+#  1) Atomicity.  The TARGET is either complete or absent. (As long as
+#     BUILD_TMP is on the same filesystem as TARGET and that fs supports atomic
+#     renames). This means no half finished files for Make to trip on.
+#  2) Isolation.  Various old teles *always* check ~/.tsh for a logged in cluster
+#     and use it if available.  There is no way to disable this behavior.  Thus we
+#     sandbox these processes in a container where ~/.tsh doesn't exist.
+#
+# The following environment variables must be specified by the caller:
+#
+# IMAGE - A docker cotainer that the build will run in
```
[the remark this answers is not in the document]
wadells: Fixed in 32dd236.

assets/robotest/config/pr.sh
wadells: I've filed https://github.com/gravitational/gravity/issues/2147 tracking this work item and will triage it as I asses next workitems for https://github.com/gravitational/gravity/projects/5

assets/robotest/Makefile
```diff
@@ -0,0 +1,306 @@
...
+# ROBOTEST_IMAGE_SRC are the files that affect the robotest application.  If any
+# of these change, all robotest images must be rebuilt.
+ROBOTEST_IMAGE_SRC := $(shell find $(ROBOTEST_IMAGE_SRCDIR) -type f)
+# ROBOTEST_IMAGE_MANIFEST is the gravity application manifest to build into robotest images.
+ROBOTEST_IMAGE_MANIFEST = $(ROBOTEST_IMAGE_SRCDIR)/app.yaml
+
+# ROBOTEST_UPGRADE_BASE_IMAGE_SRCDIR is the directory containing all sources to build the
+# robotest upgrade base application. It is mounted into a build container.
+ROBOTEST_UPGRADE_BASE_IMAGE_SRCDIR = $(TOP)/upgrade-base
+# ROBOTEST_UPGRADE_BASE_IMAGE_SRC is the gravity application manifest to build into robotest images.
```
a-palchikov, line 101: from the implementation it looks as a list of files in the `$ROBOTEST_UPGRADE_BASE_IMAGE_SRCDIR` directory - copy/paste comment?

assets/robotest/Makefile
```diff
@@ -0,0 +1,306 @@
...
+ROBOTEST_OPSCENTER_IMAGE = $(ROBOTEST_IMAGEDIR)/opscenter-$(GRAVITY_VERSION).tar
+# ROBOTEST_TELEKUBE_IMAGE is robotest's copy of the telekube image, used for install testing.
+ROBOTEST_TELEKUBE_IMAGE = $(ROBOTEST_IMAGEDIR)/telekube-$(GRAVITY_VERSION).tar
+# ROBOTEST_UPGRADE_VERSIONS is a list of gravity semvers that will be used in upgrade testing.
+ROBOTEST_UPGRADE_VERSIONS := $(shell cd $(TOP)/config && $(ROBOTEST_CONFIG_SCRIPT) upgradeversions)
+# ROBOTEST_TELE_BINARIES is the list of build specific tele binaries needed to
+# build all robotest upgrade images.
+ROBOTEST_TELE_BINARIES := $(addprefix $(ROBOTEST_TELE_CACHE)/tele-, $(ROBOTEST_UPGRADE_VERSIONS))
+# ROBOTEST_UPGRADE_IMAGES is the list of build specific upgrade images, suitable
+# for mounting into the robotest docker container. These are not shared stated.
```
a-palchikov, line 118: stated -> state?

assets/robotest/Makefile
```diff
@@ -0,0 +1,306 @@
...
+#
+#   [ERROR]: failed to sync packages for runtime version 7.0.16
+#       package(gravitational.io/bandwagon:6.0.1) already exists
+#
+# Possibly due to https://github.com/gravitational/gravity/issues/2060
+#
+# Because we must support old tele binaries to build old image versions, these races seem
+# unavoidable.
+#
+# Interestingly, I've never seen sync races happen with oss tele and the s3Syncer
```
a-palchikov, line 211: This is b/c oss syncer implementation always downloads into an abritrary temporary directory.

assets/robotest/README.md
```diff
@@ -0,0 +1,165 @@
...
+```
+make -C assets/robotest images
+```
+
+This may take a while. When complete `build/<gravity>/robotest/images` will be
+populated with cluster images needed to run robotest.
+
+### Running Robotest
+
+Before runnning robotest, set the following environment variables needed to
```
a-palchikov, line 36: runn~~n~~ing

assets/robotest/README.md
```diff
@@ -0,0 +1,165 @@
...
+Robotest image construction uses the following strategy:
+
+For the release under test, build using `TELE_OUT` and `PACKAGES_DIR` from
+the Gravity Makefiles.  This *must* occur after a successful `make telekube`
+(which populates `PACKAGES_DIR` with a base image).
+
+For upgrade to/from versions:
+
+ 1. Download & cache appropriate tele from the gravity release infrastructure.
+ 2. Build the the robotest image using the `upgrade-base` manifest
```
a-palchikov, line 147: Build the ~~the~~

assets/robotest/README.md
```diff
@@ -0,0 +1,165 @@
...
+## Tips & Tricks
+
+Building robotest images and running robotest is time consuming, 10-20 of minutes
+for a clean build and an hour for a build & PR run. To improve the development process
+try the following:
+
+### Using Alternative Configurations
+
+By default, all targets use the `pr` configuration found in the `config/`
+directory. A `noop` configuration this does not rely on any upgrade images is
```
a-palchikov, line 62: this -> that

assets/robotest/README.md
```diff
@@ -0,0 +1,165 @@
...
+(including `tele`, `gravity`, and the package cache):
+
+```
+make -C assets/robotest GRAVITY_BUILDDIR=build/current images
+```
+
+## File Structure
+
+ * `Makefile` is the entrypoint. Run `make help` to see targets.
+ * `config/` contain branch specific suite configurations. See the README.md therein.
```
a-palchikov, line 110: contain**s**

assets/robotest/config/README.md
```diff
@@ -0,0 +1,41 @@
...
+6.1.38 7.0.7 7.0.12 7.0.16
+```
+
+`./script.sh configuration` will have several environment variables set to
+allow the test configuration to refer to installer tars:
+ * OPSCENTER_URL
+ * INSTALLER_URL
+
+`./script.sh configuration` must return a robotest suite configuration as described in
+https://github.com/gravitational/robotest/blob/master/suite/README.md.  For instance:
```
a-palchikov, line 24: nit: I would use a specific tag/commit to refer to the suite README.md if possible. I keep running in situations with other repositories where links go stale pretty fast.

assets/robotest/download_tele.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+#
+# This file is used to download tele binaries.  It provies the following benefits:
```
a-palchikov, line 3: provides

assets/robotest/download_tele.sh
```diff
@@ -0,0 +1,28 @@
...
+#
+#  1) Atomicity.  The TARGET is either complete or absent. (So long as long as
```
a-palchikov, line 5: `So long` or `as long as`

assets/robotest/download_tele.sh
```diff
@@ -0,0 +1,28 @@
...
+#     BUILD_TMP is on the same filesystem as TARGET and that fs supports atomic
+#     renames). This means no half finished files for Make to trip on.
+#  2) The file can be overridden/replaced in the make file to allow enterprise
```
a-palchikov, line 8: nit: make file -> Makefile

assets/robotest/download_tele.sh
```diff
@@ -0,0 +1,28 @@
...
+set -o pipefail
+set -o xtrace
+
+TMP=$(mktemp -d -p "$BUILD_TMP")
+trap "rm -rf \"$TMP\"" exit
+
+wget --no-verbose "https://get.gravitational.com/gravity-$VERSION-linux-x86_64-bin.tar.gz" --directory-prefix "$TMP"
+tar -xzf "$TMP/gravity-$VERSION-linux-x86_64-bin.tar.gz" -C "$TMP"
+# touch because unpacking from gnu tar keeps the original mtime
+touch "$TMP/tele"
```
a-palchikov, line 27: I'm pretty sure tar supports `--touch`

assets/robotest/run.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readonly TARGET=${1:?Usage: [path to config]}
+
+GRAVITY_URL=${GRAVITY_URL:?Set GRAVITY_URL to the current gravity binary}
+IMAGEDIR_MOUNTPOINT=${IMAGEDIR_MOUNTPOINT:? Set IMAGEDIR_MOUNTPOINT to the mountpoint where imags will be availabe inside the container}
```
a-palchikov, line 9: imag**e**s

assets/robotest/run.sh
```diff
@@ -0,0 +1,118 @@
...
+INSTALLER_URL=${INSTALLER_URL:?Set INSTALLER_URL to the container local default robotest image}
+OPSCENTER_URL=${OPSCENTER_URL:?Set OPSCENTER_URL to the container local opscenter image}
+TELEKUBE_URL=${TELEKUBE_URL:?Set TELEKUBE_URL to the container local opscenter image}
```
a-palchikov, line 12: opscenter -> telekube

assets/robotest/run.sh
```diff
@@ -0,0 +1,118 @@
...
+# what should happen with provisioned VMs on individual test success or failure
+DESTROY_ON_SUCCESS=${DESTROY_ON_SUCCESS:-true}
+DESTROY_ON_FAILURE=${DESTROY_ON_FAILURE:-true}
+
+EXTRA_VOLUME_MOUNTS="-v $IMAGEDIR:$IMAGEDIR_MOUNTPOINT"
+
+export INSTALLER_URL TELEKUBE_URL OPSCENTER_URL IMAGEDIR_MOUNTPOINT
+SUITE=$($TARGET configuration)
+
+# GRAVTIY_FILE/GRAVITY_URL specify the location of the up-to-date gravity binary
```
a-palchikov, line 43: GRAV**IT**Y_FILE

assets/robotest/upgrade-base/app.yaml
a-palchikov: Just to chime in on the topic of yaml templating - if it's ever a thing, I think [cue](github.com/cuelang/cue/) would be perfect for the job.

a-palchikov (approved):
This is great work, Walt.
lgtm with just a couple of typos.

Jenkinsfile
```diff
@@ -7,95 +7,111 @@ def propagateParamsToEnv() {
...
+    if (env.KEEP_PARAMETERS == 'true') {
+      echo "KEEP_PARAMETERS detected. Ignoring Jenkins job parameters from Jenkinsfile."
+    } else {
+      echo "Overwriting Jenkins job parameters with parameters from Jenkinsfile."
+      setRobotestParameters()
+      propagateParamsToEnv()
+    }
+  }
+  runRobotest = (env.RUN_ROBOTEST == 'run')
+    stage('build robotest images') {
```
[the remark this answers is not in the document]
wadells: I'm not fully sure I understand what you're asking.

I put the build in a separate step to help tracking timing & visualization.  The `run` target will build all images necessary if they're not already complete -- and we could call it directly without this pre-bui.d But I found it useful see whether building images or running robotest was the problem.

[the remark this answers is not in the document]
a-palchikov: No, it just looked as if `runRobotest = ...` was starting a new scope where the following `stage` was nested. I just meant it could be on the same indentation level:
```
runRobotest = (env.RUN_ROBOTEST == 'run')
stage('build robotest images') {
...
```

Makefile
```diff
@@ -530,28 +522,13 @@ $(TELEKUBE_OUT): packages
...
-#
-# builds robotest installer
-#
-.PHONY: robotest-image
-robotest-image: GRAVITY=$(GRAVITY_OUT) --state-dir=$(PACKAGES_DIR)
-robotest-image: $(ROBOTEST_OUT)
-
-ROBOTEST_TAR_SRC=$(shell find $(ASSETSDIR)/robotest/ -type f)
+.PHONY: robotest-run-suite
+robotest-run-suite: # depends on: telekube opscenter $(TELE_OUT) $(GRAVITY_OUT)
```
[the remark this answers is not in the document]
wadells: Yes.

These .`.PHONY` upstream targets don't track dependency relations between sources and artifacts, meaning if `robotest-run-suite` depends on them directly it will always rebuild `production telekube opscenter`. This is tedious for anything but an automated builder that does it from scratch each time.  `robotest-run-suite` used to not declare any dependencies (which is a lie) -- and these commented out dependencies are my compromise.

We might similarly ask why all the `packages` targets don't depend on `$(GRAVITY_OUT)`? Or why doesn't  `$(TELE_OUT)` doesn't depend on `$(TELE_OUT)`?  They should, but to do that we need proper dependency relations between `$(GRAVITY_OUT)` and `$(TELE_OUT)` and their upstreams.  This quickly balloons into fixing the entire build.

I felt fixing latent broken dependency relationships  was out of scope. Especially as make's timestamp based tracking leaves a lot to be desired.  If we're going to invest this sort of energy into the build, it should be coupled with a modern build system IMO.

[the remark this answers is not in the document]
a-palchikov: Agree - we should better focus on mage progress instead of fixing makefiles.

assets/robotest/Makefile
```diff
@@ -0,0 +1,306 @@
...
+endif
+GRAVITY_BUILDDIR ?= $(BUILDDIR)/$(GRAVITY_VERSION)
+TELE_OUT ?= $(GRAVITY_BUILDDIR)/tele
+GRAVITY_OUT ?= $(GRAVITY_BUILDDIR)/gravity
+OPSCENTER_OUT ?= $(GRAVITY_BUILDDIR)/opscenter.tar
+TELEKUBE_OUT ?= $(GRAVITY_BUILDDIR)/telekube.tar
+PACKAGES_DIR ?= $(GRAVITY_BUILDDIR)/packages
+# ROBOTEST_DOWNLOAD_TELE_SCRIPT allows the e Makefile to override how tele binaries are sourced.
+ROBOTEST_DOWNLOAD_TELE_SCRIPT ?= $(TOP)/download_tele.sh
+# ROBOTEST_CACHE_FLAVOR may be overriden to specify a non-compatible cache.
```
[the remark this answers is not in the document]
wadells: It is just the name of the cache sub directory.  I was having issues with a long-lived cache that lived outside `$(GRAVITY_BUILDDIR)` without this -- oss teles & enterprise teles were getting confused.

Jenkinsfile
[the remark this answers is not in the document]
wadells: Thanks! Fixed in 4720bf4.

assets/robotest/download_tele.sh
wadells: Changed in c2a13c4.

assets/robotest/config/README.md
wadells: Pinned in dcc3d40.

assets/robotest/Makefile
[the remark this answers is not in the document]
wadells: I updated the comment to hopefully clarify this usage in d2cfbf4.

assets/robotest/Makefile
```diff
@@ -0,0 +1,306 @@
...
+ROBOTEST_CACHEDIR ?= $(ROBOTEST_CACHE_ROOT)/$(ROBOTEST_CACHE_FLAVOR)
+endif
+# ROBOTEST_GRAVITY_PKG_CACHE is a gravity state dir storing packages
+# downloaded from the default distribution infrastructure.
+# ROBOTEST_GRAVITY_PKG_CACHE may be shared between many parallel robotest builds
+# in CI infra. A reasonable override for personal use is sharing with:
+#   $(HOME)/.gravity
+ROBOTEST_GRAVITY_PKG_CACHE ?=  $(ROBOTEST_CACHEDIR)/gravity-state
+# ROBOTEST_CACHE_TMP is a staging directory for building/copying/unpacking content
+# before an atomic rename into the. Must be on the same filesystem as ROBOTEST_TELE_CACHE
```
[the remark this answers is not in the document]
wadells: Fixed in d2cfbf4.

assets/robotest/Makefile
```diff
@@ -0,0 +1,306 @@
...
+ROBOTEST_CACHE_TMP = $(ROBOTEST_CACHEDIR)/tmp
+# ROBOTEST_TELE_CACHE is a filesytem cache storing prebuilt tele binaries for
+# different gravity versions. ROBOTEST_TELE_CACHE may be shared between many
+# parallel robotest builds in CI infra.
+ROBOTEST_TELE_CACHE ?= $(ROBOTEST_CACHEDIR)/tele
+# ROBOTEST_CACHES are all the directories on this machine used to speed up
+# upgrade artifact contruction by avoiding redownloading/rebuilding released
```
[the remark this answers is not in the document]
wadells: Fixed in d2cfbf4.

assets/robotest/Makefile
wadells: Fixed in d2cfbf4.

assets/robotest/Makefile
wadells: Fixed in d2cfbf4.

assets/robotest/README.md
wadells: Fixed in d2cfbf4.

assets/robotest/README.md
wadells: Fixed in d2cfbf4.

assets/robotest/README.md
wadells: Fixed in d2cfbf4.

assets/robotest/README.md
wadells: Fixed in d2cfbf4.

assets/robotest/download_tele.sh
wadells: Fixed in d2cfbf4.

assets/robotest/download_tele.sh
wadells: Fixed in d2cfbf4.

assets/robotest/download_tele.sh
wadells: Fixed in d2cfbf4.

assets/robotest/run.sh
wadells: Fixed in d2cfbf4.

assets/robotest/run.sh
wadells: Fixed in d2cfbf4.

assets/robotest/run.sh
wadells: Fixed in d2cfbf4.

PR #2138, opened 2020-09-18
Reconcile serf cluster

bernardjkim:
## Description

This PR adds a serf cluster reconcile loop to the planet agent. Planet will periodically check for missing serf members and re-join them to the serf cluster.

Reverts https://github.com/gravitational/gravity/pull/2111. Previous PR degraded system status if an offline node was detected. This change causes the system status to temporarily degrade during shrink operations. The serf member is evicted before the member is removed from gravity's list of peers and the system status will be degraded until gravity removes the member from the list of peers.
I think it might be better to rely on this serf reconciler that is being added in this PR and allow the cluster to be healthy with offline nodes temporarily instead of having the cluster fall into a degraded state during shink operations?
## Type of change

* Revert https://github.com/gravitational/gravity/pull/2111

* Refs https://github.com/gravitational/gravity/issues/2100

* Requires https://github.com/gravitational/planet/pull/752

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

**Verify serf cluster is reconciled after network partition**
- Install 3 node cluster.
- Configure serf `reconnect_timeout` so `failed` serf nodes are kicked from the cluster after 30s.
```
{
        "reconnect_timeout": "30s"
}
```
- Create network partition using iptables.
```
node-3:/$ iptables -A OUTPUT -s 172.28.128.101,172.28.128.102 -j DROP
node-3:/$ iptables -A INPUT -s 172.28.128.101,172.28.128.102 -j DROP
```
- Wait for serf node to be kicked from the serf cluster.
```
node-1:/$ serf members
172_28_128_102.dev.test  172.28.128.102:7496  alive  publicip=172.28.128.102,role=master
172_28_128_101.dev.test  172.28.128.101:7496  alive  role=master,publicip=172.28.128.101

node-2:/$ serf members
172_28_128_102.dev.test  172.28.128.102:7496  alive  role=master,publicip=172.28.128.102
172_28_128_101.dev.test  172.28.128.101:7496  alive  role=master,publicip=172.28.128.101

node-3:/$ serf members
172_28_128_103.dev.test  172.28.128.103:7496  alive  role=master,publicip=172.28.128.103
```
- Remove iptables rules & verify serf cluster is reconciled (reconcile interval set for 10 min).


lib/status/status.go
```diff
@@ -522,15 +522,6 @@ func fromPlanetAgent(ctx context.Context, local bool, servers []storage.Server)
 		nodes = fromSystemStatus(*status)
 	}
 
-	// Collected system status may exclude the status of some nodes due to
-	// network partition. If offline nodes are detected, degrade system status.
-	for _, node := range nodes {
```
r0mant, line 527: Is this change intended?

bernardjkim: Yeah. While testing shrink op I realized that this added check causes the system status to fall into a degraded state temporarily during the shrink operation.

Since we are adding the serf reconciler, I figured it would be better to have the system status show healthy cluster status temporarily if serf is out of sync than to have the system status fall into a degraded state during shrink ops.

a-palchikov (approved):

r0mant (approved):

PR #2142, opened 2020-09-21
LIcense documentation steps update

stevenGravy:
## Description
Change numbers (1. 2. 3....) to Step 1, step 2... to prevent the same number from repeating in display.

## Type of change
documentation


a-palchikov (approved):
lgtm I guess - could you elaborate which number is repeating that this chage prevents?

stevenGravy:
@a-palchikov  you would see 1. 1. 1.... instead of 1. 2. 3.

Issue #2140, opened 2020-09-19
Upgrade can fail with prereqs check if etcd is slow


r0mant:
### Description

**What happened**:

If etcd is having issues during an upgrade (e.g. experiencing timeouts), a phase can fail with an error like:

```
Sep 18 23:31:35 controller-1 gravity[6629]: 2020-09-18T23:31:35Z INFO [FSM:REMOT] "[ERROR]: required phase \"/app\" is not completed\n" CMD:/var/lib/gravity/site/update/agent/gravity#1 gravity:[plan execute --phase /gc/worker-3 --operation-id 08ab9825-62a2-4a3c-a934-c2524913f5a2] server:worker-3/172.31.3.7 client/agent.go:143
```

The reason is that when executing the phase on `worker-3`, it still doesn't "see" the most up-to-date plan because of etcd issues and thus thinks that the pre-req phase hasn't been executed yet while in fact it had.

When executing a phase and reconciling the plan, we should try to account for such errors and retry pre-reqs check / reconciliation several times.

### Environment

 - Gravity version [e.g. 7.0.11]: All

PR #2144, opened 2020-09-21
[6.1.x] use etcd quorum reads for querying the operation plan

a-palchikov:
## Description

Use consistent (quorum) reads when querying the operation plan to avoid reading stale values.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/2140

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

This PR is under testing

### DIFF
--- a/lib/fsm/format.go
+++ b/lib/fsm/format.go
@@ -59,6 +59,7 @@ func FormatOperationPlanJSON(w io.Writer, plan storage.OperationPlan) error {
 func FormatOperationPlanText(w io.Writer, plan storage.OperationPlan) {
 	var t tabwriter.Writer
 	t.Init(w, 0, 10, 5, ' ', 0)
+	fmt.Fprintln(&t, "Operation: ", OperationString(plan))
 	common.PrintTableHeader(&t, []string{"Phase", "Description", "State", "Node", "Requires", "Updated"})
 	for _, phase := range plan.Phases {
 		printPhase(&t, phase, 0)
--- a/lib/storage/keyval/etcd.go
+++ b/lib/storage/keyval/etcd.go
@@ -35,7 +35,6 @@ import (
 	"github.com/gravitational/coordinate/leader"
 	"github.com/gravitational/trace"
 	"github.com/jonboulle/clockwork"
-	"github.com/sirupsen/logrus"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 )
@@ -108,9 +107,9 @@ func LocalEtcdConfig(retryTimeout time.Duration) (*ETCDConfig, error) {
 // EtcdBackend enables access to etcd-specific features
 type EtcdBackend interface {
 	storage.Backend
-	// CopyWithOptions creates a copy of this backend
+	// CloneWithOptions creates a shallow copy of this backend
 	// with the specified options applied
-	CopyWithOptions(opts ...EtcdOption) storage.Backend
+	CloneWithOptions(opts ...EtcdOption) storage.Backend
 }
 
 // EtcdOption is a functional option to configure an etcd backend
@@ -120,12 +119,12 @@ type EtcdOption func(*etcdOptions)
 // e.g. return the latest committed value applied in a quorum of members
 func WithReadQuorum(quorum bool) EtcdOption {
 	return func(config *etcdOptions) {
-		logrus.WithField("quorum", quorum).Info("Specify quorum reads.")
 		config.GetOptions.Quorum = quorum
 	}
 }
 
 type etcdOptions struct {
+	// GetOptions specifies the options for Get API
 	GetOptions client.GetOptions
 }
 
--- a/lib/storage/keyval/etcd.go
+++ b/lib/storage/keyval/etcd.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/gravitational/gravity/lib/defaults"
 	"github.com/gravitational/gravity/lib/state"
+	"github.com/gravitational/gravity/lib/storage"
 	"github.com/gravitational/gravity/lib/utils"
 
 	"github.com/cenkalti/backoff"
@@ -34,6 +35,7 @@ import (
 	"github.com/gravitational/coordinate/leader"
 	"github.com/gravitational/trace"
 	"github.com/jonboulle/clockwork"
+	"github.com/sirupsen/logrus"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 )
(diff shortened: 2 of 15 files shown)


lib/update/reconciler.go
```diff
@@ -65,34 +67,31 @@ func (r *reconciler) ReconcilePlan(ctx context.Context, plan storage.OperationPl
...
 	}
-
-	return nil
+	return trace.Wrap(r.syncChangelog(ctx, r.backend, r.localBackend))
 }
 
 // syncChangelog will sync changelog entries from src to dst storage
-func (r *reconciler) syncChangelog(src storage.Backend, dst storage.Backend) error {
-	return trace.Wrap(SyncChangelog(src, dst, r.cluster, r.operationID))
+func (r *reconciler) syncChangelog(ctx context.Context, src storage.Backend, dst storage.Backend) error {
```
r0mant, line 88: I'm wondering if retrying reconciliation would actually help. I don't recall seeing reconciliation errors in the logs so I wonder if the problem may actually be a lever deeper, with etcd itself returning stale data due to its leader election and other timeouts? Maybe it would be better to retry the pre-requisites check itself several times to account for situations where etcd needs some time to figure out its leader elections and start returning correct data under load?

a-palchikov: Not sure how retrying the pre-requisites check can help mitigate a stale read? If we're seeing a stale value from etcd, then we will not be able to tell whether it is stale or up-to-date unless we request the read to be from the quorum.

lib/fsm/format.go
```diff
@@ -59,6 +59,7 @@ func FormatOperationPlanJSON(w io.Writer, plan storage.OperationPlan) error {
 func FormatOperationPlanText(w io.Writer, plan storage.OperationPlan) {
 	var t tabwriter.Writer
 	t.Init(w, 0, 10, 5, ' ', 0)
+	fmt.Fprintln(&t, "Operation: ", OperationString(plan))
```
a-palchikov, line 62: When display operation plan, I always missed the details about the operation itself.

lib/update/reconciler.go
```diff
@@ -84,22 +87,28 @@ func (r *reconciler) trySyncChangelogFromEtcd(ctx context.Context) error {
...
+		return trace.Wrap(r.syncChangelog(ctx, r.backend, r.localBackend))
 	}
 
 	return nil
 }
 
 // syncChangelog will sync changelog entries from src to dst storage
-func (r *reconciler) syncChangelog(src storage.Backend, dst storage.Backend) error {
-	return trace.Wrap(SyncChangelog(src, dst, r.cluster, r.operationID))
+func (r *reconciler) syncChangelog(ctx context.Context, src, dst storage.Backend) error {
```
a-palchikov, line 97: I've wrapped the query in a retry loop but left the original precondition that skips the reads if the cluster is not healthy. 

Initially I tried to work it around by first introducing a stricter check for when to skip the reads by testing for the etcd service being masked (which is the case once it has been shutdown on the node) but it does not work in a multi-node setting until all etcd instances have been shut down.
I then attempted to explicitly skip plan syncs by configuring it as an attribute of the plan step and checking it in the engine but it required a lot of changes and I failed to come up with a way to pair it with the FSM as it would require FSM to know about the reconciler and be able to provide custom handling during upgrades.
In the end, I figured it was all not worth the effort and the changes.

tool/gravity/cli/rpcagent.go
```diff
@@ -544,6 +545,28 @@ func getGravityPackage() loc.Locator {
 	}
 }
 
+func (r deployAgentsRequest) String() string {
```
a-palchikov, line 548: I added this to avoid the request logging binary data so we only log the relevant parts.

lib/storage/keyval/etcd.go
```diff
@@ -101,6 +105,30 @@ func LocalEtcdConfig(retryTimeout time.Duration) (*ETCDConfig, error) {
...
+}
+
+// EtcdOption is a functional option to configure an etcd backend
+type EtcdOption func(*etcdOptions)
+
+// WithReadQuorum specifies that reads should go through the quorum,
+// e.g. return the latest committed value applied in a quorum of members
+func WithReadQuorum(quorum bool) EtcdOption {
+	return func(config *etcdOptions) {
+		logrus.WithField("quorum", quorum).Info("Specify quorum reads.")
```
r0mant, line 123: Looks more like a Debug message than Info?

lib/storage/keyval/leader.go
```diff
@@ -48,7 +50,17 @@ func (b *electingBackend) StepDown() {
 	b.Leader.StepDown()
 }
 
+func (b *electingBackend) CopyWithOptions(opts ...EtcdOption) storage.Backend {
```
r0mant, line 53: Nit: godoc.

lib/update/cluster/engine.go
```diff
@@ -316,8 +316,9 @@ func (f *engine) reconcilePlan(ctx context.Context) error {
 	}
 	f.plan = *plan
 	var buf bytes.Buffer
+	type s string
 	fsm.FormatOperationPlanText(&buf, f.plan)
-	f.Debugf("Reconciled plan: %v.", buf.String())
+	f.WithField("plan", s(buf.String())).Debug("Reconciled plan.")
```
r0mant, line 321: Why is the type conversion needed here?

lib/update/reconciler.go
```diff
@@ -84,22 +87,29 @@ func (r *reconciler) trySyncChangelogFromEtcd(ctx context.Context) error {
...
+	b := backoff.NewExponentialBackOff()
+	b.MaxInterval = 1 * time.Minute
```
r0mant, line 100: Maybe reduce max interval to potentially get more retry attempts?

lib/update/reconciler.go
```diff
@@ -84,22 +87,29 @@ func (r *reconciler) trySyncChangelogFromEtcd(ctx context.Context) error {
 	}
 
 	if shouldSync {
-		return trace.Wrap(r.syncChangelog(r.backend, r.localBackend))
+		consistentSrc := r.backend.CopyWithOptions(keyval.WithReadQuorum(true))
```
r0mant, line 90: I would add a comment explaining why we're doing this here.

lib/storage/keyval/leader.go
```diff
@@ -48,7 +50,17 @@ func (b *electingBackend) StepDown() {
...
+	engine := b.engine.copyWithOptions(opts...)
+	return &electingBackend{
+		Backend: b.backend,
+		Leader:  b.Leader,
+		backend: b.backend,
```
r0mant, line 58: Hmm, what's the second "backend" parameter here for? Looks like everywhere it's used it's always set to the same value as "Backend"? Or is that just to have access to the actual type, not the interface?

r0mant (approved):

lib/update/cluster/engine.go
a-palchikov: Ah, I've been meaning to comment that - will do. Basically, it is an ugly workaround to avoid plan output being escaped when going through the text log formatter.

lib/storage/keyval/leader.go
a-palchikov: Sure, I'll comment on that in code as well. The second copy of the backend so it can be manipulated since it needs access to the concrete value. I've kept the `Backend` as well since it's required for `electingBackend` to implicitly satisfy the `storage.Backend` interface.

PR #2146, opened 2020-09-23
[master] g1k scaling

knisbet:
## Description

This PR presents a number of optimizations for working with large (1k+ node) gravity clusters. This was done as a scaling project, so lots of small changes got lumped in together.

1. In empty gravity application, create a new "cloud" installation profile, with masters properly labelled. When in AWS this allows the masters ASG to properly scale.
2. Accept gravitational node-role labels in addition to kubernetes node role labels to set the service role within gravity. This is to workaround an issue when properly labelling master nodes for large numbers of masters that can't be handled by our "auto" node logic. The AWS load balancer will not target any master nodes. This is only really needed temporarily until we can rely on https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/2019-07-16-node-role-label-use.md#motivation
3. Make join concurrency relative to the number of CPU cores on the master. This allows more parallel joins when the master has plenty of CPU cores. It's not perfect, as if the masters are already heavily loaded, it's possible to trigger significant load. I wasn't able to detect any major issues though even if the CPU load was beyond the number of cores.
4. Tune the backoff timer for joins to backoff more quickly and to have a higher cap. When booting up a couple hundred nodes, the join loop itself created significant CPU load just to fail due to the concurrency control and retry.
5. Introduce server side filtering and caching to the retrieval of cluster operations. A cluster that has been scaled up/down a few times is noticeably slow, even if all workers are removed. There are two main reasons for this, API clients, such as `gravity status` have a tendency to retrieve all operations, in order to just show the last operation. So I added a bit of a hacky filter, so clients can indicate the filters to the server side, and only retrieve operations they're interested in. 

Also, gravity-site now caches operations that are not expected to change. The operations logic would retrieve a list of operations, and then have to individually query every operation for retrieval. So a cluster with lots of nodes, that have been scaled up/down, would trigger potentially thousands of queries to retrieve just one operation. For consistency, the cache only caches operations unlikely to change, so the query's aren't totally eliminated, but it's way faster.

6. Also implement caching for the join token. This was a similar problem to above, where there would be potentially thousands of tokens (from the node joins), and we would query all tokens to find a particular token to return. The pointer to which token is the join token is now cached, so only the relevant token can be re-retrieved if needed.

6. Fix an issue with moving core-dns from daemonset to deployment. The upgrade logic checks for a coredns pod on the node to make progress, which is no longer expected. Instead I rejigged the logic to ensure the expected coredns endpoints have atleast one endpoint, but overall the check may need to be rethought out. 

7. Edit: Increase the default teleport connection limit to support large clusters.

## Type of change

Enterprise PR: https://github.com/gravitational/gravity.e/pull/4337

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

Tested on my scaling cluster, up to 1200 nodes with some signs of satellite false positives in the 1000-1200 node range. Also requires the scaling PRs from the satellite, monitoring-app, planet, repos to hit those targets.

## Additional information

The logging-app isn't capable of scaling to these numbers. It should be disabled / removed on large clusters at present.


lib/ops/ops.go
```diff
@@ -855,6 +855,143 @@ type Operations interface {
...
+	// Complete indicates to only return completed operations
+	Complete bool
+
+	// Finished indicates to only return finished operations (complete or failed)
+	Finished bool
+
+	// Active indicate to only return active operations
+	Active bool
+
+	// Type indicates to only return a type of operation
```
r0mant, line 875: The comment is a little confusing - do you mean only return a particular type of operation? If so, maybe also make this a slice to allow to return multiple types of operations?

lib/ops/utils.go
```diff
@@ -34,118 +34,165 @@ import (
...
 		return nil, nil, trace.Wrap(err)
 	}
 	return op, progress, nil
 }
 
 // GetLastUninstallOperation returns the last uninstall operation for the specified siteKey
 func GetLastUninstallOperation(siteKey SiteKey, operator Operator) (op *SiteOperation, progress *ProgressEntry, err error) {
-	op, progress, err = MatchOperation(siteKey, operator, MatchByType(OperationUninstall))
+	operations, err := operator.GetSiteOperations(siteKey, OperationsFilter{
+		Type: OperationInstall,
```
r0mant, line 59: OperationUninstall?

lib/ops/ops.go
```diff
@@ -855,6 +855,143 @@ type Operations interface {
...
+		if len(t) > 0 {
+			f.Type = t[0]
+		}
+	}
+
+	return
+}
+
+// Filter takes a list of operations and filters the results based on the set filter parameters
+func (filter OperationsFilter) Filter(in SiteOperations) SiteOperations {
```
r0mant, line 942: A unit test for this wouldn't hurt.

lib/ops/ops.go
```diff
@@ -855,6 +855,143 @@ type Operations interface {
...
+	if len(in) == 0 {
+		return nil
+	}
+
+	filtered := in
+
+	if filter.Type != "" || filter.Active || filter.Complete || filter.Finished {
```
a-palchikov, line 949: Invert the condition and exit early:
```go
if filter.Type == "" && !filter.Active && !filter.Complete && !filter.Finished {
	return nil
}
```

lib/ops/utils.go
```diff
@@ -34,118 +34,165 @@ import (
...
 	}
-	return lastOperation, progress, nil
+
+	return op, progress, nil
 }
 
 // GetLastCompletedOperations returns the cluster's last completed operation
-func GetLastCompletedOperation(key SiteKey, operator Operator) (*SiteOperation, *ProgressEntry, error) {
-	operations, err := operator.GetSiteOperations(key)
+func GetLastCompletedOperation(siteKey SiteKey, operator Operator) (op *SiteOperation, progress *ProgressEntry, err error) {
```
r0mant, line 139: Nit: I would rename this to `GetLastFinishedOperation` as that's what it's actually searching for.

lib/ops/ops.go
```diff
@@ -855,6 +855,143 @@ type Operations interface {
...
+		filtered = SiteOperations{}
+
+		for _, value := range in {
+			drop := false
+
+			if filter.Type != "" && filter.Type != value.Type {
+				drop = true
```
a-palchikov, line 956: Just use `continue` - no need for an additional variable

lib/storage/keyval/operations.go
```diff
@@ -66,6 +74,14 @@ func (b *backend) GetSiteOperation(siteDomain, operationID string) (*storage.Sit
 	}
 	utils.UTC(&op.Created)
 	utils.UTC(&op.Updated)
+
+	// Operations that are not expected to change in the future are the only operations that are safe to cache
+	if op.State == storage.OperationPhaseStateCompleted || op.State == storage.OperationPhaseStateRolledBack {
```
r0mant, line 79: This should be comparing with the operation states (e.g. `ops.SiteOperationStateCompleted`), not operation phase states. Also, the operation cannot be "rolled back" (it's an individual phase state). The only terminal operation state is `ops.SiteOperationStateCompletes` IIRC (since if it's "failed", it can still become "completed").

lib/install/engine/interactive/wizard.go
```diff
@@ -115,7 +115,7 @@ func (r *executor) waitForOperation() (operation *ops.SiteOperation, err error)
 			return trace.NotFound("no clusters created yet")
 		}
 		cluster := clusters[0]
-		operations, err := r.Operator.GetSiteOperations(cluster.Key())
+		operations, err := r.Operator.GetSiteOperations(cluster.Key(), ops.OperationsFilter{})
```
a-palchikov, line 118: Have you considered introducing a new API - this seems like a breaking change.
Or are you aiming at making this change transparent so it can equally be serviced by the older implementations during the upgrade?

lib/ops/opsclient/opsclient.go
```diff
@@ -443,16 +443,20 @@ func (c *Client) CheckSiteStatus(ctx context.Context, key ops.SiteKey) error {
...
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
 	var ops ops.SiteOperations
 	if err := json.Unmarshal(out.Bytes(), &ops); err != nil {
 		return nil, trace.Wrap(err)
 	}
+
+	// Servers that haven't been upgraded may not have filtered the results.
+	ops = f.Filter(ops)
```
a-palchikov, line 458: I would really handle this in another API since otherwise, we're always handling it twice when communicating with a remote operator.

lib/storage/keyval/etcd.go
```diff
@@ -63,6 +64,8 @@ func NewETCD(cfg ETCDConfig) (*electingBackend, error) {
 		Backend: &backend{
 			Clock:    clock,
 			kvengine: engine,
+
+			cachedCompleteOperations: make(map[string]*storage.SiteOperation),
```
r0mant, line 68: I would do this as another Backend implementation, for example `CachingBackend` instead of sticking this into the regular backend. The CachingBackend would embed Backend and only reimplement GetSiteOperations method adding the caching functionality.

lib/ops/opsservice/service.go
```diff
@@ -165,6 +165,11 @@ type Operator struct {
 
 	// FieldLogger allows this operator to log messages
 	log.FieldLogger
+
+	// cachedProvisioningToken holds an in memory cache of which token is the cluster provisioning token.
+	cachedProvisioningToken string
+	// cachedProvisioningTokenMutex provides a mutex on use of the cached provisioning token.
+	cachedProvisioningTokenMutex sync.RWMutex
```
a-palchikov, line 172: nit: I think it's more common to declare the mutex first and then the fields it is supposed to protect

lib/ops/opsservice/service.go
```diff
@@ -433,6 +438,20 @@ func (o *Operator) CreateProvisioningToken(token storage.ProvisioningToken) erro
...
+		token, err := o.backend().GetProvisioningToken(o.cachedProvisioningToken)
+		if err != nil && !trace.IsNotFound(err) {
+			return nil, trace.Wrap(err)
+		}
+
+		if token != nil {
+			return token, nil
+		}
+	}
+	o.cachedProvisioningTokenMutex.RUnlock()
```
a-palchikov, line 453: This might not unlock the reading side of the mutex in case of errors.

tool/gravity/cli/operation.go
```diff
@@ -315,39 +309,145 @@ func getActiveOperation(localEnv *localenv.LocalEnvironment, environ LocalEnviro
...
+
+	if clusterOp != nil && clusterOp.Created.After(operations[0].Created) {
+		return clusterOp
 	}
+
+	return &operations[0]
 }
 
-func (r *backendOperations) List(localEnv *localenv.LocalEnvironment, environ LocalEnvironmentFactory) {
+func (r *backendOperations) getOperationByIdFromCluster(localEnv *localenv.LocalEnvironment, operationId string) (*clusterOperation, error) {
```
r0mant, line 393: Nit: Id -> ID (both in function name and the passed in argument)

lib/ops/utils.go
```diff
@@ -162,57 +209,45 @@ func GetOperationWithProgress(opKey SiteOperationKey, operator Operator) (*SiteO
...
-}
 
-// MatchOperation returns an operation that matches given match function.
-// Returns trace.NotFound if no operation matches
-func MatchOperation(siteKey SiteKey, operator Operator, match OperationMatcher) (op *SiteOperation, progress *ProgressEntry, err error) {
-	operations, err := operator.GetSiteOperations(siteKey)
-	if err != nil {
-		return nil, nil, trace.Wrap(err)
+	if len(operations) == 0 {
+		return nil, trace.NotFound("no shrink operation found for %v", siteKey)
```
a-palchikov, line 243: ~~shrink~~

lib/storage/keyval/etcd.go
a-palchikov, line 68: I would also consider factoring this out into some other type and let backend handle the retrieval. `cachingBackend` for instance can add caching non-invasively

tool/gravity/cli/operation.go
```diff
@@ -315,39 +309,145 @@ func getActiveOperation(localEnv *localenv.LocalEnvironment, environ LocalEnviro
...
 	}
-	sort.Slice(result, func(i, j int) bool {
-		return result[i].Created.After(result[j].Created)
+
+	sites, err := clusterEnv.Operator.GetSites(defaults.SystemAccountID)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	if len(sites) == 0 {
+		return nil, trace.NotFound("no sites found")
```
a-palchikov, line 333: `clusters`

tool/gravity/cli/operation.go
```diff
@@ -290,18 +290,12 @@ func completeClusterOperationPlan(localEnv *localenv.LocalEnvironment, operation
...
-	log.WithField("operations", operationList(operations).String()).Debug("Fetched backend operations.")
-	if len(operations) == 0 {
-		if operationID != "" {
-			return nil, newOperationNotFound("no operation with ID %v found", operationID)
-		}
-		return nil, newOperationNotFound("no operation found")
+	b := newBackendOperations()
+
+	if operationID != "" {
+		return b.GetOperationById(localEnv, environ, operationID), nil
```
a-palchikov, line 296: This can return `nil` - is this taken into account on the calling site?

tool/gravity/cli/operation.go
```diff
@@ -315,39 +309,145 @@ func getActiveOperation(localEnv *localenv.LocalEnvironment, environ LocalEnviro
...
+	if clusterEnv == nil {
+		return nil, trace.NotFound("clusterEnv not available")
+	}
+
+	sites, err := clusterEnv.Operator.GetSites(defaults.SystemAccountID)
+	if err != nil {
+		return nil, trace.Wrap(err)
 	}
+	if len(sites) == 0 {
+		return nil, trace.NotFound("no sites found")
```
a-palchikov, line 408: `clusters`

lib/install/engine/interactive/wizard.go
knisbet: Correct, introducing a new API leaves a lot of room for miss-matched client and servers through the upgrade or other ports. It also implies this particular API endpoint should probably be deprecated. So by passing optional arguments, the server is just free to ignore those arguments, in which case it returns all results.

lib/ops/opsclient/opsclient.go
knisbet: Why does that matter?

lib/ops/ops.go
knisbet: I can add some. I didn't bother since the operation system unit tests do largely cover this filtering through the util methods.

lib/defaults/defaults.go
```diff
@@ -1302,6 +1300,9 @@ var (
 	// sets the registry variable to the address of internal cluster registry:
 	// image.registry=leader.telekube.local:5000/
 	HelmRegistryVar = fmt.Sprintf("%v=%v/", ImageRegistryVar, DockerRegistry)
+
+	// MaxExpandConcurrency is the number of servers that can be joining the cluster concurrently
+	MaxExpandConcurrency = (runtime.NumCPU() / 3) + 1
```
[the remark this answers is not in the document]
knisbet: Yes.

lib/expand/join.go
```diff
@@ -763,7 +762,15 @@ type operationContext struct {
 // For a local gravity cluster, it will attempt to start the expand operation
 // and will return an operation context wrapping a new expand operation.
 func (p *Peer) connectLoop() (*operationContext, error) {
-	ticker := backoff.NewTicker(leader.NewUnlimitedExponentialBackOff())
+
+	// Lots of joining nodes create load on the gravity-site controller cycling on creating join operations
+	// Set a high maximum so lots of queued joins don't create too much load
+	b := backoff.NewExponentialBackOff()
```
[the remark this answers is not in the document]
knisbet: I removed NewUnlimitedExponentialBackoff because the parameters it sets are not appropriate for this fall back, and the abstraction hide the details of what's actually being configured.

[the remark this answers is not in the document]
a-palchikov: Not sure I follow. Here's the implementation of `NewUnlimitedExponentialBackoff`:
```go
func NewUnlimitedExponentialBackOff() *backoff.ExponentialBackOff {
	b := backoff.NewExponentialBackOff()
	b.MaxElapsedTime = 0
	return b
}
```
and can be used to set any additional parameters:
```go
b := utils.NewUnlimitedExponentialBackoff()
b.Multiplier = 2
b.MaxInterval = time.Minute
```
my concern was specifically with `b.MaxElapsedTime = 0` which is not as clear as `UnlimitedExponentialBackoff`

lib/storage/keyval/etcd.go
knisbet: I'm not sure I see the advantage to that approach.

lib/defaults/defaults.go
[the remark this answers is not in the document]
knisbet: I can add 2 nodes as a minimum +2, but join operations are surprisingly expensive to gravity-site on the master, and presumably the hardcode of 5 previously was just a guestimate.

lib/expand/join.go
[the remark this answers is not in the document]
knisbet: The previous function must've had a different name, but it just buried the parameters given to the backoff. So I find it non intuitive to wrap everything because now you need added context of what some random utility function does.

lib/storage/keyval/etcd.go
knisbet: I don't really see how adding another backend that acts as a shim other than caching one function (or maybe more in the future) makes this less invasive. 

I think it also creates a crossing of concerns, because GetOperations which is the frequent call, can't be passed through with caching. The logic of how to GetOperations within the cache would drive a backend change to list just the operation keys, so the caching layer can re-implement GetOperations. GetOperations now is duplicated and tightly coupled to the backend implementation, and presumably need to be kept in sync. 

So I don't see why just getting caching within the key value store is an issue here.

a-palchikov (approved):

lib/update/utils.go
```diff
@@ -67,20 +66,30 @@ func WaitForEndpoints(ctx context.Context, client corev1.CoreV1Interface, server
...
+			// this point.
+			return trace.NewAggregate(errors...)
+		}
+
+		if err := hasEndpoints(client, kubednsLabels); err != nil {
+			errors = append(errors, trace.Wrap(err))
 		}
-		return trace.NotFound("endpoints not ready")
+
+		if server.Role == string(schema.ServiceRoleNode) {
```
a-palchikov, line 86:
```go
server.ClusterRole == string(schema.ServiceRoleNode)
```

r0mant (approved):

Issue #2156, opened 2020-09-25
gravity resume doesn't work if gravity-site/etcd is down


r0mant:
### Description

**What happened**:

We have recently added a check to gravity resume command that tries to determine if all upgrade agents are deployed. This check results in the failure if gravity-site or etcd is down:

```
$ ./gravity plan resume --block
[ERROR]: failed to collect agent status
	dial tcp 10.100.80.237:3009: connect: connection refused
```

This basically makes it impossible to use the resume command in such scenarios.

The workaround is to execute phases manually using gravity plan execute until etcd/gravity-site are available.

**What you expected to happen**:

- [x] gravity resume should just print a warning if it can't determine agents status and resume anyway in such cases.
- [x] Same thing for gravity rollback as well.

**Relevant Debug Logs If Applicable**

```
2020-09-25T15:00:09Z WARN             Command failed. error:[
ERROR REPORT:
Original Error: *trace.ConnectionProblemError dial tcp 10.100.80.237:3009: connect: connection refused
Stack Trace:
        /home/deemok/go/_/src/github.com/gravitational/gravity/vendor/github.com/gravitational/teleport/lib/httplib/httplib.go:110 github.com/gravitational/gravity/vendor/github.com/gravitational/teleport/lib/httplib.ConvertResponse
        /home/deemok/go/_/src/github.com/gravitational/gravity/lib/ops/opsclient/opsclient.go:1671 github.com/gravitational/gravity/lib/ops/opsclient.(*Client).Get
        /home/deemok/go/_/src/github.com/gravitational/gravity/lib/ops/opsclient/opsclient.go:333 github.com/gravitational/gravity/lib/ops/opsclient.(*Client).GetLocalSite
        /home/deemok/go/_/src/github.com/gravitational/gravity/tool/gravity/cli/rpcagent.go:489 github.com/gravitational/gravity/tool/gravity/cli.collectAgentStatus
        /home/deemok/go/_/src/github.com/gravitational/gravity/tool/gravity/cli/rpcagent.go:501 github.com/gravitational/gravity/tool/gravity/cli.verifyOrDeployAgents
        /home/deemok/go/_/src/github.com/gravitational/gravity/tool/gravity/cli/clusterupdate.go:251 github.com/gravitational/gravity/tool/gravity/cli.executeOrForkPhase
[...]

PR #2157, opened 2020-09-25
Continue operation in case gravity-site or etcd is down

bernardjkim:
## Description

Previous PR https://github.com/gravitational/gravity/pull/2017 added a check in upgrade resume and rollback operations that will abort if unable to verify agent status or re-deploy agents. This check made it impossible to resume/rollback if gravity-site or etcd is down. More details found in https://github.com/gravitational/gravity/issues/2156. This PR will allow resume/rollback continue operation in these situations.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/2156, https://github.com/gravitational/gravity/pull/2017

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

**Verify upgrade resume executes if gravity-site or etcd is down**

**Verify rollback executes if gravity-site or etcd is down**

### DIFF
--- a/tool/gravity/cli/clusterupdate.go
+++ b/tool/gravity/cli/clusterupdate.go
@@ -237,7 +237,9 @@ func executeUpdatePhaseForOperation(env *localenv.LocalEnvironment, environ Loca
 func executeOrForkPhase(env *localenv.LocalEnvironment, updater updater, params PhaseParams, operation ops.SiteOperation) error {
 	if params.isResume() {
 		if err := verifyOrDeployAgents(env); err != nil {
-			return trace.Wrap(err)
+			// Continue operation in case gravity-site or etcd is down. In these
+			// cases the agent status may not be retrievable.
+			log.WithError(err).Warn("Failed to verify or deploy agents.")
 		}
 	}
 
--- a/tool/gravity/cli/rpcagent.go
+++ b/tool/gravity/cli/rpcagent.go
@@ -476,15 +476,16 @@ func collectAgentStatus(env *localenv.LocalEnvironment) (statusList rpc.StatusLi
 func verifyOrDeployAgents(env *localenv.LocalEnvironment) error {
 	statusList, err := collectAgentStatus(env)
 	if err != nil {
+		env.Println("Failed to verify agents. Ensure gravity-site and etcd are up and running.")
 		return trace.Wrap(err, "failed to collect agent status")
 	}
 	if statusList.AgentsActive() {
 		return nil
 	}
 	if err := rpcAgentDeploy(env, deployOptions{}); err != nil {
-		log.WithError(err).Error("Failed to deploy agents.")
 		env.Println(statusList.String())
-		return trace.BadParameter("some agents are offline; ensure all agents are deployed with `./gravity agent deploy`")
+		env.Println("Some agents are offline. Ensure all agents are deployed with `./gravity agent deploy`.")
+		return trace.Wrap(err, "failed to deploy agents")
 	}
 	return nil
 }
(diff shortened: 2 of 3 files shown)


tool/gravity/cli/clusterupdate.go
```diff
@@ -237,7 +237,9 @@ func executeUpdatePhaseForOperation(env *localenv.LocalEnvironment, environ Loca
 func executeOrForkPhase(env *localenv.LocalEnvironment, updater updater, params PhaseParams, operation ops.SiteOperation) error {
 	if params.isResume() {
 		if err := verifyOrDeployAgents(env); err != nil {
-			return trace.Wrap(err)
+			// Continue operation in case gravity-site or etcd is down. In these
+			// cases the agent status may not be retrievable.
+			log.WithError(err).Warn("Failed to verify or deploy agents.")
```
a-palchikov, line 242: Technically we can test whether agents are active using the static database that we create and keep as the source of truth during the upgrade on each node. This is a more involved change though as one would have to rethink the way we query the cluster state to get the list of servers. 

 * `collectAgentStatus` only requires the operator to get the list of servers which is available in the static database.
 * `rpcDeployAgent` does rely on `localenv.LocalEnvironment` and `localenv.ClusterEnvironment` for credentials handling but I believe this is a design mistake. If we want to keep the upgrade operation resilient then pulling (and even upserting as I've seen!) credentials from/to the cluster package service is not a viable option. The secrets should continue to be extracted as files and used only as such during the operation (they are not both - extracted as files and pulled from the cluster package service).

When the operation is initialized, it performs the following:
  1. create operation/plan
  1. sync operation plan into a local database file which is kept as a source of truth
  1. extract RPC creds on each node for the agent
  1. deploy agents
Then, depending on the operational mode, it either starts executing the plan or pauses (e.g. manual mode)
Given all these prerequisites, it should be possible to verify the state of agents and redeploy agents if necessary at any given time using just the static database.

I'd really like that we take some time to consider this carefully and make improvements to the current state of flux.

r0mant: I'm not sure how we can redeploy agents without gravity-site running because agents deployment is done through teleport.

Also, maybe we can redesign this somehow in master if we want, but when backporting to LTS branches let's keep the changeset to a minimum and not make any redesigns without a good reason to reduce risk.

tool/gravity/cli/rpcagent.go
```diff
@@ -476,15 +476,16 @@ func collectAgentStatus(env *localenv.LocalEnvironment) (statusList rpc.StatusLi
 func verifyOrDeployAgents(env *localenv.LocalEnvironment) error {
 	statusList, err := collectAgentStatus(env)
 	if err != nil {
+		env.Println("Failed to verify agents. Ensure gravity-site and etcd are up and running.")
```
r0mant, line 479: "Couldn't verify upgrade agents status. If some are offline, they won't be redeployed automatically".

Also, can we make this a colored warning e.g. yellow? And as a small nit, I don't think we add dots to the end of our progress messages.

tool/gravity/cli/rpcagent.go
```diff
@@ -476,15 +476,16 @@ func collectAgentStatus(env *localenv.LocalEnvironment) (statusList rpc.StatusLi
...
 		return trace.Wrap(err, "failed to collect agent status")
 	}
 	if statusList.AgentsActive() {
 		return nil
 	}
 	if err := rpcAgentDeploy(env, deployOptions{}); err != nil {
-		log.WithError(err).Error("Failed to deploy agents.")
 		env.Println(statusList.String())
-		return trace.BadParameter("some agents are offline; ensure all agents are deployed with `./gravity agent deploy`")
+		env.Println("Some agents are offline. Ensure all agents are deployed with `./gravity agent deploy`.")
```
r0mant, line 487: "Some upgrade agents are offline and redeployment attempt failed. Ensure agents on all nodes are deployed using ..."

Make it yellow, remove final dot?

a-palchikov (approved):
Sure, we can give it a shot at redesigning later.

PR #2163, opened 2020-09-29
(7.0) Disable source/dest check for instances

r0mant:
## Description

Update our AWS autoscaler controller loop with additional responsibility to periodically check if any of the cluster instances have source/destination check enabled and disable it. The check should be disabled in AWS cloud integration mode.

Our autoscaler actually already disables this parameter for new instances that come up when it receives events from ASG, but it doesn't work for pre-existing instances or those created outside of auto-scaling groups.

I kept scaling in mind when adding this, there should not be any downsides in that regard because I'm using a single "describe instances" call with filters (once every 10 minutes) to get information specifically about instances that have that check enabled which should not be many.

## Type of change

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

Provisioned instances using our terraform and made sure that gravity-site disables source/dest check. Also, manually enabled it back and made sure it was disabled again.

### DIFF
--- a/lib/autoscale/aws/autoscaler.go
+++ b/lib/autoscale/aws/autoscaler.go
@@ -160,6 +160,27 @@ func (a *Autoscaler) DescribeInstance(ctx context.Context, instanceID string) (*
 	return resp.Reservations[0].Instances[0], nil
 }
 
+// DescribeInstancesWithSourceDestinationCheck returns all instances from the
+// specified list that have source/destination check enabled.
+func (a *Autoscaler) DescribeInstancesWithSourceDestinationCheck(ctx context.Context, instanceIDs []string) (result []*ec2.Instance, err error) {
+	resp, err := a.Cloud.DescribeInstancesWithContext(ctx, &ec2.DescribeInstancesInput{
+		InstanceIds: aws.StringSlice(instanceIDs),
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("source-dest-check"),
+				Values: aws.StringSlice([]string{"true"}),
+			},
+		},
+	})
+	if err != nil {
+		return nil, utils.ConvertEC2Error(err)
+	}
+	for _, reservation := range resp.Reservations {
+		result = append(result, reservation.Instances...)
+	}
+	return result, nil
+}
+
 // WaitUntilInstanceTerminated blocks until the instance with the specified ID
 // is terminated.
 //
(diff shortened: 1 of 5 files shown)


a-palchikov (approved):

bernardjkim (approved):

lib/autoscale/aws/autoscaler.go
```diff
@@ -160,6 +160,27 @@ func (a *Autoscaler) DescribeInstance(ctx context.Context, instanceID string) (*
 	return resp.Reservations[0].Instances[0], nil
 }
 
+// DescribeInstancesWithSourceDestinationCheck returns all instances from the
+// specified list that have source/destination check enabled.
+func (a *Autoscaler) DescribeInstancesWithSourceDestinationCheck(ctx context.Context, instanceIDs []string) (result []*ec2.Instance, err error) {
+	resp, err := a.Cloud.DescribeInstancesWithContext(ctx, &ec2.DescribeInstancesInput{
+		InstanceIds: aws.StringSlice(instanceIDs),
```
knisbet, line 167: Not so much on the response side, but on the request side I'm curious how AWS responds when passed 1k instances?

I don't see any specific notes in the AWS API docs, except when not using instance id's and instead using maxResults, the range can be 5-1000 (Source: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html). 

It might be worth considering splitting the list into separate calls in groups of a couple hundred just incase?

lib/autoscale/aws/discovery.go
```diff
@@ -46,15 +48,50 @@ func (a *Autoscaler) PublishDiscovery(ctx context.Context, operator ops.Operator
...
+			a.Warnf("%v doesn't have cloud instance id.", node)
+		}
+	}
+	instances, err := a.DescribeInstancesWithSourceDestinationCheck(ctx, instanceIDs)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	for _, instance := range instances {
+		a.Infof("Instance %v has source/dest check enabled, disabling.",
+			aws.StringValue(instance.InstanceId))
```
knisbet, line 88: Do you need to convert to an aws String for logging?

knisbet (approved):

lib/autoscale/aws/discovery.go
r0mant: Yeah that's what I'm basically doing here, `aws.String()` actually returns a string pointer. `InstanceId` field is a string pointer so `aws.StringValue()` dereferences it and returns a value.

PR #2168, opened 2020-09-29
[6.1.x] take mount overrides into account in preflight checks

a-palchikov:
## Description

Take mount overrides into account when running local pre-flight checks during expand and install.

## Type of change

* Refs github.com/gravitational/gravity/issues/2167

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

Edited the application manifest to add a volume and tested with a modified `check` that accepted a list of mount overrides:
```
$ ./gravity check -p node --mounts=data=/path/to/new/volume
```

### DIFF
--- a/lib/checks/checks.go
+++ b/lib/checks/checks.go
@@ -59,32 +59,50 @@ func New(config Config) (*checker, error) {
 	return &checker{Config: config}, nil
 }
 
-// ValidateManifest verifies the specified manifest against the host environment.
+// Validate verifies the specified manifest against the host environment.
 // Returns list of failed health probes.
-func ValidateManifest(
-	manifest schema.Manifest,
-	profile schema.NodeProfile,
-	dockerConfig storage.DockerConfig,
-	stateDir string,
-) (failedProbes []*agentpb.Probe, err error) {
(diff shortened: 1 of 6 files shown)


lib/checks/checks.go
```diff
@@ -59,34 +59,52 @@ func New(config Config) (*checker, error) {
...
-	failed, err := schema.ValidateRequirements(profile.Requirements, stateDir)
+	requirements := r.Profile.Requirements
+	if len(r.Mounts) != 0 {
+		requirements = overrideMounts(requirements, r.Mounts)
+	}
+	failed, err := schema.ValidateRequirements(ctx, requirements, r.StateDir)
 	if err != nil {
 		errors = append(errors, trace.Wrap(err,
-			"error validating profile requirements, see syslog for details"))
+			"error validating profile requirements, see log file for details"))
```
r0mant, line 73: Nit: If this error is user-facing, it might be confusing, I would say which log file at least (/var/log/gravity-system.log?)

r0mant (approved):

PR #2178, opened 2020-10-01
[6.1.x] explicit control over gRPC logging

a-palchikov:
## Description

Expose control over gRPC logging which is implicitly disabled by the etcd client package. The logging is on if a gravity command is executed with `--debug`, otherwise - the logging can be turned on using these environment variables:
 * `GRPC_GO_LOG_SEVERITY_LEVEL` - control the severity level, takes one of `error`, `warning` or `info`
 * `GRPC_GO_LOG_VERBOSITY_LEVEL` - arbitrary verbosity level as an integer
See https://pkg.go.dev/google.golang.org/grpc@v1.32.0/grpclog#pkg-overview for details.

## Type of change

* Requires https://github.com/gravitational/gravity.e/pull/4340

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

### DIFF
--- a/lib/rpc/server/peers.go
+++ b/lib/rpc/server/peers.go
@@ -227,7 +227,6 @@ func (r *peers) reconnectPeer(p Peer, reqCh <-chan chan clientUpdate, doneCh cha
 			err := utils.RetryWithInterval(r.ctx, r.Backoff(), func() (err error) {
 				clt, err = p.Reconnect(r.ctx)
 				if err != nil {
-					log.WithError(err).Warn("Unable to reconnect.")
 					return r.ShouldReconnect(err)
 				}
 				return nil
--- a/lib/rpc/server/peers.go
+++ b/lib/rpc/server/peers.go
@@ -227,6 +227,7 @@ func (r *peers) reconnectPeer(p Peer, reqCh <-chan chan clientUpdate, doneCh cha
 			err := utils.RetryWithInterval(r.ctx, r.Backoff(), func() (err error) {
 				clt, err = p.Reconnect(r.ctx)
 				if err != nil {
+					log.WithError(err).Warn("Unable to reconnect.")
 					return r.ShouldReconnect(err)
 				}
 				return nil
--- a/lib/utils/logging.go
+++ b/lib/utils/logging.go
@@ -40,42 +40,42 @@ func InitLogging(level log.Level, logFile string) {
 
 // InitGRPCLoggerFromEnvironment configures the GRPC logger if any of the related environment variables
 // are set.
-func InitGRPCLoggerFromEnvironment() {
+func InitGRPCLoggerFromEnvironment(severityDefault, verbosityDefault string) {
 	const (
 		envSeverityLevel  = "GRPC_GO_LOG_SEVERITY_LEVEL"
 		envVerbosityLevel = "GRPC_GO_LOG_VERBOSITY_LEVEL"
 	)
-	severityLevel := os.Getenv(envSeverityLevel)
-	verbosityLevel := os.Getenv(envVerbosityLevel)
-	var verbosity int
-	if verbosityOverride, err := strconv.Atoi(verbosityLevel); err == nil {
-		verbosity = verbosityOverride
-	}
+	severityLevel := GetenvWithDefault(envSeverityLevel, severityDefault)
+	verbosityLevel := GetenvWithDefault(envVerbosityLevel, verbosityDefault)
 	if severityLevel == "" && verbosityLevel == "" {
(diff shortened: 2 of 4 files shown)


lib/utils/logging.go
```diff
@@ -20,23 +20,70 @@ import (
...
+	setLoggingOptions(level)
 }
 
-func setLoggingOptions() {
-	log.SetLevel(log.DebugLevel)
+// InitGRPCLoggerFromEnvironment configures the GRPC logger if any of the related environment variables
+// are set.
+func InitGRPCLoggerFromEnvironment() {
+	const (
+		envSeverityLevel  = "GRPC_GO_LOG_SEVERITY_LEVEL"
```
r0mant, line 45: Can you please add information about using these environment variables to our docs, e.g. troubleshooting section. I never remember what these variables are and how to use them, let's document it once and for all.

lib/utils/logging.go
```diff
@@ -20,23 +20,70 @@ import (
...
+	if severityLevel == "" && verbosityLevel == "" {
+		// Nothing to do
+		return
+	}
+	InitGRPCLogger(severityLevel, verbosity)
+}
+
+// InitGRPCLoggerWithDefaults configures the GRPC logger with debug defaults.
+func InitGRPCLoggerWithDefaults() {
+	InitGRPCLogger("info", 10)
```
r0mant, line 63: Nit: I'd use a constant for clarity, e.g. 

```go
const maxVerbosity = 10
InitiGRPCLogger("info", maxVerbosity)
```

lib/utils/logging.go
```diff
@@ -20,23 +20,70 @@ import (
...
+		envSeverityLevel  = "GRPC_GO_LOG_SEVERITY_LEVEL"
+		envVerbosityLevel = "GRPC_GO_LOG_VERBOSITY_LEVEL"
+	)
+	severityLevel := os.Getenv(envSeverityLevel)
+	verbosityLevel := os.Getenv(envVerbosityLevel)
+	var verbosity int
+	if verbosityOverride, err := strconv.Atoi(verbosityLevel); err == nil {
+		verbosity = verbosityOverride
+	}
+	if severityLevel == "" && verbosityLevel == "" {
```
r0mant, line 54: Nit: I'd place this condition right after the two getenv calls, the logic would be a bit clearer.

r0mant (approved):

lib/utils/logging.go
a-palchikov: Ok, will do. On a side note, given that these are sort of implementation detail, this changeset will  automatically turn on gRPC logging if `--debug` has been specified. Now, this might be something to take with a grain of salt since these logs are output directly to stdout - so if we decide it is not worth it, I can remove that bit so the logs can only be controlled via environment.

lib/rpc/server/peers.go
```diff
@@ -227,6 +227,7 @@ func (r *peers) reconnectPeer(p Peer, reqCh <-chan chan clientUpdate, doneCh cha
 			err := utils.RetryWithInterval(r.ctx, r.Backoff(), func() (err error) {
 				clt, err = p.Reconnect(r.ctx)
 				if err != nil {
+					log.WithError(err).Warn("Unable to reconnect.")
```
knisbet, line 230: Nit: to me unable to reconnect I think implies a permanent failure, IE I get the impression from the wording that the connection won't necessarily work in the future. Which presumably is what Should reconnect is trying to determine after this.

So I think it might be a bit clearer to indicate whether a reconnection will be attempted. Ie: "Failed to connect. Retry in 6s", probably with feedback from whether the code is actually treating this as a permanent failure or not.

tool/gravity/cli/run.go
```diff
@@ -70,6 +70,12 @@ func Run(g *Application) (err error) {
 		return trace.Wrap(err)
 	}
 
+	if *g.Debug {
+		utils.InitGRPCLoggerWithDefaults()
+	} else {
+		utils.InitGRPCLoggerFromEnvironment()
```
knisbet, line 76: So the debug flag and GRPC logger env variables cannot be used together?

lib/rpc/server/peers.go
a-palchikov: The reconnect attempts are already being logged with the corresponding amount of backoff. I'll remove this entry since it's superfluous.

tool/gravity/cli/run.go
a-palchikov: True that - I've restructured the code so it was not the case.

PR #2185, opened 2020-10-01
[5.5.x] Add check for rollback of latest unstarted phase

bernardjkim:
## Description

This PR updates the check for `gravity plan rollback` command. The check now ensures that phases of an operation are being rolled back in order. This check can be bypassed by providing the `--force` flag.

## Type of change

* Ports https://github.com/gravitational/gravity/pull/2184

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

**Verify rollback phase fails if out of order and `--force` flag is not provided**

**Verify rollback phase executes if out of order and `--force` flag is provided**

**Verify latest phase can still be rolled back**

**Verify plan can be rolled back completely**

### DIFF
--- a/lib/fsm/fsm.go
+++ b/lib/fsm/fsm.go
@@ -257,12 +257,14 @@ func (f *FSM) RollbackPhase(ctx context.Context, p Params) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	err = CanRollback(plan, p.PhaseID, p.DryRun)
-	if err != nil {
-		if !p.Force {
-			return trace.Wrap(err)
+	if !p.DryRun {
+		err = CanRollback(plan, p.PhaseID)
+		if err != nil {
+			if !p.Force {
+				return trace.Wrap(err)
+			}
+			f.WithError(err).Warn("Forcing rollback.")
 		}
-		f.WithError(err).Warn("Forcing rollback.")
 	}
 	phase, err := FindPhase(plan, p.PhaseID)
 	if err != nil {
--- a/lib/fsm/fsm.go
+++ b/lib/fsm/fsm.go
@@ -257,6 +257,7 @@ func (f *FSM) RollbackPhase(ctx context.Context, p Params) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
+	// No need to verify if phase can be rolled back during dry runs.
 	if !p.DryRun {
 		err = CanRollback(plan, p.PhaseID)
 		if err != nil {
--- a/lib/fsm/fsm.go
+++ b/lib/fsm/fsm.go
@@ -257,7 +257,7 @@ func (f *FSM) RollbackPhase(ctx context.Context, p Params) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	err = CanRollback(plan, p.PhaseID)
+	err = CanRollback(plan, p.PhaseID, p.DryRun)
 	if err != nil {
 		if !p.Force {
 			return trace.Wrap(err)
--- a/lib/fsm/fsm.go
+++ b/lib/fsm/fsm.go
@@ -329,13 +329,28 @@ You can redeploy upgrade agents on all cluster nodes using "./gravity agent depl
 }
 
 func (f *FSM) rollbackPhaseRemotely(ctx context.Context, p Params, phase storage.OperationPhase, server storage.Server) error {
-	return f.RunCommand(ctx, f.Runner, server, Params{
+	err := f.RunCommand(ctx, f.Runner, server, Params{
 		PhaseID:  p.PhaseID,
 		Force:    p.Force,
 		Resume:   p.Resume,
 		Rollback: true,
 		Progress: p.Progress,
 	})
+	if err == nil {
+		// if the remote phase rollback is successful, we need to mark it in our local database
+		// because etcd might not be available to synchronize the changes back to us
+		err = f.ChangePhaseState(ctx,
+			StateChange{
+				Phase: phase.ID,
+				State: storage.OperationPhaseStateRolledBack,
+			})
+	}
+
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	return nil
 }
 
 // SetPreExec sets the hook that's called before phase execution
--- a/lib/fsm/utils.go
+++ b/lib/fsm/utils.go
@@ -18,6 +18,7 @@ package fsm
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/gravitational/gravity/lib/constants"
 	"github.com/gravitational/gravity/lib/ops"
@@ -28,6 +29,17 @@ import (
 	"github.com/gravitational/trace"
 	"github.com/sirupsen/logrus"
 )
+
+// rollbackDependentsErrorMsg returns an error message for when a phase is being
+// rolled back, but has dependent phases that have not yet been rolled back.
+func rollbackDependentsErrorMsg(phaseID string, dependents []string) string {
+	const msg = `phase %[1]s cannot be rolled back because some phases that depend on it haven't been rolled back yet. Please rollback the following phases first:
+
+	%[2]s
+
+you can pass --force flag to override this check and force phase %[1]s rollback`
+	return fmt.Sprintf(msg, phaseID, strings.Join(dependents, "\n\t"))
+}
 
 // CanRollback checks if specified phase can be rolled back
 func CanRollback(plan *storage.OperationPlan, phaseID string) error {
@@ -52,26 +64,17 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
 			AddField("phase", phase.ID)
 	}
 
-	if !dependentsRolledBack(plan, phase.ID) {
-		return trace.BadParameter(
-			"rollback subsequent phases before rolling back phase %q", phase.ID)
+	requiresRollback := getRequiresRollback(plan, phase.ID)
+	if len(requiresRollback) != 0 {
+		return trace.BadParameter(rollbackDependentsErrorMsg(phase.ID, requiresRollback))
 	}
+
 	return nil
 }
 
-// dependentsRolledBack returns true if all phases that are dependent on the
-// phase specified by phaseID are unstarted or have been rolled back.
-func dependentsRolledBack(plan *storage.OperationPlan, phaseID string) bool {
-	// required will be nil if an invalid phaseID is provided.
-	required := getRequired(plan.Phases, phaseID)
-	if required == nil {
-		return true
-	}
-	return dependentsRolledBackHelper(required, plan.Phases)
-}
-
 // getRequired constructs the initial set of required phases. This set includes
-// the phase specified by phaseID and its parent phases.
+// the phase specified by phaseID and its parent phases. Returns nil if phases
+// does not contain a phase with phaseID.
 //
 // Given a list of phases like:
 //
@@ -100,28 +103,39 @@ func getRequired(phases []storage.OperationPhase, phaseID string) map[string]str
 	return nil
 }
 
-// dependentsRolledBackHelper is a recursive helper function that returns true
-// if all phases that are dependent on the required set are unstarted or have
-// been rolled back.
-func dependentsRolledBackHelper(required map[string]struct{}, phases []storage.OperationPhase) bool {
+// getRequiresRollback returns a list of phases that need to be rolled back
+// before the phase specified by phaseID can be rolled back.
+func getRequiresRollback(plan *storage.OperationPlan, phaseID string) (dependents []string) {
+	// required will be nil if an invalid phaseID is provided.
+	required := getRequired(plan.Phases, phaseID)
+	if required == nil {
+		return dependents
+	}
+	return getRequiresRollbackHelper(required, plan.Phases)
+}
+
+// getRequiresRollbackHelper is a recursive helper function that returns a list
+// of dependent phases that have been started and have not been rolled back.
+func getRequiresRollbackHelper(required map[string]struct{}, phases []storage.OperationPhase) (dependents []string) {
 	if len(phases) == 0 {
-		return true
+		return dependents
 	}
 
 	for _, phase := range phases {
 		if isDependent(required, phase) {
 			if !phase.IsUnstarted() && !phase.IsRolledBack() {
-				return false
+				// Append phase to list of dependents that need to be rolled back.
+				dependents = append(dependents, phase.ID)
 			}
+			// Add phase to the required set. Phases dependent on this phase are
+			// also dependents of the original set of required phases.
 			required[phase.ID] = struct{}{}
 		}
-
-		if !dependentsRolledBackHelper(required, phase.Phases) {
-			return false
-		}
+		// Append any dependent sub phases that need to be rolled back.
+		dependents = append(dependents, getRequiresRollbackHelper(required, phase.Phases)...)
 	}
 
-	return true
+	return dependents
 }
 
 // isDependent returns true if the phase requires any of the phases contained in
--- a/lib/fsm/utils.go
+++ b/lib/fsm/utils.go
@@ -33,11 +33,11 @@ import (
 // rollbackDependentsErrorMsg returns an error message for when a phase is being
 // rolled back, but has dependent phases that have not yet been rolled back.
 func rollbackDependentsErrorMsg(phaseID string, dependents []string) string {
-	const msg = `phase %[1]s cannot be rolled back because some phases that depend on it haven't been rolled back yet. Please rollback the following phases first:
+	const msg = `Phase %[1]s cannot be rolled back because some phases that depend on it haven't been rolled back yet. Please rollback the following phases first:
 
 	%[2]s
 
-you can pass --force flag to override this check and force phase %[1]s rollback`
+You can pass --force flag to override this check and force phase %[1]s rollback.`
 	return fmt.Sprintf(msg, phaseID, strings.Join(dependents, "\n\t"))
 }
 
--- a/lib/fsm/utils.go
+++ b/lib/fsm/utils.go
@@ -43,32 +43,42 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
 		return trace.BadParameter(
 			"phase %q has already been rolled back", phase.ID)
 	}
-	if !latestRollback(plan.Phases, phaseID) {
+
+	leafPhases := plan.GetLeafPhases()
+
+	// TODO: Rollback of non-leaf phases may currently not be implemented
+	// properly. Roll back starts top-down, and not in reverse order.
+	if !containsPhase(leafPhases, phaseID) {
+		return trace.BadParameter(
+			"attempting to rollback non-leaf phase %q", phase.ID)
+	}
+	if !latestRollback(leafPhases, phaseID) {
 		return trace.BadParameter(
 			"rollback subsequent phases before rolling back phase %q", phase.ID)
 	}
 	return nil
 }
 
+// containsPhase returns true if phases contains a phase with ID matching
+// phaseID.
+func containsPhase(phases []storage.OperationPhase, phaseID string) bool {
+	for _, phase := range phases {
+		if phase.ID == phaseID {
+			return true
+		}
+	}
+	return false
+}
+
 // latestRollback returns true if the phase specified by phaseID is next in line
 // to be rolled back.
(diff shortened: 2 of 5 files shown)


lib/fsm/utils.go
```diff
@@ -41,9 +41,42 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
 		return trace.BadParameter(
 			"phase %q has already been rolled back", phase.ID)
 	}
+	if !dryRun && !latestRollback(plan.Phases, phaseID) {
```
bernardjkim, line 44: Rolling back an operation as a dry run does not change the state of operation phases. So dry run rollbacks would always fail this check. For now, simple workaround was to just ignore this check during dry runs.

lib/fsm/fsm.go
```diff
@@ -257,7 +257,7 @@ func (f *FSM) RollbackPhase(ctx context.Context, p Params) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	err = CanRollback(plan, p.PhaseID)
+	err = CanRollback(plan, p.PhaseID, p.DryRun)
```
a-palchikov, line 260: No need to pass `dryRun` into CanRollback since it only makes sense to evaluate in case of an actual rollback:
```go
if !p.DryRun {
	err = CanRollback(plan, p.PhaseID)
	if err != nil {
		if !p.Force {
			return trace.Wrap(err)
		}
		f.WithError(err).Warn("Forcing rollback.")
	}
}
```

lib/fsm/utils.go
```diff
@@ -41,9 +41,42 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
...
+		return trace.BadParameter(
+			"rollback subsequent phases before rolling back phase %q", phase.ID)
+	}
 	return nil
 }
 
+// latestRollback returns true if the phase specified by phaseID is next in line
+// to be rolledback.
```
a-palchikov, line 52: nit: rolled back

lib/fsm/utils.go
```diff
@@ -41,9 +41,42 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
...
+	// latest keeps track of the latest phase that has not yet been rolled back.
+	var latest storage.OperationPhase
+	for _, phase := range phases {
+		if latestRollback(phase.Phases, phaseID) {
+			return true
+		}
+		switch {
+		// if we run into an unstarted phase, all later phases should also be unstarted
+		case phase.IsUnstarted():
+			break
```
a-palchikov, line 67: this is equivalent to the empty `case` clause (as it terminates the innermost `switch`):
```go
switch {
case phase.IsUnstarted():
case phase.IsInProgress(), phase.IsFailed(), phase.IsCompleted():
...
}
```
if you needed to terminate the loop - use a label.

lib/fsm/utils_test.go
```diff
@@ -148,3 +148,123 @@ func (s *FSMUtilsSuite) TestDiffPlanNoPrevious(c *check.C) {
...
+					},
+				},
+			},
+			phaseID:  "/masters",
+			expected: true,
+		},
+	}
+	for _, tc := range tests {
+		comment := check.Commentf(tc.comment)
+		c.Assert(latestRollback(tc.phases, tc.phaseID), check.Equals, tc.expected, comment)
```
a-palchikov, line 268: I would actually write a test for the exported `CanRollback` rather than exercising the private latestCallback.
Also, I'm not seeing the case where there's a rolled back phase is at the end (i.e. prior to an unstarted one) - how is it supposed to behave in this case?

lib/fsm/utils.go
bernardjkim: Oops. Yes, we would want to terminate the loop here.

lib/fsm/fsm.go
```diff
@@ -257,12 +257,14 @@ func (f *FSM) RollbackPhase(ctx context.Context, p Params) error {
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	err = CanRollback(plan, p.PhaseID)
-	if err != nil {
-		if !p.Force {
-			return trace.Wrap(err)
+	if !p.DryRun {
```
r0mant, line 260: Add a comment on why these checks aren't executed in dry-run mode?

lib/fsm/utils.go
```diff
@@ -41,9 +41,44 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
...
+	// latest keeps track of the latest phase that has not yet been rolled back.
+	var latest storage.OperationPhase
+
+L:
+	for _, phase := range phases {
+		if latestRollback(phase.Phases, phaseID) {
+			return true
+		}
+		switch {
+		// if we run into an unstarted phase, all later phases should also be unstarted
```
r0mant, line 67: That's technically not true, a phase can be executed "out of band" using plan execute command so I don't think we can make this assumption.

lib/fsm/utils.go
```diff
@@ -41,9 +41,44 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
...
+	if len(phases) == 0 {
+		return false
+	}
+
+	// latest keeps track of the latest phase that has not yet been rolled back.
+	var latest storage.OperationPhase
+
+L:
+	for _, phase := range phases {
+		if latestRollback(phase.Phases, phaseID) {
```
r0mant, line 63: Why do we need recursion here? I would make the algorithm very simple, all we need to do is to "unwrap" the plan (we have the method called "FlattenPlan" somewhere, there's also a method called GetLeafPhases on the plan itself IIRC) and then check if any of the phases that appear after phaseID have been rolled back or are unstarted. Sounds simple enough and easy to reason about, what do you think?

lib/fsm/utils.go
bernardjkim: Ah, thanks for pointing that out. Will update.

lib/fsm/utils.go
r0mant: And I think this is good. If I'm not mistaken, rolling back top-level phases is not implemented properly at the moment, as in it starts rollback top-down, not in reverse order. Or it may be disabled completely, I don't recall.

bernardjkim: Oh I see. Okay, we can update to only allow rollback of leaf phases.

lib/fsm/utils.go
```diff
@@ -41,9 +41,54 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
 		return trace.BadParameter(
 			"phase %q has already been rolled back", phase.ID)
 	}
+
+	leafPhases := plan.GetLeafPhases()
+
+	// TODO: Rollback of non-leaf phases may currently not be implemented
```
knisbet, line 47: Nit: "Rollback of non-leaf phases may currently not be implemented" -> Rollback of non-leaf phases is not supported. 

Although I think this get's a bit strange in gravity, because there is a rollback command that can rollback entirely. So may be difficult to convey that this applies specifically here.

lib/fsm/utils.go
```diff
@@ -41,9 +41,54 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
...
+	// properly. Roll back starts top-down, and not in reverse order.
+	if !containsPhase(leafPhases, phaseID) {
+		return trace.BadParameter(
+			"attempting to rollback non-leaf phase %q", phase.ID)
```
knisbet, line 51: Nit: I don't think a customer would understand why they receive an error like this. So I would consider going something more along the lines of:

`return trace.BadParameter("unable to rollback non-leaf phase, only leaf phases can be rolled back.").AddField("phase", phase.ID)`

lib/fsm/utils.go
```diff
@@ -41,9 +41,54 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
...
+		if phase.ID == phaseID {
+			return true
+		}
+	}
+	return false
+}
+
+// latestRollback returns true if the phase specified by phaseID is next in line
+// to be rolled back.
+func latestRollback(phases []storage.OperationPhase, phaseID string) bool {
```
knisbet, line 73: I'm a bit worried about this algorithm to determine dependence, especially since we kind of support parallel phase execution which I'm probably going to turn on for upgrades of large clusters. By looking at the latest phase from the flattened structure, it creates a dependency that doesn't necessarily exist between phases. IE if in an upgrade I want to rollback / retry bootstrap on node-1, I have to rollback nodes 2-800. 

I think identifying when a rollback can happen would be a closer fit if any depends on operations are in progress, failed, completed. So it's the same readiness to move forwards as it is to move backwords. Although this is a much much more complicated algorithm with the way dependencies are structures, because they can point to a group phase or leaf phase.

r0mant: > IE if in an upgrade I want to rollback / retry bootstrap on node-1, I have to rollback nodes 2-800.

This is exactly the situation that we want to prevent customers from getting in, unless they know what they're doing. In a recent support scenario, someone rolled back one of masters without rolling back any of subsequent phases which put cluster in a bad state. For "power users" there should be a --force flag to override this check and allow you to rollback anything. @bernardjkim - btw I don't see this flag implemented?

r0mant: Nevermind, I see the --force flag is handled above.

bernardjkim: @r0mant @knisbet 
I've updated the implementation in 0bf4934 to be more accommodating towards rolling back parallel phases.

lib/fsm/utils.go
```diff
@@ -41,9 +41,54 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
 		return trace.BadParameter(
 			"phase %q has already been rolled back", phase.ID)
 	}
+
+	leafPhases := plan.GetLeafPhases()
+
+	// TODO: Rollback of non-leaf phases may currently not be implemented
+	// properly. Roll back starts top-down, and not in reverse order.
+	if !containsPhase(leafPhases, phaseID) {
```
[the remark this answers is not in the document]
bernardjkim: Looks like we can make it even simpler and just check if `phase.HasSubPhases`

lib/fsm/utils_test.go
```diff
@@ -148,3 +149,248 @@ func (s *FSMUtilsSuite) TestDiffPlanNoPrevious(c *check.C) {
...
+						State: storage.OperationPhaseStateCompleted,
+						Phases: []storage.OperationPhase{
+							{
+								ID:    "/nodes/node-2",
+								State: storage.OperationPhaseStateCompleted,
+							},
+							{
+								ID:       "/nodes/node-3",
+								State:    storage.OperationPhaseStateCompleted,
+								Requires: []string{"/nodes-node-2"},
```
r0mant, line 358: Typo? `/nodes-node-2` -> `/nodes/node-2`

lib/fsm/utils.go
```diff
@@ -43,9 +43,85 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
...
+	required := getRequired(plan.Phases, phaseID)
+	if required == nil {
+		return true
+	}
+	return dependentsRolledBackHelper(required, plan.Phases)
+}
+
+// getRequired constructs the initial set of required phases. This set includes
+// the phase specified by phaseID and its parent phases.
+func getRequired(phases []storage.OperationPhase, phaseID string) map[string]struct{} {
```
r0mant, line 74: I would update the godoc comment with an example of what it returns, to simplify understanding. I.e.

```
Given plan like:

/init
/masters
  * /node-1
    * /system-upgrade
  * /node-2

for "/masters/node-1/system-upgrade" phase id it will return:

["/masters", "/masters/node-1", "/masters/node-1/system-upgrade"]
```

Also, why does this return a map, doesn't the order matter?

lib/fsm/utils.go
```diff
@@ -43,9 +43,85 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
...
+		if required != nil {
+			required[phase.ID] = struct{}{}
+			return required
+		}
+	}
+	return nil
+}
+
+// dependentsRolledBackHelper is a recursive helper function that returns true
+// if all phases that are dependent on the required set is unstarted or has
```
r0mant, line 91: *are* unstarted, *have* been

lib/fsm/utils.go
```diff
@@ -43,9 +43,85 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
...
+	if len(phases) == 0 {
+		return true
+	}
+
+	for _, phase := range phases {
+		if isDependent(required, phase) {
+			if !phase.IsUnstarted() && !phase.IsRolledBack() {
+				return false
+			}
+			required[phase.ID] = struct{}{}
```
r0mant, line 103: Hmm, do we need to add this phase to the "required" set as well? I wonder if it'll be simpler to just leave the original set as it is, this way the algo will be just one dependency "deep", right? So for example, in case of the following plan:

```
/init
/masters <--depends on /init
/gc <-- depends on /masters
```

when you try to rollback /init and /masters and /gc are in progress, it would only tell you to rollback /masters first. Then you try to rollback /masters and it will tell you to rollback /gc first.

We could probably even get rid of recursion in this case, just iterate over all phases and see if any of them depend on those in the "required" set.

What do you think?

lib/fsm/utils_test.go
```diff
@@ -148,3 +149,248 @@ func (s *FSMUtilsSuite) TestDiffPlanNoPrevious(c *check.C) {
...
+func (s *FSMUtilsSuite) TestCanRollback(c *check.C) {
+	tests := []struct {
+		comment  string
+		plan     *storage.OperationPlan
+		phaseID  string
+		expected string
+	}{
+		{
+			comment: "Rollback latest phase",
+
```
r0mant, line 162: Nit: Extra new line.

lib/fsm/utils_test.go
```diff
@@ -148,3 +149,248 @@ func (s *FSMUtilsSuite) TestDiffPlanNoPrevious(c *check.C) {
...
+						},
+					},
+				},
+			},
+			phaseID:  "/masters",
+			expected: "unable to rollback non-leaf phase, only leaf phases can be rolled back",
+		},
+		{
+			comment: "Top level phase has dependent phases that have not been rolled back",
+			plan: &storage.OperationPlan{
```
r0mant, line 335: Can you add a few more test cases with more complicated plans, i.e. multiple-level deep phases, multiple "requires", etc., more like our real plan. E.g.

```
/init
/masters <-- depends on /init
  * /node-1
    * /drain
    * /system-upgrade <-- depends on /masters/node-1/drain
  * /node-2
    * /drain
    * /system-upgrade <-- depends on /masters/node-2/drain
/nodes <-- depends on /masters
  * /node-3
    * /drai
    * /system-upgrade
/runtime <-- depends on /masters
  * /monitoring
  * /site
/gc <-- depends on /runtime
```

... you get the idea... and try different scenarios.

r0mant (commented):
I actually quite like the algorithm, just have one simplification suggestion. @a-palchikov @knisbet can you guys take a look too.

lib/fsm/utils.go
bernardjkim: That approach would definitely be a lot simpler, but I'm not sure if the check will function properly if a phase is executed out of band. Since phases can be executed out of order, wouldn't something like this be a valid state of the plan?
```
/Init { state: completed }
/masters { state: unstarted, requires: { /init } }
/gc { state: completed, requires: { /masters } }
```
In this situation all phases that require /init are either un-started or rolled back.
Would /gc in this case even have been able to be executed?

lib/fsm/utils.go
bernardjkim: Order doesn't matter, we just need the set to check if any other phases 'requires' any of the phases in this set.

lib/fsm/utils_test.go
bernardjkim: Oops, good catch.

lib/fsm/utils.go
r0mant: Yes, you're right. Let's keep it as-is then, but add a comment here (as it's also a side-effect to the passed in parameter which is not great) and a test case for this scenario.

lib/fsm/utils.go
```diff
@@ -43,9 +43,85 @@ func CanRollback(plan *storage.OperationPlan, phaseID string) error {
...
+	// TODO: Rollback of non-leaf phases is not currently supported.
+	// Rollback starts top-down, and not in reverse order.
+	if phase.HasSubphases() {
+		return trace.BadParameter(
+			"unable to rollback non-leaf phase, only leaf phases can be rolled back").AddField("phase", phase.ID)
+	}
+
+	if !dependentsRolledBack(plan, phase.ID) {
+		return trace.BadParameter(
+			"rollback subsequent phases before rolling back phase %q", phase.ID)
```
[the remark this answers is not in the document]
bernardjkim: The check short circuits on the first dependent phase that needs to be rolled back, so we would only be able to display that first phase. I think this could be more confusing than helpful if we only display that one phase.

We could extend the check for the purpose of collecting all dependent phases. I don't think it would affect performance that much, and I think better ux might be worth it.

a-palchikov (approved):
As long as other comments have been address - especially the extended error message listing the phases that prevent the rollback of a particular phase.

lib/fsm/utils.go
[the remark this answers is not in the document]
r0mant: Yeah let's do it :+1:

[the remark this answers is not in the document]
bernardjkim: Updated in 447efa2

Example output:
```
[vagrant@node-1 installer]$ sudo ./gravity plan rollback --phase /checks
[ERROR]: phase /checks cannot be rolled back because some phases that depend on it haven't been rolled back yet. Please rollback the following phases first:

        /pre-update
        /bootstrap

you can pass --force flag to override this check and force phase /checks rollback
```

lib/fsm/fsm.go
```diff
@@ -329,13 +329,28 @@ You can redeploy upgrade agents on all cluster nodes using "./gravity agent depl
...
 		PhaseID:  p.PhaseID,
 		Force:    p.Force,
 		Resume:   p.Resume,
 		Rollback: true,
 		Progress: p.Progress,
 	})
+	if err == nil {
+		// if the remote phase rollback is successful, we need to mark it in our local database
+		// because etcd might not be available to synchronize the changes back to us
+		err = f.ChangePhaseState(ctx,
```
bernardjkim, line 342: `gravity rollback` would fail in the middle of rolling back a plan whenever a phase was rolled back remotely. 
```
[vagrant@node-1 installer]$ sudo ./gravity rollback
You are about to rollback the following operation:
Type                 ID                                       State                  Created
----                 --                                       -----                  -------
operation_update     c8e539eb-23ef-4e7b-999c-e022ce04963b     update_in_progress     2020-10-07 01:07
Consider checking the operation plan and using --dry-run flag first to see which actions will be performed.
You can suppress this warning in future by providing --confirm flag.
Proceed? (yes/no):
yes
Wed Oct  7 01:14:20 UTC Rolling back "/masters/node-3/untaint" locally
Wed Oct  7 01:14:24 UTC Rolling back "/masters/node-3/endpoints" locally
Wed Oct  7 01:14:31 UTC Rolling back "/masters/node-3/uncordon" locally
Wed Oct  7 01:14:39 UTC Rolling back "/masters/node-3/taint" locally
Wed Oct  7 01:14:45 UTC Rolling back "/masters/node-3/system-upgrade" on node node-3
        Still rolling back "/masters/node-3/system-upgrade" on node node-3 (10 seconds elapsed)
        Still rolling back "/masters/node-3/system-upgrade" on node node-3 (20 seconds elapsed)
        Still rolling back "/masters/node-3/system-upgrade" on node node-3 (30 seconds elapsed)
        Still rolling back "/masters/node-3/system-upgrade" on node node-3 (40 seconds elapsed)
Wed Oct  7 01:15:31 UTC operation(update(c8e539eb-23ef-4e7b-999c-e022ce04963b), cluster=dev.test, created=2020-10-07 01:07) finished in 1 minute
[ERROR]: phase /masters/node-3/drain cannot be rolled back because some phases that depend on it haven't been rolled back yet. Please rollback the following phases first:
        /masters/node-3/system-upgrade
you can pass --force flag to override this check and force phase /masters/node-3/drain rollback
```
This was because `rollbackPhaseRemotely` was not updating the local plan after rolling back a phase.

r0mant (approved):

PR #1975, opened 2020-08-06
[7.0.x] service CIDR update

a-palchikov:
## Description

This PR implements changes to fix support for updating service CIDR on a live cluster.

## Type of change

* Regression fix (non-breaking change which fixes a regression)
* This change has a user-facing impact

## Linked tickets and other PRs

* Updates http://github.com/gravitational/gravity/issues/936
* Requires https://github.com/gravitational/gravity.e/pull/4327, https://github.com/gravitational/planet/pull/722.
* Ports https://github.com/gravitational/gravity/pull/1839, https://github.com/gravitational/gravity/pull/1864.

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Write documentation
- [ ] Address review feedback

## Testing done

Install `6.1.33` 3-node cluster.

Original configuration:
```
$ gravity resource get clusterconfiguration
Configuration
=============
Pod IP Range:         10.244.0.0/16
Service IP Range:     10.100.0.0/16
Cloud
-----
Provider:     onprem
```

Update service CIDR to `10.200.0.0/16`:

```yaml
kind: ClusterConfiguration
version: v1
spec:
  global:
    # CIDR range for Services in Cluster
    serviceCIDR: "10.200.0.0/16"
```
```
$ gravity resource create servicecidr.yaml  --confirm
...
$ gravity resource get clusterconfiguration
Configuration
=============
Pod IP Range:         10.244.0.0/16
Service IP Range:     10.200.0.0/16
Cloud
-----
Provider:     onprem
```

Upgrade to the version from this branch and update service CIDR again:

```
kind: ClusterConfiguration
version: v1
spec:
  global:
    # CIDR range for Services in Cluster
    serviceCIDR: "10.203.0.0/16"
```

```
$ gravity resource create servicecidr-203.yaml  --confirm
...
$ gravity resource get clusterconfiguration
Configuration
=============
Pod IP Range:         10.244.0.0/16
Service IP Range:     10.203.0.0/16
Cloud
-----
Provider:     onprem

$ kubectl get svc --all-namespaces
NAMESPACE     NAME                     TYPE           CLUSTER-IP       EXTERNAL-IP   PORT(S)                      AGE
default       kubernetes               ClusterIP      10.203.0.1       <none>        443/TCP                      46m
kube-system   bandwagon                ClusterIP      10.203.87.106    <none>        80/TCP                       46m
kube-system   gravity-site             LoadBalancer   10.203.224.97    <pending>     3009:32009/TCP               46m
[...]

PR #2204, opened 2020-10-07
Check for conflicts between host and overlay or service networks

### LINKED ISSUE #867: [FEATURE] Add check for conflicts between host network config and overlay network
From a slack chat:
```
Hi, Can you please consider to add a check while gravity installation so if my host has the same ip address like the '—service-cidr' or ‘pod-network-cidr’ it will exit with error
For example, I had the default service cidr which is ‘10.100.0.0/16’ and my host was in 10.100.0.21 which cause to several issues like dns problem and in gravity status: ‘docker-registry (healtz check failed: Get https://leader.telekube.local:5000/v2/: net/http: request canceled while waiting for connection (client.Timeout exceeded while awaiting headers))’ (edited) 
```

lenko-d:
## Description

* This PR adds a check for overlapping values between host and overlay or service networks.
* The check is performed during install and aborts the install if the host IP is in the range of the service or overlay networks.
* The check is also performed during resource config update operation performed on a live cluster
* The check is needed because having overlapping values creates problems. Example problems could be 
 DNS issues or healtz check failure.

## Type of change

* Closes https://github.com/gravitational/gravity/issues/867
* Requires https://github.com/gravitational/gravity/pull/2208

## TODOs

- [X] Self-review the change
- [X] Write tests
- [X] Perform manual testing
- [X] Address review feedback

## Testing done

Verified manually from the command line. Passed values that are in the range of the cidrs and got the expected error messages below. 
Also tested with the default values of the cidrs.

```
sudo ./gravity install    --config=config.yaml             --token=secret     

Starting installer
[ERROR]: The selected advertise address 172.16.76.130 conflicts with the service network CIDR range 172.16.76.128/26. Please specify a different service CIDR.


config.yaml:

kind: ClusterConfiguration
version: v1
spec:
  global:
    serviceCIDR:  "172.16.76.129/16"
    podCIDR: "173.16.76.129/16"



```

```
$ sudo gravity resource create config.yaml
[ERROR]: The advertise address 172.31.78.148 conflicts with the service network CIDR range 172.31.78.148/16. Please specify a different service CIDR.


the config.yaml file has the following:

kind: ClusterConfiguration
version: v1
spec:
  global:
    serviceCIDR:  "172.31.78.148/16"
    podCIDR: "173.31.78.148/16"

``` 

---

```
sudo ./gravity install         --advertise-addr=172.31.74.157         --token=secret         --service-cidr=172.31.74.128/26         --pod-network-cidr=172.31.74.128/26         --cloud-provider=aws
Thu Oct  8 14:53:43 UTC	Starting installer
[ERROR]: The selected advertise address 172.31.74.157 conflicts with the service network CIDR range 172.31.74.128/26. Please specify a different service CIDR via --service-cidr flag.

```
---

```
sudo ./gravity install         --advertise-addr=172.31.74.157         --token=secret         --service-cidr=172.31.75.128/26         --pod-network-cidr=172.31.74.128/26         --cloud-provider=aws
Thu Oct  8 15:00:00 UTC	Starting installer
[ERROR]: The selected advertise address 172.31.74.157 conflicts with the pod network CIDR range 172.31.74.128/26. Please specify a different pod CIDR via --pod-network-cidr flag.

```


tool/gravity/cli/config.go
```diff
@@ -335,6 +335,11 @@ func (i *InstallConfig) CheckAndSetDefaults() (err error) {
 	if err := checkLocalAddr(i.AdvertiseAddr); err != nil {
 		return trace.Wrap(err)
 	}
+
+	if err := checkOverlayNet(i.AdvertiseAddr, i.ServiceCIDR, i.PodCIDR); err != nil {
```
a-palchikov, line 339: These command line options is not the only way to configure service/pod subnets - ClusterConfiguration resource is another way to set these up.
During installation these options get merged with an optional cluster configuration resource which then gets validated as a whole. I now spotted a problem that the [PR](https://github.com/gravitational/gravity/pull/1975/) which introduced these changes was not ported to master which I would do before doing these changes.

tool/gravity/cli/config.go
```diff
@@ -1352,6 +1357,33 @@ func checkLocalAddr(addr string) error {
...
+	_, serviceSubnet, err := net.ParseCIDR(serviceCIDR)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	ip := net.ParseIP(advertiseAddr)
+	if ip == nil {
+		return trace.BadParameter("unable to parse advertiseAddr(%s) ", advertiseAddr)
+	}
+	if serviceSubnet.Contains(ip) {
+		return trace.BadParameter("advertiseAddr(%s) is in range of the serviceCIDR %s",
```
r0mant, line 1371: Keep in mind that these errors will be shown in a terminal so try to make them easily understandable by an average user and helpful by providing possible resolution advice, for example:

```
The selected advertise address %v conflicts with the service network CIDR range %v.
Please specify a different service CIDR via --service-cidr flag.
```

Same for other error messages here.

tool/gravity/cli/config.go
```diff
@@ -1352,6 +1357,33 @@ func checkLocalAddr(addr string) error {
 		addr, strings.Join(availableAddrs, ", "))
 }
 
+// checkOverlayNet verifies that the host network address is not in the range of the serviceCIDR or podCIDR
+func checkOverlayNet(advertiseAddr string, serviceCIDR string, podCIDR string) error {
```
r0mant, line 1361: I think that in addition to checking the advertise address itself we also need to check other network interfaces on the node in case of the multi-homed node? @knisbet What do you think?

tool/gravity/cli/config.go
r0mant, line 339: Nit: "overlay" usually refers only to the pod network, not service network, so I'd rename this method to something more generic, e.g. `checkNetworkConflict` or something.

r0mant (commented):
When we settle on the final implementation, please also write a unit test for this conflict checking logic, I think it'd be useful to have.

tool/gravity/cli/config.go
a-palchikov: I submitted https://github.com/gravitational/gravity/pull/2208 to backport changes and I'd like this PR to go after.

a-palchikov:
I've merged the https://github.com/gravitational/gravity/pull/2208 with service CIDR changes so this PR can be updated to verify the unified CIDR configuration.

tool/gravity/cli/config.go
knisbet: I think this gets into rabbit hole territory in the context of detecting conflicts via cluster configuration on a presumably running cluster. For overlay network conflicts it's probably feasible, since the overlay network doesn't get created on any virtual interfaces in the host network namespace. Although there might be alternate implementations out their that would show an interface in the host NS we don't expect. 

Pod cidr would be a whole lot harder, as on a running cluster there may be interfaces assigned within the range on the host (think the flannel bridge). While we can probably control for our own interfaces, this would conflict if anyone tries to use their own CNI plugin. 

So my 2 cents would be to go with the minimal implementation for now. I think to properly check all network interfaces, would require some attention to detail, to only filter to say non virtual / non internal interfaces.

lib/validate/net.go
```diff
@@ -22,6 +22,25 @@ import (
 	"github.com/gravitational/trace"
 )
 
+// NetworkOverlap verifies that ipAddr is not in the range of the subnetCIDR
+func NetworkOverlap(ipAddr string, subnetCIDR string, errMsg string) error {
```
r0mant, line 26: Nit: Could you shorter form here since all arguments are of the same type:

```go
func NetworkOverlap(ipAddr, subnetCIDR, errMsg string) error {
```

lib/validate/net.go
```diff
@@ -22,6 +22,25 @@ import (
...
+	_, subNet, err := net.ParseCIDR(subnetCIDR)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	ip := net.ParseIP(ipAddr)
+	if ip == nil {
+		return trace.BadParameter("Invalid IP address (%v).", ipAddr)
```
r0mant, line 34: Nit: Usually, we don't capitalize/dot error messages for better embedding ([error strings](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)). An exception would be when the error represents a full message that will be displayed to a user.

```go
return trace.BadParameter("invalid IP address (%v)", ipAddr)
```

tool/gravity/cli/clusterconfig.go
```diff
@@ -280,7 +281,26 @@ func validateClusterConfig(localEnv *localenv.LocalEnvironment, update libcluste
...
 	}
-	return validate.ClusterConfiguration(existing, update)
+	err = validate.ClusterConfiguration(existing, update)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	server, err := findLocalServer(cluster.ClusterState.Servers)
+	if err != nil {
+		return trace.NotFound("unable to find server info")
```
r0mant, line 291: I would make the error more helpful for potential troubleshooting i.e. "unable to find local node among cluster state servers: %v".

tool/gravity/cli/clusterconfig.go
```diff
@@ -280,7 +281,26 @@ func validateClusterConfig(localEnv *localenv.LocalEnvironment, update libcluste
...
+	}
+	if err := validate.NetworkOverlap(server.AdvertiseIP, update.GetGlobalConfig().ServiceCIDR, fmt.Sprintf("The advertise address %v conflicts with the service network CIDR range %v. Please specify a different service CIDR.",
+		server.AdvertiseIP, update.GetGlobalConfig().ServiceCIDR)); err != nil {
+		return trace.Wrap(err)
+	}
+
+	if err := validate.NetworkOverlap(server.AdvertiseIP, update.GetGlobalConfig().PodCIDR, fmt.Sprintf("The advertise address %v conflicts with the pod network CIDR range %v. Please specify a different pod CIDR.",
```
r0mant, line 298: Nit: I would move the message into a separate variable so the lines are shorter and code is easier to read. As a rule of thumb, I use 80-100 characters as a cut-off point for a line length.

```go
message := fmt.Sprintf("xxx", ...)
if err := validate.NetworkOverlap(xxx, yyy, message); err != nil {
```

This applies to all 4 places where this is called.

r0mant (approved):
Mostly stylistic nits, otherwise lgtm.

r0mant:
@lenko-d Once merged, let's please backport this to 7.0 as well.

lib/validate/net.go
a-palchikov, line 34: nit: `Invalid` should not be capitalized in the error message.

lib/validate/net_test.go
```diff
@@ -73,3 +73,40 @@ func (*S) TestValidateKubernetesSubnets(c *check.C) {
...
+			ipAddr:      "11.100.0.111",
+			subnetCIDR:  "10.100.0.0/16",
+			ok:          true,
+			description: "Outside the range.",
+		},
+	}
+	for _, tc := range testCases {
+		err := NetworkOverlap(tc.ipAddr, tc.subnetCIDR, "errMsg")
+		if tc.ok {
+			c.Assert(err, check.IsNil, check.Commentf(tc.description))
```
a-palchikov, line 107: Ideally this should verify that the err is also an expected one. You could used ErrorMatches checker instead.

tool/gravity/cli/config.go
```diff
@@ -349,6 +349,17 @@ func (i *InstallConfig) CheckAndSetDefaults(validator resources.Validator) (err
 	if err := checkLocalAddr(i.AdvertiseAddr); err != nil {
 		return trace.Wrap(err)
 	}
+
+	if err := validate.NetworkOverlap(i.AdvertiseAddr, i.ServiceCIDR, fmt.Sprintf("The selected advertise address %v conflicts with the service network CIDR range %v. Please specify a different service CIDR via --service-cidr flag.",
```
a-palchikov, line 353: Like I hinted at previously, this will miss the service/pod CIDR configuration coming from a resource file. I'd look into [validateResources](https://github.com/gravitational/gravity/blob/version/7.0.x/tool/gravity/cli/config.go#L650) instead.

a-palchikov (approved):

PR #2209, opened 2020-10-08
Certificate changes according to the latest MacOS specific requirements.

### LINKED ISSUE #868: [BUG] Gravity self signed certificates don't meet requirements MacOS Catalina
Gravity: 6.1.11
Browser: Opera or Chrome on MacOS Catalina

image

On MacOS this is generating error about the certificate being revoked in several browsers. Doing some very quick research, this appears to be due to requirements changes on self signed certs in MacOS catalina.

https://superuser.com/questions/1492207/neterr-cert-revoked-in-chrome-chromium-introduced-with-macos-catalina

Macos cataline has some new requirements:
https://support.apple.com/en-us/HT210176

## Implementation details

Self-signed cert is generated in https://github.com/gravitational/gravity/blob/version/7.0.x/lib/process/process.go#L2320.

lenko-d:
## Description

This PR changes the gravity's self-signed certificate properties in order to make it work according to the latest MacOS specific requirements. Automated cert rotation mechanism is implemented in order to mitigate the impact of the reduced validity period. 

## Type of change

* Closes https://github.com/gravitational/gravity/issues/868

## TODOs

- [X] Self-review the change
- [X] Write tests
- [X] Perform manual testing
- [X] Address review feedback

## Implementation

Copy-pasted existing function from Teleport's code base(GenerateSelfSignedCert). Modified the function according to the new requirements:
* Reduced the validity period from 10 years to 825 days
* Added Extended Key Usage extension with value x509.ExtKeyUsageServerAuth

### Verification of the web UI
Verified that the gravity installer web page works when opened from a Mac laptop and when having the new certificate. 

### Verification of the certificate parameters
Used openssl from the command line to dump the properties of the new certificate and verified that they are according to the expectations. The expected new properties are described in [this article](http://blog.nashcom.de/nashcomblog.nsf/dx/more-strict-server-certificate-handling-in-ios-13-macos-10.15.htm).

```
$ openssl x509 -text -noout -in  localhost.pem 
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            2d:94:80:4d:03:5d:b7:94:e7:a0:83:b0:b7:e3:10:4e
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: C = US, O = localhost, CN = localhost
        Validity
            Not Before: Oct  8 06:27:07 2020 GMT
            Not After : Jan 11 06:27:07 2023 GMT
        Subject: C = US, O = localhost, CN = localhost
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                RSA Public-Key: (2048 bit)
                Modulus:
                    00:da:d4:ae:b4:20:6b:f2:b2:bc:c2:ae:2a:80:e9:
                    8e:2c:80:d8:66:e5:6b:77:c9:46:80:f4:e5:ff:14:
                    88:73:d9:67:3d:6c:12:f8:9b:3e:04:a5:9e:9e:bd:
                    d4:81:d5:13:48:3f:14:b8:5c:cc:42:ce:68:d7:f3:
                    84:c7:dc:a9:63:55:a8:3a:cd:57:b1:8c:7b:3c:15:
                    82:ac:54:e5:c0:dc:ab:4b:47:b3:51:7b:bf:b3:59:
                    da:e7:a3:87:bb:20:27:ff:22:12:aa:f2:a7:1b:b9:
                    6c:b5:88:cd:c7:66:61:3b:71:d8:a5:5f:2b:c4:be:
                    5e:b5:6c:0d:24:98:31:4f:69:a1:80:d5:6a:92:50:
                    36:65:41:7a:fc:d7:a9:0a:52:ae:09:f0:29:b7:c6:
                    85:ea:a7:79:8b:9e:f1:f9:e8:42:15:d9:ba:97:c5:
                    99:b9:e2:d9:66:44:1f:60:67:68:5b:0d:51:57:44:
                    03:f1:b5:08:38:ca:17:48:05:5f:e9:fd:7f:45:f9:
                    fd:e0:09:0e:a4:f0:fa:34:46:c1:be:1c:ae:24:4c:
                    fc:83:0d:a1:45:35:ea:6f:c5:2c:a6:19:0c:f3:6d:
                    94:dc:30:6d:b3:7c:28:b2:d3:15:55:47:28:98:f7:
                    09:f2:18:75:02:20:ef:79:5f:c3:b5:a3:c4:0f:72:
                    0e:c7
                Exponent: 65537 (0x10001)
        X509v3 extensions:
            X509v3 Key Usage: critical
                Digital Signature, Key Encipherment, Certificate Sign
            X509v3 Extended Key Usage: 
                TLS Web Server Authentication
            X509v3 Basic Constraints: critical
                CA:TRUE
            X509v3 Subject Alternative Name: 
                DNS:osboxes, DNS:localhost, DNS:10.0.2.15, DNS:localhost.local, IP Address:127.0.0.1
    Signature Algorithm: sha256WithRSAEncryption
.......

```

### Testing of the auto rotation functionality.
Reduced the validity period of the cert to 5 min and the expiration checker's period to 2 minutes. The logs below show the rotation events:
```

2020-10-16T00:58:24Z INFO [PROCESS]   The cert with SerialNumber=22039170944473623024545722472042953135 will expire soon. Replacing it with a new one... mode:site process/watch.go:89
2020-10-16T00:58:24Z INFO [PROCESS]   Got event {"ClusterCertificateUpdated" <nil>}, restarting listener 0.0.0.0:3009. mode:site process/process.go:1760
2020-10-16T00:58:24Z INFO [PROCESS]   Successfully rotated the self-signed cluster certificate. mode:site process/watch.go:106
2020-10-16T00:58:24Z INFO [PROCESS]   Cluster certificate is already initialized. mode:site process/process.go:1947
2020-10-16T00:58:24Z INFO [PROCESS]   Serving on 0.0.0.0:3009. mode:site process/process.go:1808

``` 

### A user uploads their own cert (no rotation):

```
Uploaded a certificate that doesn't have "Gravitational" as the OrganisationalUnit:
$ sudo gravity resource create  my_certs2.yaml
Updated TLS keypair

Verified in the logs that the uploaded certificate doesn't get rotated:

2020-10-16T22:01:46Z INFO [PROCESS]   Running self signed certificate expiration watch. mode:site process/watch.go:69
2020-10-16T22:01:46Z INFO [PROCESS]   Skipping expiration check for customer provided certificate. mode:site process/watch.go:92

```

### Web UI continues working after the cert has expired (Firefox on Linux):
Loaded the cluster management URL and the web page was working.

### Cluster join command works when the cert has expired:
```
Last completed operation:
    * Join node lenko-HP (192.168.1.102) as node
      ID:		b64e467f-f6e4-42d3-953d-35f2c91a1ca3
      Started:		Sun Oct 18 00:39 UTC (5 minutes ago)
      Completed:	Sun Oct 18 00:41 UTC (3 minutes ago)

```

### In MacOS the web page still works after cert expiration.
Opened the Ops center web page and it still works after the self signed certificate has expired.


lib/utils/tls.go
```diff
@@ -54,6 +62,74 @@ type CertificateName struct {
 	OrganizationalUnit []string `json:"org_unit"`
 }
 
+// TLSCredentials keeps the typical 3 components of a proper HTTPS configuration
+type TLSCredentials struct {
```
r0mant, line 66: Let's just reuse this struct from Teleport codebase. I.e. copy/update the method, but we don't have to copy this struct, just use Teleport's one.

lib/utils/tls.go
```diff
@@ -54,6 +62,74 @@ type CertificateName struct {
...
+	// PublicKey in PEM format
+	PublicKey []byte
+	// PrivateKey in PEM format
+	PrivateKey []byte
+	Cert       []byte
+}
+
+// GenerateSelfSignedCert generates a self signed certificate that
+// is valid for given domain names and ips, returns PEM-encoded bytes with key and cert
```
r0mant, line 75: Let's mention in the godoc that this method is supposed to generate certificate compatible with macOS requirements and insert a link to that Apple's KB article.

lib/utils/tls.go
```diff
@@ -54,6 +62,74 @@ type CertificateName struct {
...
+	}
+
+	template := x509.Certificate{
+		SerialNumber:          serialNumber,
+		Issuer:                entity,
+		Subject:               entity,
+		NotBefore:             notBefore,
+		NotAfter:              notAfter,
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
```
r0mant, line 103: Add a comment here to indicate this is a part of the macOS cert requirements, same as you did above for the validity duration?

lib/utils/tls.go
```diff
@@ -18,15 +18,23 @@ package utils
...
 	"io/ioutil"
+	"math/big"
+	"net"
 	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/gravitational/gravity/lib/archive"
 	"github.com/gravitational/gravity/lib/defaults"
+	"github.com/sirupsen/logrus"
```
r0mant, line 37: Nit: Not a huge deal, but 3rd party imports usually go into the 3rd group according to Go code review comments document, so move it to the group below?

r0mant (commented):
I would also consider writing a unit test that will make sure the requirements are satisfied (duration/extkeyusage).

a-palchikov (approved):

lib/utils/tls.go
```diff
@@ -54,6 +63,67 @@ type CertificateName struct {
...
+// is valid for given domain names and ips, returns PEM-encoded bytes with key and cert
+// Generates a certificate that is compatible with the MacOS requirements described at:
+// https://support.apple.com/en-us/HT210176
+func GenerateSelfSignedCert(hostNames []string) (*teleutils.TLSCredentials, error) {
+	priv, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	notBefore := time.Now()
+	notAfter := notBefore.Add(time.Hour * 24 * 825) // 825 days or fewer is the required validity period for MacOS
```
r0mant, line 76: So I'm thinking, the new validity period for the certificate is just over 2 years and I think a lot of folks just go with never updating the certificate to their own, which means that it will stop working for them after 2 years. It will also affect not just the web UI but operations since various gravity commands also call HTTP APIs.

I wonder if gravity should also detect if the certificate is about to expire and reissue it. Or do you think we will migrate to something like K8s CSR API by then so it'll no longer be an issue? @knisbet @a-palchikov what do you guys think?

a-palchikov: I think it's a valid point and should not be too hard to implement even if we move to CSR APIs in the future.

knisbet: 1. Yea, I think we'll need some way to rotate the certificate automatically. This will be tricky, because we probably don't want to replace a user provided certificate even if it's expired. 
    1. It really depends on how the client reacts to an expired self signed cert. If it produces a similar alert to a self signed cert, this only really needs to be manual/cli driven flow. If it's a hard error like happens to chrome currently, then it should be automatic sometime before expiry. Because requirements / browsers may change in the future, it's probably safer to automate though. In which case I'd probably even go even tighter than 825 days.
    1. For user provided certs, probably should have an alert that the cert is about to expire as well. Can treat this as a separate feature though.
    
2. I'm not as sure gravity operations will be effected on present builds. AFAIK we don't properly configure the trust store anywhere for the internal comms, and we have lots of hard code for setting insecure on all TLS connections which I haven't had the opportunity to fix. This is awful, but however that gets solved should account for the ability for certs to rotate, or use separate long term trust certs for m2m connections.
    1. My knee jerk thought is long term the internal comms should probably hang off their own internal CA/Cert that the web server presents using SNI, and you know, setup the trust anchors.

lib/process/watch.go
```diff
@@ -18,23 +18,98 @@ package process
...
+func (p *Process) runCertExpirationWatch(client *kubernetes.Clientset) clusterService {
+	p.Info("Starting self signed certificate expiration watch.")
+
+	return func(ctx context.Context) {
+		ticker := time.NewTicker(time.Hour * 24)
+		defer ticker.Stop()
+		for {
+			err := p.replaceCertIfAboutToExpire(client)
+			if err != nil {
+				p.Errorf("Failed to check for certificate expiration: %v.", trace.DebugReport(err))
```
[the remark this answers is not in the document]
r0mant: > Also, I'm a bit wary about letting this loop run on each gravity-site replica - it would make more sense to run on just the leader.

Unless I'm missing anything, it will run on leader as implemented (since it's launched via `p.startService`).

Agree about retrying on error.

lib/process/watch.go
```diff
@@ -18,23 +18,98 @@ package process
...
+	}
+
+	if len(cert.Issuer.OrganizationalUnit) == 0 || !strings.Contains(cert.Issuer.OrganizationalUnit[0], utils.SelfSignedCertOrg) {
+		p.Info("Skipping expiration check for customer provided certificate.")
+		return nil
+	}
+
+	// This is the time window to replace the certificate before expiration.
+	// Lets Encrypt recommends to renew 30 days before expiration.
+	periodBeforeExpire := time.Now().Add(time.Hour * 24 * 30)
```
r0mant, line 88: I'd move this period to lib/defaults.

lib/process/watch.go
```diff
@@ -18,23 +18,98 @@ package process
...
+	}
+
+	block, _ := pem.Decode(clusterCert)
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	if len(cert.Issuer.OrganizationalUnit) == 0 || !strings.Contains(cert.Issuer.OrganizationalUnit[0], utils.SelfSignedCertOrg) {
+		p.Info("Skipping expiration check for customer provided certificate.")
```
r0mant, line 82: Nit: I think this should be a Debug level log entry.

lib/utils/tls.go
```diff
@@ -31,7 +38,13 @@ import (
 	cfsslhelpers "github.com/cloudflare/cfssl/helpers"
 	dockerarchive "github.com/docker/docker/pkg/archive"
 	"github.com/gravitational/license/authority"
+	teleutils "github.com/gravitational/teleport/lib/utils"
 	"github.com/gravitational/trace"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	SelfSignedCertOrg = "Gravitational"
```
r0mant, line 47: Missing godoc. Also, we usually put things like this in lib/defaults.

lib/process/watch.go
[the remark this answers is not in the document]
knisbet: I think with renewing 30 days before expiry, this leaves plenty of attempts at the 24 hour timer, so my 2 cents is we don't need the added complexity of a rapid retry on error.

Although my corner case concern may be a bit different, I wonder if we have any clusters where the elections are unstable, and I wonder if there is a possibility a cluster won't stay the leader for 24 hours.

lib/utils/tls.go
```diff
@@ -54,6 +67,71 @@ type CertificateName struct {
...
+// GenerateSelfSignedCert generates a self signed certificate that
+// is valid for given domain names and ips, returns PEM-encoded bytes with key and cert
+// Generates a certificate that is compatible with the MacOS requirements described at:
+// https://support.apple.com/en-us/HT210176
+func GenerateSelfSignedCert(hostNames []string) (*teleutils.TLSCredentials, error) {
+	priv, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	notBefore := time.Now()
```
knisbet, line 79: IIRC it's generally recommended to set this time slightly in the past, to ensure the cert will be accepted by clients with clock drift errors. I believe letsencrypt backdates by 1 hour.

lib/utils/tls.go
knisbet, line 47: I don't know what the best bet is here. I don't really have an issue with gravitational. The only advantage I see with using something else, is when looking at the certification and/or trying to search the sources it might be easier to locate if it was a string that isn't so common. So an org like "Gravitational Self-Signed Web Access" would create a distinction if someone looks at the cert, and is also easier to grep out of the code base.

lib/utils/tls.go
```diff
@@ -54,6 +67,71 @@ type CertificateName struct {
...
+		NotAfter:              notAfter,
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth}, // MacOS specific requirement
+		BasicConstraintsValid: true,
+		IsCA:                  true,
+	}
+
+	// collect IP addresses localhost resolves to and add them to the cert. template:
+	template.DNSNames = append(hostNames, "localhost.local")
+	ips, _ := net.LookupIP("localhost")
```
knisbet, line 112: Need to be careful of a very particular oddity of golang and some base images missing /etc/nsswitch.conf. I checked and at least on my test cluster we're missing the nsswitch configuration.

Several versions of golang when missing nsswitch configuration, will use DNS without local host resolution. This means this lookup could query external DNS servers and be subject to a number of unexpected results.

Just one of those very peculiar quirks of go. I would consider just hard coding the response to 127.0.0.1 instead of relying on name resolution of the localhost name.

knisbet (approved):
I think my only real concern is the weirdness in golang when missing nssswitch, so I'd change that before merging.

lib/process/watch.go
```diff
@@ -35,6 +41,100 @@ import (
...
+				continue
+			}
+
+			cert, err := tlsca.ParseCertificatePEM(clusterCert)
+			if err != nil {
+				p.WithError(err).Error("Failed to parse the certificate.")
+				continue
+			}
+
+			if len(cert.Issuer.OrganizationalUnit) == 0 || !strings.Contains(cert.Issuer.OrganizationalUnit[0], defaults.SelfSignedCertWebOrg) {
```
r0mant, line 95: Any particular reason this is using `strings.Contains()` instead of `==`?

lib/utils/tls_test.go
```diff
@@ -71,3 +74,22 @@ func (s *TLSSuite) TestCertArchive(c *C) {
...
 }
+
+// TestGenerateSelfSignedCert verifies that MacOS specific requirements are met according to:
+// https://support.apple.com/en-us/HT210176
+func (s *TLSSuite) TestGenerateSelfSignedCert(c *C) {
+	hostNames := []string{"localhost"}
+	credentials, err := GenerateSelfSignedCert(hostNames)
+	c.Assert(err, IsNil)
+
+	block, _ := pem.Decode([]byte(credentials.Cert))
```
r0mant, line 85: I think `block` may be `nil` which will result in a panic below. Not a huge deal for the test, but I would assert that block isn't nil anyway.

lib/process/watch.go
```diff
@@ -35,6 +41,100 @@ import (
...
+
+	for {
+		select {
+		case tm := <-ticker.C:
+			if tm.IsZero() {
+				return trace.ConnectionProblem(nil, "timed out waiting while checking for certificate expiration")
+			}
+			clusterCert, _, err := opsservice.GetClusterCertificate(client)
+			if err != nil {
+				p.WithError(err).Error("Failed to retrieve the certificate from k8s.")
```
r0mant, line 85: In the log messages here and below, I would specify what certificate is being referred to exactly, otherwise it may not be clear from the logs. I.e. "cluster web UI certificate" or something like this.

lib/process/process.go
```diff
@@ -1972,6 +1973,11 @@ func (p *Process) initClusterCertificate(ctx context.Context, client *kubernetes
 		return trace.Wrap(err)
 	}
 
+	err = p.replaceCertIfAboutToExpire(ctx, p.client)
```
r0mant, line 1976: Hmm, if the intention of this is to detect an old expiring certificate upon process start, then shouldn't this be called above? Otherwise this function (`initClusterCertificate`) will detect in the beginning that the certificate is already initialized and exit, thus this code will never really be reached.

lenko-d: The intention of this is to handle the case of the cert being deleted using:
```
 gravity resource rm tls keypair
```
In this case the cert will be loaded from the file and this check will verify that the cert from the file hasn't expired. 

I will also add it above to verify the cert on processed restart.

r0mant (approved):

PR #2225, opened 2020-10-13
Release notes for 6.1.43

knisbet:
I'm not sure how we want to communicate that we don't intend to support the 1k node support through the entire 6.1 lifecycle.


r0mant (approved):
> I'm not sure how we want to communicate that we don't intend to support the 1k node support through the entire 6.1 lifecycle.

@knisbet One idea is instead of saying "support up to 1000 nodes" say something more generic, like "Make a series of performance and scalability enhancements to allow building larger clusters".

Issue #1191, opened 2020-03-03
Add support for incremental upgrades


r0mant:
A common scenario is when a user needs to quickly patch a running application, for example to update a Docker image.

With generic clusters, it’s just a matter of pushing the image to the appropriate registry but with Gravity the images pushed to the active Docker registry out-of-band are not synced among registries on other master nodes and are removed during the garbage collection operation.

To handle this scenario Gravity can provide a way to build an incremental upgrade image (a “patch”) that can only be applied to the version it was built off of:

```
# Build an incremental image that only contains layers different from the specified version.
$ tele helm build ./mattermost --base/since/from=mattermost-2.2.0.tar --output=mattermost-2.2.1.tar

# Upgrade the release running 2.2.0.
$ gravity helm upgrade <release-name> mattermost-2.2.1.tar
```

PR #2235, opened 2020-10-15
Incremental upgrades

r0mant:
## Description

I'm still going through some testing but decided that it wouldn't hurt to open PR to start gathering feedback as I don't expect the implementation to change significantly.

This is a pretty large changeset but what it does basically comes down to a couple of things:

* `tele build` (and `tele helm build`) accepts an `--upgrade-from` flag with the path to an installer tarball to build an incremental upgrade image from. The difference of the incremental image is that it only contains Docker images that are not present in the "base" tarball. It also doesn't include any of the system packages/apps, only user application.
* `tele build` also accepts a `--diff` flag which, combined with `--upgrade-from`, displays a diff table b/w images in the base tarball vs the one being built which looks kinda like this:

```
$ tele build robot-shop-app-2/resources/app.yaml --upgrade-from=robot-shop-1.1.0.tar -f --diff
                              robot-shop v1.1.0         robot-shop v1.2.0
                              -----------------         -----------------
gravitational/debian-tall     buster                    buster
rabbitmq                      3.7-management-alpine     3.7-management-alpine
redis                         4.0.6 (removed)           4.0.7 (added)
robotshop/rs-cart             latest                    latest
...
robotshop/rs-shipping         latest                    latest
robotshop/rs-user             latest                    latest
robotshop/rs-web              0.4.20 (removed)          0.4.30 (added)
```

In the example above, only the changed images (`redis:4.0.7` and `robotshop/rs-web:0.4.30`) will be vendored in the incremental image.

Installation using a partial image is prohibited. When it comes to upgrading, the process is exactly the same as using a "full" cluster/application image, it is basically the same as the scenario where runtime doesn't change and there's only application upgrade. This works out nicely because all Docker images missing from the tarball are supposed to be present in the cluster's registry. There is one gotcha described in the implementation section.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/1191.

## TODOs

- [x] Self-review the change
- [x] Write tests
- [ ] Perform manual testing
- [ ] Write documentation
- [ ] Address review feedback

## Implementation

One implementation detail worth mentioning is how an incremental upgrade image is handled when transferred to the cluster. Because it only contains a subset of all required Docker images, uploading it to the cluster as-is would be problematic because after GC the missing images would be gone from the registry.

For this reason, I "reconstruct" the incremental image's registry layers from the cluster's registry before uploading it to the cluster. This way it basically becomes a "full" image and as such no changes to garbage collection or anything else are required. It also lets validate the "integrity" of the application because the cluster registry is supposed to have all non-vendored images present and the upload will fail if any of them are missing (which would otherwise have resulted in a failed upgrade).

Another detail, is I introduced a "status" field in our manifest object, similar to K8s, to store some runtime kind of information. It is currently populated during tele build and stores information about Docker images the application depends on and vendors which is needed for some logic. I also tried another approach with storing it in a package envelope metadata but the manifest approach turned out to be cleaner and easier to work with.

## Performance/Scaling

Does not apply.

## Testing done

Lots of scenarios, gonna keep the list here for reference. I've tested most (if not all) of these before, but will be checking those off as I retest them:

### Build

- [x] Build an incremental upgrade cluster image
  - [x] Make sure --diff produces correct output
- [ ] Build an incremental upgrade app image
  - [ ] Make sure --diff produces correct output
- [ ] Building cluster image from cluster image with different base is prohibited

### Install/upgrade

- [x] Install cluster image, perform incremental upgrade
  - [x] Validate can join nodes and they pull the app properly
  - [x] Validate can run GC and everything’s preserved
  - [ ] Validate can export the app from cluster
- [ ] Install application image, perform incremental upgrade
- [x] Attempt to install incremental image should return an error
- [ ] Attempt to upgrade using incremental image from version other than the one if was built off of should show a warning
- [x] Attempt to upgrade the cluster with different base version should fail

### DIFF
--- a/lib/app/apps.go
+++ b/lib/app/apps.go
@@ -20,10 +20,12 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"time"
 
 	"github.com/gravitational/gravity/lib/app/hooks"
 	"github.com/gravitational/gravity/lib/defaults"
 	"github.com/gravitational/gravity/lib/loc"
+	"github.com/gravitational/gravity/lib/pack"
 	"github.com/gravitational/gravity/lib/schema"
 	"github.com/gravitational/gravity/lib/utils"
 
@@ -174,4 +176,33 @@ func GetDirectDeps(app Application) (deps []loc.Locator, err error) {
 		return nil, trace.Wrap(err)
 	}
 	return append(manifest.Dependencies.GetApps(), app.Package), nil
+}
+
+// ImportApplication imports the specified stream of application data into
+// the provided package/application services.
+func ImportApplication(stream io.ReadCloser, packages pack.PackageService, apps Applications) (*Application, error) {
+	progressC := make(chan *ProgressEntry)
+	errorC := make(chan error, 1)
+	err := packages.UpsertRepository(defaults.SystemAccountOrg, time.Time{})
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	operation, err := apps.CreateImportOperation(&ImportRequest{
+		Source:    stream,
+		ProgressC: progressC,
+		ErrorC:    errorC,
+	})
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	for range progressC {
+	}
+	if err := <-errorC; err != nil {
+		return nil, trace.Wrap(err)
+	}
+	app, err := apps.GetImportedApplication(*operation)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	return app, nil
 }
--- a/lib/app/service/layer.go
+++ b/lib/app/service/layer.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2020 Gravitational, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package service
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/gravitational/gravity/lib/app"
+	"github.com/gravitational/gravity/lib/blob/fs"
+	"github.com/gravitational/gravity/lib/defaults"
+	"github.com/gravitational/gravity/lib/pack"
+	"github.com/gravitational/gravity/lib/pack/layerpack"
+	"github.com/gravitational/gravity/lib/pack/localpack"
+	"github.com/gravitational/gravity/lib/storage/keyval"
+
+	"github.com/gravitational/trace"
+)
+
+// LayeredApps contains layered package and application services.
+type LayeredApps struct {
+	// Package is a layered package service.
+	Packages pack.PackageService
+	// Apps is app service based on the layered package service.
+	Apps app.Applications
+	// dir is the directory with the new write layer data.
+	dir string
+}
+
+// Cleanup removes the write layer directory.
+func (l LayeredApps) Cleanup() error {
+	if l.dir != "" {
+		return os.RemoveAll(l.dir)
+	}
+	return nil
+}
+
+// NewLayeredApps returns layered package and application services where
+// the provided package service serves as a read-only layer.
+func NewLayeredApps(readPackages pack.PackageService) (*LayeredApps, error) {
+	dir, err := ioutil.TempDir("", "")
+	if err != nil {
+		return nil, trace.ConvertSystemError(err)
+	}
+	backend, err := keyval.NewBolt(keyval.BoltConfig{
+		Path:  filepath.Join(dir, defaults.GravityDBFile),
+		Multi: true,
+	})
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	objects, err := fs.New(filepath.Join(dir, defaults.PackagesDir))
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	packages, err := localpack.New(localpack.Config{
+		Backend:     backend,
+		UnpackedDir: filepath.Join(dir, defaults.PackagesDir, defaults.UnpackedDir),
+		Objects:     objects,
+	})
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	layeredPackages := layerpack.New(readPackages, packages)
+	layeredApps, err := New(Config{
+		StateDir:    filepath.Join(dir, defaults.ImportDir),
+		Backend:     backend,
+		Packages:    layeredPackages,
+		UnpackedDir: filepath.Join(dir, defaults.PackagesDir, defaults.UnpackedDir),
+	})
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	return &LayeredApps{
+		Packages: layeredPackages,
+		Apps:     layeredApps,
+		dir:      dir,
+	}, nil
+}
--- a/lib/app/service/layer.go
+++ b/lib/app/service/layer.go
@@ -32,63 +32,80 @@ import (
 	"github.com/gravitational/trace"
 )
 
-// LayeredApps contains layered package and application services.
-type LayeredApps struct {
-	// Package is a layered package service.
+// LayeredAppsConfig is the configuration for the layered service.
+type LayeredAppsConfig struct {
+	// Packages is the package service to use as a read layer.
 	Packages pack.PackageService
-	// Apps is app service based on the layered package service.
-	Apps app.Applications
-	// dir is the directory with the new write layer data.
-	dir string
+	// Dir is the directory where write layer will reside.
+	Dir string
 }
 
-// Cleanup removes the write layer directory.
-func (l LayeredApps) Cleanup() error {
-	if l.dir != "" {
-		return os.RemoveAll(l.dir)
+// CheckAndSetDefaults validates the config and sets default values.
+func (c *LayeredAppsConfig) CheckAndSetDefaults() (err error) {
+	if c.Packages == nil {
+		return trace.BadParameter("missing read layer Packages")
+	}
+	if c.Dir == "" {
+		if c.Dir, err = ioutil.TempDir("", ""); err != nil {
+			return trace.ConvertSystemError(err)
+		}
 	}
 	return nil
 }
 
+// LayeredApps contains layered package and application services.
+type LayeredApps struct {
+	// LayeredAppsConfig is the service configuration.
+	LayeredAppsConfig
+	// Package is a layered package service.
+	Packages pack.PackageService
+	// Apps is app service based on the layered package service.
+	Apps app.Applications
+}
+
+// Cleanup removes the write layer directory.
+func (l LayeredApps) Cleanup() error {
+	return os.RemoveAll(l.Dir)
+}
+
 // NewLayeredApps returns layered package and application services where
 // the provided package service serves as a read-only layer.
-func NewLayeredApps(readPackages pack.PackageService) (*LayeredApps, error) {
-	dir, err := ioutil.TempDir("", "")
-	if err != nil {
-		return nil, trace.ConvertSystemError(err)
+func NewLayeredApps(config LayeredAppsConfig) (*LayeredApps, error) {
+	if err := config.CheckAndSetDefaults(); err != nil {
+		return nil, trace.Wrap(err)
 	}
 	backend, err := keyval.NewBolt(keyval.BoltConfig{
-		Path:  filepath.Join(dir, defaults.GravityDBFile),
+		Path:  filepath.Join(config.Dir, defaults.GravityDBFile),
 		Multi: true,
 	})
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
-	objects, err := fs.New(filepath.Join(dir, defaults.PackagesDir))
+	objects, err := fs.New(filepath.Join(config.Dir, defaults.PackagesDir))
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
 	packages, err := localpack.New(localpack.Config{
 		Backend:     backend,
-		UnpackedDir: filepath.Join(dir, defaults.PackagesDir, defaults.UnpackedDir),
+		UnpackedDir: filepath.Join(config.Dir, defaults.PackagesDir, defaults.UnpackedDir),
 		Objects:     objects,
 	})
 	if err != nil {
 		return nil, trace.Wrap(err)
 	}
-	layeredPackages := layerpack.New(readPackages, packages)
+	layeredPackages := layerpack.New(config.Packages, packages)
 	layeredApps, err := New(Config{
-		StateDir:    filepath.Join(dir, defaults.ImportDir),
+		StateDir:    filepath.Join(config.Dir, defaults.ImportDir),
 		Backend:     backend,
 		Packages:    layeredPackages,
-		UnpackedDir: filepath.Join(dir, defaults.PackagesDir, defaults.UnpackedDir),
+		UnpackedDir: filepath.Join(config.Dir, defaults.PackagesDir, defaults.UnpackedDir),
 	})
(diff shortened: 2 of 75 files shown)


lib/docker/testhelpers.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"context"
+
+	"github.com/gravitational/gravity/lib/utils"
+
+	dockerapi "github.com/fsouza/go-dockerclient"
+)
+
+// MockDocker is a mock Docker client implementation used in unit tests.
+type MockDocker struct{}
```
a-palchikov, line 12: Would it make sense to use the real docker? We run tests in an environment that has docker access in at least one of the branches (might be master). What is the benefit of this mock?

lib/loc/diff.go
```diff
@@ -0,0 +1,75 @@
...
+	allRepos = utils.Deduplicate(allRepos)
+	sort.Strings(allRepos)
+	for _, repo := range allRepos {
+		repoTags := append(left.Tags(repo), right.Tags(repo)...)
+		repoTags = utils.Deduplicate(repoTags)
+		sort.Strings(repoTags)
+		result := RepositoryDiff{Repository: repo}
+		for _, tag := range repoTags {
+			if left.Has(repo, tag) && right.Has(repo, tag) {
+				result.Tags = append(result.Tags, TagDiff{
```
a-palchikov, line 55:
```go
result.Tags = append(result.Tags, TagDiff{
	Tag: tag,
	Left: left.Has(repo, tag),
	Right: right.Has(repo, tag),
})
```

lib/loc/diff.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/gravitational/teleport/lib/utils"
+)
+
+// RepositoryDiff represents the difference in Docker images for a particular repository.
+type RepositoryDiff struct {
+	// Repository is the Docker image repository.
+	Repository string
+	// Tags are the tags for this repository.
+	Tags []TagDiff
```
a-palchikov, line 30: Have you considered left/right groups of tags instead?
```go
type RepositoryDiff struct {
	// Repository is the Docker image repository.
	Repository string
	// Left/Right define the differences between repositories
	Left, Right []string
}
```

lib/loc/loc.go
```diff
@@ -172,6 +172,11 @@ func (l Locator) String() string {
 	return str
 }
 
+// Human returns user-friendly string representation of the locator.
+func (l Locator) Human() string {
```
a-palchikov, line 176: I mean just `Human` carries a bit different connotation if you ask me, how about `Description` or `Readable`?

tool/gravity/cli/clusterupdate.go
```diff
@@ -428,6 +431,29 @@ func (r *clusterInitializer) validatePreconditions(localEnv *localenv.LocalEnvir
...
+		return trace.Wrap(err)
+	}
+	apps, err := env.AppServiceCluster()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	err = app.VerifyDependencies(upgradeApp, apps, packages)
+	if err != nil {
+		log.WithError(err).Errorf("Failed to verify %v dependencies.", upgradeApp)
+		return trace.BadParameter(`There was an issue trying to verify %v cluster image integrity.
```
a-palchikov, line 448: `Some required dependencies are missing from the cluster store for image %v. Check logs for more details.` should suffice

tool/gravity/cli/commands.go
```diff
@@ -654,6 +654,8 @@ type UpdateUploadCmd struct {
 	*kingpin.CmdClause
 	// OpsCenterURL is cluster URL
 	OpsCenterURL *string
+	// Force allows to override certain preconditions.
```
a-palchikov, line 657: The uncertainty is beginning to make me feel uneasy. Maybe we should break the preconditions down to a list and provide clear override rules for each or just evaluate whether a generic override makes sense for all?

tool/gravity/cli/config.go
```diff
@@ -578,9 +577,23 @@ func (i *InstallConfig) shouldBootstrapSELinux() bool {
...
+			"Make sure you're running the installer from the directory with the contents "+
+			"of the installer tarball")
+	}
+	upgradeFrom, err := application.LabelAsLocator(pack.UpgradeFromLabel)
+	if err != nil && !trace.IsNotFound(err) {
+		return trace.Wrap(err)
+	}
+	if upgradeFrom != nil {
+		return trace.BadParameter("This cluster image was built as an incremental "+
+			"upgrade from %v. It cannot be used to perform full installation as "+
```
a-palchikov, line 593: as **it** only ...

tool/gravity/cli/ops.go
```diff
@@ -81,33 +82,87 @@ func uploadUpdate(ctx context.Context, tarballEnv *localenv.TarballEnvironment,
...
 
-	env.PrintStep("Importing application %v v%v", appPackage.Name, appPackage.Version)
-	_, err = appservice.PullApp(appservice.AppPullRequest{
-		SrcPack: tarballEnv.Packages,
-		SrcApp:  tarballEnv.Apps,
-		DstPack: clusterPackages,
-		DstApp:  clusterApps,
-		Package: *appPackage,
+	var registries []string
+	err = utils.Retry(defaults.RetryInterval, defaults.RetryLessAttempts, func() error {
```
a-palchikov, line 91: Frankly i never understood the retryattemps/lessattempts semantics - can we just define the timeout values for that?
 Also, do you think it might make sense to implement it with a backoff?

tool/gravity/cli/ops.go
```diff
@@ -81,33 +82,87 @@ func uploadUpdate(ctx context.Context, tarballEnv *localenv.TarballEnvironment,
...
+	}
+	images, err := docker.NewClusterImageService(registries[0])
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	sourcePackages := tarballEnv.Packages
+	sourceApps := tarballEnv.Apps
+
+	// Before importing check if the new version is an incremental upgrade
```
a-palchikov, line 109: Before importing, check

tool/gravity/cli/ops.go
```diff
@@ -81,33 +82,87 @@ func uploadUpdate(ctx context.Context, tarballEnv *localenv.TarballEnvironment,
...
+	err = utils.Retry(defaults.RetryInterval, defaults.RetryLessAttempts, func() error {
+		registries, err = getRegistries(ctx, env, cluster.ClusterState.Servers)
+		return trace.Wrap(err)
 	})
 	if err != nil {
-		if !trace.IsAlreadyExists(err) {
+		return trace.Wrap(err)
+	}
+	if len(registries) == 0 {
+		return trace.NotFound("could not determine cluster registry addresses")
```
a-palchikov, line 99: this sounds like something getRestries should handle - i.e. return a corresponding error for? In this case, the error context might be closer to the error.

tool/gravity/cli/ops.go
```diff
@@ -81,33 +82,87 @@ func uploadUpdate(ctx context.Context, tarballEnv *localenv.TarballEnvironment,
...
+	// Before importing check if the new version is an incremental upgrade
+	// and if so, check whether the currently installed version can be
+	// upgraded to it.
+	upgradeFrom, err := application.LabelAsLocator(pack.UpgradeFromLabel)
+	if err != nil && !trace.IsNotFound(err) {
+		return trace.Wrap(err)
+	}
+	if upgradeFrom != nil {
+		if !cluster.App.Package.IsEqualTo(*upgradeFrom) && !force {
+			return trace.BadParameter(`This cluster image was built as an incremental upgrade from %v.
```
a-palchikov, line 118: This does not state what the problem is - I'd be puzzled at what is expected.
`This cluster image was built as an incremental upgrade from %v which differs from the currently installed cluster image %v. Upgrading over a different image might break the cluster ...`

tool/tele/cli/buildparameters.go
```diff
@@ -0,0 +1,129 @@
...
+
+// BuildParameters represents the arguments provided for building an application
+type BuildParameters struct {
+	// StateDir is build state directory, if was specified
+	StateDir string
+	// SourcePath is the path to a manifest file or a Helm chart to build image from
+	SourcePath string
+	// OutPath holds the path to the installer tarball to be output
+	OutPath string
+	// Overwrite indicates whether or not to overwrite an existing installer file
```
a-palchikov, line 33: nit: whether to overwrite

tool/tele/cli/buildparameters.go
```diff
@@ -0,0 +1,129 @@
...
+	Overwrite bool
+	// SkipVersionCheck indicates whether or not to perform the version check of the tele binary with the application's runtime at build time
+	SkipVersionCheck bool
+	// Silent is whether builder should report progress to the console
+	Silent bool
+	// Verbose turns on more detailed progress output
+	Verbose bool
+	// Insecure turns on insecure verify mode
```
a-palchikov, line 41: turns off TLS verification

tool/tele/cli/commands.go
```diff
@@ -86,6 +86,10 @@ type HelmBuildCmd struct {
 	Values *[]string
 	// Pull allows to force-pull Docker images even if they're already present.
 	Pull *bool
+	// UpgradeFrom is a path to the image to build an incremental image off of.
+	UpgradeFrom *string
+	// Diff shows differences between two images without building the image.
+	Diff *bool
```
a-palchikov, line 92: This sounds like a different command tbh

lib/app/service/layer.go
```diff
@@ -0,0 +1,94 @@
...
+func (l LayeredApps) Cleanup() error {
+	if l.dir != "" {
+		return os.RemoveAll(l.dir)
+	}
+	return nil
+}
+
+// NewLayeredApps returns layered package and application services where
+// the provided package service serves as a read-only layer.
+func NewLayeredApps(readPackages pack.PackageService) (*LayeredApps, error) {
```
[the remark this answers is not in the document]
r0mant: It would introduce some circular dependencies in the restore.go where it's used, unless I also move that elsewhere. I think it's fine where it is, by analogy with the layered package service which is under lib/pack.

lib/app/service/vendor.go
```diff
@@ -143,96 +205,153 @@ func (v *vendorer) VendorTarball(ctx context.Context, tarball io.ReadCloser, req
...
+	_, err = v.VendorDir(ctx, unpackedDir, req)
+	if err != nil {
+		return "", trace.Wrap(err)
+	}
+	return unpackedDir, nil
+}
+
+// Images returns a list of Docker image references extracted from the resources.
+func (v *vendorer) Images(unpackedDir string, req VendorRequest) (result []loc.DockerImage, err error) {
+	err = req.CheckAndSetDefaults(unpackedDir)
```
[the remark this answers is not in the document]
r0mant: I don't think it would help simplify things much.

lib/app/service/vendor.go
```diff
@@ -143,96 +205,153 @@ func (v *vendorer) VendorTarball(ctx context.Context, tarball io.ReadCloser, req
...
+		// Rewrite app/package dependencies according to --set-dep flags.
+		makeRewriteDepsFunc(req.SetDeps),
+		// Rewrite packages with meta versions.
+		makeRewritePackagesMetadataFunc(v.packages),
+		// Rewrite image metadata.
+		makeRewriteAppMetadataFunc(req.Repository, req.PackageName, req.PackageVersion),
+	}
+}
+
+// getManifestRewrites returns a list of transformations to apply to Docker
```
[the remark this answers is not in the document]
r0mant: I think getManifestRewrites is a more suitable name because it returns manifest-specific rewrites.

lib/app/service/vendor.go
```diff
@@ -143,96 +205,153 @@ func (v *vendorer) VendorTarball(ctx context.Context, tarball io.ReadCloser, req
...
 
 // VendorDir creates an application tarball from the unpackedDir using configuration specified in req.
 //
 // It will detect and import missing docker images and rewrite image references in all resource files
 // to point to a fixed docker registry address.
-func (v *vendorer) VendorDir(ctx context.Context, unpackedDir string, req VendorRequest) error {
-	if req.ProgressReporter == nil {
-		req.ProgressReporter = utils.DiscardProgress
+func (v *vendorer) VendorDir(ctx context.Context, unpackedDir string, req VendorRequest) (*VendorResponse, error) {
+	log.Infof("%#v", req)
```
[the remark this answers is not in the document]
r0mant: Yeah, less typing :)

lib/app/service/vendor_test.go
```diff
@@ -40,6 +43,100 @@ type VendorSuite struct{}
...
+	sort.Sort(ByRepository(result.VendoredImages))
+	sort.Sort(ByRepository(result.Images))
+	c.Assert(result, DeepEquals, &VendorResponse{
+		VendoredImages: []loc.DockerImage{
+			{Repository: "install-hook", Tag: "latest"},
+			{Repository: "mysql", Tag: "5.7"},
+			{Repository: "nginx"},
+			{Repository: "wordpress", Tag: "5.4-apache"},
+		},
+		Images: []loc.DockerImage{
```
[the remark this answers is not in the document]
r0mant: It contains all discovered images.

lib/builder/inspector.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	manifest, err := generateApplicationImageManifest(chart)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	vendorer, err := service.NewVendorer(service.VendorerConfig{})
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	images, err := vendorer.Images(path, vendor)
```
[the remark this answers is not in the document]
r0mant: It relies on much of the same logic as vendorer (to perform rewrites), hence I made it a part. I was actually originally thinking about this as a "dry run" mode for vendoring so it made sense to place it here.

lib/docker/testhelpers.go
r0mant: Just mocks it in the new vendoring test I added where we don't need Docker.

lib/loc/diff.go
r0mant: That was a nice one :)

lib/loc/diff.go
r0mant: The current data structure is more convenient to operate with on the caller side.

tool/gravity/cli/commands.go
r0mant: Yeah good call. Renamed this to `--skip-version-check`.

tool/gravity/cli/ops.go
r0mant: Not sure I fully follow the proposal here. I still need to check the returned list length to be sure that I can take the first element?

tool/tele/cli/commands.go
r0mant: I thought about it. On one hand, it does feel like a separate command, but on the other I think the UX with the flag is a little better. I.e. you run tele build in a sort of dry run mode to see what will be packaged, then remove the flag and build the actual image. I don't have strong points in favor of one approach vs another though, @knisbet @bernardjkim what do you guys think?

lib/app/service/vendor.go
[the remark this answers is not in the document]
a-palchikov: I don't follow - this method is called `getImagesRewrites`, not `getManifestRewrites` - I simply pointed out the name inconsistency.

tool/tele/cli/commands.go
bernardjkim: I'm in favor of keeping the diff flag. I agree with @r0mant that this provides a slightly better UX as it is now.

lib/app/apps.go
```diff
@@ -175,3 +177,32 @@ func GetDirectDeps(app Application) (deps []loc.Locator, err error) {
...
+	}
+	operation, err := apps.CreateImportOperation(&ImportRequest{
+		Source:    stream,
+		ProgressC: progressC,
+		ErrorC:    errorC,
+	})
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	for range progressC {
```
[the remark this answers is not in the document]
a-palchikov: It is to achieve blocking mode as application import is asynchronous to the caller.

a-palchikov (approved):
`getImagesRewrites` is still referred to as `getManifestRewrites` in the doc-comment and some unit tests are failing.

wadells:
Closing until we allocate priority to pick this work up again.

Issue #2009, opened 2020-08-14
Gravity 7.0.10, 6.1.x does not install with SELinux enabled


pablodcar:
### Description

**What happened**:
When installing gravity in a RHEL 8.1 selinux is enabled by default and the following error appears:

```
[ec2-user@ip-10-1-0-13 installer]$ sudo ./gravity install --advertise-addr=10.1.0.13 --token=secr3t-token --cluster=auto-install-three --cloud-provider=generic --flavor=three --license="$(cat license.pem)"  --role=general_node
Wed Aug 12 18:29:46 UTC	Starting enterprise installer
To abort the installation and clean up the system,
press Ctrl+C two times in a row.
If you get disconnected from the terminal, you can reconnect to the installer
agent by issuing 'gravity resume' command.
If the installation fails, use 'gravity plan' to inspect the state and
'gravity resume' to continue the operation.
See https://gravitational.com/gravity/docs/cluster/#managing-an-ongoing-operation for details.
Wed Aug 12 18:29:47 UTC	Connecting to installer
[ERROR]: error enabling the service
	failed to enable /home/ec2-user/installer/.gravity/gravity-installer.service: Failed to enable unit: Access denied
		exit status 1
```

**What you expected to happen**:
It should install as with gravity 5.5 which works without disabling selinux by hand. Using `setenforce 0` made the install to progress.

**How to reproduce it (as minimally and precisely as possible)**:
Try to install gravity 7.0.10 in a AWS EC2 instance using the AMI RHEL-8.1.0_HVM-20191029-x86_64-0-Hourly2-GP2.

### Environment

 - Gravity version [e.g. 7.0.11]: 7.0.10
 - OS [e.g. Redhat 7.4]: Red Hat 8.1
 - Platform [e.g. Vmware, AWS]: AWS, AMI RHEL-8.1.0_HVM-20191029-x86_64-0-Hourly2-GP2

**Browser environment**

- Browser Version (for UI-related issues):
- Install tools:
- Others:

**Relevant Debug Logs If Applicable**
```
status 1 stderr:"Failed to enable unit: Access denied\n" stdout: utils/logging.go:103
2020-08-12T21:15:15Z WARN             Command failed. error:[
ERROR REPORT:
Original Error: *exec.ExitError exit status 1
Stack Trace:
	/gopath/src/github.com/gravitational/gravity/lib/utils/exec.go:216 github.com/gravitational/gravity/lib/utils.ExecWithInput
[...]

PR #2240, opened 2020-10-15
[7.0.x] service state

a-palchikov:
## Description

This PR fixes the installer/expand/upgrade operations on CentOS 8/RHEL by moving the gravity binary to the system location intended for binary files. This is important on SELinux-enabled systems where `init_t` (systemd) domain does not have permissions to work with executables in arbitrary locations (e.g. user home directories or `/var/lib/gravity`).
This prompted a series of changes that mainly concern the agent service's operational state to maintain it the same way it was maintained previously. The main problem is supporting both installer/agent services on a single box (when the installer is running in `remote` mode - i.e. does not itself represent a joining agent) which requires that the state is separate for each.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/2009
* Requires https://github.com/gravitational/gravity.e/pull/4345

## TODOs

- [x] Self-review the change
- [x] Perform manual testing (see below)
- [ ] Write documentation
- [ ] Address review feedback

## Testing done

 * [x] Installs on a single node
 * [x] Expands
 * [ ] Node reconfiguration
 * [ ] Upgrades


tool/gravity/cli/operation.go
```diff
@@ -508,7 +510,8 @@ func (r *backendOperations) listUpdateOperation(environ LocalEnvironmentFactory)
 }
 
 func (r *backendOperations) listJoinOperation(environ LocalEnvironmentFactory) error {
-	env, err := environ.NewJoinEnv()
+	// TODO(dmitri): assumes local state inside process's working directory
```
r0mant, line 513: What does this mean? Join-related plan commands have to be run from a specific directory? Is that a change in behavior?

lib/state/state.go
```diff
@@ -144,12 +144,40 @@ func LogDir(baseDir string, suffixes ...string) string {
 	return filepath.Join(append(elems, suffixes...)...)
 }
 
+// GravityInstallerServicePath returns the path to the installer service
+// as configured in the state directory rooted at baseDir.
+func GravityInstallerServicePath(baseDir string) (path string, err error) {
```
r0mant, line 149: Is this function actually used anywhere? Couldn't find it.

tool/gravity/cli/install.go
```diff
@@ -683,6 +704,10 @@ func completePlanFromService(
 	operation ops.SiteOperation,
 	connecting, connected string,
 ) error {
+	strategy, err := newResumeStrategy(utils.Exe.WorkingDir)
```
r0mant, line 707: Could you please explain a bit what kind of impact this change has on UX? Are all of these commands now depend on the working dir of the binary somehow?

I.e. will users still be able to just execute `gravity join`, `gravity plan`, etc. commands from wherever on cluster nodes?

tool/gravity/cli/register.go
```diff
@@ -129,6 +130,7 @@ func RegisterCommands(app *kingpin.Application) *Application {
 	g.AutoJoinCmd.Token = g.AutoJoinCmd.Flag("token", "Unique token to authorize this node to join the cluster.").Hidden().String()
 	g.AutoJoinCmd.SELinux = g.AutoJoinCmd.Flag("selinux", "Run with SELinux support. Default 'false'.").Default("false").Envar(defaults.GravitySELinuxEnv).Bool()
 	g.AutoJoinCmd.FromService = g.AutoJoinCmd.Flag("from-service", "Run in service mode.").Hidden().Bool()
+	g.AutoJoinCmd.StateDir = g.AutoJoinCmd.Arg("path", "Path to the directory with local state. Defaults to working directory.").String()
```
r0mant, line 133: What are these flags for? How does this change behavior for existing users (if it does)?

lib/defaults/defaults.go
```diff
@@ -371,6 +374,12 @@ const (
 	// StatBin is stat executable path inside planet
 	StatBin = "/usr/bin/stat"
 
+	// AlternativeBinDir defines the default location for binaries on Ubuntu Core
```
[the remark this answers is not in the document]
a-palchikov: This will simplify things but not much. On the good side, we'll be throwing away code - not changing it.

tool/gravity/cli/operation.go
a-palchikov: No, this is not a change in behavior - it is note to self when we need to change this behavior.

lib/state/state.go
a-palchikov: Yes, it is used [here](https://github.com/gravitational/gravity/blob/dmitri/7.0.x/2009-service-state/tool/gravity/cli/install.go#L954)

tool/gravity/cli/install.go
a-palchikov: No, I'm simply pulling the previously implicit state configuration to the top so it's easier to change if necessary - there is no change in behavior.

tool/gravity/cli/register.go
a-palchikov: These flags specify to the service where the operation-specific state is.
Now that I'm thinking about it, we can continue using the process's working directory for the service since this is how we set up installer/join agents in the first place.

a-palchikov:
It looks like using process's working directory does not work in all cases  specifically when the operation-specific phases are remotely executed by individual calls to `gravity plan execute`. The plan now is to attempt and move operation-specific state to a fixed location (probably under state directory) so this needs a bit more work.

lib/ops/opsservice/instructions.go
```diff
@@ -36,32 +36,33 @@ var (
 	// gravityTemplateSource is a bash script that downloads gravity binary
 	// from an Ops Center and installs it into /usr/bin
 	gravityTemplateSource = `
+
```
r0mant, line 39: Is this an empty template now?

r0mant (approved):
lgtm if operations are working as expected

lib/ops/opsservice/instructions.go
a-palchikov: Forgot to remove it.

a-palchikov:
Going to merge as the failures seem to be unrelated to robotest (which ran fine).

Issue #2229, opened 2020-10-14
Old tele binaries will bundle newer gravity packages/binaries if they're in the package cache


wadells:
## Description

**What happened**:
I build a cluster image with the 7.0.12 tele binary, and found out that it contained the 7.0.20 gravity binary.

**What you expected to happen**:
I expected 7.0.12 tele to use its version as the runtime base if an alternative was not specified via `--with-base-image`

**How to reproduce it (as minimally and precisely as possible)**:
1) clear the tele build cache (or create a new one and supply it with `--state-dir`)
```
$ rm -rf ~/.gravity               
```
2) build a cluster image with the 7.0.12 tele binary -- check the image's gravity version (7.0.12)
<details><summary>7.0.12 tele build & gravity version</summary>

```
$ /home/walt/.cache/robotest/ent/tele/tele-7.0.12 version && /home/walt/.cache/robotest/ent/tele/tele-7.0.12 build ~/git/gravity/assets/robotest/upgrade-base/app.yaml --version=7.0.12 --output=robotest-7.0.12.tar
Edition:        enterprise
Version:        7.0.12
Git Commit:     8a1dbd9c0cfa4ae4ce11a63e3b4a8d6c7b9226e7
Helm Version:   v2.15
Wed Oct 14 20:13:28 UTC Building cluster image robotest 7.0.12
Wed Oct 14 20:13:28 UTC Selecting base image version
        Will use base image version 7.0.12
Wed Oct 14 20:13:28 UTC Downloading dependencies from https://get.gravitational.io
        Still downloading dependencies from https://get.gravitational.io (10 seconds elapsed)
        Still downloading dependencies from https://get.gravitational.io (20 seconds elapsed)
        Still downloading dependencies from https://get.gravitational.io (30 seconds elapsed)
        Still downloading dependencies from https://get.gravitational.io (40 seconds elapsed)
        Still downloading dependencies from https://get.gravitational.io (50 seconds elapsed)
        Still downloading dependencies from https://get.gravitational.io (1 minute elapsed)
Wed Oct 14 20:14:29 UTC Embedding application container images
        Using local image quay.io/gravitational/debian-tall:buster
        Using local image quay.io/gravitational/debian-tall:buster
        Using local image k8s.gcr.io/echoserver:1.0
        Vendored image gravitational/debian-tall:buster
        Still embedding application container images (10 seconds elapsed)
        Still embedding application container images (20 seconds elapsed)
[...]

PR #2247, opened 2020-10-16
[7.0.x] save manifest file prior to vendoring

a-palchikov:
lib/builder: always write the manifest file in the vendor directory to persist the overridden runtime version. If the original manifest runtime version uses a template like `0.0.0+latest`, it might lead to the wrong (e.g. more recent) version of runtime to be actually used for building if the cache directory has a recent version of packages.

Updates https://github.com/gravitational/gravity/issues/2229.

## Type of change

* Refs #2229

## TODOs

- [x] Self-review the change
- [ ] Write tests
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

So far I've tested manually on patched 7.0.12/7.0.20. Will see if I can add a manual test.

### DIFF
--- a/lib/builder/builder.go
+++ b/lib/builder/builder.go
@@ -88,8 +88,6 @@ type Config struct {
 
 	// manifestDir is the fully-qualified directory path where manifest file resides
 	manifestDir string
-	// manifestFilename is the name of the manifest file
-	manifestFilename string
 }
 
 // CheckAndSetDefaults validates builder config and fills in defaults
@@ -115,8 +113,7 @@ func (c *Config) CheckAndSetDefaults() error {
 			return trace.Wrap(err)
 		}
 		c.manifestDir = filepath.Dir(manifestAbsPath)
-		c.manifestFilename = filepath.Base(manifestAbsPath)
-		if c.manifestFilename != defaults.ManifestFileName {
+		if filepath.Base(manifestAbsPath) != defaults.ManifestFileName {
 			return trace.BadParameter("manifest filename should be %q",
 				defaults.ManifestFileName)
 		}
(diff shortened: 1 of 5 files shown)


lib/builder/builder.go
```diff
@@ -259,19 +256,10 @@ func (b *Builder) Vendor(ctx context.Context, dir string) (io.ReadCloser, error)
...
-		data, err := yaml.Marshal(b.Manifest)
-		if err != nil {
-			return nil, trace.Wrap(err)
-		}
-		err = ioutil.WriteFile(manifestPath, data, defaults.SharedReadMask)
-		if err != nil {
-			return nil, trace.Wrap(err)
-		}
+	manifestPath := filepath.Join(dir, defaults.ResourcesDir, defaults.ManifestFileName)
+	err = writeManifest(manifestPath, b.Manifest)
```
r0mant, line 260: Do we actually update base version to the selected runtime version in the manifest anywhere? I can't see it, so I'm not sure just copying the manifest does anything?

wadells:
Some documentation about the expected behavior of:, unspecified, `0.0.0`, `0.0.0+latest` (or `+stable`?) wouldn't go awry. I'm could be convinced `+latest` should get a more recent *patch* base image version if available.  However unspecified or `0.0.0` should get the the version specified with `--with-base-version` or the `tele`'s version.

https://github.com/gravitational/gravity/blob/8499843e7463a377975c29e0d9d099bc150082e8/lib/loc/loc.go#L33-L45

Docs probably are not needed until the master port.

wadells (approved):
Thanks for adding the test case!

knisbet (approved):

lib/schema/manifest.go
```diff
@@ -169,7 +169,12 @@ func (m *Manifest) SetBase(locator loc.Locator) {
 // WithBase returns a copy of this manifest with the specified base application
 func (m *Manifest) WithBase(locator loc.Locator) Manifest {
 	result := *m
-	result.SetBase(locator)
+	if m.BaseImage != nil {
+		result.BaseImage = &BaseImage{Locator: locator}
+	}
+	result.SystemOptions = &SystemOptions{
```
r0mant, line 175: Won't this basically override the entire systemOptions field the manifest may have set, rather than just setting the runtime locator?

a-palchikov: It's a good catch - I'll update it conditionally.

PR #2263, opened 2020-10-22
[5.2] bump planet

knisbet:
Bumping 5.2.x since it's required for intermediate upgrades.


a-palchikov (approved):

a-palchikov:
It fails for the same reason version/6.1.x builds were failing. I ported the necessary [changes](https://github.com/gravitational/gravity/pull/2268) to version/5.2.x as well so we need to rebase once they're merged.

Issue #1005, opened 2020-01-08
OpenEBS integration enhancements


r0mant:
Gravity (starting from release 7.0) has an out-of-the-box persistent storage support by integrating with [OpenEBS](https://docs.openebs.io/). Here's [documentation](https://gravitational.com/gravity/docs/storage/) on persistent storage integration and usage.

There are a couple of things that our integration is lacking in its current state.

## OpenEBS upgrades are not supported

Gravity currently ships with OpenEBS 1.7 (via [storage-app](https://github.com/gravitational/storage-app)) and upgrading it is not currently supported. We cannot just easily bump versions to the latest 2.1 release because upgrading OpenEBS requires an upgrade not just of the control plane components but of the data plane pods as well.

So, in addition to upgrading the control plane components via our [upgrade script](https://github.com/gravitational/storage-app/tree/master/images/hook), we'd also need to modify our upgrade operation to perform upgrades of the data plane pods using the instructions provided by OpenEBS.

Success criteria:

- [ ] Gravity (in master branch) uses the latest OpenEBS 2.1.0 (via the storage-app).
- [ ] The upgrade operation upgrades data plane pods in addition to control plane pods.
  - [ ] We might need to do a bit of research and find out what would happen if we don't upgrade data plane pods? Talk to the OpenEBS team (they're in K8s Slack now).
- [ ] We can rollback in case of unsuccessful upgrade.

OpenEBS upgrade-related documentation: https://docs.openebs.io/docs/next/upgrade.html.

## iSCSI tools are not delivered in planet

Planet has iSCSI tools installed in order for OpenEBS cStor engine to work but iscsid service is disabled - it relies on iscsid running on host at the moment. This causes issues if the version of iscsid running on host is incompatible with whatever is installed in planet. For example:

```
# host
$ iscsiadm -V
iscsiadm version 2.0-873
```
```
# planet
$ iscsiadm -V
iscsiadm version 2.0-874
```

With this configuration, when iscsid is running on host, cStor volumes are not attaching to pods with the error similar to:

```
failed for volume "pvc-11cb12e7-5060-11e8-8d24-9600000a1650" : failed to get any path for iscsi disk, last err seen:
iscsi: failed to attach disk: Error: iscsiadm: Could not login to [iface: default, target: iqn.2016-09.com.openebs.jiva:pvc-11cb12e7-5060-11e8-8d24-9600000a1650, portal: 10.43.53.198,3260].
[...]

PR #2288, opened 2020-10-27
OpenEBS enhancements

lenko-d:
## Description

This PR provides the following OpenEBS related enhancements:
* Automated upgrade of OpenEBS data and control plane components.
* Pre-flight checks to make sure that iscsid is not present on the host.
* Iscsid to be setup in Planet in such a way that it only starts when OpenEBS enabled deployment is present.

## Type of change

* Updates https://github.com/gravitational/gravity/issues/1005

## TODOs

- [X] Self-review the change
- [ ] Write tests
- [ ] Perform manual testing
- [ ] Address review feedback

### Pre-flight checks
Enabled and started the iscsid.service and iscsid.socket systemd units on the host. 
Verified that the install is aborted by the pre-flight checks:

```
[ERROR]: The following pre-flight checks failed:
	[×] potentially conflicting programs running: [iscsid], note this is an issue only before Gravity is installed
	[×] Found conflicting systemd service: iscsid.socket. Please stop and mask this service and try again.
	[×] Found conflicting systemd service: iscsid.service. Please stop and mask this service and try again.
```

### Upgrade of control and data plane components
Verified that upgrade of data plane and control plane components works on a single host cluster.


lib/checks/checks.go
```diff
@@ -109,9 +109,9 @@ type ManifestValidator struct {
 
 // RunBasicChecks executes a set of additional health checks.
 // Returns list of failed health probes.
-func RunBasicChecks(ctx context.Context, options *validationpb.ValidateOptions) (failed []*agentpb.Probe) {
+func RunBasicChecks(ctx context.Context, options *validationpb.ValidateOptions, openEBSEnabled bool) (failed []*agentpb.Probe) {
```
r0mant, line 112: Can this parameter be propagated via existing `ValidateOptions`?

tool/gravity/cli/status.go
```diff
@@ -94,7 +94,11 @@ func status(env *localenv.LocalEnvironment, printOptions printOptions) error {
 			env.Println("Failed to query Gravity cluster status. Running additional checks")
 		}
 
-		failed = checks.RunBasicChecks(ctx, nil)
+		localSite, err := clusterOperator.GetLocalSite(ctx)
+		if err != nil {
+			log.WithError(err).Warn("Failed to get local site.")
+		}
+		failed = checks.RunBasicChecks(ctx, nil, localSite.App.Manifest.OpenEBSEnabled())
```
r0mant, line 101: What will happen if the above call returns an error?

r0mant: And also, why is this called here, in the `gravity status` command?

lenko-d: > And also, why is this called here, in the `gravity status` command?

In order to handle the case of a user installing iscsid on the host after the installation is complete. In this case the status command will show that there is a conflict with iscsid. I will remove it If we don't want to handle this scenario.

lenko-d: > What will happen if the above call returns an error?

I will remove the OpenEBS related checks from here because it seems that we don't want to handle the case of a user installing iscsid on the host after the deployment is complete.

lib/checks/checks.go
lenko-d: > Can this parameter be propagated via existing `ValidateOptions`?

The problem is that there is a call with nil for ValidateOptions:

```
failed = checks.RunBasicChecks(ctx, nil)
```

And RunBasicChecks invokes:

```
		monitoring.DefaultProcessChecker(openEBSEnabled),
```

The above expects a bool and in this case I don't have the needed information if the propagation happens via ValidateOptions.

tool/gravity/cli/config.go
```diff
@@ -540,9 +540,10 @@ func (i *InstallConfig) RunLocalChecks() error {
 		Docker:   i.Docker,
 		Mounts:   i.Mounts,
 		Options: &validationpb.ValidateOptions{
-			VxlanPort: int32(i.VxlanPort),
-			DnsAddrs:  i.DNSConfig.Addrs,
-			DnsPort:   int32(i.DNSConfig.Port),
+			VxlanPort:    int32(i.VxlanPort),
+			DnsAddrs:     i.DNSConfig.Addrs,
+			DnsPort:      int32(i.DNSConfig.Port),
+			CheckOpenebs: app.Manifest.OpenEBSEnabled(),
```
a-palchikov, line 546: I'd drop the `Check` prefix - the struct itself already hints at the purpose.

lib/checks/checks.go
```diff
@@ -822,17 +847,29 @@ func currentServerTime(currentTime, heartbeatTime, serverTime time.Time) time.Ti
...
+	checkers := []health.Checker{
+		monitoring.NewIPForwardChecker(),
+		monitoring.NewBridgeNetfilterChecker(),
+		monitoring.NewMayDetachMountsChecker(),
+		monitoring.DefaultProcessChecker(processesToCheck),
+		defaultPortChecker(options),
+		monitoring.DefaultBootConfigParams(),
+	}
+	if checkISCSI {
+		checkers = append(checkers, monitoring.NewISCSIChecker(failedISCSIProbeMessage))
```
a-palchikov, line 869: It made sense to move the process list away from the process checker since it was a detail that was specific to a gravity version but factoring out an error message seems like an overkill? Do you think another error message would be used elsewhere?

lenko-d: Moving the error message out of Satellite is needed because the text in the error message contradicts the goal of the Satellite project. Satellite should become a standalone project independent of Gravity. The error message shows "Gravity" and as such if left in Satellite it will make it very specific to Gravity.

The other option is to remove the "Gravity" text from the error message but this contradicts the goal of showing a user friendly error message.  The current error message shown below makes it clear that this is specific to the Gravity application. Embedding a Gravity specific error message into the Satellite project will make it look like they are very tightly coupled. 

Moving the error message out will allow the iscsi checker in Satellite to be used as a general purpose checker. It will make it more independent by allowing possible future usage for other purposes.

For reference below is the Gravity specific message that was moved out of Satellite:

```
failedISCSIProbeMessage = "Found conflicting systemd service: %v. " +
		"If this service is present on the host it will interfere " +
		"with OpenEBS enabled applications running in Gravity." +
		"Please stop and mask this service and try again."

```

a-palchikov: Ok - my concern is that customizing via an error message is pretty fragile esp. here since it's also a placeholder. It would be more appropriate to move this to a configuration struct or accept a function instead - this way we can at least verify type safety and ensure intent:
```go
monitoring.NewISCSIChecker(fmt FormatISCSIError) Checker { ... }
type FormatISCSIError func(unitName string) string
```

a-palchikov (approved):

knisbet (approved):

lib/checks/checks.go
```diff
@@ -47,7 +47,36 @@ import (
...
+)
+
+var (
+	fmtISCSICheckFailedMsg = func(unitName string) string {
+		return fmt.Sprintf(failedISCSIProbeMessage, unitName)
+	}
+
+	// DefaultProcessesToCheck is the default list of processes to verify that are not running on the host.
+	// This list should be passed to the DefaultProcessChecker function.
+	DefaultProcessesToCheck = []string{
```
r0mant, line 64: Isn't this the same list as in satellite? Why not just use the one from there?

lib/expand/join.go
```diff
@@ -725,6 +725,7 @@ func (p *Peer) runLocalChecks(ctx operationContext) error {
 			VxlanPort: int32(installOperation.GetVars().OnPrem.VxlanPort),
 			DnsAddrs:  ctx.Cluster.DNSConfig.Addrs,
 			DnsPort:   int32(ctx.Cluster.DNSConfig.Port),
+			Openebs:   ctx.Cluster.App.Manifest.OpenEBSEnabled(),
```
r0mant, line 728: Nit: This field is generated from the proto file, right? Just curious if there's a way to make it `OpenEBS` (with capital abbreviation) so it's consistent. Don't worry about it if it's not trivial.

r0mant (commented):
Have you verified how this check behaves on the upgrade? I.e. if someone is upgrading from 6.1 to 7.0 with OpenEBS enabled.

lib/checks/checks.go
lenko-d: This list was moved from Satellite to here because it is Gravity specific. Satellite still has a default list but the obvious Gravity specific processes are removed from that list and the list is parametrized. This is needed because the goal of the Satellite project is to be independent of gravity. After the new changes it will be possible to use the functionality provided by Satellite as a general purpose process list checker.

knisbet: Satellite is not really an independent project from gravity at this point in time, and satellite would need a significant investment to work independently / with non gravity clusters.

lib/update/cluster/builder.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 Gravitational, Inc.
+Copyright 2020 Gravitational, Inc.
```
knisbet, line 2: We're not terribly consistent and should really script this at some point, but we use a date range to indicate the content was published and amended over time. So in this case for this particular file, it would be first_publication_year-last_publication_year which would be `Copyright 2019-2020 Gravitational, Inc.`

Example: https://github.com/gravitational/gravity/blob/master/tool/gravity/cli/register.go#L2

lib/update/cluster/builder.go
```diff
@@ -271,6 +273,68 @@ func (r phaseBuilder) openEBS(leadMaster storage.UpdateServer) *update.Phase {
...
+	}
+}
+
+func buildOpenEBSUpgradePhaseData(dataPlaneComponent string, fromVer string, toVer string) string {
+	return dataPlaneComponent + " " + fromVer + " " + toVer
+}
+
+func openEBSComponentToVer(storageAppVer string, openEBSComponentName string, openEBSComponentFromVer string) string {
+	if storageAppVer == "0.0.4" {
+		openEBSComponentToVer := "2.2.0"
```
knisbet, line 317: This value seems a bit buried, what if an upstream patch comes out and we need to bump this to 2.2.1. Ideally we would have a Makefile variable so it's easy to locate at bump a patch, and as long as the upgrade behaviour doesn't change, we can just update the variable / storage-app version and release.

lib/update/cluster/phases/openebs.go
```diff
@@ -0,0 +1,315 @@
...
+type PoolUpgrade struct {
+	Pool        string
+	FromVersion string
+	ToVersion   string
+	JobName     string
+}
+
+func (p *PhaseUpgradePool) execPoolUpgradeCmd(ctx context.Context) error {
+	jobName := k8sutil.MakeJobName(k8sJobPrefix, p.Pool)
+	out, err := k8sutil.ExecJob(ctx, jobName, k8sNamespace, poolUpgradeJobTemplate, &PoolUpgrade{Pool: p.Pool,
```
knisbet, line 85: Have you tested whether this is re-entrant? Does the phase work if run twice?

lib/update/cluster/phases/openebs.go
```diff
@@ -0,0 +1,315 @@
...
+        env:
+        - name: OPENEBS_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        tty: true
+
+        # the image version should be same as the --to-version mentioned above
+        # in the args of the job
+        image: openebs/m-upgrade:{{.ToVersion}}
```
knisbet, line 247: This looks like it's a reference to the dockerhub project, and if I follow the logic I don't see this get replaced anywhere. You'll need to include a reference to the container image within the storage-app resources, so they get vendored to the local registry, and update the path to point to the registry.

knisbet (changes requested):
Request changes for:
- I would make sure the image refs are updated to the local registry. Gravity clusters shouldn't be connecting to the internet to pull images.
- I would prefer to uses client-go for creating jobs instead of saving to a file and using kubectl to apply them unless there is a really good reason not to.

lib/kubernetes/jobs.go
```diff
@@ -0,0 +1,91 @@
...
+func ExecJob(ctx context.Context, jobSpecYAML []byte, client *kubernetes.Clientset) (string, error) {
+	jobSpecJSON, err := yaml.YAMLToJSON(jobSpecYAML)
+	if err != nil {
+		return "", trace.Wrap(err)
+	}
+
+	var jobSpec batchv1.Job
+	err = json.Unmarshal(jobSpecJSON, &jobSpec)
+	if err != nil {
+		return "", trace.Wrap(err)
```
knisbet, line 61: There actually a conversion buried in the rigging project to help convert kubernetes errors to the trace package equivalents: https://github.com/gravitational/rigging/blob/f933ef9d8a6c8786905755d274361d06d7d60ab7/utils.go#L232-L263

Example: https://github.com/gravitational/gravity/blob/master/lib/app/hooks/hooks.go#L254-L257

wadells:
Closing this until such time as we find priority to allocate someone to carry the work forward.

PR #2291, opened 2020-10-28
Remove the extraneous word "hours"

ulysseskan:
## Description
As it is now, the message displayed is, for example:
**Signup token has been created and is valid for 8h0m0s hours. Share this URL with the user:**

Since the amount of time is already in %v (in this case 8h), I suggest removing the extraneous word "hours".

## Type of change

* Closes #, refs #

* Requires #

* Ports #

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback


tool/gravity/cli/user.go
```diff
@@ -163,7 +163,7 @@ func inviteUser(env *localenv.LocalEnvironment, username string, roles []string,
 		return trace.Wrap(err)
 	}
 
-	fmt.Printf("Signup token has been created and is valid for %v hours. Share this URL with the user:\n%v\n\nNOTE: make sure this URL is accessible!\n",
+	fmt.Printf("Signup token has been created and is valid for %v  Share this URL with the user:\n%v\n\nNOTE: make sure this URL is accessible!\n",
```
r0mant, line 166: I think you didn't mean to remove the dot.

ulysseskan: If I left the dot, it would say "8h0m0s ." , right? That doesn't look quite right either, due to the extra space. Would concatenating the strings like this work in golang?

`... valid for %v" + ". Share this URL ...`

r0mant: No, just leave the dot after `%v`.

```go
fmt.Printf("Signup token has been created and is valid for %v. Share ...")
```

ulysseskan: Done, thanks.

r0mant (approved):

a-palchikov (approved):

lenko-d (approved):

knisbet (approved):

Issue #2314, opened 2020-11-11
Tracking: gravity-agent.service / others inheret systemd file limits


knisbet:
### Description

**What happened**:
gravity-agent failed in an upgrade, with errors about not being able to open file descriptors. It appears we're not explicitly setting this in our systemd units created on the hosts for gravity services. The defaults applied by systemd may not be sufficient for processes that need to network with others, or for large clusters.

TODO, explicitly set FD and Proc limits in generated systemd units placed on the host systemd.

knisbet:
Fixed on each branch

PR #2315, opened 2020-11-12
set 100000 fds on systemd unit creation if unspecified

knisbet:
## Description

Some distros / versions of systemd default to 1024 open fds. So by default if unspecified, create systemd units from gravity with 100000 fds as 1024 is far too low for gravity daemons by default.

## Type of change

Updates #2314

## TODOs

- [x] Self-review the change
- [x] Write tests
- [ ] Address review feedback

## Implementation

I just set this to infinity if not specified, to avoid any defaults / limits. 

## Performance/Scaling

I suppose in theory this could cause issues with lots of open connections that will present in a different way than being blocked due to max open fds.

## Testing done

Created an equivelant change and tested it on a test cluster. New Unit tests. Relying on robotest for this.


lenko-d (commented):
I think that we should make it the same as the current value in Planet. This is for the sake of consistency and also security. 

From a security point of view having a resource that is unlimited means that an attacker could potentially bring the system to its knees by opening too many fds. Having some resource limits is the simplest form of protection.

The value in Planet should be big enough:

```
cat /proc/sys/fs/file-max
1619372

```

knisbet:
> I think that we should make it the same as the current value in Planet. This is for the sake of consistency and also security.
> 
> From a security point of view having a resource that is unlimited means that an attacker could potentially bring the system to its knees by opening too many fds. Having some resource limits is the simplest form of protection.
> 
> I think that we should put the same value as in Planet. It should be big enough:
> 
> ```
> cat /proc/sys/fs/file-max
> 1619372
> ```

Setting to unlimited in the systemd configuration on LimitNoFile doesn't actually make the fds unlimited in the kernel.
```
root@kevin-test1:~# cat /proc/28039/limits
Limit                     Soft Limit           Hard Limit           Units
Max cpu time              unlimited            unlimited            seconds
Max file size             unlimited            unlimited            bytes
Max data size             unlimited            unlimited            bytes
Max stack size            8388608              unlimited            bytes
Max core file size        0                    unlimited            bytes
Max resident set          unlimited            unlimited            bytes
Max processes             60041                60041                processes
Max open files            1048576              1048576              files
Max locked memory         65536                65536                bytes
Max address space         unlimited            unlimited            bytes
Max file locks            unlimited            unlimited            locks
Max pending signals       60041                60041                signals
Max msgqueue size         819200               819200               bytes
Max nice priority         0                    0
Max realtime priority     0                    0
Max realtime timeout      unlimited            unlimited            us
```

This is not actually file-max, so I don't know where it pulls from.

lenko-d:
> Setting to unlimited in the systemd configuration on LimitNoFile doesn't actually make the fds unlimited in the kernel.

What is more important is that you are declaring an intention to have a resource that is unlimited. It may not end up being unlimited in reality but the fact that you define it like that doesn't sound good from a security point of view.

I'm not saying that Gravity will not pass a security audit because it has defined unlimited resource that could possibly enable denial of service attacks. I'm just saying that it doesn't look good to me to declare a resource as unlimited.

knisbet:
> > Setting to unlimited in the systemd configuration on LimitNoFile doesn't actually make the fds unlimited in the kernel.
> 
> What is more important is that you are declaring an intention to have a resource that is unlimited. It may not end up being unlimited in reality but the fact that you define it like that doesn't sound good from a security point of view.
> 
> I'm not saying that Gravity will not pass a security audit because it has defined unlimited resource that could possibly enable denial of service attacks. I'm just saying that it doesn't look good to me to declare a resource as unlimited.


That's fair, sometimes perceived problems are worth addressing, I changed it to 100k which should only be hit if there is a legit problem.

lenko-d (approved):

lib/systemservice/systemd_test.go
```diff
@@ -120,7 +125,91 @@ TasksMax=infinity
 [Install]
 WantedBy=test.target
 
-`)
+`,
+		},
+		{
+			description: "unspecified file limits set to infinity",
```
lenko-d, line 131: Infinity is not exactly correct.

lenko-d (changes requested):

lenko-d (approved):

r0mant (approved):

lib/systemservice/systemd_test.go
```diff
@@ -120,7 +125,91 @@ TasksMax=infinity
...
+WantedBy=test.target
+
+`,
+		},
+	}
+
+	for _, test := range tt {
+		buf := &bytes.Buffer{}
+		err := serviceUnitTemplate.Execute(buf, test.in)
+		c.Assert(err, IsNil)
```
bernardjkim, line 210: nit: Include comment

bernardjkim (approved):

Issue #2317, opened 2020-11-12
Tracking: etcd upgrade watch failure in planet-agent


knisbet:
### Description

**What happened**:
During 5.5 upgrades that include a new etcd release, after the etcd upgrade worker nodes may no longer correctly track which node is the current planet leader (leader.telekube.local). 

This issue has been isolated and occurs because the watches setup by the planet-agent process to monitor election changes break during the etcd upgrade, and no longer receive change events to the election key.

PR #2330, opened 2020-11-16
[5.5] restore restart phase on workers that reset hung etcd watches

knisbet:
## Description

Rolls back portions of #2086 which are still required on the 5.5.x branch (but not 7.0.x). Basically, call the restart etcd on planet, which will also restart etcd clients that may have broken watches.

Also does a small tweak, that when calling the planet cli, we pass debug so we can capture more thorough upgrade logs.

## Type of change

Partial rollback #2086 
Updates #2317

## TODOs

- [x] Self-review the change
- [x] Write tests
- [ ] Perform manual testing
- [ ] Address review feedback

## Testing done

In Progress - manually run an upgrade, and failover the leader election. Observe whether clients are receiving leadership change events.


lib/update/cluster/steps.go
```diff
@@ -668,6 +669,13 @@ func (r updateStep) etcdPhase(leadMaster storage.Server, otherMasters []storage.
 		restartMasters.AddWithDependency(builder.DependencyForServer(upgradeServers, server), p)
 	}
 
+	// The etcd restart phase resets any etcd clients that may hang watches to the etcd cluster. So ensure the restart
+	// phase is called on each worker to ensure all watches are properly reset.
+	for _, server := range workers {
+		p := r.etcdRestartPhase(server)
+		restartMasters.AddParallel(p)
```
r0mant, line 676: Nit: I'd just do `restartMasters.AddParallel(r.etcdRestartPhase(server))` to avoid an extra variable.

r0mant (approved):

PR #2350, opened 2020-11-27
[master] disable elections and drain node during shrink

### LINKED ISSUE #2349: Control plane doesn't failover on controller removal
### Description

**What happened**:
During gravity uninstall on the node that is currently elected planet leader, the kubernetes control plane does not stop when the etcd member is removed from the etcd cluster.

On the controllers, the etcd clients are only configured to connect to the local node. So when etcd goes down, the kubernetes api isn't able to serve records, and planet-agent is no longer able to participate in elections. Planet-agent appears to leave the control plane running in this situation, despite not successfully participating in the election, and the apiserver also doesn't appear to pass this on to any clients that are watching the api.

This extends to a cluster-wide issue, as kube-proxy has it's watches open to the apiserver, and as such does not update any services, including kubernetes.default to point to the new apiserver that won the election on the remainder of the cluster. As soon as the apiserver does shutdown, kube-proxy is forced to reconnect to the newly elected planet leader, and all clients of kubernetes.default service resume normal operation.

The reason we found this particular case, is the above appeared to occur, but normally after removing the etcd member from the cluster, planet is uninstalled which will shutdown the control plane. In our abnormal case though, the planet uninstall failed for unknown reasons.

1. Uninstall to disable control plane (planet) elections on the node ahead of removing the etcd member
2. If the planet elections are not working or otherwise failed, the control plane should be stopped, as other cluster members may be able to form a cluster / control plane.
3. Alternatively, the etcd clients on controllers should be reconfigured to connect to all etcd members the same way a worker does.  This would allow the etcd clients on the controller nodes to be more resilient in the rare case where only etcd is shutdown on a control plane node for various reasons.

1. Build a multi-node control plane, I used 3 controllers and 2 workers.
2. On the currently elected planet leader (`getent hosts leader.telekube.local` within planet), remove the corresponding member from the etcd cluster. Observer the cluster go unhealthy, kubectl commands fail, etc.

To Resolve:
1. Stop kube-apiserver on the member where etcd was uninstalled, observer the cluster return to normal status.

knisbet:
## Description

1. Disables the control plane on controllers when uninstalling
2. Drain the node when doing a non-force shrink operation

## Type of change

Updates https://github.com/gravitational/gravity/issues/2349
Reqiures https://github.com/gravitational/planet/pull/800

## TODOs

- [x] Self-review the change
- [ ] Write tests
- [x] Perform manual testing
- [ ] Address review feedback

## Implementation

Fixes #2349 by ensuring we disable the node from planet elections ahead of removal from etcd. When used in conjunction with the planet PR (https://github.com/gravitational/planet/pull/800), this shuts down / fails over the control plane to another controller, which should prevent sustained control plane interruptions. The control plane will still see a several second glitch (up to 10 seconds + apiserver initialization time), since the planet election logic doesn't support removal of the election key when a node is demoted and forcing an immediate election. 

Also, when not doing a force remove, I also hooked in the drain logic, to ensure any pods are properly removed from the node. For now I'm not draining during `--force`, since the pods will be uncleanly terminated later in the uninstall process anyways, and I'm not sure we should block a `--force` removal if a pod is not being evicted cleanly. 

## Testing done

Ran through a number of `gravity removes` on the currently elected leader, and monitored for interruptions to the apiserver, and other failures.


lenko-d (approved):

lib/ops/opsservice/shrink.go
```diff
@@ -289,15 +306,28 @@ func (s *site) shrinkOperationStart(opCtx *operationContext) (err error) {
...
-	s.reportProgress(opCtx, ops.ProgressEntry{
-		State:      ops.ProgressStateInProgress,
-		Completion: 30,
-		Message:    "removing the node from the cluster",
-	})
-
 	// if the node is online, it needs to leave the serf cluster to
 	// prevent joining back
 	if online {
+		if !force {
```
r0mant, line 312: Since we know that drain in a production setting can take a very long time, I wonder if adding this step to shrink would just lead to it failing more often for users. Maybe consider keeping the existing behavior with not doing drain by default, and adding a `--drain` flag to remove/leave commands for users who explicitly want it?

knisbet: So my thought on this was, I think when doing a remove command on a node that is online / running, the expectation would be that the uninstall would do a graceful drain of the node if it can. And maybe it should block / fail if the cluster isn't able to accommodate the request.

In which case I can see it being useful, to have a sort of reverse of the flag, uninstall the node without running the drain, which is what I sort of left `--force` as. It will force remove the node regardless of constraints / failures, so that's where I ended up.

So yea, it's definitely difficult, because it does change the behaviour, which causes problems with expectations. But my suspicion is that if we were writing this from scratch, we'd want to include the drain in the uninstall process, so I sort of ended up on changing current behavior, as annoying as that is.

knisbet: @r0mant is this still a concern?

r0mant: Okay, that's reasonable I think. IIRC we already do drain during rolling upgrade so this wouldn't really be much different.

r0mant (approved):

lib/ops/opsservice/shrink.go
helgi: Not to add on the PR as such, what about if I want to do `--force` but want it to attempt a graceful drain first, but am fine with it not being able to. Something like `--force --drain`?

knisbet: My knee jerk thought is why do you want to start with a force remove? Wouldn't that be just to trigger a normal uninstall... and then if that fails to repeat with the force flag?

helgi: That's fair - Depends on the retry logic, i.e. do you retry the current command or do an "if fail do Y" - no problem with that as such. 

For what I described, in the scope of AWS terminations with lifecycle hooks then the gravity side would have to implement it that way since it would be nice to drain gracefully if at all possible but also handling the overall state would be good (so that etcd/serf/etc aren't in a funky state).

knisbet: IIRC, the uninstall hook today just call's a non-force uninstall. But you're right, we're probably not careful that we retry everything for temporary errors, and the force uninstall doesn't really either, it's just if it encounters a temporary error it continues on. The process itself also isn't re-entrant if gravity-site is interrupted, so a long drain could also lead to a weird failure mode.

Probably part of the question should also be what should the behaviour be if the drain fails (IE a pod disruption budget can't be met by the cluster). IIRC the life cycle hook can only keep the node up for some amount of time before it's terminated regardless of the keepalives.

helgi: Agreed, lots to think about. My inclination is that a PDB should be respected but at some point the node gots to go, even if AWS terminates it, even if it means waiting out the max time. Kubernetes should be doing its best to accommodate the change, especially if you have something like `cluster-autoscaler` installed as well.

Perhaps when this lands in beta/stable https://github.com/kubernetes/enhancements/issues/2000 this will be less of an issue since the kubelet gets more involved and is handling draining, and such.

On the `gravity-site` issue is that due to loss of connectivity or what role does the interruption play here?

knisbet: the uninstall operation isn't checkpointed and the state isn't saved, so if the process is killed off and a new one starts, the new one doesn't look for open / running operations to continue to work on.

PR #2352, opened 2020-12-03
Add note about 6.1.x and SELinux support.

wadells:
## Description
This adds documentation about the current state of RHEL SELinux support in 6.1.x. It formalizes (at least for now) our intention to leave Gravity 6.1.x +RHEL 8 + SELinux as a unsupported use case because the fix isn't high ROI compared to other work.

## Type of change
* Documentation

* Refs #2009

## TODOs
- [x] Write documentation :wink: 
- [x] Address review feedback

## Testing done
None.

### DIFF
--- a/docs/6.x/requirements.md
+++ b/docs/6.x/requirements.md
@@ -5,19 +5,24 @@ Kubernetes and Gravity.
 
 ## Linux Distributions
 
-Gravity supports the following Linux distributions:
+Gravity 6.1 supports the following Linux distributions:
 
-| Linux Distribution        | Version         | Docker Storage Drivers                 |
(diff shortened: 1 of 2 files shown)


knisbet (approved):
Although it's weird / not really correct, I would also leave this note on the 7.x docs, just because I'm sure most customers only refer to the latest docs even for 6.x / 5.5.

wadells:
> Although it's weird / not really correct, I would also leave this note on the 7.x docs, just because I'm sure most customers only refer to the latest docs even for 6.x / 5.5.

Added to `docs/7.x` in 865d0e7.

bernardjkim (approved):

PR #2355, opened 2020-12-04
bump planet to 7.0.55-11709

knisbet:


a-palchikov (approved):

wadells (approved):
Code looks straightforward.  Can you link to the changes this introduced in the description?

Also:  the commit message is incorrect, it says 7.1.20-11709 instead of 7.0.55-11709. I presume this will be fixed in the squash.

lenko-d (approved):

knisbet:
> Code looks straightforward. Can you link to the changes this introduced in the description?

This is one of those tough things about gravity, the changes are really just changing the version of a dependency (planet). If this were a third party component, like etcd, we wouldn't be linking to the individual changes.
 
> Also: the commit message is incorrect, it says 7.1.20-11709 instead of 7.0.55-11709. I presume this will be fixed in the squash.

Correct,

Issue #2293, opened 2020-10-29
Add ImagePolicyWebhook admission controller


aelkugia:
Add ImagePolicyWebhook admission controller by request of a customer.

https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#imagepolicywebhook

PR #2362, opened 2020-12-10
Optional admission plugins

bernardjkim:
## Description

Additional Kubernetes admission plugins can be enabled by creating a custom Planet image and specifying the `KUBE_ADMISSION_PLUGINS` variable in the `/etc/optional-environment` file. 

## Type of change

* Refs https://github.com/gravitational/gravity/issues/2293

* Requires https://github.com/gravitational/planet/pull/809

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Write documentation
- [x] Address review feedback

## Testing done

- [x] Verify `--enable-admission-plugins` startup flag can be configured by specifying `KUBE_ADMISSION_PLUGINS` in `/etc/optional-environment` in the Planet container.


docs/7.x/installation.md
```diff
@@ -127,6 +127,7 @@ Flag               | Description
 `--vxlan-port`       | _(Optional)_ Specify custom overlay network port. Default is `8472`.
 `--remote` | _(Optional)_ Excludes this node from the Cluster, i.e. allows to bootstrap the Cluster from a developer's laptop, for example. In this case the Kubernetes master will be chosen randomly.
 `--selinux` | _(Optional)_ Turns on SELinux support. Defaults to `false`. See [SELinux](selinux.md) for more details.
+`--admission-plugins` | _(Optional)_ List of Kubernetes admission plugins to enable.
```
a-palchikov, line 130: You should mention how this option affects the default list - whether it overwrites or appends to?

a-palchikov:
I have not been following the discussion about details on adding this, but I wonder why this is being added as a static cluster metadata instead of being part of `ClusterConfiguration` resource?

bernardjkim:
@a-palchikov
I guess it does make more sense to make this config a part of `ClusterConfiguration` resource. I'll go ahead and update.

wadells (commented):
I'll review after the next round of changes.  Code & testing looks good for this round, no concerns from me.

lib/validate/clusterconfig.go
```diff
@@ -94,3 +99,28 @@ func ClusterConfiguration(existing, update clusterconfig.Interface) error {
 func isCloudConfigEmpty(global clusterconfig.Global) bool {
 	return global.CloudProvider == "" && global.CloudConfig == ""
 }
+
+// validateAdmissionPlugins verifies that the provided Kubernetes admission
+// plugins are valid.
+func validateAdmissionPlugins(plugins []string) error {
```
a-palchikov, line 105: nit: is modifying source code the only way to enable additional plugins? I'm not sure this is in scope, but kubernetes [has](https://github.com/kubernetes/kubernetes/blob/v1.17.9/pkg/kubeapiserver/options/plugins.go#L61) a list of all admission plugins which could be extracted/used for validation.

bernardjkim:
@a-palchikov, @wadells
Sorry for jumping around with different implementations. After chatting with @knisbet, we're considering an alternate approach to enabling admission controllers. 

We could modify the `kube-apiserver.service` to read environment variables from an optional environment file , let's say `optional.env`. `KUBE_ADMISSION_PLUGINS` will be read from this file and appended to the default list of admission plugins. When users create a custom Planet image, they can create this env file and specify additional plugins to enable.

The benefit of this approach is that once the custom Plant image has been created with the desired admission controllers, the cluster can now be installed on multiple environments without having to define ClusterConfiguration each time.

This mechanism can be used for additional configuration when creating custom Planet images.

How does this approach sound compared to updating the `ClusterConfiguration` resource?

a-palchikov:
@bernardjkim it sounds like a workaround really but the benefit is that it's dead simple and if that solves the problem at hand - I have no objections.

knisbet:
> @bernardjkim it sounds like a workaround really but the benefit is that it's dead simple and if that solves the problem at hand - I have no objections.

Yea, sorry Dima. This is of limited appeal to make sure we unblock a specific customer, so starting with the dead simple approach that does place some of the maintenance burden on them with custom planet containers, but also doesn't require replacing / editing files we ship as defaults.

wadells (approved):

a-palchikov (approved):
Don't forget to bump planet to release when merging

PR #1231, opened 2020-03-11
Docs build improvements

wadells:
### Summary

This PR introduces a new make target `docs-lint` which runs `milv -ignore-external` on all of our docs. As well, the regular `make docs` and `make run-docs` now use the `--strict` flag to catch issues.  Both of these changes are building towards enforcement of these targets on all pull requests.

Why?  We had tens to hundreds of broken **_internal_** links in our customer facing documentations that @benarent and I cleaned up in #1210. Providing tooling to catch link rot and enforcing that will prevent issues like this from creeping back in.  This also establishes an engineering norm that docs quality is enforced by automatic tools.

In addition to the necessary changes, I also refactored our docs Makefile substantially, to better use partial builds and caching.  See individual commit messages for more information about those changes.

Contributes to fixing #1220.

### Testing Done

`make docs` and `make docs-lint` on this branch before rebase:
[docs-build-dirty.txt](https://github.com/gravitational/gravity/files/4315484/docs-build-dirty.txt)

`make docs` and `make docs-lint` on this branch after rebase (updated):
[docs-build-clean.txt](https://github.com/gravitational/gravity/files/4320063/docs-build-clean.txt)

I also have a PR job basically ready to run these new targets against every PR to gravity.  See https://jenkins.gravitational.io/job/gravity-pr-docs/ and the test PRs here:

https://github.com/wadells/gravity/pull/1
https://github.com/wadells/gravity/pull/2
https://github.com/wadells/gravity/pull/3

We just need to flip it over from my personal repo to the main repo and configure webhook pushes to make it go live.

### Removal of sass

@alex-kovoy @benarent: I've removed sass from our docker file in 7528d552. This should be safe @alex-kovoy removed the last scss/sass files in 2e1724ed7e3c7742c7c32a114f028108f77bda7d:

```
walt@work:~/git/gravity/docs$ git log --all --oneline -- "*.sass"                                                                                                                             
walt@work:~/git/gravity/docs$ git log --all --oneline -- "*.scss"                                                                                                                             
2e1724ed (docs) cleanup                                                                                                                                                                       
b580aeac Wolfe/docs (#79)                                                                                                                                                                     
ff8241d9 (tag: 5.2.0-alpha.2) Move all OSS code to this repository.                                                                                                                           
45b74e98 (tag: 5.2.0-alpha.0) Add OSS code.                                                                                                                                                   
walt@work:~/git/gravity/docs$ git log --stat -1 2e1724ed                                                                                                                                      
                                                                                                                                                                                              
[...]

PR #2368, opened 2020-12-18
Add Drone CI docs build/lint on PRs.

wadells:
## Description
This dips our toes into the drone water, replacing the smallest gravity Jenkins
job -- after this PR we'll use Drone validate docs on each PR.

## Type of change

* Builds off of https://github.com/gravitational/gravity/pull/1231, reimplements the jenkins portions 

## TODOs
- [x] Self-review the change
- [x] Perform manual testing
- [x] Update documentation
- [x] Address review feedback
- [x] Turn off the jenkins job

## Testing done
The PR build is sufficient.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,59 @@
+---
+kind: pipeline
+type: kubernetes
+name: docs-pr
+
+trigger:
+  event:
+  - pull_request
+
+steps:
+  - name: wait for docker
+    image: docker
+    commands:
+      - timeout 15s /bin/sh -c 'while [ ! -S /var/run/docker.sock ]; do sleep 1; done'
+      - docker version
+    volumes:
+      - name: dockersock
+        path: /var/run
+  - name: build buildbox
+    image: docker:git
+    commands:
+      - apk add --no-cache make
+      - make -C docs bbox
+    volumes:
+      - name: dockersock
+        path: /var/run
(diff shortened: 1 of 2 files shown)


benarent (approved):
Looks good. We are building a newer stack for docs so I wouldn't invest too much time in changes for now, but incremental improvements are always welcome.

a-palchikov (approved):

.drone.yml
```diff
@@ -0,0 +1,59 @@
...
+    commands:
+      - apk add --no-cache make
+      - make -C docs docs
+    volumes:
+      - name: dockersock
+        path: /var/run
+  - name: lint
+    image: docker:git
+    commands:
+      - apk add --no-cache make
```
knisbet, line 38: Does drone cache patterns like this? Adding a package as a command in a particular step?

knisbet (approved):

.drone.yml
wadells: There isn't a cache internal to drone.  The right thing to do here would be to build, upload, and use a `docker:git-make` tag to our quay.  This would save us a call to the alpine cdn and ~1s of time.  In my opinion this isn't worth the extra investment.  Though as this pattern gets repeated across more repos, it may be worth revisiting the idea. Thoughts?

PR #2371, opened 2020-12-24
Disable OpsCenter and WebInstallerWizard via a property in app.yaml

### LINKED ISSUE #1766: Provide ability to disable wizard install and web UI
The web based installation and management interface have significant surface area and add significant complexity to the implementation of security controls. So at least for the short term, we can drop the UI based install and management from the product for FISMA builds.

lenko-d:
## Description

This PR implements the ability to disable OpsCenter and WebInstallerWizard via a property in the app.yaml.
A single property (extensions  -> opsCenter) disables them both.

## Type of change

* Closes #1766 

## TODOs

- [X] Self-review the change
- [X] Perform manual testing
- [X] Address review feedback

## Testing done

Tested via the command lline and WebUI. Verified that when a ./install is attempted the command line shows a message that the UI is disabled. Verified that the logs show a message that the web handler is disabled for web requests.

Verified that the web UI doesn't work when opsCenter property is disabled in the app.yaml.


lib/webapi/webhandler.go
```diff
@@ -444,45 +444,60 @@ func (h *WebHandler) notFound() http.HandlerFunc {
 	}
 }
 
-func (h *WebHandler) noLogin(handle webHandle) httprouter.Handle {
+func noOperationHandler() func(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
 	return func(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
-		handle(w, r, p, session{Session: base64.StdEncoding.EncodeToString([]byte("{}"))})
+		log.WithFields(log.Fields{"urlPath": r.URL.Path, "method": r.Method,
```
a-palchikov, line 449: Isn't this supposed to return some indication of a no service to the client?

lib/webapi/webhandler.go
```diff
@@ -634,6 +649,21 @@ func getWebConfigAuthSettings(cfg WebHandlerConfig) telewebui.WebConfigAuthSetti
 	return authSettings
 }
 
+func (h *WebHandler) disableUI() bool {
```
a-palchikov, line 652: This reads as an action but is, in fact, a query - I'd rename this to `disabledUI` or `isDisabled`

lib/install/engine/interactive/wizard.go
```diff
@@ -77,6 +78,14 @@ func (r *Engine) execute(ctx context.Context, installer install.Interface, confi
 	if err != nil {
 		return trace.Wrap(err)
 	}
+
+	if e.config.App.Manifest.OpsCenterDisabled() {
```
a-palchikov, line 82: This is too late to check it here. Ideally, this validation should run in the client process before it gets to the service and I'd move it [here](https://github.com/gravitational/gravity/blob/349f0824c66d74450fcabb451e332a40099b320d/tool/gravity/cli/config.go#L377) where the manifest is also already available.

lib/webapi/webhandler.go
```diff
@@ -444,45 +444,60 @@ func (h *WebHandler) notFound() http.HandlerFunc {
...
-		handle(w, r, p, session{Session: base64.StdEncoding.EncodeToString([]byte("{}"))})
+		log.WithFields(log.Fields{"urlPath": r.URL.Path, "method": r.Method,
+			"remoteAddr": r.RemoteAddr}).Warn("Not handling this request because the UI is disabled.")
+	}
+}
+
+func (h *WebHandler) noLogin(handle webHandle) httprouter.Handle {
+	if h.disableUI() {
+		return noOperationHandler()
+	} else {
```
a-palchikov, line 457: Remove `else`

a-palchikov (commented):
I'm not super certain leaving the web handler with UI disabled is a good idea - I'd look into a possibility of removing it entirely unless it is serving a route that I have not accounted for.

lib/webapi/webhandler.go
```diff
@@ -444,13 +444,30 @@ func (h *WebHandler) notFound() http.HandlerFunc {
 	}
 }
 
+func noOperationHandler() func(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+	return func(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+		log.WithFields(log.Fields{"urlPath": r.URL.Path, "method": r.Method,
+			"remoteAddr": r.RemoteAddr}).Warn("Not handling this request because the UI is disabled.")
+
+		replyError(w, "Web UI is disabled", http.StatusBadRequest)
```
knisbet, line 452: Nit: I'm not sure if BadRequest is the best error as I believe it's meant to imply a client error. what about Forbidden?

lib/webapi/webhandler.go
```diff
@@ -634,6 +651,21 @@ func getWebConfigAuthSettings(cfg WebHandlerConfig) telewebui.WebConfigAuthSetti
 	return authSettings
 }
 
+func (h *WebHandler) disabledUI() bool {
+	sites, err := h.cfg.Operator.GetSites(defaults.SystemAccountID)
```
knisbet, line 655: If I remember correctly, each call to GetSites does require an etcd read, so I think there might be a risk that unauthenticated access can cause high query load on etcd. So similar to Dima, I would consider an approach where this is configured when setting up the handler as opposed to when the various handlers are used.

knisbet (commented):
@lenko-d similar to Dima, I think I would consider an approach that checks this on setup as opposed to on use.

knisbet (approved):

lib/process/process.go
```diff
@@ -1545,10 +1545,19 @@ func (p *Process) initService(ctx context.Context) (err error) {
...
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		disabledUI = site.DisabledWebUI
+	}
 	p.handlers.Web = web.NewHandler(web.WebHandlerConfig{
 		AssetsDir:      assetsDir,
 		Mode:           p.mode,
 		Wizard:         p.mode == constants.ComponentInstaller,
+		DisabledUI:     disabledUI,
```
a-palchikov, line 1560: Can we remove the web handler in this case entirely?

lib/webapi/webhandler.go
```diff
@@ -444,6 +444,15 @@ func (h *WebHandler) notFound() http.HandlerFunc {
 	}
 }
 
+func noOperationHandler(message string) func(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
```
a-palchikov, line 447: This looks unused - remove?

a-palchikov (approved):

PR #2387, opened 2021-01-12
[7.0.x] Fix coredns-worker podAntiAffinity to correctly select on kube-dns-worker

burdzwastaken:
## Description
Currently, the podAntiAffinity on the coredns-worker Deployment matches on the label that is only on the coredns masters. The coredns masters & workers are already separated using node selectors. 

This PR fixes it to correctly select on `k8s-app: kube-dns-worker` that is present on the workers' Deployment to make sure they don't all end up on the same node.

## Type of change

* Closes #, refs #

* Requires #

* Ports #2386

## TODOs

- [ ] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback


knisbet (approved):

knisbet:
Note: you also need to bump the DNS_APP_TAG in the makefile or the changes won't be applied during the upgrade: https://github.com/gravitational/gravity/blob/version/7.0.x/Makefile#L57

PR #2396, opened 2021-01-20
Provide a minimal, hardened base image

### LINKED ISSUE #1774: Provide a minimal, hardened base image
We should have a minimal base Gravity image with all optional components disabled by default (like OpenEBS, nginx ingress, etc.)

From original ticket:

The default base image, called “gravity”, is a development-friendly image that is easy to get started with. It has the following properties:

* Privileged containers are enabled.
* Development features are enabled: OpenEBS integration, ingress controller, service mesh.

Another base image is hardened, “gravity-strict”, that is more secure by default:

* Privileged containers are disabled by default. Can be enabled via manifest.
* Default integrations (OpenEBS, ingress, mesh) are disabled. Can be enabled via manifest.
* Wireguard?

lenko-d:
## Description

This PR provides the following benefits:
* Hardened base image
Implements the ability to remove artifacts that are not needed from the application image. This improves security because if the artifact are present they could be exploited even if not used.
* Reduced size of the application image
Example:
  telekube.tar size:
    -  before this PR: 2.2 Gb
    - after the changes and with several apps disabled: 1.1 Gb
 

## Type of change

* Closes #1774

## TODOs

- [X] Self-review the change
- [X] Perform manual testing
- [X] Write documentation
- [ ] Address review feedback

## Implementation

Filtering of application dependencies in the tele application.

## Testing done

Disabled several applications in the app.yaml and verified that after the deployment there are no issues visible on the UI.
Verified the same for the upgrade path.

### DIFF
--- a/docs/7.x/cluster.md
+++ b/docs/7.x/cluster.md
@@ -410,6 +410,15 @@ drwxr-xr-x. 5 user user 4.0K Jan 14 09:08 packages
 You can execute the `./upgrade` script to upload and upgrade in one go,
 or you can upload the update and then execute `gravity upgrade` command.
 
+!!! note "Disabling existing functionality when upgrading"
+Currently, `gravity upgrade` will not uninstall existing applications that have been disabled in the upgrade manifest.
+Example:
+We have a cluster that has the monitoring app enabled.
+We perform an upgrade, and the upgrade image has the monitoring app as disabled in the manifest.
+In this case the monitoring app will not be stopped, and it will keep on working as before.
+The existing artifacts of the monitoring app will not be cleaned up. The docker images 
+will not be removed from the internal repository.
+
 Using `gravity upgrade` gives more Control over the upgrade procedure.
 
 ```bash
--- a/docs/7.x/cluster.md
+++ b/docs/7.x/cluster.md
@@ -411,7 +411,7 @@ You can execute the `./upgrade` script to upload and upgrade in one go,
 or you can upload the update and then execute `gravity upgrade` command.
 
 !!! note "Disabling existing functionality when upgrading"
-Currently, `gravity upgrade` will not uninstall existing applications that have been disabled in the upgrade manifest.
+`gravity upgrade` will not uninstall existing applications that have been disabled in the upgrade manifest.
 Example:
 We have a cluster that has the monitoring app enabled.
 We perform an upgrade, and the upgrade image has the monitoring app as disabled in the manifest.
--- a/lib/app/app.go
+++ b/lib/app/app.go
@@ -374,7 +374,9 @@ type ImportRequest struct {
 	// a new tag to update for an image matching the specified registry/repository
 	SetImages []loc.DockerImage `json:"set_images"`
 	// SetDeps defines a list of package dependencies that will be set to the specified version
-	SetDeps []loc.Locator `json:"set_deps"`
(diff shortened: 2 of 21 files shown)


docs/7.x/cluster.md
```diff
@@ -410,6 +410,15 @@ drwxr-xr-x. 5 user user 4.0K Jan 14 09:08 packages
 You can execute the `./upgrade` script to upload and upgrade in one go,
 or you can upload the update and then execute `gravity upgrade` command.
 
+!!! note "Disabling existing functionality when upgrading"
+Currently, `gravity upgrade` will not uninstall existing applications that have been disabled in the upgrade manifest.
```
a-palchikov, line 414: I'd remove `currently` as it does not provide a useful time frame - when this is possible, I would just remove the warning.

a-palchikov (commented):
I think this should research an approach with rewriting the resulting (from the build step) manifest's dependencies to remove the dependencies that were specified as disabled instead of dealing with them at each step of the way after the build. I'm not 100% certain this will work but I believe it's worth a try and will remove the need for a lot of potentially fragile changes.

lib/app/app.go
```diff
@@ -375,6 +375,8 @@ type ImportRequest struct {
 	SetImages []loc.DockerImage `json:"set_images"`
 	// SetDeps defines a list of package dependencies that will be set to the specified version
 	SetDeps []loc.Locator `json:"set_deps"`
+	// ExcludeApps defines a list of app dependencies that will be excluded
+	ExcludeApps []loc.Locator `json:"exclude_apps"`
```
r0mant, line 379: Nit: Add `omitempty`?

lib/app/app.go
```diff
@@ -393,6 +395,9 @@ type Application struct {
 	PackageEnvelope pack.PackageEnvelope `json:"envelope"`
 	// Manifest defines the application install configuration
 	Manifest schema.Manifest `json:"manifest"`
+	// ExcludeApps defines a list of app dependencies
+	// that will be excluded from the build pipeline
+	ExcludeApps []loc.Locator `json:"exclude_apps"`
```
r0mant, line 400: Nit: Add `omitempty`?

lib/app/dependency.go
```diff
@@ -130,6 +144,32 @@ func getDependencies(app *Application, apps Applications, state *state) error {
...
+			excludeApps = append(excludeApps, app)
+		}
+	}
+
+	return excludeApps
+}
+
+// AppsToExcludeFromManifest returns a list of dependencies that should be excluded
+func AppsToExcludeFromManifest(manifestPath string) ([]loc.Locator, error) {
+	if manifestPath != "" {
```
r0mant, line 161: This is a bit weird condition, can anything actually call this function with empty manifest path, and why?

r0mant (commented):
I would agree with Dima, just threading the exclude field through all the code doesn't seem like a very scalable and robust approach to be honest... I would try to find a better approach, maybe indeed look into the possibility of doing this on the tele build step, the potential issue though might be the fact that dependencies are actually defined in the base app manifest which gets merged with the user manifest.

lenko-d:
Closing because replaced by PR 2397

lenko-d:
Closing because replaced by PR 2397

PR #2397, opened 2021-01-27
Ability to provide a minimal, hardened base image

### LINKED ISSUE #1774: Provide a minimal, hardened base image

lenko-d:
## Description

This PR provides the following benefits:
* Hardened base image
Implements the ability to remove artifacts that are not needed from the application image. This improves security because if the artifacts are present they could be exploited even if not used.
* Reduced size of the application image
Example:
  telekube.tar size:
    -  before this PR: 2.2 Gb
    - after the changes and with several apps disabled: 1.1 Gb
 

## Type of change

* Closes #1774

## TODOs

- [X] Self-review the change
- [X] Perform manual testing
- [X] Write documentation
- [X] Address review feedback

## Implementation

Filtering of application dependencies in the tele application.

## Testing done

Disabled several applications in the app.yaml and verified that after the deployment there are no issues visible on the UI.
Verified the same for the upgrade path.


a-palchikov (approved):

docs/7.x/cluster.md
```diff
@@ -410,6 +410,15 @@ drwxr-xr-x. 5 user user 4.0K Jan 14 09:08 packages
 You can execute the `./upgrade` script to upload and upgrade in one go,
 or you can upload the update and then execute `gravity upgrade` command.
 
+!!! note "Disabling existing functionality when upgrading"
+`gravity upgrade` will not uninstall existing applications that have been disabled in the upgrade manifest.
+Example:
+    We have a cluster that has the monitoring app enabled.
```
knisbet, line 416: Are you sure the indented docs will render correctly? Atleast in github markdown it doesn't seem to render, I'm not sure about our docs engine.

knisbet (approved):

docs/7.x/cluster.md
lenko-d: the note in the markdown renders the same as the other notes in the doc. I have followed the example of the other notes.

a-palchikov (approved):

knisbet (approved):

PR #2415, opened 2021-02-16
Add Drone CI build for Pull Requests

wadells:
## Description
This PR adds a Drone CI pipeline that will replace https://jenkins.gravitational.io/job/Gravity-PR-5.2/

This includes:
 - Building the tooling
 - Building a variety of cluster images

Furthermore, there is one extra quality of life feature: The full build & robotest will no longer run when a change consists entirely of  `.md` or items in `docs/` edits.

## Type of change

* Contributes to https://github.com/gravitational/ops/issues/139

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Write documentation
- [x] Address review feedback

## Implementation
Moving robotest tele builds out of a container decreases the reliability of the jenkins job (removing `~jenkins/.gravity` isolation) but I expect that won't matter shortly when the jenkins job is turned off.

It was needed for Drone because we already run docker-in-docker (dind) inside of kubernetes.  Nesting docker a further layer was a recipe for madness.

## Performance/Scaling
We take a slight performance hit because there is no longer parallelism between the unit test and robotest stages.  This is more than made up for by beefy nodes in the Drone kubernetes cluster and autoscaling to help with many builds at the same time.

## Testing done
Most importantly, see the PR build for this PR.

For short-circuiting docs changes, see: https://github.com/gravitational/gravity/pull/2414.  The push build can be ignored -- I cancelled that one to save a build because push builds:
1) aren't part of the final changeset
2) don't always have a reasonable diff (in the case of a rebase)

All of the `walt/drone-build jobs` here: https://drone.gravitational.io/gravitational/gravity/ are further intermediate testing ... though they're probably not of much interest.

## Secrets

There are several secrets available to this PR job:

* `GITHUB_RO_KEY_PR` is a ssh key set up as a deploy key for https://github.com/gravitational/selinux/settings/keys.  It does not have access to anything else
* `AWS_S3_RO...` is a token associated with an Read Only S3 IAM user provisioned here: https://github.com/gravitational/cloud-terraform/pull/132
* `GCP_ROBOTEST_CREDENTIALS` is a json token associated with a GCP service user. This was hand provisioned and uses the standard Robotest policy.

All of the above are also saved in 1Password.

## Notes
I've not included any of the publishing work in this changeset -- though that is coming soon.  I'm trying to keep the PRs smaller and relatively digestable -- and also to give some soak time to catch issues.

This work loses some functionality that was present in Jenkins.  Particularly, robotest saves log tarballs to the local filesystem.  Since the local FS is now in a kubernetes pod this data is immediately lost.  The right answer here is likely to update robotest to save those logs to an appropriate data dump (s3) and set up a housekeeping process to remove old logs.  However, I'm not aware of anyone who uses these logs other than Dima and I -- so I've posponed on this task as lower priority work.  See https://github.com/gravitational/robotest/issues/281.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,7 +1,107 @@
 ---
 kind: pipeline
 type: kubernetes
-name: docs-pr
+name: pr build
+
+trigger:
+  event:
+  - pull_request
+
+steps:
+  # https://discourse.drone.io/t/how-to-exit-a-pipeline-early-without-failing/3951
+  # If a change is entirely documentation, skip the expensive build & test steps.
+  - name: short circuit docs changes
+    image: docker:git
+    commands:
+    -  ./build.assets/drone/diff-is-all-docs.sh $DRONE_COMMIT_BEFORE..$DRONE_COMMIT_AFTER && exit 78 || exit 0
+  - name: fetch tags
+    image: docker:git
+    commands:
+      - git fetch --tags
+  - name: wait for docker
+    image: docker
+    commands:
+      - timeout 15s /bin/sh -c 'while [ ! -S /var/run/docker.sock ]; do sleep 1; done'
+      - docker version
+    volumes:
+      - name: dockersock
+        path: /var/run
+  - name: build
+    image: docker:git
+    environment:
+      GITHUB_RO_KEY_PR:
+        from_secret: GITHUB_RO_KEY_PR
+      AWS_ACCESS_KEY_ID:
+        from_secret: AWS_S3_RO_ACCESS_KEY_ID
+      AWS_SECRET_ACCESS_KEY:
+        from_secret: AWS_S3_RO_SECRET_ACCESS_KEY
+    commands:
+      - apk add --no-cache make bash libc6-compat aws-cli
+      - mkdir -m 0700 /root/.ssh && echo "$GITHUB_RO_KEY_PR" > /root/.ssh/id_ed25519 && chmod 600 /root/.ssh/id_ed25519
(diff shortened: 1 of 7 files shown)


lenko-d (approved):

.drone.yml
```diff
@@ -1,7 +1,107 @@
...
+  event:
+  - pull_request
+
+steps:
+  # https://discourse.drone.io/t/how-to-exit-a-pipeline-early-without-failing/3951
+  # If a change is entirely documentation, skip the expensive build & test steps.
+  - name: short circuit docs changes
+    image: docker:git
+    commands:
+    -  ./build.assets/drone/diff-is-all-docs.sh $DRONE_COMMIT_BEFORE..$DRONE_COMMIT_AFTER && exit 78 || exit 0
```
bernardjkim, line 16: Is there any significance in using exit code 78?

bernardjkim (approved):

.drone.yml
wadells: Yeah, https://discourse.drone.io/t/how-to-exit-a-pipeline-early-without-failing/3951 has more info.  I (or rather Gus) tried to cover that in the comment above, but perhaps it is too far from the exit to be contextually relevant.

UnderGreen (approved):
LGTM

.drone.yml
bernardjkim: Ah, thanks for the clarification.

PR #2420, opened 2021-02-23
Drone CI: Add post-merge builds

wadells:
## Description
These are a heavier version of our test suite, formerly "nightlies".

Now they run upon push to master (because we're typically seeing less
than one change to master a day so running nightly would waste more
than it saves).

This replaces the Gravity-Nightly Jenkins job.

## Type of change

* Contributes to gravitational/ops#139
* Builds off https://github.com/gravitational/gravity/pull/2415

## TODOs
- [x] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback

## Implementation
This is essentially a duplication of the PR build, with the following changes:

* Triggered upon merge/push to master
* Unit tests are skipped
* ROBOTEST_CONFIG=nightly

I skipped unit testing, as that is always checked at the PR phase -- no need to repeat the effort.  Everything else is needed to produce the artifacts for nightly testing.

Unfortunately, there isn't a great way do dedupe the shared code without going to something like Starlark or https://github.com/gravitational/teleport/pull/5644. See https://github.com/gravitational/teleport/issues/5533 for further discussion.  I don't want to invent this wheel in parallel, so I'm sticking to vanilla drone yaml for the moment.

I didn't take the time to refactor out the "nightly" terminology in favor of something more trigger agnostic.   I may revisit this later, but there isn't much ROI.

## Performance/Scaling
"Nightly"  runs are pretty consistent between 2-3 hours.  I currently have the job timeout set at 4 hours.  Although the runtime is long, the load on Drone is minimal, as this skips the  CPU expensive unit tests.

## Testing done
See https://drone.gravitational.io/gravitational/gravity/119 for a build I triggered by including this branch.

The real test will occur after this merges, and we see the master trigger kick off.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -101,6 +101,101 @@ volumes:
 ---
 kind: pipeline
 type: kubernetes
+name: post-merge build
+
+trigger:
+  event:
+  - push
+  branch:
+  - master
+
+steps:
+  - name: short circuit docs changes
+    image: docker:git
+    commands:
+      # If a change is entirely documentation, skip the expensive build & test steps.
+      # https://discourse.drone.io/t/how-to-exit-a-pipeline-early-without-failing/3951
+      - ./build.assets/drone/diff-is-all-docs.sh $DRONE_COMMIT_BEFORE..$DRONE_COMMIT_AFTER && exit 78 || exit 0
+  - name: fetch tags
+    image: docker:git
+    commands:
+      - git fetch --tags
+  - name: wait for docker
+    image: docker
+    commands:
+      - timeout 15s /bin/sh -c 'while [ ! -S /var/run/docker.sock ]; do sleep 1; done'
+      - docker version
+    volumes:
+      - name: dockersock
+        path: /var/run
+  - name: build
+    image: docker:git
+    environment:
+      GITHUB_RO_KEY_PR:
+        from_secret: GITHUB_RO_KEY_PR
+      AWS_ACCESS_KEY_ID:
+        from_secret: AWS_S3_RO_ACCESS_KEY_ID
+      AWS_SECRET_ACCESS_KEY:
+        from_secret: AWS_S3_RO_SECRET_ACCESS_KEY
+    commands:
+      - apk add --no-cache make bash libc6-compat aws-cli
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -98,6 +98,101 @@ volumes:
...
+        from_secret: GITHUB_RO_KEY_PR
+      AWS_ACCESS_KEY_ID:
+        from_secret: AWS_S3_RO_ACCESS_KEY_ID
+      AWS_SECRET_ACCESS_KEY:
+        from_secret: AWS_S3_RO_SECRET_ACCESS_KEY
+    commands:
+      - apk add --no-cache make bash libc6-compat aws-cli
+      - mkdir -m 0700 /root/.ssh && echo "$GITHUB_RO_KEY_PR" > /root/.ssh/id_ed25519 && chmod 600 /root/.ssh/id_ed25519
+      - ssh-keyscan -H github.com > /root/.ssh/known_hosts 2>/dev/null && chmod 600 /root/.ssh/known_hosts
+      - make -C e production telekube opscenter
```
webvictim, line 144: Might to be explicit about deleting the SSH key from the storage once the operation it's used for has completed, although I presume this is effectively done already because the container terminates after this step and the key isn't written to the shared workspace?

webvictim (approved):
LGTM

.drone.yml
wadells: This key is conditionally used for a variety of clones of public repos.
For example:

https://github.com/gravitational/gravity/blob/3cb2505cf2a5fc7f2bb23114e8d6c7c4b1193f3d/build.assets/Makefile#L697
https://github.com/gravitational/gravity/blob/3cb2505cf2a5fc7f2bb23114e8d6c7c4b1193f3d/build.assets/Makefile#L168

> I presume this is effectively done already because the container terminates after this step and the key isn't written to the shared workspace?

Yeah, I was implicitly leveraging the transient nature of /root to scope the key to just this container.  Even if we deleted it from the FS, it would still be present in the environment -- though I suppose I could unset it there too.

Unlike some of the other credentials, this key doesn't have access to any private repos -- thus it is quite low sensitivity.  I needed it because I didn't want to rewite a bunch of build scripts to use public https clones instead of git clones.

bernardjkim (approved):

Issue #2200, opened 2020-10-06
Gracefully remove node from cluster on AWS node termination


r0mant:
### Feature Request
Describe in as much detail as possible what feature you would like to see added and how it should work. 

### Motivation
Describe why we should work on this feature request.

### Who's it for?
OSS User, Enterprise

helgi:
How different is this from https://github.com/gravitational/gravity/issues/942 ? 

Right now the SQS handles the removal and passes to `gravity remove`, eventually, it gets a bit janky atm due to how join/remove can't happen in parallel atm.

Curious of the scope here

r0mant:
@helgi The use-case targeted here is when someone terminates the instance directly via AWS control panel so it doesn't go through ASG lifecycle hooks. Not sure yet if there's a way to tell this scenario from a regular node shutdown and launch shrink in this case, needs some research.

Issue #1289, opened 2020-03-26
Ability to tell the upgrade script to only upgrade master


helgi:
The `./upgrade` script that comes in the tarball, I would like to be able to instruct it to only upgrade masters with `--masters-only` (or similar) flag with the intent of rolling `workers` myself with a different mechanism.

I can see this working well with compatible gravity and kubernetes versions but perhaps use the upgrade path support logic that was developed to determine if 5.1 can upgrade to 6.3 (with its in-between versions). In the case of incompatible behaviour fail the `upgrade` or have a flag to do a full upgrade?

r0mant:
This is an interesting idea. I'm curious, what is the specific use-case for using different mechanism for upgrading workers? To me it sounds like if the current algorithm is not good enough, Gravity should do a better job at rolling them then :)

helgi:
The way it works for me is to update `gravity` we bake a new AMI and if the new AMI has a newer gravity we kick off the upgrade (it compares cluster gravity with AMI gravity) as part of `cloud-init`. 

The new AMI also causes a full autoscaling group (ASG) roll which terminates nodes, it will pause when it sees `gravity status` turning to `upgrading` so it isn't being disruptive. Gravity upgrade will drain, upgrade and uncordon nodes; a 20 worker cluster can take ~1 hour to just do that. And then the nodes will be drained and rolled again due to the ASG picking back up.

Due to how we roll AMIs it is more efficient for me to bring up net new workers when possible as I get the same end result :) But I also made the caveat about the version disparity might be an issue (i.e. 5.1 -> 6.1 or some)

The issues with the current gravity approach you could consider to improve:
1. Better information when a plan fails, inform people `gravity upgrade trigger` with flags can help recover/continue
2. The `gravity plan` for upgrade is static, if a node is shutdown (variety of reasons, including AWS just plain retiring a node) can make an update fail as it is expecting to talk to node 1234. It should be fine if a node is no longer available, somehow.
3. `gravity plan` could be more dynamic; spin-off sub-operations per grouping (f.ex. each node upgrade)
4. Ability to upgrade masters and workers independently (expansion on this ticket)

To summerize I think there are opportunity for improve algo around this but also providing more granular control of order of operations for cases like I describe above, a more "cloudy" roll if you will :)

helgi:
another additive, as `cluster-autoscaler` can be the culprit sometimes in scaling nodes down (if it isn't AWS retiring, etc, etc)

[...]

Issue #942, opened 2019-12-07
Drain node before removing it during a shrink operation


helgi:
Looking at the shrink operation (https://github.com/gravitational/gravity/blob/master/lib/ops/opsservice/shrink.go#L164) it does not seem like a `kubectl drain` (or the library variant that upstream is working on) is being called but rather various gravity package/system uninstall and then `kubectl delete` on the node.

It would be good to include `drain` in there to allow for more graceful shrinks as I don't think `delete` handles that for nodes in the same way as it handles graceful deletes for pods.

The issue with drain will be the various flags one has to provide to handle volumes and other things; these could be all configurable somehow/somewhere in gravity. i.e. `--delete-local-data`, `--grace-period`, `--force` and others

This would allow services to properly finish what they are doing and be rescheduled to other nodes with minimal disruptions.

PR #2426, opened 2021-03-08
[master] g1k part2

knisbet:
Note: Sorry to the reviewers, but this is a bunch of changes rolled into a single PR, since it's tuning a bunch of changes need to improve life on large clusters which is way easier to test in one go. This is an extension of the g1k work done last year, that is more centered around upgrades.

Customer M is also volunteering to help check out these changes in some of their environments, so we can release without 100% validation. Also, most of this was written after midnight over the last couple of weeks :(

## Description

1. Introduces a hidden flag that eliminates workers from upgrade plan. This is a feature for Customer M who prefers to replace VMs on a regular basis, so after an upgrade they would replace all VMs with new ones that will join with the latest version. This should be relatively safe, since not a lot changes in node compatibility otherwise nodes would fail during an upgrade, but is a feature that is for use at their own risk as it is not something we will include in our test matrix.
2. Introduce a hidden flag to allow parallel execution of certain upgrade phases. This was mostly an adjustment on already present code that allowed certain install steps to occur in parallel, however, I adjusted it to cap the number of parallel phases to prevent runaway CPU usage. When passed as a flag this also allows a customer to shoot themselves in the foot since too much capacity could be taken offline in a single AZ for example, so keeping it hidden to use at their own risk.
Note: I did introduce logic to always parallelize non-impacting steps, like the bootstrap phases, certain etcd steps, etc.
3. Introduce caching of plan phases. The gravity plan works sort of on an event sourcing concept, but every plan changelog entry previously required an individual query to etcd, where to view plan might take thousands of etcd queries. 
4. Adjust agent deployment timeout, I experienced some timeouts in large cluster testing
5. parallelize the checks code within gravity for remote node checks.
6. Rewrite the event handler for AWS integrations. This is a long standing issue where we didn't hook in properly to the ASG to allow a proper install, so the logic was a bit reversed. Now, upon receiving a shrink event, we begin a heartbeat to AWS to leave the node online until an uninstall is performed, which also drains the node. And when the uninstall has completed, send AWS a notification so that it can then shutdown the node.
7. Fix an issue where tailing an operation plan would only show 100 previous events
8. Add debug logging + response time calculation to the internal API client
9. Fix degraded cluster performance when lots of nodes are waiting to join by running an early check and exit if the expand operation won't be accepted.
10. Add compression to large plans so that we don't exceed the etcd object size
11. Retry failed chown's in a retry loop. This is a hack ontop of a hack at this point, but hopefully will cut down on what appears to be spurious failures inside the chown. This really needs to be redesigned in the packaging logic, but I don't know when we'd be able to get to that. I did attempt a more targeted logic, but broke installs and upgrades that now expect this chown to have occurred, so a hack is what we get for now.
12. Allow multiple shrink operations to be created instead of just one, and allows shrink and expand to coexist as long as there are not changes on masters.

## Type of change

Updates https://github.com/gravitational/gravity/issues/2200
Updates https://github.com/gravitational/gravity/issues/942
Updates https://github.com/gravitational/gravity/issues/1289
Requires https://github.com/gravitational/satellite/pull/297

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [ ] Write documentation
- [ ] Address review feedback

## Performance/Scaling

All of these changes are introduced to make gravity more resilient on larger clusters.

## Testing done

Mainly building / rebuilding / upgrading my 1k test cluster.


lib/update/cluster/plan.go
```diff
@@ -47,13 +47,19 @@ import (
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
+type UserConfig struct {
```
bernardjkim, line 50: Missing godocs.

tool/gravity/cli/clusterconfig.go
```diff
@@ -219,6 +219,7 @@ func (r configInitializer) newOperationPlan(
 	localEnv, updateEnv *localenv.LocalEnvironment,
 	clusterEnv *localenv.ClusterEnvironment,
 	leader *storage.Server,
+	userConfig interface{},
```
bernardjkim, line 222: nit: If parameter is unused, can we use blank identifiers instead?

tool/gravity/cli/environ.go
```diff
@@ -207,6 +207,7 @@ func (environInitializer) newOperationPlan(
 	localEnv, updateEnv *localenv.LocalEnvironment,
 	clusterEnv *localenv.ClusterEnvironment,
 	leader *storage.Server,
+	userConfig interface{},
```
bernardjkim, line 210: nit: Same here, `userConfig` is unused.

tool/gravity/cli/clusterconfig.go
knisbet: I'm matching the function signature to the interface for newOperationPlan

tool/gravity/cli/environ.go
knisbet: I'm matching the function signature to the interface for newOperationPlan

lib/storage/keyval/backend.go
```diff
@@ -115,10 +150,87 @@ func (*v1codec) DecodeFromString(val string, in interface{}) error {
...
+// compress will gzip compress the input data if the slice is above a threshold. Data below the threshold will
+// be returned without modification.
+func compress(in []byte) ([]byte, error) {
+	if len(in) < 2 {
+		return in, nil
+	}
+
+	// if the data is already compressed (with gzip) don't compress a second time
+	// insert our own magic number to indicate to return the value as is
+	if (in[0] == 0x1f && in[1] == 0x8b) || (in[0] == 0x1f && in[1] == 0x8c) {
```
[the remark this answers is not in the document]
knisbet: it's unlikely, usually magic numbers at specific positions are chosen as unlikely to collide with normal text or other file formats, although this is a relatively old practice which is why only 16 bits are used for magic numbers.

lib/storage/keyval/backend.go
```diff
@@ -115,10 +150,87 @@ func (*v1codec) DecodeFromString(val string, in interface{}) error {
...
+
+	var buf bytes.Buffer
+	gz := gzip.NewWriter(&buf)
+
+	_, err := gz.Write(in)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	if err := gz.Close(); err != nil {
```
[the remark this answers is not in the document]
knisbet: it appears to be the case, Close only flushes data buffers, which we're discarding by returning the error and not using the data.

lib/storage/keyval/backend.go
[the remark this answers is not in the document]
knisbet: link added

lib/autoscale/aws/events.go
```diff
@@ -148,7 +168,203 @@ func (a *Autoscaler) ensureInstanceTerminated(ctx context.Context, event HookEve
...
+	a.WithError(err).WithFields(log.Fields{
+		"instance": event.InstanceID,
+		"asg_name": event.AutoScalingGroupName,
+	}).Info("notified AWS of completed uninstall")
+
+	return trace.Wrap(err)
+}
+
+func (a *Autoscaler) startHeartbeatLoop(event HookEvent) context.CancelFunc {
+	ctx, cancel := context.WithTimeout(context.Background(), eventTimeout)
```
wadells, line 321: Should this inherit from the `RemoveInstance` context? It looks like you instead return the cancel fn and use it in a defer in the parent.  It seems like similar functionality -- is there a reason to prefer the latter here?

lib/checks/checks.go
```diff
@@ -371,11 +373,40 @@ func (r *checker) Check(ctx context.Context) (failed []*agentpb.Probe) {
...
+	wg.Add(numRoutines)
+
+	for i := 0; i < numRoutines; i++ {
+		go func() {
+			for server := range serverC {
+				f := r.CheckNode(ctx, server)
+
+				mutex.Lock()
+				failed = append(failed, f...)
+				mutex.Unlock()
```
wadells, line 401: Mixing new & old style concurrency feels weird here.  Elsewhere (e.g. executeSubphasesConcurrently) you use a channel to aggregate errors.  I don't think it needs to be fixed because the logic isn't wrong -- just something that my brain flagged while reading the code.

lib/update/cluster/plan.go
```diff
@@ -47,13 +47,20 @@ import (
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
+// UserConfig holds configuration parameters provided by the user triggering the update operation.
+type UserConfig struct {
+	ParallelWorkers int
```
wadells, line 52: Validation wise, we probably want this to be a positive integer.  Setting 0 ornegative integers are accepted, but silently coerced to serial execution. It doesn't break anything perse, just leaves undefined behavior in the UX.

tool/gravity/cli/register.go
```diff
@@ -205,8 +205,12 @@ func RegisterCommands(app *kingpin.Application) *Application {
 	g.UpdateTriggerCmd.Manual = g.UpdateTriggerCmd.Flag("manual", "Manual operation. Do not trigger automatic update.").Short('m').Bool()
 	g.UpdateTriggerCmd.SkipVersionCheck = g.UpdateTriggerCmd.Flag("skip-version-check", "Bypass version compatibility check.").Hidden().Bool()
 	g.UpdateTriggerCmd.Force = g.UpdateTriggerCmd.Flag("force", "Force update operation even if some nodes have active warnings.").Bool()
+	g.UpdateTriggerCmd.SkipWorkers = g.UpdateTriggerCmd.Flag("skip-workers", "Skip upgrade of worker nodes").Hidden().Bool()
+	g.UpdateTriggerCmd.ParallelWorkers = g.UpdateTriggerCmd.Flag("parallel-workers", "Number of worker nodes to upgrade in parallel (Default: 1)").Int()
```
wadells, line 209: Where is the default set to 1?

wadells (approved):
:ship: 

No major concerns, just some questions.

tool/gravity/cli/register.go
knisbet: The default for the variable isn't really set to 1, as you noted above, the default behaviour is serial execution, which is basically the same as parallel execution where only 1 is allowed at a time.

lib/checks/checks.go
knisbet: Yea, I see the conundrum. I'm really not hugely tied to one approach or the other tbh, and this approach was tested so will probably just leave it for now.

lib/autoscale/aws/events.go
knisbet: Yea, you're right, the code doesn't quite make sense, as the parent context was indirectly tied to the heartbeat context because if the parent context is cancelled, the defer would cancel the heartbeat.

I'm not sure why I didn't originally tie the heartbeat to the parent getting cancelled, other than it's one of the context's that we use to signal service shutdown, which is a pattern used in the gravity code but doesn't actually tend to work as specified. 

We still want a separate context for the heartbeat, so that it times out around the same time as the node, and doesn't keep hitting the AWS API and respective rate limits if the heartbeat has no chance of progressing, but removeNode can still be running after that point.

tool/gravity/cli/register.go
knisbet: ended up tweaking this in 22eb5a3b292090bd33fb5ba4a7a9a371132283f2

lib/update/cluster/plan.go
knisbet: Ended up tweaking this in 22eb5a3b292090bd33fb5ba4a7a9a371132283f2

Although I did the validation in the CLI code, I think it's fine for the library code to treat 0 or 1 as serial execution.

tool/gravity/cli/register.go
knisbet: Apparently I brain farted and kingpin doesn't really reflect an unset value, so I had to tweak this to only detect negative numbers in aaf1809dc1b2b16438f8170ccf4b2c7e2831bf94 and otherwise keep the logic to do serial invocation for a 0 or 1.

PR #2431, opened 2021-03-11
[5.5] fix problem with loading influxdb client

knisbet:
## Description

Fixes an issue observed by customer M that after a controller is rebooted the WebUI configuration page for Metrics Retention may not work. This occurs if during startup gravity-site isn't able to reach influxdb, it would exit with a nil client that would throw exceptions.

- Fixes a typo that allowed the influx client to be nil without returning an error
- Lazy initializes the influx client to avoid hanging the gravity-site startup on influx

## Type of change

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Implementation

This can probably race in a rare scenario, but that will just lead to a dead/empty client so I'm consciously allowing the race under this circumstance to avoid a larger rewrite.

### DIFF
--- a/lib/ops/monitoring/influxdb.go
+++ b/lib/ops/monitoring/influxdb.go
@@ -34,15 +34,13 @@ import (
 
 type influxDB struct {
 	*roundtrip.Client
+
+	kubeClient *kubernetes.Clientset
 }
 
 // NewInfluxDB returns a new InfluxDB monitoring provider
 func NewInfluxDB(kubeClient *kubernetes.Clientset) (Monitoring, error) {
-	client, err := newInfluxDBClient(kubeClient.CoreV1())
-	if err != nil {
-		return nil, trace.Wrap(err)
-	}
-	return &influxDB{Client: client}, nil
+	return &influxDB{kubeClient: kubeClient}, nil
 }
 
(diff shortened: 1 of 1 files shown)


lib/ops/monitoring/influxdb.go
```diff
@@ -94,6 +92,14 @@ func Get(client *roundtrip.Client, endpoint string, params url.Values) (*roundtr
 
 // GetRetentionPolicies returns a list of retention policies for the site
 func (i *influxDB) GetRetentionPolicies() ([]RetentionPolicy, error) {
+	if i.Client == nil {
+		var err error
```
a-palchikov, line 96: nit: move this to `i.initClient`?

a-palchikov (approved):

lib/ops/monitoring/influxdb.go
```diff
@@ -34,15 +34,13 @@ import (
 
 type influxDB struct {
 	*roundtrip.Client
+
+	kubeClient *kubernetes.Clientset
 }
 
 // NewInfluxDB returns a new InfluxDB monitoring provider
 func NewInfluxDB(kubeClient *kubernetes.Clientset) (Monitoring, error) {
```
bernardjkim, line 42: Nit: We can remove `error` from return values.

bernardjkim (approved):

PR #2433, opened 2021-03-13
Upgrade k8s client lib to 1.19.8

bernardjkim:
## Description

Upgrade client-go to `v1.19.8`. Relevant changes can be found in 4d800b5. These mainly consist of passing context to Kubernetes API calls. 

The `GetKubeClient` logic has been slightly modified in 888269d. If kubeconfig is not specified, call `InClusterConfig` to get config. This is to avoid printing out warnings which would result in robotest failures.

Upgrade helm to `v3.4.2`. Relevant changes can be found in 363969f. Helm 3 no longer requires Tiller, and setup of directories and repositories is automated and does not need to be handled on initialization. The helm actions/cmds have been rewritten. They have been ported from https://github.com/helm/helm/tree/v3.4.2/cmd/helm with some minor tweaks.

## Type of change

* Requires https://github.com/gravitational/planet/pull/827, https://github.com/gravitational/rigging/pull/100

## TODOs

- [ ] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback


wadells:
Looked through everything but the helm changes today.  I'll check those out further tomorrow.

Robotest does use a helm chart in its standard image, but it is quite basic.

wadells:
Helm 3 note:

https://helm.sh/docs/faq/#name-or---generate-name-is-now-required-on-install seems like a gotcha.  We should make sure our documentation and examples are all updated for this.

bernardjkim:
> https://helm.sh/docs/faq/#name-or---generate-name-is-now-required-on-install seems like a gotcha. We should make sure our documentation and examples are all updated for this.

Good catch. I've updated the install action to default to using `--generate-name=true` if a release name is not provided. 
c98d72e

bernardjkim:
@wadells @knisbet
Can I move on with this change? It doesn't seem like helm integration is used much and shouldn't have a critical impact on the core functionality of gravity.

wadells:
> @wadells @knisbet
> Can I move on with this change? It doesn't seem like helm integration is used much and shouldn't have a critical impact on the core functionality of gravity.

Sorry for the delay. The helm changeset is pretty large so I kept procrastinating on the review.

Aside: We should work on a better team process for reviewing large diffs -- e.g. some pointers in the PR description (look at section X, section Y doesn't need much attention due to Z) or doing a synchronous walk through. E.g. it seems like large chunks of 363969f was cut & pasted verbatim, and thus probably doesn't deserve the same level of scrutiny as tailor made new sections.  I can see you started to head down this road with your description, but I still was getting a bit lost within a single ~1K SLOC patch  I end up pulling the code locally and then reviewing with the convenience of go-pls, but then it became difficult to tell where your changes end -- as someone new to both this patch and the helm part of the codebase.

wadells (approved):
I checked through all the helm code and didn't see any local errors.  I don't feel confident saying I didn't miss any larger/architectural errors as I'm not particularly familiar with this workflow.

Would love to see before 7.1 GA:

 - a hand pass on testing the helm commands, checking alternate & default namespaces
   - install
   - list
   - upgrade
   - rollback
   - uninstall
 - examples & documentation updated for helm 3 (the actual work here is probably splitting 7.0 and 7.1 documentation)

However, I think we should get this in, and do another alpha for testing -- we can pick these items up later.

bernardjkim:
> Aside: We should work on a better team process for reviewing large diffs -- e.g. some pointers in the PR description (look at section X, section Y doesn't need much attention due to Z) or doing a synchronous walk through

I agree, I try to minimize the scope of my PRs, but sometimes its difficult to do. 

In future PRs with large changes, I'll try to include more comments and walk through important changes.

PR #2458, opened 2021-04-03
Fetch SSM service url and token in a single API call

helgi:
## Description
Logic was to do 2x SSM API calls every 15 seconds until service URL is reachable. Booting up a new cluster or joining a lot of nodes they will all hit the SSM API without any jitter which in turn will throttle the clients (https://aws.amazon.com/premiumsupport/knowledge-center/ssm-parameter-store-rate-exceeded). This halfs the requests.

Adding jitter could help and doing progressively longer backoffs (without being exponential)

## Type of change

* Closes #, refs #

* Requires #

* Ports #

## TODOs

- [ ] Self-review the change
- [ ] Write tests
- [ ] Perform manual testing
- [ ] Write documentation
- [ ] Address review feedback

### DIFF
--- a/lib/autoscale/aws/autoscaler.go
+++ b/lib/autoscale/aws/autoscaler.go
@@ -203,32 +203,33 @@ func (a *Autoscaler) WaitUntilInstanceTerminated(ctx context.Context, instanceID
 	return trace.Wrap(err)
 }
 
-// GetJoinToken fetches and decrypts cluster join token from SSM parameter
-func (a *Autoscaler) GetJoinToken(ctx context.Context) (string, error) {
-	name := a.tokenParam()
-	a.Debugf("GetJoinToken(%v)", name)
-	resp, err := a.SystemsManager.GetParameterWithContext(ctx, &ssm.GetParameterInput{
-		Name:           aws.String(name),
+// GetServiceURLAndJoinToken fetches and decrypts cluster service URL and join token from SSM parameter
+func (a *Autoscaler) GetServiceURLAndJoinToken(ctx context.Context) (string, string, error) {
+	token := a.tokenParam()
+	url := a.serviceURLParam()
+	names := aws.StringSlice([]string{token, url})
+	a.Debugf("GetServiceURLAndJoinToken(%v)", names)
+	resp, err := a.SystemsManager.GetParametersWithContext(ctx, &ssm.GetParametersInput{
+		Names:          names,
 		WithDecryption: aws.Bool(true),
 	})
 	if err != nil {
-		return "", ConvertError(err)
+		return "", "", ConvertError(err)
 	}
-	return aws.StringValue(resp.Parameter.Value), nil
-}
 
-// GetServiceURL returns service URL
-func (a *Autoscaler) GetServiceURL(ctx context.Context) (string, error) {
-	name := a.serviceURLParam()
-	a.Debugf("GetServiceURL(%v)", name)
-	resp, err := a.SystemsManager.GetParameterWithContext(ctx, &ssm.GetParameterInput{
-		Name:           aws.String(name),
-		WithDecryption: aws.Bool(false),
-	})
-	if err != nil {
-		return "", ConvertError(err)
+	// find the values within the response
+	var rtoken, rurl string
+	for _, params := range resp.Parameters {
+		if aws.StringValue(params.Name) == token {
+			rtoken = aws.StringValue(params.Value)
+		}
+
+		if aws.StringValue(params.Name) == url {
+			rurl = aws.StringValue(params.Value)
+		}
 	}
-	return aws.StringValue(resp.Parameter.Value), nil
+
+	return rurl, rtoken, nil
 }
 
 func (a *Autoscaler) publishServiceURL(ctx context.Context, serviceURL string, force bool) error {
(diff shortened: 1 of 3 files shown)


lib/autoscale/aws/autoscaler.go
```diff
@@ -203,32 +203,33 @@ func (a *Autoscaler) WaitUntilInstanceTerminated(ctx context.Context, instanceID
...
 }
 
-// GetJoinToken fetches and decrypts cluster join token from SSM parameter
-func (a *Autoscaler) GetJoinToken(ctx context.Context) (string, error) {
-	name := a.tokenParam()
-	a.Debugf("GetJoinToken(%v)", name)
-	resp, err := a.SystemsManager.GetParameterWithContext(ctx, &ssm.GetParameterInput{
-		Name:           aws.String(name),
+// GetServiceURLAndJoinToken fetches and decrypts cluster service URL and join token from SSM parameter
+func (a *Autoscaler) GetServiceURLAndJoinToken(ctx context.Context) (string, string, error) {
```
UnderGreen, line 207: I would probably add here variable names for output parameters, otherwise, it is not clear what is string one and string two.

lib/autoscale/aws/autoscaler.go
```diff
@@ -203,32 +203,33 @@ func (a *Autoscaler) WaitUntilInstanceTerminated(ctx context.Context, instanceID
...
-	name := a.serviceURLParam()
-	a.Debugf("GetServiceURL(%v)", name)
-	resp, err := a.SystemsManager.GetParameterWithContext(ctx, &ssm.GetParameterInput{
-		Name:           aws.String(name),
-		WithDecryption: aws.Bool(false),
-	})
-	if err != nil {
-		return "", ConvertError(err)
+	// find the values within the response
+	var rtoken, rurl string
```
UnderGreen, line 221: I believe `responseToken` and `responseURL` is more intuitive for the reader and better fits into Go conventions.

lib/autoscale/aws/autoscaler.go
helgi: Good point. Fixed.

lib/autoscale/aws/autoscaler.go
helgi: Agreed, addressed.

wadells (commented):
Branch naming: Typically we use `username/topic` as opposed to `category/topic`.  This helps us determine ownership and clean up old branches if they're abandoned.  No need to fix here -- just something to keep in mind for future PRs.  Alternatively, fork and name your branch whatever in the fork, as that won't linger in the commons.

lib/autoscale/aws/autoscaler.go
```diff
@@ -203,32 +203,32 @@ func (a *Autoscaler) WaitUntilInstanceTerminated(ctx context.Context, instanceID
...
+			responseToken = aws.StringValue(params.Value)
+		}
+
+		if aws.StringValue(params.Name) == url {
+			responseURL = aws.StringValue(params.Value)
+		}
 	}
-	return aws.StringValue(resp.Parameter.Value), nil
+
+	return responseURL, responseToken, nil
```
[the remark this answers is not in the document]
helgi: https://docs.aws.amazon.com/sdk-for-go/api/service/ssm/#SSM.GetParameters will return an error if you get `InvalidParameters` which in turn allows you to introspect on `InvalidParameters`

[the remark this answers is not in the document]
wadells: > https://docs.aws.amazon.com/sdk-for-go/api/service/ssm/#SSM.GetParameters will return an error if you get `InvalidParameters` which in turn allows you to introspect on `InvalidParameters`

It won't return an error.   Requestion parameters that don't exist results in a 200 and items in the InvalidPameters field.

<details><summary>demo</summary>

```go
package main

import (
    "log"

    "github.com/aws/aws-sdk-go/aws"
    "github.com/aws/aws-sdk-go/aws/session"
    "github.com/aws/aws-sdk-go/service/ssm"
)

func main() {
    region := "us-west-2"
    sess, err := session.NewSession(&aws.Config{
        Region: &region,
    })  
    if err != nil {
        log.Fatal(err)
    }   
    sm := ssm.New(sess)
    resp, err := sm.GetParameters(&ssm.GetParametersInput{
        Names: aws.StringSlice([]string{"/test/exists", "/test/doesnotexist"}),
    })  
    if err != nil {
        log.Println("Response returned error:")
        log.Println(err)
    } else {
        log.Println("Response did not return error")
    }   
    log.Println("Valid:")
    for _, param := range resp.Parameters {
        log.Println(aws.StringValue(param.Name))
    }   
    log.Println("Invalid:")
    for _, invalid := range resp.InvalidParameters {
        log.Println(*invalid)
    }   
}
```

Results in:

```
$ go run main.go
2021/04/06 12:32:19 Response did not return error
2021/04/06 12:32:19 Valid:
2021/04/06 12:32:19 /test/exists
2021/04/06 12:32:19 Invalid:
2021/04/06 12:32:19 /test/doesnotexist
```

Try it with no valid parameters at all for "fun".
</details>

Previously this code path bubbled that up as an error.  The proposed implementation does not -- it silently passes the zero value for the non-existant parameters afaict.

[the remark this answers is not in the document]
helgi: Addressed and used `trace.NotFound` - let me know if I should a different way of bubbling it up

wadells (approved):
LGTM!

bernardjkim (approved):

PR #2465, opened 2021-04-12
Bump logging application to 7.1.2

UnderGreen:
## Description

* Forward-port of Run containers with normal users [#82](https://github.com/gravitational/logging-app/pull/82)
* Fix panic in parseCSVIntoMap [#79](https://github.com/gravitational/logging-app/pull/79)

## Type of change

* Breaking change (fix or feature that would cause existing functionality to not work as expected)
* This change has a user-facing impact

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Update documentation about persistence problem

PR #2466, opened 2021-04-12
Use https to clone public repos instead of ssh

wadells:
## Description
SSH requires a unique deploy key per repo (or a dedicated service user),
wheras https will happily clone unauthenticated. This simplifies CI as well
as the new user build experience, as a github ssh key no longer needs to
be set up to build.

## Type of change

Unblocks https://github.com/gravitational/gravity/pull/2465

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [x] Address review feedback

## Testing done
Tested in my 7.0 drone backport where I initially discovered this issue: https://drone.gravitational.io/gravitational/gravity/299/1/5.  Before it looked like this: https://drone.gravitational.io/gravitational/gravity/295/1/5

## Additional information
The clone code paths are typically short-circuited because published artifacts are cached in clientbuilds s3 bucket and the build prefers to download those over building from source code.  We could also consider adding a tag build to the various `*-app` projects that pushes to clientbulilds to preempt failures like this.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -29,16 +29,12 @@ steps:
   - name: build
     image: docker:git
     environment:
-      GITHUB_RO_KEY_PR:
-        from_secret: GITHUB_RO_KEY_PR
       AWS_ACCESS_KEY_ID:
         from_secret: AWS_S3_RO_ACCESS_KEY_ID
       AWS_SECRET_ACCESS_KEY:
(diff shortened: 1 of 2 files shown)


UnderGreen (approved):
Very useful change. 👍 
LGTM.

bernardjkim (approved):

PR #2469, opened 2021-04-14
Add 7.1.0-alpha.6 changelog

wadells:
## Description
Add 7.1.0-alpha.6 changelog

## Type of change
* Documenation

## TODOs
- [ ] Address review feedback

## Testing Done
Eyeballed the links and format here: https://github.com/gravitational/gravity/blob/c1e4524e614573ea38be46f2cd972aa735db7f7d/docs/7.x/changelog.md

### DIFF
--- a/docs/7.x/changelog.md
+++ b/docs/7.x/changelog.md
@@ -11,7 +11,7 @@ Find the latest Open Source Gravity releases at [Gravity Downloads](https://grav
 
 | Version             | Latest Patch  | LTS | Release Date         | Latest Patch Date    | End of Support *        | Kubernetes Version   | Teleport Version |
 | ------------------- | ------------- | --- | -------------------- | -------------------- | ----------------------- | -------------------- | ---------------- |
-| [7.1](#71-releases) | 7.1.0-alpha.5 | No  | pre-release          | March 16, 2021       | Set upon release        | 1.19.8               | 3.2.17-gravity   |
+| [7.1](#71-releases) | 7.1.0-alpha.6 | No  | pre-release          | April 14, 2021       | Set upon release        | 1.19.8               | 3.2.17-gravity   |
 | [7.0](#70-releases) | 7.0.30        | Yes | April 3, 2020        | January 15, 2021     | July 9, 2022            | 1.17.9               | 3.2.14-gravity   |
 | [6.1](#61-releases) | 6.1.48        | Yes | August 2, 2019       | March 23, 2021       | November 10, 2021       | 1.15.12              | 3.2.14-gravity   |
 | [5.5](#55-releases) | 5.5.59        | Yes | March 8, 2019        | March 30, 2021       | March 8, 2021           | 1.13.11              | 3.0.7-gravity    |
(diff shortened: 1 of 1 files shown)


docs/7.x/changelog.md
```diff
@@ -54,6 +54,21 @@ extend updates past End of Support through customer agreements if required.
...
+* Improve performance of AWS autoscaling integration ([#2458](https://github.com/gravitational/gravity/pull/2458))
+* Improve satellite cluster health monitoring on large clusters up to 1000 nodes ([#2439](https://github.com/gravitational/gravity/pull/2439), [planet#828](https://github.com/gravitational/planet/pull/828), [satellite#297](https://github.com/gravitational/satellite/pull/297))
+
+#### Bugfixes
+* Fix ownership of monitoring and logrange directories ([#2448](https://github.com/gravitational/gravity/pull/2448))
+* A kubernetes config is no longer required for `gravity app sync` ([#2438](https://github.com/gravitational/gravity/pull/2438))
+
+#### Internal Changes
+* Upgrade internal kubernetes client go to v1.19.8 and helm to v3.4.2. Removed helm2. ([#2433](https://github.com/gravitational/gravity/pull/2433), [planet#827](https://github.com/gravitational/planet/pull/827), [rigging#100](https://github.com/gravitational/rigging/pull/100), [satellite#299](https://github.com/gravitational/satellite/pull/299))
```
wadells, line 69: @bernardjkim I may have made up that helm 2 was removed.  Can you check this note for me, please?

docs/7.x/changelog.md
```diff
@@ -54,6 +54,21 @@ extend updates past End of Support through customer agreements if required.
 
 7.1 is currently pre-release.
 
+### 7.1.0-alpha.6 (April 14, 2021)
+
+#### Improvements
+* Add GCE Alias IP support for flannel ([#2439](https://github.com/gravitational/gravity/pull/2439), [planet#830](https://github.com/gravitational/planet/pull/830), [flannel#10](https://github.com/gravitational/flannel/pull/10))
```
knisbet, line 60: I think we put `.` at the end of each entry, although I think alpha.5 is also missing this.

knisbet (approved):

docs/7.x/changelog.md
bernardjkim: Helm v2 binary is still included in gravity. The integrated helm/app command has just been upgraded to use helm v3

bernardjkim: I still don't really know the use cases, so I'm not sure if we can just remove helm v2 binary or not.

docs/7.x/changelog.md
wadells: Added to this changelog in 6aa4ac2.  I'll catch the others in a later PR as I don't want to include unrelated changes from a diff tracking perspective.

docs/7.x/changelog.md
wadells: Helm 2 is no longer mentioned in 6aa4ac2.  My impression is without tiller, the helm 2 binary is somewhat useless.  Customers would have to run their own tiller (in which case they could probably bootstrap their own helm 2 in the hooks containers).

If we do remove helm 2, it should be before 7.1.0 releases.  It'd be bad form to remove something like that in a patch release.

bernardjkim: Hmm, that's a good point. I'll do some clean up before 7.1.0 release.

PR #2479, opened 2021-04-20
Create new release docs

bernardjkim:
## Description

Gravity is no longer in active development. I don't think it's necessary to maintain separate docs for each new major version of Gravity. Most likely, we'll only be bumping the major version when upgrading Kubernetes version.

Instead of creating a new dir for version `8.x`, I've created a `release` dir that we can write all future changelogs. Maybe we can change the naming to `latest` instead of `release`?

EDIT:
Sticking with the current approach to documentation. This PR adds 8.x and 9.x docs and includes changelogs for `8.0.0-alpha.0` and `9.0.0-alpha.0`.

## Type of change

* Documentation

## TODOs

- [x] Self-review the change
- [x] Write documentation
- [ ] Address review feedback


wadells:
> Instead of creating a new dir for version `8.x`, I've created a `release` dir that we can write all future changelogs. Maybe we can change the naming to `latest` instead of `release`?

I still like the idea of maintaining 7.0, 8.0, 9.0 docs directories.

Why?  The current documentation *is* useful internally -- I reference it regularly and wouldn't want it to fall out of sync with the code if we plan on using gravity.  We'll hopefully be removing features from 9.0 that won't necessarily be removed from previous versions.  E.g. 9.0 will need stuff cut from the docs that won't necessarily apply to 8.0, because we wouldn't backport those removals -- no ROI.

In practice, I think the only difference is we make two copies of 7.x.  One for 8.0 that will basically never be touched again and one for 9.0/latest.  8.0 will probably only exist as a vestigial upgrade path between 7.0 and 9.0/latest with no real updates.  Changelogs will only go into 9.0/latest.  We could even remove the changelog file from previous versions -- thought that would probably require fixing some links.

wadells:
You'll want to update the makefile to pick up these versions in `CFGS` and also select a `LATEST_CFG`

bernardjkim:
>  We'll hopefully be removing features from 9.0 that won't necessarily be removed from previous versions. E.g. 9.0 will need stuff cut from the docs that won't necessarily apply to 8.0, because we wouldn't backport those removals -- no ROI.

Once we deploy the latest version of gravity, I don't think we'd be very likely to fall back to a previous version, and so I don't think it will really be necessary to maintain docs for all previous versions.

I'll go ahead and stick with how we've been doing it so far and create 8.0/9.0 docs. We can always re-evaluate this later and make large refactoring changes to the docs if necessary.

wadells (approved):

PR #2484, opened 2021-04-26
[7.0.x] Fix terraform deprecation warnings and ansible module for 2.8+

UnderGreen:
## Description

Fixed some deprecation warnings for terraform > 0.12.x. Ansible unarchive module in the latest versions does not have `force` parameter and that parameter didn't make sense in the first place(usually it is safer to recreated nodes from scratch to avoid any already running gravity binaries on nodes).

## Type of change

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback
- [ ] Backport to 5.5.x and forward-port to master
-

### DIFF
--- a/terraform-dev/cluster.tf
+++ b/terraform-dev/cluster.tf
@@ -1,25 +1,25 @@
 variable "image_name" {
-  type = "string"
+  type = string
   default = "ubuntu-18.04-server-cloudimg-amd64.img"
 }
 
 variable "disk_size" {
-  type = "string"
+  type = string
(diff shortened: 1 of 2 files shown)


bernardjkim (approved):

wadells:
@UnderGreen: Robotest is turned off due to a security concern.  I'm currently working on it.

wadells:
Does this need to be forward ported? Will it result in developer facing regressions between 7.x and later versions?

Issue #2459, opened 2021-04-05
gravity restores node labes declared in app.yaml


UnderGreen:
### Description

**What happened**:
gravity restores custom node label declared in the application manifest.
```
telekube0:/$ kubectl get nodes -Lstolon-keeper -w
NAME           STATUS   ROLES   AGE   VERSION   STOLON-KEEPER
172.28.128.3   Ready    node    15h   v1.17.9   yes
172.28.128.4   Ready    node    15h   v1.17.9
172.28.128.5   Ready    node    15h   v1.17.9   yes
172.28.128.4   Ready    node    15h   v1.17.9   yes
172.28.128.5   Ready    node    15h   v1.17.9   yes
```
app.yaml content:
```
nodeProfiles:
  - name: node
    description: "stolon database node"
    labels:
      role: "node"
      stolon-keeper: "yes"
```
**What you expected to happen**:
Behavior existed in 5.5 and lower versions: manual manipulation with node labels are not affected by gravity.

**How to reproduce it (as minimally and precisely as possible)**:
Create a cluster with custom node labels declared in app.yaml and delete that label from any node.

### Environment

 - Gravity version [e.g. 7.0.11]: 7.0.30
 - OS [e.g. Redhat 7.4]:
 - Platform [e.g. Vmware, AWS]:

**Relevant Debug Logs If Applicable**
Changing label value is not affected.
```
telekube0:/$ kubectl get nodes -Lstolon-keeper -w
NAME           STATUS   ROLES   AGE   VERSION   STOLON-KEEPER
172.28.128.3   Ready    node    15h   v1.17.9   yes
172.28.128.4   Ready    node    15h   v1.17.9   no
172.28.128.5   Ready    node    15h   v1.17.9   yes
172.28.128.3   Ready    node    15h   v1.17.9   yes
172.28.128.5   Ready    node    15h   v1.17.9   yes
172.28.128.4   Ready    node    15h   v1.17.9   no
172.28.128.3   Ready    node    15h   v1.17.9   yes
172.28.128.5   Ready    node    15h   v1.17.9   yes
172.28.128.4   Ready    node    15h   v1.17.9   no
```

PR #2486, opened 2021-05-12
Add possibility to disable label reconciliation

whiterm:
## Description

User can change node label `gravitational.io/reconcile` to change  behavior  for label reconciliation.
Available values:
* EnsureExists - checking for existence only. Users can edit the labels as they want. (default value)
* True - If the value of the label on the node is not equal to the value from the NodeProfile, the value will be overwritten.
* False - reconciliation is disabled for current node

## Type of change
New feature (non-breaking change which adds functionality)

* Closes #, refs #2459 

* Requires #

* Ports #

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [x] Write documentation
- [ ] Address review feedback


lib/process/process.go
```diff
@@ -731,36 +736,65 @@ func (p *Process) reconcileNode(client *kubernetes.Clientset, cluster ops.Site,
...
+		}
 	}
 	return nil
 }
 
-func getMissingLabels(cluster ops.Site, server storage.Server, node v1.Node) (map[string]string, error) {
-	profile, err := cluster.App.Manifest.NodeProfiles.ByName(server.Role)
+func reconcileLabels(logger logrus.FieldLogger, currentLabels, requiredLabels map[string]string) (map[string]string, bool) {
+	labels := make(map[string]string)
+	// Copy from the currentLabels
```
a-palchikov, line 771: nit: only comment things that aren't obvious

lib/process/process.go
```diff
@@ -731,36 +736,65 @@ func (p *Process) reconcileNode(client *kubernetes.Clientset, cluster ops.Site,
...
-	requiredLabels := server.GetNodeLabels(profile.Labels)
 	for key, val := range requiredLabels {
-		if _, ok := node.Labels[key]; !ok {
-			missingLabels[key] = val
+		currentVal, ok := labels[key]
+		if !ok {
+			labels[key] = val
+			needUpdate = true
+		}
+		if reconcileMode == defaults.ReconcileModeEnabled && currentVal != val {
```
a-palchikov, line 792:
```suggestion
		if !ok || reconcileMode == defaults.ReconcileModeEnabled && currentVal != val {
```

lib/process/process_test.go
```diff
@@ -324,3 +326,162 @@ type importerSuite struct {
...
+			},
+			expectedLabels: map[string]string{
+				"gravitational.io/reconcile": "EnsureExists",
+				"label1":                     "value1",
+				"label2":                     "value2",
+			},
+			needUpdate: true,
+		},
+		{
+			name: "reconciliation mode = EnsureExists. Different value of labels",
```
a-palchikov, line 416: Maybe add a hint at the intent (to emphasize the difference between `Enabled` and `EnsureExists`):
```
name: reconciliation mode = EnsureExists. Different value of labels and no change is expected.
```

lib/process/process_test.go
```diff
@@ -324,3 +326,162 @@ type importerSuite struct {
...
+			},
+			expectedLabels: map[string]string{
+				"gravitational.io/reconcile": "EnsureExists",
+				"label1":                     "value1",
+				"label2":                     "value2",
+			},
+			needUpdate: true,
+		},
+		{
+			name: "reconciliation mode is incorrect. Different value of labels",
```
a-palchikov, line 455: ditto here:
```
name: "reconciliation mode is incorrect. Different value of labels and no change is expected",
```
Consider adding a test that an invalid mode does not behave as `EnsureExists` (e.g. will not add labels)

vagrant/ansible/upload.yaml
```diff
@@ -15,44 +15,6 @@
   - set_fact: gopath="{{ lookup('env','GOPATH') }}"
   - debug: msg="gopath {{gopath}}"
 
-  - name: Creating /vagrant/installer
```
a-palchikov, line 18: Can you also provide a bit of detail why this is being removed?

lib/defaults/labels.go
```diff
@@ -0,0 +1,71 @@
...
+	KubernetesRoleLabel = "gravitational.io/k8s-role"
+
+	// KubernetesAdvertiseIPLabel is the kubernetes node label of the advertise IP address
+	KubernetesAdvertiseIPLabel = "gravitational.io/advertise-ip"
+
+	// RunLevelLabel is the Kubernetes node taint label representing a run-level
+	RunLevelLabel = "gravitational.io/runlevel"
+
+	// KubernetesReconcileLabel is the kubernetes node label to control the reconcile process for the node
+	KubernetesReconcileLabel = "gravitational.io/reconcile"
```
a-palchikov, line 35: I have a small concern with a too broad of a name for this label - it should hint at labels as both it and the values are node label-specific - e.g. `reconcile-labels` or even flipping it and embedding the context in the label name: `node-labels.gravitational.io/mode` and naming the modes `reconcile`, `ensureexists` and empty or `none` (for no reconciliation)?

Or did you want to reserve the name for a more general application that might potentially extend to future types of reconciliation loops?

lib/defaults/defaults.go
```diff
@@ -25,9 +25,8 @@ import (
 	"runtime"
 	"time"
 
-	"github.com/gravitational/gravity/lib/constants"
-
 	"github.com/coreos/go-semver/semver"
+	"github.com/gravitational/gravity/lib/constants"
```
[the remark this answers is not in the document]
whiterm: Groups are usually grouped by project. In the original version, this was not.

lib/defaults/labels.go
whiterm: I thought about more general behavior that might potentially extend to future types of reconciliation loops. And I thought about the name `reconcile.gravitational.io/mode`, maybe it would be better. What do you think about it?

vagrant/ansible/upload.yaml
whiterm: This file has duplicate code which is not needed here

lib/process/process_test.go
whiterm: Sure

lib/process/process_test.go
whiterm: Sure

lib/process/process.go
```diff
@@ -731,36 +736,65 @@ func (p *Process) reconcileNode(client *kubernetes.Clientset, cluster ops.Site,
...
+	if err != nil {
+		return trace.Wrap(err)
 	}
-	if _, err := client.CoreV1().Nodes().Update(context.TODO(), &node, metav1.UpdateOptions{}); err != nil {
-		return rigging.ConvertError(err)
+
+	requiredLabels := server.GetNodeLabels(profile.Labels)
+	needUpdate := false
+	node.Labels, needUpdate = reconcileLabels(p, node.Labels, requiredLabels)
+	if needUpdate {
```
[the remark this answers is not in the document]
whiterm: Sure

lib/defaults/labels.go
a-palchikov: I think it could be confusing if this had an additional application (outside of labels) as none of the current values are label-specific. So you would either have to embed the label detail inside of the mode values - e.g. `EnsureLabelExists` or `ReconcileLabels` or go with a more specific label name: `node-label.gravitational.io/mode` similar to [addonmanager](https://github.com/kubernetes/kubernetes/tree/4f254674fa7f3752fcf25cbd17ad6dc2485c124c/cluster/addons/addon-manager)

whiterm: I suggest not to use the word "label" in the name and use a more generic name, because later we can add an additional type of reconciliation that will rely on this label.

knisbet:
I'm mostly curious about why this approach. As I understand #2459 the problem is certain labels want to be controlled by stolon, but initialized by gravity at install time. As I understand this change, it seems more related to turning reconciliation off / change behaviour for all labels configured in the app.yaml. 

I'm not sure if upgrades still use a separate path or just use the reconciliation logic when an upgrade introduces or changes labels. Might be unexpected interaction with the way this function works, unless I'm missing something.

So why this approach vs something like updating the app.yaml to specify specific labels to be ignored. For example from Sergei's example:
```
nodeProfiles:
  - name: node
    description: "stolon database node"
    labels:
      role: "node"
      stolon-keeper: "yes"
    disableLabelReconciler:
    - stolon-keeper
```

whiterm:
@knisbet Your approach makes sense when we want permanent behavior. I thought it needed as temporary. You remove the labels, do something and want to put them back. 
For example, you want to evict the pods from some node and you know that the `drain` command cannot help you because this pods is controlled by a DaemonSet. And the best way to evict these pods is to remove the labels from the node.

UnderGreen:
From Kevin's example looks like I need to set specific label in node profile, what means it could be hard change something on running cluster. I didn't dig into changes, but if Roman's change allows me to do anything(temporarily) on running cluster without interacting with a manifest then I would like to have that one.

whiterm:
@knisbet I would change your approach a little. Like this:
```
nodeProfiles:
  - name: node
    description: "stolon database node"
    labels:
      role: "node"
    initLabels:
      stolon-keeper: "yes"
```
In this case, you do not need to duplicate the same label name in two different places in `labels` and `disableLabelReconciler `.
And I think that this approach might be more difficult to implement, because we have to write somewhere which labels have already been initialized. This can help determine if this label has been removed or has not yet been initialized.

knisbet:
Ok hmm, I might've miss-understood what sergei was noting in the original issue. So the question is, whether this is meant to be used during troubleshooting / resolving an outage, in which case the proposed approach probably makes more sense where a particular node no longer gets any reconciliation of labels as long as the admin remembers to re-enable after resolving the problem.

Or if the reconciliation is interfering with normal use of some of the labels where it's not expected for an automated system to also control those labels, so it should just ignore labels that are normally tweaked through normal operation of the cluster.

My suggestion was for the second issue, not really the first.

@UnderGreen please clarify which problem you're trying to solve.

UnderGreen:
For my case, it is about temporary troubleshooting/maintenance.

lib/process/process.go
```diff
@@ -718,49 +723,74 @@ func (p *Process) reconcileNodeLabels(ctx context.Context, client *kubernetes.Cl
...
+func reconcileLabels(logger logrus.FieldLogger, currentLabels, requiredLabels map[string]string) (map[string]string, bool) {
+	labels := make(map[string]string)
+	for key, value := range currentLabels {
+		labels[key] = value
+	}
+	needUpdate := false
+	reconcileMode, err := defaults.ParseReconcileMode(labels[defaults.KubernetesReconcileLabel])
 	if err != nil {
-		return nil, trace.Wrap(err)
+		logger.WithError(err).Error("Unable to get reconcile mode, will reset to EnsureExists.")
```
knisbet, line 778: Because this is primarily a troubleshooting tool, I'm curious about the tradeoffs of this approach vs treating a missing label as enabled. 

I suspect this approach is slightly easier to use, since someone can just look at the label to find the right label and edit the mode. On the other hand it's another label that under normal circumstances wouldn't really need to be set since it's really for troubleshooting.

In the wild too, I could see some of the on-prem teams struggling to use this a bit, which might lead to some cli options to set/remove the label / show the status. In which case the label could just be missing since the main UX would be a CLI helper. 

I think both approaches are valid, so if you want to stick with this approach I'm fine with it.

knisbet (approved):
I think the only thing I'd ask for yet is to add a note in the docs somewhere about this label and what it does. Doesn't have to be super complex, but some way to refer to it would be great.

And then to consider whether a missing label should have the label added, or get interpreted as reconciler enabled. Since this is primarily a debug tool, it's probably mostly about which is less noisy and easier to use.

lib/process/process.go
whiterm: In my opinion, if the label is missing, it should be added automatically. This helps the operator team not to look for the label name in the documentation, but simply execute: `kubectl get no --show-labels`

knisbet: Yea, makes sense.

knisbet:
@wadells please rerun the CI once we've got robotest setup again.

wadells:
> @wadells please rerun the CI once we've got robotest setup again.

@whiterm can you rebase or merge to pick up 018b75d7d826a271a25673d3712513e02153647b from master?  This should fix CI.  I'd typically do this for you, but since we're new to working together, I didn't want to make unexpected changes to your branch.

wadells (approved):

docs/9.x/cluster.md
```diff
@@ -1440,6 +1440,33 @@ nodeProfiles:
...
+nodeProfiles:
+  - name: node
+    description: "Gravity Node"
+    labels:
+      label1: value1
+      label2: value2
+```
+
+Gravity has three modes of label reconciliation, which the user can change at any time:
+ - `EnsureExists` (default) - Compares the labels in the node with what is written in the profile, the value of the labels is not compared.
```
a-palchikov, line 1457:
```suggestion
 - `EnsureExists` (default) - Ensures that the labels from the node profile exist recreating the missing ones. The label values from the node profile are not enforced.
```

docs/9.x/cluster.md
```diff
@@ -1440,6 +1440,33 @@ nodeProfiles:
...
+   - Users can edit the labels as they want.
+   - Users cannot delete the label specified in the node profile.
+ - `Enabled` or `true` - Compares the labels in the node with what is written in the profile, the values of the labels are also compared.
```
a-palchikov, line 1460:
```suggestion
 - `Enabled` or `true` - Ensures that the labels from the node profile exist recreating the missing ones. Also ensures that the label values match those from the node profile.
```

docs/9.x/cluster.md
```diff
@@ -1440,6 +1440,33 @@ nodeProfiles:
...
+   - Users cannot change the labels specified in the node profile.
+   - Users cannot delete the label specified in the node profile.
+ - `Disabled` or `false` - Reconciliation of labels is disabled.
+   - Users can edit the labels as they want.
+   - Users can delete labels.
+
+Reconciliation mode is assigned to the node and controlled by the label `label-reconciler.gravitational.io/mode`.
```
a-palchikov, line 1467:
```suggestion
Reconciliation mode on a node is controlled via the label named `label-reconciler.gravitational.io/mode`.
```

a-palchikov (approved):

PR #2495, opened 2021-05-20
Linter configuration

a-palchikov:
## Description

This PR implements just the configuration required to set up linting as requested in https://github.com/gravitational/gravity/pull/2488#pullrequestreview-663470519

## Type of change

* Internal change (not necessarily a bug fix or a new feature)

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Implementation
I flipped the `golint` target (and correspondingly `grpc` as the version definitions moved to build.assets/Makefile) so that the root Makefile builds on host and docker-based targets are available in build.assets/Makefile - this is the way teleport's makefiles are structured and I find it more meaningful.

PR #2496, opened 2021-05-21
Address lint warnings #2

a-palchikov:
## Description

This is a continuation of lint warning fixes.

## Type of change

## TODOs

- [x] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,14 @@ K8S_VER_SUFFIX := $(shell printf "%d%02d%02d" $(shell echo $(K8S_VER) | sed "s/\
 GOLFLAGS ?= -w -s
 GOLINT ?= golangci-lint
 GOLINT_PACKAGES ?= \
-	./lib/httplib/... ./lib/localenv/... ./lib/system/...
+	./lib/httplib/... \
+	./lib/localenv/... \
+	./lib/system/... \
+	./lib/archive/... \
+	./lib/blob/... \
+	./lib/clients/... \
+	./lib/helm/... \
+	./lib/kubernetes/...
 
 GOPATH ?= $(shell go env GOPATH)
 
--- a/lib/archive/archive.go
+++ b/lib/archive/archive.go
@@ -153,7 +153,7 @@ func HasFile(tarballPath, filename string) error {
 	err = TarGlob(tar.NewReader(file), ".", []string{filename},
 		func(match string, file io.Reader) error {
 			hasFile = true
-			return Abort
+			return ErrAbort
 		})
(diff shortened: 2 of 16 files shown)


lib/install/dispatcher/dispatcher.go
```diff
@@ -99,12 +103,14 @@ type Status byte
...
+	StatusUnknown Status = iota
 	// StatusCompleted indicates a completed operation
-	StatusCompleted Status = iota
+	StatusCompleted
 	// StatusCompletedPending indicates a completed operation
 	// but with installer still active
-	StatusCompletedPending Status = iota
+	StatusCompletedPending
+	// StatusAborted indicates an aborted operation
+	StatusAborted
```
wadells, line 113: This (and the related case) seems like new code?

lib/pack/localpack/packageserver.go
```diff
@@ -383,8 +384,12 @@ func (p *PackageServer) UnpackedPath(loc loc.Locator) (string, error) {
 }
 
 func (p *PackageServer) Unpack(loc loc.Locator, targetDir string) error {
-	var err error
+	return p.UnpackWithOptions(loc, targetDir, nil)
+}
 
+// UnpackWithOptions unpacks the package specified with loc in targetDir
+// using opts for extraction
+func (p *PackageServer) UnpackWithOptions(loc loc.Locator, targetDir string, opts *dockerarchive.TarOptions) (err error) {
```
wadells, line 392: What is this new function for? It doesn't seem used except by the old `Unpack` -- in which case why add a new public function?

wadells (approved):

lib/pack/localpack/packageserver.go
wadells: Looks like this change shouldbe included in https://github.com/gravitational/gravity/pull/2498 instead -- that PR has the rest of the context.

lib/install/dispatcher/dispatcher.go
a-palchikov: Yes, it is. I'll move this to another PR.

lib/pack/localpack/packageserver.go
a-palchikov: Removed.

a-palchikov:
@wadells I addressed all your comments, PTAL.

wadells (approved):
Please take #2495 first, and enable the appropriate files/linters when merging this PR -- so that we're consistent with our enforcement.

PR #2498, opened 2021-05-21
Address lint warnings #4

a-palchikov:
## Description

This is a continuation of lint warning fixes.

## Type of change

## TODOs

- [x] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,14 @@ GOLINT_PACKAGES ?= \
 	./lib/helm/... \
 	./lib/kubernetes/... \
 	./lib/report/... \
-	./lib/schema/...
+	./lib/schema/... \
+	./e/lib/events/... \
+	./lib/processconfig/... \
+	./lib/state/... \
+	./lib/systeminfo/... \
+	./lib/testutils/... \
+	./lib/network/... \
+	./lib/pack/...
 
 GOPATH ?= $(shell go env GOPATH)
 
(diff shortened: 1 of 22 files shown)


e/tool/tele/cli/helm.go
```diff
@@ -61,12 +61,19 @@ func (o *repoAddOptions) repoAdd() error {
...
+	if !locked {
+		return trace.LimitExceeded("unable to obtain file lock")
+	}
+	defer func() {
+		errUnlock := fileLock.Unlock()
+		if err != nil {
+			return
+		}
+		err = errUnlock
+	}()
```
wadells, line 76: It feels ugly that the unlock error is conditionally ignored.  Typically in this case we'd log, e.g:

https://github.com/gravitational/gravity/pull/2488/files#diff-1e407a6245149757c630df2b55ac3e9d17db91779569d981c4d77f5536b9a0afR127-R135

Why isn't that the approach here?

wadells (approved):

e/tool/tele/cli/helm.go
a-palchikov: The linked diff implements an error handler that logs the error - here, it makes sense to fail the function as a result. If you mean that the error unlocking should be logged in case another error is active - I'll update that.

a-palchikov:
@wadells I addressed all your comments, PTAL.

wadells (approved):
Please take #2495 first, and enable the appropriate files/linters when merging this PR -- so that we're consistent with our enforcement.

PR #2514, opened 2021-06-03
Linter warnings #8

a-palchikov:
## Description

This PR focuses on lint changes in the builder package.

## Type of change

## TODOs

- [x] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,12 @@ GOLINT_PACKAGES ?= \
 	./lib/process/... \
 	./e/lib/process/... \
 	./lib/vacuum/... \
-	./lib/expand/...
+	./lib/expand/... \
+	./e/lib/builder/... \
+	./e/lib/catalog/... \
+	./lib/builder/... \
+	./lib/catalog/... \
+	./lib/hub/...
 
 GOPATH ?= $(shell go env GOPATH)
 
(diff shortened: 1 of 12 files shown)


wadells (approved):
Both #2513 and this have a bunch of churn related to formerly-embedded-now-private fields of newly public structs.  What is the advantage of making the struct public, over an interface that specifies only the public methods needed outside the namespace?  Better gopls integration? More idiomatic? Are there other solutions you considered?

Not a critique, just curious as I build my go skills.

a-palchikov:
> Both #2513 and this have a bunch of churn related to formerly-embedded-now-private fields of newly public structs. What is the advantage of making the struct public, over an interface that specifies only the public methods needed outside the namespace? Better gopls integration? More idiomatic? Are there other solutions you considered?
> 
> Not a critique, just curious as I build my go skills.

Reasons are somewhat hectic tbh:
  * if there was an interface, I'd use it (but not consistently)
  * If there was no interface, the type would be exported and all exported fields would be unexported

Pragmatic reasons why one would favor an exported type over an interface would be that e.g. a unit test relied on accessing state in which case I'd always prefer having a way of constructing a concrete value instead of having to typecast an interface value. But you can argue further that tests should ideally be in a separate package and hence work exclusively against an exported interface and figure out whether you actually need to access the state in the test. So I don't have a compelling reason to always favor one over the other.

PR #2517, opened 2021-06-04
Lint warnings #10

a-palchikov:
## Description

This PR focuses on lint changes in lib/rpc.

## Type of change

## TODOs

- [x] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback

### DIFF
--- a/lib/rpc/server/testing.go
+++ b/lib/rpc/server/testing.go
@@ -60,16 +60,21 @@ func TestClientCredentials(c *C) credentials.TransportCredentials {
 	if !cp.AppendCertsFromPEM(certFile) {
 		c.Error("failed to append certificates")
 	}
-	//nolint:gosec // TODO: set MinVersion
-	return credentials.NewTLS(&tls.Config{ServerName: "agent", RootCAs: cp})
+	return credentials.NewTLS(&tls.Config{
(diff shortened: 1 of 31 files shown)


lib/rpc/server/testing.go
```diff
@@ -60,18 +60,20 @@ func TestClientCredentials(c *C) credentials.TransportCredentials {
 	if !cp.AppendCertsFromPEM(certFile) {
 		c.Error("failed to append certificates")
 	}
+	//nolint:gosec // TODO: set MinVersion
```
knisbet, line 63: it's test code, so may as well just set the min version?

lib/rpc/tls.go
```diff
@@ -194,6 +194,7 @@ func ClientCredentials(secretsDir string) (credentials.TransportCredentials, err
 		return nil, trace.BadParameter("failed to add CA to pool")
 	}
 
+	//nolint:gosec // TODO: set MinVersion
```
knisbet, line 197: these ones are probably more important, even if not part of this PR I would probably set atleast TLS 1.2 and probably TLS1.3 soon.

knisbet (approved):

PR #2519, opened 2021-06-04
Lint warnings #12

a-palchikov:
## Description

This PR focuses on lint changes in lib/app.
Note: there is a bit of redundancy to other lint PRs - this is intentional and will be resolved at merge time.

## Type of change

## TODOs

- [x] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback

### DIFF
--- a/lib/schema/template.go
+++ b/lib/schema/template.go
@@ -18,6 +18,7 @@ package schema
 
 import (
 	"bytes"
+	"context"
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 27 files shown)


lib/schema/template.go
```diff
@@ -176,8 +177,13 @@ func valueFromFile(path, basePath string) (data []byte, mimeType string, err err
...
 // and its MIME type
-func valueFromHTTP(url string) (data []byte, mimeType string, err error) {
-	response, err := http.Get(url)
+func valueFromHTTP(ctx context.Context, url string) (data []byte, mimeType string, err error) {
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return nil, "", trace.Wrap(err)
+	}
+	req = req.WithContext(ctx)
+	response, err := (&http.Client{}).Do(req)
```
knisbet, line 186: Instead of creating an empty client it might be preferable to use http.DefaultClient? Alternatively might want to initialize the client with a Timeout, although the context should largely cover that, but might help if someone just passes a TODO context.

knisbet (approved):

lib/schema/template.go
a-palchikov: `http.DefaultClient` is also the empty `http.Client` value - e.g. it has the same defaults. I'll update to use `lib/httplib` for timeouts though.

PR #2529, opened 2021-06-07
Lint warnings #15

a-palchikov:
## Description

This PR focuses on lint changes in lib/users.

## Type of change

## TODOs

- [x] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,8 @@ GOLINT_PACKAGES ?= \
 	./lib/storage/... \
 	./lib/app/... \
 	./lib/docker/... \
-	./lib/loc/...
+	./lib/loc/... \
+	./lib/users/...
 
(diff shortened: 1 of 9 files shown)


a-palchikov:
Can any of the admins review this?

wadells (approved):
What is the reason for the `u` -> `c` reciever change?  Typically they change to match the first letter of the receiver type, but in this case it changes away?

a-palchikov:
> What is the reason for the `u` -> `c` reciever change? Typically they change to match the first letter of the receiver type, but in this case it changes away?

Not sure why I went with `c` but it was original a tie:
```
ST1016: methods on the same type should have the same receiver name (seen 86x "c", 89x "u")
```

PR #2542, opened 2021-06-10
Remove call to ComponentStatus API

bernardjkim:
## Description

ComponentStatus API has been deprecated in Kubernetes 1.19+. We might want to re-implement this check using a different method, but the API check will be removed for now.  See [CHANGELOG](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md#deprecation-5)
## Type of change

## TODOs

- [ ] Self-review the change
- [ ] Perform manual testing
- [ ] Address review feedback

### DIFF
--- a/lib/install/phases/postsystem.go
+++ b/lib/install/phases/postsystem.go
@@ -90,11 +90,6 @@ func (p *waitExecutor) waitForAPI(ctx context.Context, done chan bool) {
 	for {
 		select {
 		case <-timer.C:
-			if err := p.tryQueryAPI(); err != nil {
-				p.Infof("Waiting for Kubernetes API to start: %v", err)
-				continue
-			}
-			p.Debug("Kubernetes API is available.")
 			if err := p.tryQueryNamespace(); err != nil {
 				p.Infof("Waiting for kube-system namespace: %v", err)
 				continue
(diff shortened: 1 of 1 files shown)


wadells (approved):
What is the motivation for this change?  Is it from a test failure, or something you saw in upgrade logs?

Per [Chesterson's Fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence), do you understand why this check was initially added, and are confident it isn't valuable?  Looks like it came from @knisbet's work in #136 / 2fa20d25d33384cd3fb6c63dd2bbc654bd35a9fe.

IMO it seems fine given the following kube ns call can do double duty and verify the API.

a-palchikov:
You can here that these checks are basically HTTP probes on `127.0.0.1:<port>/healthz` (for scheduler `127.0.0.1:10251/healthz`): https://github.com/kubernetes/kubernetes/blob/release-1.19/pkg/registry/core/rest/storage_core.go#L345-L346. So they could be replaced, for example, by an agent API to do the same probe on the leader node in non-HA or on all masters in HA mode.
Additionally, the HTTP insecure port could also be closed (there's already 10259 with tls which allows insecure on `/healthz`)

Issue #2536, opened 2021-06-09
Gravity Drone jobs failing due to insufficent ephemeral-storage


wadells:
### Description

**What happened**:

A PR build failed on the new drone.teleport.dev infrastrucutre with the following error:

```
[ERROR]: write /drone/src/e/build/cache/gravity-state/7.0.12/packages/tmp/blob181123562: no space left on device
```

https://drone.teleport.dev/gravitational/gravity/68/1/9

Although different symptoms, I believe the following has the same root cause:

https://drone.teleport.dev/gravitational/gravity/70/1/9
https://drone.teleport.dev/gravitational/gravity/73/1/9

**What you expected to happen**:

I expect CI jobs to run to completion without infrastructral issues.

**How to reproduce it (as minimally and precisely as possible)**:

Queue up several Gravity PRs at the same time

wadells:
When I looked into this, I found some of the pods were being evicted:

```
Status:       Failed                                                                                                                                                                          
Reason:       Evicted  
Message:      The node was low on resource: ephemeral-storage. Container drone-4ig7p4tf79bebcqfw0x9 was using 2204Ki, which exceeds its request of 0. Container drone-rv43li9h56l51t5j128p was using 224Ki, which exceeds its request of 0.
```

Worse, the `/var/lib/docker` FS had filled up causing docker to get in a strange state were it was up and systemd though it was healthy, but it wouldn't actually turn on any pods due to a stale socket.

Ultimately, this is because gravity builds (with robotest artifacts) weigh in at about 30GB of data, and the current node pool has 100GB of ephemeral storage.  3 builds on the same node and everything falls over.

wadells:
drone-runner-kube doesn't allow a node to specify its ephemeral storage requirements:

https://github.com/drone-runners/drone-runner-kube/blob/5abe9d7f0d2ddcc0ff4b2fd8c2ae6dfa2a727def/engine/resource/pipeline.go#L188-L203

So we'll need to use something else to proxy for that for the time being.  CPU is probably fine -- have gravity jobs request < 50% of a node's CPU and no two will ever get scheduled on the same node.

wadells:
Also, I'm moving gravity CI to its own nodepool, which will have larger root volumes (200gb).  It is a notoriously bad neighbor to other jobs, and we don't want to decrease the overall QoS because of one ill behaved customer.

[...]

PR #2550, opened 2021-06-21
[7.0.x] Add ability to cache docker images

whiterm:
## Description
This PR adds a new `--image-cache-dir` flag. This path will correspond to the file store for the docker registry. When `tele build` creates a new tarball, each docker image will be put into this registry. And this folder is copied to `vendor/registry` when building target app.
This cache registry will contain all tagged images from the previous build. But after this directory is copied to `vendor/ registry`, the cleanup procedure will remove all unnecessary images.

These are not breaking changes. If you do not use this flag, then everything works as before.

## Type of change
* New feature (non-breaking change which adds functionality)act

* Closes #, refs #

* Requires #

* Ports #2540

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [ ] Write documentation
- [ ] Address review feedback

### DIFF
--- a/build.assets/Makefile
+++ b/build.assets/Makefile
@@ -124,7 +124,7 @@ endif
 #
 .PHONY: test
 test: testbox test-etcd
-	docker run --net=host --rm=true $(NOROOT) \
+	docker run --net=host --rm=true -u root  \
 		-v $(TOP):$(SRCDIR) \
 		-v /var/run/docker.sock:/var/run/docker.sock \
 		--tmpfs /tmp:rw,nosuid,nodev,exec,relatime \
--- a/e/tool/tele/cli/run.go
+++ b/e/tool/tele/cli/run.go
@@ -84,6 +84,7 @@ func Run(tele Application) error {
 			PackageVersion:         *tele.BuildCmd.Version,
 			ResourcePatterns:       *tele.BuildCmd.VendorPatterns,
 			IgnoreResourcePatterns: *tele.BuildCmd.VendorIgnorePatterns,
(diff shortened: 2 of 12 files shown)


whiterm:
@wadells Could you take a look at the issue with building the robotest? Maybe the same problem as last time, not enough disk space.

wadells:
> @wadells Could you take a look at the issue with building the robotest? Maybe the same problem as last time, not enough disk space.

I agree with your assesment.  The symptom is different than the more explicit one documented in https://github.com/gravitational/gravity/issues/2536

```
rpc error: code = Unknown desc = unable to inspect docker image "sha256:7303fde5f20624623382d7b52fcd3c5f82d10035dbd93c9790385c2cc5e46ac7" while inspecting docker container "6e51e9f9aef994ac9764b1f11548f5c5f1e28ee53a7f5959d3d2f01a906677d4": error during connect: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.40/images/sha256:7303fde5f20624623382d7b52fcd3c5f82d10035dbd93c9790385c2cc5e46ac7/json": read unix @->/var/run/docker.sock: read: connection reset by peer
```

But after preliminary research on this, it looks space related.

wadells:
I'm working on some worker changes that should improve the behavior here (fewer jobs per node, as well as more space) --  I'll update this PR once those are rolled out.

wadells:
> I'm working on some worker changes that should improve the behavior here (fewer jobs per node, as well as more space) -- I'll update this PR once those are rolled out.

I've made the changes and retriggered this build. This (and future builds) shouldn't hit any issues related to full volumes.

wadells (approved):

PR #2551, opened 2021-06-22
allow gravity workers to install into separate regions

knisbet:
## Description

Quick and dirty gravity patch that allows worker nodes to join from a separate region in AWS.


wadells (approved):
Can you show how you tested this?  Code looks reasonable.

knisbet:
> Can you show how you tested this? Code looks reasonable.

https://github.com/gravitational/cloud/pull/740 - good luck following that though :) 

Basically just built a multi-region test cluster in the dev account with those terraform changes I was working on and needed this tweak to get the scripted join to work.

Issue #2557, opened 2021-06-23
tele build error "failed to export image layers"


wadells:
### Description

**What happened**:

Users reported `tele build` failing on 8.0.0-beta.1 when it previously worked on 8.0.0-beta.0.

```
[2021-06-23T19:19:31.855Z] Wed Jun 23 19:19:31 UTC	Build finished in 4 seconds
[2021-06-23T19:19:31.855Z] [ERROR]: failed to export image layers
[2021-06-23T19:19:31.855Z] 	Error response from daemon: Get https://leader.telekube.local:5000/v2/: dial tcp: lookup leader.telekube.local: Temporary failure in name resolution
[2021-06-23T19:19:31.855Z]
```

**What you expected to happen**:

The same build succesfully created an installer tar with 8.0.0-beta.0.

### Environment

 - Tele version : 8.0.0-beta.1
 - OS :

wadells:
@helgi: Can you add minimal public repro steps here?

a-palchikov:
It looks like some (docker) image references might be pointing to `leader.telekube.local` (internal cluster registry) during the build? The `build` rewrites the image references to point to the internal registry before they're placed into the tarball but the original references should not reference the internal registry or these resources should not be looked at by `build`. This would break on both versions though.
Debug logs will help here (`tele build --debug`).

wadells:
> It looks like some (docker) image references might be pointing to `leader.telekube.local` (internal cluster registry) during the build?

I found out that this happened during a re-pack of the opscenter tarball.  Apparently the workflow is something like:

* download opscenter.tar
* unpack the installer
* build a new tarball using an edited copy of the vendored app.yaml

Apparently this worked in builds prior to 8.0.0-beta.0.  It isn't a workflow that we test at Teleport.

PR #2558, opened 2021-06-24
Do not export images starting with leader.telekube.local

whiterm:
## Description
Images defined for the internal registry starting with `leader.telekube.local: 5000` must be excluded from the export list.

## Type of change

* Refs #2557

* Requires #

* Ports #

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [ ] Write documentation
- [ ] Address review feedback

### DIFF
--- a/lib/app/service/docker.go
+++ b/lib/app/service/docker.go
@@ -69,6 +69,22 @@ func exportLayers(
 	return nil
 }
 
+// excludeImagesStartingWith excludes the items witch start with prefix
+// It works with zero memory allocation and changes the input slice
+func excludeImagesStartingWith(images []string, prefix string) []string {
+	if prefix == "" {
+		return images
+	}
+	x := 0
+	for _, image := range images {
+		if !strings.HasPrefix(image, prefix) {
+			images[x] = image
+			x++
+		}
+	}
+	return images[:x]
+}
+
 // parseImageNameTag parses the specified image reference into name/tag tuple.
 // The returned name will include domain/path parts merged in a way to conform
 // to telekube package name syntax.
--- a/lib/app/service/docker_test.go
+++ b/lib/app/service/docker_test.go
@@ -18,6 +18,8 @@ package service
 
 import (
 	"fmt"
+	"reflect"
+	"testing"
 
 	"github.com/gravitational/gravity/lib/compare"
 	"github.com/gravitational/gravity/lib/loc"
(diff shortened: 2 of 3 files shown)


lib/app/service/docker_test.go
```diff
@@ -82,3 +84,63 @@ func (s *VendorSuite) TestGeneratesProperPackageName(c *C) {
 		c.Assert(*runtimePackage, compare.DeepEquals, testCase.result, comment)
 	}
 }
+
+func Test_excludeImagesStartingWith(t *testing.T) {
+	type args struct {
+		images []string
+		s      string
```
a-palchikov, line 91: What's an `s`? Would a `searchPattern` make for a better name?

lib/app/service/docker_test.go
```diff
@@ -82,3 +84,63 @@ func (s *VendorSuite) TestGeneratesProperPackageName(c *C) {
...
+		{
+			name: "exclude images. result is not empty",
+			args: args{
+				images: []string{"registry1/name1", "registry1/name2", "registry2/name2", "registry2/name3"},
+				s:      "registry2",
+			},
+			want: []string{"registry1/name1", "registry1/name2"},
+		},
+		{
+			name: "doesn't exclude images",
```
a-palchikov, line 131:
```suggestion
			name: "doesn't exclude images since the pattern does not match",
```

lib/app/service/docker_test.go
```diff
@@ -82,3 +84,63 @@ func (s *VendorSuite) TestGeneratesProperPackageName(c *C) {
...
+		{
+			name: "exclude all images. result is empty slice",
+			args: args{
+				images: []string{"registry1/name1", "registry1/name2"},
+				s:      "registry1",
+			},
+			want: []string{},
+		},
+		{
+			name: "exclude images. result is not empty",
```
a-palchikov, line 123:
```suggestion
			name: "exclude a subset of images (partial match)",
```

lib/app/service/docker_test.go
```diff
@@ -82,3 +84,63 @@ func (s *VendorSuite) TestGeneratesProperPackageName(c *C) {
...
+		{
+			name: "empty slice",
+			args: args{
+				images: []string{},
+				s:      "123",
+			},
+			want: []string{},
+		},
+		{
+			name: "empty pattern",
```
a-palchikov, line 107:
```suggestion
			name: "empty pattern - nothing is excluded",
```

lib/app/service/docker_test.go
```diff
@@ -82,3 +84,63 @@ func (s *VendorSuite) TestGeneratesProperPackageName(c *C) {
...
+		images []string
+		s      string
+	}
+	tests := []struct {
+		name string
+		args args
+		want []string
+	}{
+		{
+			name: "empty slice",
```
a-palchikov, line 99:
```suggestion
			name: "empty input - no changes",
```

lib/app/service/docker_test.go
```diff
@@ -82,3 +84,63 @@ func (s *VendorSuite) TestGeneratesProperPackageName(c *C) {
...
+			want: []string{"registry1/name1", "registry1/name2"},
+		},
+		{
+			name: "doesn't exclude images",
+			args: args{
+				images: []string{"registry1/name1", "registry1/name2", "registry2/name2"},
+				s:      "registry3",
+			},
+			want: []string{"registry1/name1", "registry1/name2", "registry2/name2"},
+		},
```
a-palchikov, line 137: I would add tests that mix the pattern - i.e. you have pattern in the beginning test but but no pattern in the middle and pattern as a suffix.

lib/app/service/docker.go
```diff
@@ -69,6 +69,17 @@ func exportLayers(
 	return nil
 }
 
+func excludeImagesStartingWith(images []string, s string) []string {
```
a-palchikov, line 72:
```suggestion
func excludeImagesStartingWith(images []string, prefix string) []string {
```
Also, check for empty prefix at the start:
```go
func excludeImagesStartingWith(images []string, prefix string) []string {
	if prefix == "" {
		return images
	}
	i := 0 // i is a more popular index name :)
	for _, image := range images {
		if !strings.HasPrefix(image, prefix) {
			images[i] = image
			i++
		}
	}
	return images[:i]
}
```

a-palchikov (approved):

knisbet (approved):
I'm good as long as Dima's comments get addressed.

helgi:
I noticed in prior builds when I had leader.telekube.local (hub image I extracted and repackaged) that it would magically pull it from quay.io -- Not sure which approach is better, exclude or swap with quay to re-add to the internal registry

wadells (approved):

PR #2585, opened 2021-07-26
Add GEP to change DNS balancer to a software loadbalancer

whiterm:
## Description
It is Gravity Enhancement Proposal to change DNS balancer to a software loadbalancer

## Type of change
New feature (non-breaking change which adds functionality)

### DIFF
--- a/enhancements/gep/001-loadbalancer/README.md
+++ b/enhancements/gep/001-loadbalancer/README.md
@@ -1,0 +1,119 @@
+# Gravity Enhancement Proposal
+
+## GEP-001: Change DNS balancer to a software loadbalancer (HAProxy)
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+    - [User Stories (Optional)](#user-stories-optional)
+        - [Story 1](#story-1)
+        - [Story 2](#story-2)
+    - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+    - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+    - [Test Plan](#test-plan)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+    - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+- [Implementation History](#implementation-history)
+
+<!-- /toc -->
+
+## Summary
+The main goal of this proposal is to replace dns balancing with a software loadbalancer.
+Thus, all requests to the kube-apiserver from Kubernetes components (kubelet, scheduler, control-manager, kube-proxy)
+will go through this loadbalancer. Also, requests to the docker registry must also go through the loadbalancer.
+
+## Motivation
+At the moment, Gravity is using CoreDNS to determine which master to send requests to,
+but DNS is not a good tool for balancing traffic.
+This proposal aims to provide the following improvements:
+  1. Evenly distribute the load between all apiserver replicas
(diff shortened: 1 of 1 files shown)


wadells:
Hi @whiterm: you should bring this up at one of the regular sync meetings -- to figure out what a reasonable path forward is.

enhancements/gep/001-loadbalancer/README.md
```diff
@@ -0,0 +1,119 @@
...
+To achieve this, docker registry will use port 5001 with port 5000 left to the loadbalancer.
+
+### User Stories
+#### Story 1
+As a cluster administrator, I can configure the loadbalancer when creating a cluster. If I set the type of loadbalancer to external, I can provide the external address in configuration.
+#### Story 2
+As a cluster administrator, I can change the loadbalancer settings on a running cluster.
+
+### Risks and Mitigations
+As the `leader.telekube.local` domain will not point to a master node, the applications using this address should be updated accordingly.
```
knisbet, line 61: If you're going to support changing this name, it's going to take more work as this is also part of the mTLS authentication between nodes.

enhancements/gep/001-loadbalancer/README.md
```diff
@@ -0,0 +1,119 @@
...
+kind: ClusterConfiguration
+version: v1
+spec:
+  loadbalancer:
+    # default value is internal
+    type: "external|internal"
+    # gravity uses this field when type is external
+    externalAddress: "IP address | DNS address"
+```
+The docker image references should not be changed, and they should start with `leader.telekube.local:5000/`.
```
knisbet, line 85: One of the problems gravity has, is it uses gravity-site to synchronize the registries. So on a multi-master install, I believe it only bootstraps the first registry, and then waits for gravity-site to sync the container images once it's loaded. On the current design, this causes problems if there is a failover of the leader within the first minute of install, as the registry may be empty. If this happens before gravity-site starts, then it can't start and sync the registry.

enhancements/gep/001-loadbalancer/README.md
```diff
@@ -0,0 +1,119 @@
...
+At the moment, Gravity is using CoreDNS to determine which master to send requests to,
+but DNS is not a good tool for balancing traffic.
+This proposal aims to provide the following improvements:
+  1. Evenly distribute the load between all apiserver replicas
+  1. Improve the stability and scalability of the cluster
+  1. Nodes will react faster to master failures
+  1. Cluster upgrades will be faster and more efficient
+
+### Goals
+When creating a cluster, the user can choose the load balancer to be either external or internal.
```
knisbet, line 39: It's not clear to me what the desire to support this with external load balancers is, but that could be internal requirements. However, I would just caution being careful with supporting lots of different models, if the built-in load balancer will sufficiently support the needs.

knisbet (commented):
Fede asked me to do a quick review earlier in the week. So this is my couple minute but not in-depth review at all into the proposal.

Another item you have to watch out for is tracking who is a current master in the cluster. This can happen in any cluster, say controller 1 with ip 1.1.1.1 dies and needs to be replaced. But instead of using the same IP, the new controller 1 has an IP of 1.1.1.5. All the load balancers need to discover that topology change of the controllers and update themselves. 

You might be able to leverage the code that already exists in planet that does this tracking for etcd. The etcd proxy has this same problem, so there is a bunch of code that tracks the member list and persists it to disk to survive reboots. https://github.com/gravitational/planet/blob/master/tool/planet/etcd.go#L444-L533

Infact the etcdgateway itself is a simple L4 proxy, so you might be able to consider it as an option. No idea if it will have the desired properties of haproxy though for detecting a dead node.

wadells:
It seems unlikely that we'll merge this design document to the repo (even if we do end up merging the features it describes).  Please reopen if/when you'd like more feedback.

PR #2587, opened 2021-07-26
Fix a data race in fsm plan follow tests

### LINKED ISSUE #2586: Data races in fsm tests
FSM plan follow tests are flaky, reference: https://drone.teleport.dev/gravitational/gravity/272/1/8

a-palchikov:
## Description

Reference: https://drone.teleport.dev/gravitational/gravity/272/1/8

## Type of change

* Closes https://github.com/gravitational/gravity/issues/2586

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

### DIFF
--- a/lib/fsm/follow.go
+++ b/lib/fsm/follow.go
@@ -81,46 +81,19 @@ func FollowOperationPlan(ctx context.Context, getPlan GetPlanFunc) <-chan PlanEv
 			}
 		}
 		ticker := backoff.NewTicker(getFollowStepPolicy())
-		tickerC := ticker.C
-		var errorTicker *backoff.Ticker
-		defer func() {
-			if ticker != nil {
-				ticker.Stop()
-			}
-			if errorTicker != nil {
-				errorTicker.Stop()
-			}
-		}()
-		resetBackoff := func() {
-			if errorTicker != nil {
-				errorTicker.Stop()
-				errorTicker = nil
-			}
-			ticker = backoff.NewTicker(getFollowStepPolicy())
-			tickerC = ticker.C
-		}
-		startBackoff := func() {
-			if ticker != nil {
-				ticker.Stop()
-				ticker = nil
-			}
-			errorTicker = backoff.NewTicker(getFollowBackoffPolicy())
-			tickerC = errorTicker.C
-		}
+		defer ticker.Stop()
 		defer logrus.Info("Operation plan watcher done.")
 		for {
 			select {
-			case <-tickerC:
+			case <-ticker.C:
 				nextPlan, err := getPlan()
 				if err != nil {
 					logrus.WithError(err).Error("Failed to diff plans.")
-					startBackoff()
 					continue
 				}
 				changes, err := diffPlan(plan, *nextPlan)
 				if err != nil {
 					logrus.WithError(err).Error("Failed to diff plans.")
-					startBackoff()
 					continue
 				}
 				err = sendPlanChanges(ctx, changes, *nextPlan, ch)
@@ -129,14 +102,12 @@ func FollowOperationPlan(ctx context.Context, getPlan GetPlanFunc) <-chan PlanEv
 					return
 				}
 				if err != nil {
-					startBackoff()
 					continue
 				}
 				// Update the current plan for comparison on the next cycle and
 				// reset the backoff so the ticker keeps ticking every second
 				// as long as there are no errors.
 				plan = nextPlan
-				resetBackoff()
 			case <-ctx.Done():
 				return
 			}
@@ -178,20 +149,8 @@ func getPlanEvents(changes []storage.PlanChange, plan storage.OperationPlan) (ev
 	return events
 }
 
-// getFollowBackoffPolicy returns retry backoff policy for the plan follower.
-//
-// Backoff triggers when plan reload fails.
-func getFollowBackoffPolicy() backoff.BackOff {
-	return &backoff.ExponentialBackOff{
-		InitialInterval: time.Second,
-		Multiplier:      backoff.DefaultMultiplier,
-		MaxInterval:     5 * time.Second,
-		Clock:           backoff.SystemClock,
-	}
-}
-
 // getFollowStepPolicy returns the pacing policy for the plan follower
 // on the happy path
 func getFollowStepPolicy() backoff.BackOff {
-	return &backoff.ConstantBackOff{Interval: time.Second}
+	return &backoff.ConstantBackOff{Interval: 5 * time.Second}
 }
--- a/lib/fsm/follow.go
+++ b/lib/fsm/follow.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/gravitational/gravity/lib/storage"
+	"github.com/gravitational/trace"
 
 	"github.com/cenkalti/backoff"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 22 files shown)


lib/fsm/follow.go
```diff
@@ -65,60 +66,50 @@ func (e *PlanRolledBackEvent) isTerminalEvent() bool { return true }
...
 	ch := make(chan PlanEvent, 100)
 	// Send an initial batch of events from the initial state of the plan.
-	plan, err := getPlan()
-	if err != nil {
-		logrus.WithError(err).Error("Failed to load plan.")
-	}
-	if plan != nil {
-		events := getPlanEvents(GetPlanProgress(*plan), *plan)
-		// recreate channel with enough room for all events + some room for additional events
-		ch = make(chan PlanEvent, len(events)+10)
```
a-palchikov, line 75: I removed the resizing logic in favor of moving this part into a separate function to reuse in the loop. It did not seem beneficial (e.g. no recorded evidence it was optimizing anything).

lib/fsm/follow.go
```diff
@@ -65,60 +66,50 @@ func (e *PlanRolledBackEvent) isTerminalEvent() bool { return true }
...
+						ticker = backoff.NewTicker(getFollowStepPolicy())
+						tickerC = ticker.C
+					}
+					continue
 				}
-				// Update the current plan for comparison on the next cycle and
-				// reset the backoff so the ticker keeps ticking every second
-				// as long as there are no errors.
-				plan = nextPlan
-				tickerBackoff.Reset()
```
a-palchikov, line 121: This is in violation of the package documented behavior: https://github.com/cenkalti/backoff/blob/v4/ticker.go#L26-L28 hence I added another ticker for backoffs during errors.

lib/fsm/follow.go
```diff
@@ -65,60 +66,50 @@ func (e *PlanRolledBackEvent) isTerminalEvent() bool { return true }
 func FollowOperationPlan(ctx context.Context, getPlan GetPlanFunc) <-chan PlanEvent {
 	ch := make(chan PlanEvent, 100)
 	// Send an initial batch of events from the initial state of the plan.
-	plan, err := getPlan()
-	if err != nil {
-		logrus.WithError(err).Error("Failed to load plan.")
-	}
-	if plan != nil {
```
a-palchikov, line 72: This is a minor deviation from the existing behavior but I have not found evidence where both `plan` and `err` could be `nil`.

lib/fsm/follow.go
wadells: @knisbet you added this in your scalability work (#2426). Can you double check?

wadells (changes requested):
Thanks for tackling this data race!  The test mutex makes sense, but I'm having a deal of trouble understanding the `follow.go` refactoring.

Suggestions:
* keep the `getPlan` calls in `FollowOperationPlan` -- this gives `sendNextPlan` a cleaner signature (no need to return `nextPlan`, `plan` and `nextPlan` have the same types) and logic flow (only `err` to worry about upon return). Signature:
```go
func sendNextPlan(ctx context.Context, plan, nextPlan *storage.OperationPlan, ch chan<- PlanEvent) error { }
```
* Stick with `if err != nil` fault handling instead of inverting to `if err == nil`.  Inverting the error handling  makes it quite difficult to follow the expected flow of this.

lib/fsm/follow.go
knisbet: This wasn't an optimization, it was to fix that when we had more than 100 events the follow would only show the first 100 events of history and then only show new events. So on large plans it would print the 100 events, and then jump to like step 5000 if it was still running.

lib/fsm/follow.go
```diff
@@ -127,6 +118,30 @@ func FollowOperationPlan(ctx context.Context, getPlan GetPlanFunc) <-chan PlanEv
...
+		return nil, trace.Wrap(err)
+	}
+	changes, err := DiffPlan(plan, *nextPlan)
+	if err != nil {
+		logrus.WithError(err).Error("Failed to diff plans.")
+		return nil, trace.Wrap(err)
+	}
+	for _, event := range getPlanEvents(changes, *nextPlan) {
+		select {
+		case ch <- event:
```
knisbet, line 134: Taking a very quick look, it looks to me like sendNextPlan get's called synchronously ahead of returning the `ch` to the caller on the first invocation of FollowOperationPlan, but this is a blocking call to add to ch, so if there are more than 100 events to replay won't this just block until the context expires because nothing can consume from `ch`?

knisbet (changes requested):

lib/fsm/follow.go
a-palchikov: Right, this is indeed a problem.

a-palchikov:
While working on a stress test for the follow loop i stumbled on implicit mutating behavior of the plan APIs (implemented in lib/fsm/utils.go) and decided to give it a minor refactor:
  * made all read APIs explicitly non-mutating including passing by value where the pointer is not necessary
  * mutating APIs (like [MarkCompleted](https://github.com/gravitational/gravity/blob/dmitri/fix-fsm-test-race/lib/fsm/utils.go#L174) and an exported [fillSteps](https://github.com/gravitational/gravity/blob/dmitri/fix-fsm-test-race/lib/expand/builder.go#L411) do not mutating possible source references - e.g. if a plan value is a shallow copy of another, the plan-relevant bits (phases) are cloned instead.

Added tests for changes in behavior.

lib/fsm/follow.go
```diff
@@ -59,66 +61,82 @@ func (e *PlanRolledBackEvent) isTerminalEvent() bool { return true }
 
 // FollowOperationPlan returns a channel that will be receiving phase updates
 // for the specified plan.
+// The returned channel must be served to avoid blocking the producer as
```
a-palchikov, line 64: Moved the whole logic of the `FollowOperationPlan` inside of the goroutine to do away with buffered channel and avoid the issue in https://github.com/gravitational/gravity/pull/2426 entirely. All uses of the API were already servicing the channel so there was no need for the buffer.

lib/fsm/follow.go
a-palchikov: See my comment above - I don't think the buffer is needed in the first place.

lib/update/cluster/engine_test.go
```diff
@@ -232,7 +234,14 @@ func (p *testPhase2) Rollback(context.Context) error {
 }
 
 func (r *testReconciler) ReconcilePlan(ctx context.Context, plan storage.OperationPlan) (*storage.OperationPlan, error) {
+	changes, err := r.backend.GetOperationPlanChangelog(plan.ClusterName, plan.OperationID)
```
a-palchikov, line 237: As the original plan is never modified, the reconciler needs to actually apply changes even in the test.

lib/fsm/follow.go
```diff
@@ -59,66 +61,82 @@ func (e *PlanRolledBackEvent) isTerminalEvent() bool { return true }
...
+// no events are dropped.
 //
 // The watch will stop upon entering one of the terminal operation states, for
-// example if the obtained plan is completed or fully rolled back.
+// example if the obtained plan is completed or fully rolled back or when the specified
+// context has expired. In any case, the channel will be closed to signal completion
```
knisbet, line 69: It's probably the case but I haven't tried to check it, how sure are we that the context always gets cancelled if the client drops?

lib/fsm/follow.go
```diff
@@ -59,66 +61,82 @@ func (e *PlanRolledBackEvent) isTerminalEvent() bool { return true }
...
+		}()
+		resetBackoff := func() {
+			if errorTicker != nil {
+				errorTicker.Stop()
+				errorTicker = nil
 			}
+			ticker = backoff.NewTicker(getFollowStepPolicy())
+			tickerC = ticker.C
+		}
+		startBackoff := func() {
```
knisbet, line 102: I sort of wonder if swapping the timers complexity is really required. What I mean is will this follower really reduce enough load by backing off to make a different on transient errors? And if it does, could the normal ticker just be a little bit slower?

knisbet (approved):
I had a couple of thoughts, but overall I didn't see anything major on this revision

lib/fsm/follow.go
a-palchikov: good question. I can live with a single ticker and update the step to 3-5s - wdyt?

lib/fsm/follow.go
a-palchikov: Not sure I follow - drops as in stops servicing the channel from `FollowOperationPlan` without also cancelling the context?
Currently the only client is the `plan --tail` command. Do you mean something else?

knisbet: Sorry, for some reason I thought this was the service side of the call. If this is all running client side then it shouldn't be an issue.

lib/fsm/follow.go
knisbet: Yea, while I don't see a problem with the code as implemented, I think 3-5s is fine for the behaviour and probably alot simpler.

a-palchikov:
@wadells Can you take another look - I've addressed the comments.

wadells (approved):

PR #1749, opened 2020-06-22
[7.0.x] backport RPC credentials rotation from 5.5.x

a-palchikov:
## Description

The PR does not port the RPC `rotate` sub-command as the taken approach is to refresh the RPC credentials on each cluster upgrade operation.

## Type of change

* Bug fix (non-breaking change which fixes an issue)

## Linked tickets and other PRs

* Refs https://github.com/gravitational/gravity/issues/1740.

* Ports https://github.com/gravitational/gravity/pull/1747.

## TODOs

- [x] Self-review the change
- [ ] Write tests
- [x] Perform manual testing
- [ ] Address review feedback

#### Start upgrade to `7.0.14-dev.2` (from this branch)
```sh
$ ./gravity upgrade -m
```
#### Export credentials package to test pre-upgrade state
```sh
$ gravity package unpack --insecure --ops-url=https://gravity-site.kube-system.svc.cluster.local:3009 gravitational.io/rpcagent-secrets:0.0.1 creds
$ openssl x509 -noout -text -in creds/client.cert |grep -iE 'not before|not after'
            Not Before: Jul 17 14:24:09 2020 GMT
            Not After : Jul 15 15:24:09 2030 GMT
``` 
#### Run `/init` step to update RPC credentials and make sure credentials have been updated
```
$ $ ./gravity plan execute --phase=/init
$ gravity package unpack --insecure --ops-url=https://gravity-site.kube-system.svc.cluster.local:3009 gravitational.io/rpcagent-secrets:0.0.1 creds2
$ openssl x509 -noout -text -in creds2/client.cert |grep -iE 'not before|not after'
            Not Before: Jul 17 14:29:12 2020 GMT
            Not After : Jul 15 15:29:12 2030 GMT
```
#### Roll back `/init` step to restore credentials
```
$ ./gravity plan rollback --phase=/init
$ gravity package unpack --insecure --ops-url=https://gravity-site.kube-system.svc.cluster.local:3009 gravitational.io/rpcagent-secrets:0.0.1 creds3
$ openssl x509 -noout -text -in creds3/client.cert |grep -iE 'not before|not after'
            Not Before: Jul 17 14:24:09 2020 GMT
            Not After : Jul 15 15:24:09 2030 GMT
```
#### Complete upgrade and make sure it's successful
```sh
$ gravity package unpack --insecure --ops-url=https://gravity-site.kube-system.svc.cluster.local:3009 gravitational.io/rpcagent-secrets:0.0.1 creds4
[...]

PR #2595, opened 2021-07-30
Forward port RPC credentials rotation on upgrade

a-palchikov:
## Description

Forward port of https://github.com/gravitational/gravity/pull/1749 to bring feature parity with 7.x.
Part of a bigger attempt to forward-port intermediate version support to 9.x.

## Type of change

* Closes #, refs #

* Ports https://github.com/gravitational/gravity/pull/1749

## TODOs

- [x] Self-review the change
- [ ] Write tests
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

1. Install a 3-node cluster
2. Attempt to upgrade a 3-node cluster from base version 8.0.0-beta.1.5
  2.1. Trigger manual upgrade
  2.2. Step over /init to rotate RPC credentials
  2.3. Roll back the operation (to also restore credenetials)
3. Upgrade a 3-node cluster from base version 8.0.0-beta.1.5

### DIFF
--- a/lib/blob/fs/fs.go
+++ b/lib/blob/fs/fs.go
@@ -27,17 +27,25 @@ import (
 
 	"github.com/gravitational/gravity/lib/blob"
 	"github.com/gravitational/gravity/lib/defaults"
+	"github.com/gravitational/gravity/lib/systeminfo"
 
 	"github.com/gravitational/trace"
 	log "github.com/sirupsen/logrus"
 )
 
-func New(path string) (blob.Objects, error) {
-	if path == "" {
-		return nil, trace.BadParameter("missing Path parameter")
+// New creates a new instance of the local fs blob service
+// rooted as the given path
+func New(root string) (blob.Objects, error) {
+	return NewWithConfig(Config{Path: root})
+}
+
+// NewWithConfig creates a new instance of the local fs blob service with the specified configuration
+func NewWithConfig(config Config) (blob.Objects, error) {
+	if err := config.checkAndSetDefaults(); err != nil {
+		return nil, trace.Wrap(err)
 	}
-	o := &objects{dir: path}
-	for _, d := range []string{o.tempDir(), o.blobDir()} {
+	o := &objects{config: config}
+	for _, d := range []string{config.tempDir(), config.blobDir()} {
 		if err := os.MkdirAll(d, defaults.SharedDirMask); err != nil {
 			return nil, trace.ConvertSystemError(err)
 		}
@@ -45,16 +53,31 @@ func New(path string) (blob.Objects, error) {
 	return o, nil
 }
 
+// Config defines the blob service configuration
+type Config struct {
+	// Path specifies the directory for blobs
+	Path string
+	// User optionally specifies the user context for file operations
+	User *systeminfo.User
+}
+
+func (r *Config) checkAndSetDefaults() error {
+	if r.Path == "" {
+		return trace.BadParameter("missing Path parameter")
+	}
+	return nil
+}
+
+func (r Config) tempDir() string {
+	return filepath.Join(r.Path, "tmp")
+}
+
+func (r Config) blobDir() string {
+	return filepath.Join(r.Path, "blobs")
+}
+
 type objects struct {
-	dir string
-}
-
-func (o *objects) tempDir() string {
-	return filepath.Join(o.dir, "tmp")
-}
-
-func (o *objects) blobDir() string {
-	return filepath.Join(o.dir, "blobs")
+	config Config
 }
 
 // hashDir helps us to organize the blobs in the folder -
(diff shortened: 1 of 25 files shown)


lib/localenv/clusterenv.go
```diff
@@ -122,11 +169,39 @@ func newClusterEnvironment(config clusterEnvironmentConfig) (*ClusterEnvironment
...
+	if config.nodeAddr == "" {
+		objects, err = fs.New(packagesDir)
+		if err != nil {
+			return nil, trace.Wrap(err)
+		}
+	} else {
+		// To be able to use cluster-level package service,
+		// a node address is required. This is not available on nodes prior to upgrade
+		// with an version that did not support the necessary system metadata (node address
+		// and service user).
```
wadells, line 190: When were these added?  Is this concern still relevant in 9.0?

lib/blob/fs/fs.go
```diff
@@ -27,34 +27,57 @@ import (
...
 
 	"github.com/gravitational/trace"
 	log "github.com/sirupsen/logrus"
 )
 
-func New(path string) (blob.Objects, error) {
-	if path == "" {
-		return nil, trace.BadParameter("missing Path parameter")
+// New creates a new instance of the local fs blob service
+// rooted as the given path
```
wadells, line 37:
```suggestion
// rooted at the given path
```

lib/update/cluster/phases/bootstrap.go
```diff
@@ -214,7 +224,19 @@ func (p *updatePhaseBootstrap) updateDNSConfig() error {
...
+	return p.HostLocalBackend.SetNodeAddr(p.Server.AdvertiseIP)
+}
+
+// updateServiceUser persists the service user in the local state database
+func (p *updatePhaseBootstrap) updateServiceUser() error {
+	p.Infof("Update service user as %v.", p.ServiceUser)
+	return p.HostLocalBackend.SetServiceUser(p.ServiceUser.OSUser())
+}
+
+func (p *updatePhaseBootstrap) pullSystemUpdates(ctx context.Context) error {
```
wadells, line 239: `ctx` seems unused?

wadells (commented):
Does this need to go to 8.0.x too?  7.0 -> 9.0 isn't possible with a HA config, based on my understanding.

Code looks reasonable -- but I'm not familiar with the original patches.

Testing?

lib/localenv/clusterenv.go
a-palchikov: This is still applicable to 9.x.
These changes have roots in 5.5.x in the case when RPC credentials expired if the cluster was not upgraded in a year. For 5.5.x we added a specialized command to rotate them which was not ported for various reasons (not that it needs to be). As part of the problem with working with cluster packages using CLI commands and direct access to the etcd backend (as cluster package service was out of question for various reasons, the main reason is that the command needed to be able to operate even with the cluster package service inaccessible). Additionally, when working with cluster package store, it is important to make modifications to the fs using the system user instead of the effective user context (`root` most of the time) as the changes under a different user might break fs access for the k8s workloads that did not run under `root`.

lib/storage/keyval/system.go
```diff
@@ -50,3 +50,33 @@ func (b *backend) GetSELinux() (enabled bool, err error) {
...
+	}
+	return nodeAddr, nil
+}
+
+// SetNodeAddr sets current node advertise IP
+func (b *backend) SetNodeAddr(addr string) error {
+	return b.upsertVal(b.key(systemP, nodeAddrP), addr, forever)
+}
+
+// GetServiceUser returns the current serviceo user
```
knisbet, line 69:
```suggestion
// GetServiceUser returns the current service user
```

lib/storage/keyval/system.go
```diff
@@ -50,3 +50,33 @@ func (b *backend) GetSELinux() (enabled bool, err error) {
 func (b *backend) SetSELinux(enabled bool) error {
 	return b.upsertVal(b.key(systemP, seLinuxP), &enabled, forever)
 }
+
+// GetNodeAddr returns the current node advertise IP
+func (b *backend) GetNodeAddr() (addr string, err error) {
```
knisbet, line 55: I realize this is probably just part of the port and already this way on other branches, I imagine it applies to just the local backend. I just find it funny, because the other backend is etcd, in which case each node would clobber others writes.

Not asking for any changes, just something I found interesting.

knisbet (approved):
Just looked through quickly.

lib/storage/keyval/system.go
a-palchikov: Agree - moved the system metadata APIs to a separate interface in https://github.com/gravitational/gravity/commit/c059b97e

bernardjkim (approved):

wadells (approved):

PR #2606, opened 2021-08-19
Remove sles/SUSE from robotest configs

wadells:
## Description
We not aware of any customers that use SLES, and this test is
currently broken because GCP removed the SLES disk image robotest
depends on, resulting in the following error:

   Error: Error resolving image name 'suse-cloud/sles-15-sp2-v20201014': Could not find image or family suse-cloud/sles-15-sp2-v20201014

## Type of change

* Works around https://github.com/gravitational/robotest/issues/290

## TODOs
- [x] Self-review the change
- [ ] Verify CI Passes
- [ ] Address review feedback

## Testing done
None.  If CI passes this is good to go.

## Additional information
After a discussion with @knisbet in slack, we determined we weren't getting a good ROI running
SLES tests regularly.  I'll fix/remove this upstream in Robotest at some point (see https://github.com/gravitational/robotest/issues/290), but this at least will unblock gravity merges.

This needs backports to 8.0.x and 7.0.x, but no further, as we don't test against sles on 6.1/5.5.

### DIFF
--- a/assets/robotest/config/nightly.sh
+++ b/assets/robotest/config/nightly.sh
@@ -9,8 +9,8 @@ source $(dirname $0)/lib/utils.sh
 declare -A UPGRADE_MAP
 
 # Use a fixed tag until we cut our first non-pre-release, as recommended_upgrade_tag skips pre-releases
-# UPGRADE_MAP[$(recommended_upgrade_tag $(branch 9.0.x))]="redhat:8.4 redhat:7.8 centos:8.4 centos:7.8 sles:12-sp5 sles:15-sp2 ubuntu:16 ubuntu:18 ubuntu:20 debian:9 debian:10"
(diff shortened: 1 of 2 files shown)


knisbet:
Just leaving a note, we don't know for sure that no customers use SUSE at all. But the enterprise customer who requested SUSE support and had support tickets around SUSE is no longer a customer. The current set of customers do not use SUSE as far as we are aware, which is why we're removing this from our test matrix but could still investigate specific problems that may occur.

wadells:
> Just leaving a note, we don't know for sure that no customers use SUSE at all. But the enterprise customer who requested SUSE support and had support tickets around SUSE is no longer a customer. The current set of customers do not use SUSE as far as we are aware, which is why we're removing this from our test matrix but could still investigate specific problems that may occur.

I updated my description a bit to better reflect this.

bernardjkim (approved):

PR #2637, opened 2021-09-13
[8.0.x] adjust clusterconfiguration on upgrade

a-palchikov:
## Description

Implicitly adjust cluster configuration on featuregates when upgrading to 8.x (k8s 1.19.12+).
This primarily concerns the configuration that explicitly turns `EndpointSlice` feature gate flag off as in this case, it might be problematic to upgrade as this flag needs to be paired with `EndpointSliceProxying` which is only [available](https://github.com/kubernetes/kubernetes/blob/v1.18.0/pkg/features/kube_features.go#L513-L517) since k8s `1.18`. 
The flag had been moved to Beta in k8s `1.18` and stabilized in `1.20`.  In future versions it will not be possible to turn it off (k8s `1.21+`).

## Type of change

## TODOs

- [x] Self-review the change
- [ ] Write tests
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

Tested on upgrades from 7.x to 8.x with the following configuration:
```yaml
# config.yaml
kind: ClusterConfiguration
version: v1
spec:
  global:
    featureGates: 
      AllAlpha: true
      APIResponseCompression: false
      BoundServiceAccountTokenVolume: false
      CSIMigration: false
      KubeletPodResources: false
      EndpointSlice: false
      IPv6DualStack: false
      RemoveSelfLink: false
```
and ensuring that the resulting `ClusterConfiguration` does not have the `EndpointSlice` feature gate turned off after the upgrade.

### DIFF
--- a/lib/update/cluster/phases/init.go
+++ b/lib/update/cluster/phases/init.go
@@ -34,6 +34,7 @@ import (
 	"github.com/gravitational/gravity/lib/storage"
 	"github.com/gravitational/gravity/lib/storage/clusterconfig"
 	"github.com/gravitational/gravity/lib/users"
+	"github.com/gravitational/gravity/lib/utils"
 
 	"github.com/gravitational/rigging"
 	"github.com/gravitational/trace"
@@ -78,6 +79,7 @@ type updatePhaseInit struct {
 	existingDNS                storage.DNSConfig
 	existingEnviron            map[string]string
 	existingClusterConfigBytes []byte
+	updatedClusterConfigBytes  []byte
 	existingClusterConfig      clusterconfig.Interface
 }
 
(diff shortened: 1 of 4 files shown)


bernardjkim (approved):

lib/update/cluster/phases/init.go
```diff
@@ -511,11 +536,15 @@ func removeLegacyUpdateDirectory(log log.FieldLogger) error {
...
+func (p *updatePhaseInit) Rollback(ctx context.Context) error {
+	if err := p.revertClusterConfig(); err != nil {
 		return trace.Wrap(err)
 	}
-	return nil
+
+	b := utils.NewExponentialBackOff(defaults.ShutdownTimeout)
+	return utils.RetryTransient(ctx, b, func() error {
+		return p.removeConfiguredPackages()
+	})
```
wadells, line 547: This is just opportunistic improvement, right?  Or does it somehow relate to the main changes?

wadells (approved):
LGTM.

lib/update/cluster/phases/init.go
a-palchikov: Sorry, yes - an opportunistic improvement since it fails fairly consistently on rollback.

PR #2638, opened 2021-09-15
[8.0.x] serf deprecation mode

a-palchikov:
- [Linked tickets and other PRs](#linked-tickets-and-other-prs)
- [Summary](#summary)
- [Motivation](#motivation)
  - [Goals](#goals)
- [Proposal](#proposal)
- [Design Details](#design-details)
- [Constraints and Conflict Resolution](#constraints-and-conflict-resolution)
  - [Test Plan](#test-plan)

Requires https://github.com/gravitational/satellite/pull/302, and https://github.com/gravitational/planet/pull/860.

## Summary

Gravity 7.x employs [serf](https://github.com/hashicorp/serf) as a cluster membership coordinator. Gravity 8.x removes
serf in favor of the k8s-based membership package. As no smooth transition had been
provided, upgrades of larger clusters that skip the worker nodes for efficiency, go through the state when members report membership failures which can be cumbersome to deal with or even prevent the upgrade from working.

## Motivation

In order to provide for smooth transition during upgrades, serf infrastructure layer needs to be removed in a non-destructive way that guarantees proper inter-cluster communication even during partial version bumps - e.g. when employing different upgrade strategies for subsets of the clusters.

## Goals

Allow gradual roll-out of newer versions for clusters based on Gravity 7.x.

### Reintroduce Serf

To keep the inter-node status upgrades uninterrupted, 8.x upgrade will continue to bundle the serf binary and the corresponding `systemd` service unit. The service unit will be conditionally enabled during upgrade but will not be used for new installations.

The planet environment configuration will include an environment variable `PLANET_UPGRADE_FROM7` to indicate that there's an ongoing upgrade from 7.x.

The serf agent will be used to join the existing serf cluster but will otherwise not be used for the purpose of maintaining an agent cluster - the agents will continue to use kubernetes for membership tests.

### Planet agent

New agent gRPC status attribute will be introduced to relay the version of the binary. This will usually coincide with the version of the cluster.
Additionally, a new `NodeName` attribute is added to explicitly identify the agent with a kubernetes node name. The attribute is added to both `NodeStatus` as well as `MemberStatus` messages to avoid confusion and ambiguity:

```proto
message NodeStatus {
  // Name identifies this agent to the cluster.
  // Obsolete: use NodeName instead
  string name = 1;
 
  // ... existing fields

  // Version specifies the agent version.
  Version version = 5;

  // NodeName identifies this agent to the cluster.
  string node_name = 6;
}

message MemberStatus {
  // Name identifies this agent to the cluster.
  // Obsolete: use NodeName instead
  string name = 1;
 
  // ... existing fields

  // NodeName identifies this agent to the cluster.
  string node_name = 5;
}

// Version describes the version of the agent
message Version {
    // Major specifies the major version component. E.g. 7 for 7.0.34
    int32 major = 1;
    // Minor specifies the minor version component. E.g. 0 for 7.0.34
    int32 minor = 2;
    // Patch specifies the patch version component. E.g. 34 for 7.0.34
    int32 patch = 3;
    // Prerelease specifies the pre-prelease metadata in free form
    string prerelease = 4;
}
```

The agent version might be used in the future to perform upgrade-specific customizations.

### Version Differentiation

An 8.x agent will be updated to dynamically discover the stable ID of the corresponding peer for internal management.

Currently, an 7.x agent uses `<public ip>.<cluster-name>` as the agent name while 8.x agent uses the node name as configured in the kubelet while both agents have `publicip` and `role` metadata attached to their status heartbeats.
In order to deal with this inconsistency, 8.x agents will use the `publicip` tag value as the common denominator for computing overall agent status in its own status heartbeat.

During upgrade to 8.x, the upgrade process will [store](https://github.com/gravitational/gravity/pull/2638/files#diff-632b61786da6848b5c08b4bbc63bf117534d8e025c16fb8d00a6dcf01cbe286bR263) the fact that the cluster has been upgraded from 7.x in the cluster metadata. This is to ensure uninterruptible rollout of worker nodes after upgrading the cluster master nodes to the new version. As new nodes join, they will still see that the compatibility flag is on and will continue operating in compatibility mode until all of the workers have been replaced.
The compatibility flag is only in effect on an 8.x cluster. 

### Upgrade progressions

After upgrading a cluster to 8.x, the serf service will become redundant and even though it will incur small performance overhead, the effects of this overhead are negligible and can be ignored.
After the cluster has been migrated to 9.x, the serf binary and the corresponding service will be finally removed.
As a performance optimization, new installations will have the service disabled.

### Risks

The only place that is potentially risky to change due to possible clients relying on the specific layout is the influxdb health status backend:
https://github.com/gravitational/satellite/blob/842725cda188d6c4a817b4445896fcac9e4a39de/agent/backend/influxdb/influxdb.go#L119-L121.

As the backend is not currently in use by planet, I see no risk in also updating it due to suggested changes. The node name has been also added as a tag so it could be used if necessary.

### Test Plan

Following test scenarios will have been verified for this PR:

 - New installation of 8.x
 - Upgrading an 7.x cluster to 8.x (this revision)
 - Upgrading an 7.x cluster to 8.x (this revision) with workers skipped and workers rolled out separately
 - Upgrading an 7.x cluster to 9.x (to be submitted) via 8.x (this revision)
 - Upgrading an 7.x cluster to 9.x (to be submitted) via 8.x (this revision) with workers skipped and workers rolled out separately


a-palchikov:
@gravitational/codeowners-gravity-code Let me know if this PR is too large for review and should be moved to a fork.

bernardjkim (approved):

wadells:
> @gravitational/codeowners-gravity-code Let me know if this PR is too large for review and should be moved to a fork.

I think this is worth taking -- I don't know how many large clusters we'll see outside MS upgrade from 7.0 to 8.0, but it definitely is important for folks who do that upgrade -- and doesn't present any maintenance burden since it isn't going to 9.0+

wadells (approved):
Thanks for your detailed description and implementation notes. It makes PR stacks like this a pleasure to review.  Your attention to detail is always appreciated!

Please update dependencies links before merge.

a-palchikov:
I was able to come up with a solution for the issue I've run into yesterday - namely, after upgrading only masters (via `--skip-workers`), the new worker nodes joined via `join/autojoin` would lose the knowledge about the serf compatibility, hence I've updated the upgrade step somewhat. Also added this bit in the description:

> During upgrade to 8.x, the upgrade process will store the fact that the cluster has been upgraded from 7.x in the cluster metadata. This is to ensure uninterruptible rollout of worker nodes after upgrading the cluster master nodes to the new version. As new nodes join, they will still see that the compatibility flag is on and will continue operating in compatibility mode until all of the workers have been replaced.
The compatibility flag is only in effect on an 8.x cluster.

There have been some [build issues](https://drone.teleport.dev/gravitational/gravity/441/1/8) with robotest image (`7.0.12`) I'm not sure what's up with it though.

wadells:
> There have been some [build issues](https://drone.teleport.dev/gravitational/gravity/441/1/8) with robotest image (`7.0.12`) I'm not sure what's up with it though.

I've seen this before -- it is an intermittent error.  I don't know what causes it, but I hypothesize it has to do with a download failure from get.gravitational.io.  It is typically infrequent, maybe 1/50 runs.  If you've see it repeatedly, we can likely invest some time in gathering better logs / profiling the failure.

PR #2642, opened 2021-09-17
[6.1] don't sync plan when doing plan following

### LINKED ISSUE #2641: gravity plan --tail excessive CPU usage
### Description

**What happened**:
Using `gravity plan --tail` results in excessive CPU usage, multiple cores worth of usage to follow the plan execution.

CPU usage to be mostly idle

Start a plan, such as an upgrade, and then run `gravity plan --tail`. Monitor the CPU usage of the host.

** Other Notes **
- There also appears to be a resource leak, as the tail runs, connections build up against etcd and the teleport SSH proxy.

knisbet:
## Description

1. Only initialize the environment client once and re-use for subsequent calls. Initializing the client repeatedly leads to resource leaks, as the etcd client and teleport client do not get collected when going out of scope.
2. Do not reconcile the plan when retrieving the plan during follow. Reconciling should happen automatically as phases complete, so having the CLI do so when tailing the plan leads to extra work.

The above combined were causing `./gravity plan --tail` to consume multiple cores worth of CPU usage during normal operation.

## Type of change

Fixes #2641 

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Implementation

The plan reconciliation sync the plan in the local database and etcd. This is done by the upgrade process, so having the CLI call this shouldn't be required, and it just wastes time and resource to run through this.

### DIFF
--- a/tool/gravity/cli/plan.go
+++ b/tool/gravity/cli/plan.go
@@ -27,13 +27,11 @@ import (
 	"github.com/gravitational/gravity/lib/ops"
 	"github.com/gravitational/gravity/lib/storage"
 	libenviron "github.com/gravitational/gravity/lib/system/environ"
-	"github.com/gravitational/gravity/lib/update"
 	clusterupdate "github.com/gravitational/gravity/lib/update/cluster"
 	"github.com/gravitational/gravity/lib/utils"
 
 	"github.com/fatih/color"
 	"github.com/gravitational/trace"
-	"github.com/sirupsen/logrus"
 )
 
 func initUpdateOperationPlan(localEnv, updateEnv *localenv.LocalEnvironment) error {
@@ -132,28 +130,29 @@ func displayClusterOperationPlan(localEnv *localenv.LocalEnvironment, opKey ops.
 	return outputOrFollowPlan(localEnv, getClusterOperationPlanFunc(localEnv, opKey), opts)
 }
 
-func getUpdateOperationPlanFunc(localEnv *localenv.LocalEnvironment, environ LocalEnvironmentFactory, opKey ops.SiteOperationKey) fsm.GetPlanFunc {
+func getUpdateOperationPlanFunc(environ LocalEnvironmentFactory, opKey ops.SiteOperationKey) (fsm.GetPlanFunc, error) {
+	updateEnv, err := environ.NewUpdateEnv()
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
 	return func() (*storage.OperationPlan, error) {
-		updateEnv, err := environ.NewUpdateEnv()
-		if err != nil {
-			return nil, trace.Wrap(err)
-		}
 		plan, err := fsm.GetOperationPlan(updateEnv.Backend, opKey)
 		if err != nil {
 			return nil, trace.Wrap(err)
 		}
-		reconciledPlan, err := tryReconcilePlan(context.TODO(), localEnv, updateEnv, *plan)
-		if err != nil {
-			logrus.WithError(err).Warn("Failed to reconcile plan.")
-		} else {
-			plan = reconciledPlan
-		}
+
 		return plan, nil
-	}
+	}, nil
 }
 
 func displayUpdateOperationPlan(localEnv *localenv.LocalEnvironment, environ LocalEnvironmentFactory, opKey ops.SiteOperationKey, opts displayPlanOptions) error {
-	return outputOrFollowPlan(localEnv, getUpdateOperationPlanFunc(localEnv, environ, opKey), opts)
+	f, err := getUpdateOperationPlanFunc(environ, opKey)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	return outputOrFollowPlan(localEnv, f, opts)
 }
 
 func getInstallOperationPlanFunc(opKey ops.SiteOperationKey) fsm.GetPlanFunc {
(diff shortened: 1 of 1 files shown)


tool/gravity/cli/plan.go
```diff
@@ -132,28 +132,29 @@ func displayClusterOperationPlan(localEnv *localenv.LocalEnvironment, opKey ops.
 	return outputOrFollowPlan(localEnv, getClusterOperationPlanFunc(localEnv, opKey), opts)
 }
 
-func getUpdateOperationPlanFunc(localEnv *localenv.LocalEnvironment, environ LocalEnvironmentFactory, opKey ops.SiteOperationKey) fsm.GetPlanFunc {
+func getUpdateOperationPlanFunc(localEnv *localenv.LocalEnvironment, environ LocalEnvironmentFactory, opKey ops.SiteOperationKey) (fsm.GetPlanFunc, error) {
```
wadells, line 135: `localEnv` is no longer needed without reconciliation.

tool/gravity/cli/plan.go
```diff
@@ -132,28 +132,29 @@ func displayClusterOperationPlan(localEnv *localenv.LocalEnvironment, opKey ops.
...
 	return func() (*storage.OperationPlan, error) {
-		updateEnv, err := environ.NewUpdateEnv()
-		if err != nil {
-			return nil, trace.Wrap(err)
-		}
 		plan, err := fsm.GetOperationPlan(updateEnv.Backend, opKey)
 		if err != nil {
 			return nil, trace.Wrap(err)
 		}
-		reconciledPlan, err := tryReconcilePlan(context.TODO(), localEnv, updateEnv, *plan)
```
wadells, line 145: This is the only place `tryReconcilePlan` was used, should it be deleted?

wadells (commented):
Testing?

> etcd client and teleport client do not get collected when going out of scope.

I'm not sure I understand why these aren't GC'd -- any notes for a noob like me as to why?

More generally, I'm not sure I understand enough about the `localEnv *localenv.LocalEnvironment` vs the `environment LocalEnvironmentFactory` parameters to grok the impacts of this change.  I'd be very wary of merging without thorough testing or approval from someone who knows this system better than I do.

I can come back and do a more detailed pass if we don't find a better reviewer.

Edit: I was comparing to master locally -- not 6.1.  I may need to update some of my remarks.
Edit2: Doesn't look like there were relevant differences in the immediate codepaths between 6.1 and master -- should this be ported?

knisbet:
> Testing?
> 
> > etcd client and teleport client do not get collected when going out of scope.
> 
> I'm not sure I understand why these aren't GC'd -- any notes for a noob like me as to why?

AFAICT it's all handles to the resources are launched in alternate go routines, so it all looks active to the GC although unused. AFAICT on a quick pass, all the cancellation logic isn't wired in our wrappers, so we don't have a clean way to cancel the handles, and I don't think it's worth refactoring at this point.

> More generally, I'm not sure I understand enough about the `localEnv *localenv.LocalEnvironment` vs the `environment LocalEnvironmentFactory` parameters to grok the impacts of this change. I'd be very wary of merging without thorough testing or approval from someone who knows this system better than I do.

the tail functionality has been tested.

> I can come back and do a more detailed pass if we don't find a better reviewer.
> 
> Edit: I was comparing to master locally -- not 6.1. I may need to update some of my remarks.
> Edit2: Doesn't look like there were relevant differences in the immediate codepaths between 6.1 and master -- should this be ported?

Yes I'll port, I just developed this against 6.1 as that's where the original report was, and I had to instrument the code to figure out where the CPU usage was coming from.

knisbet:
> More generally, I'm not sure I understand enough about the localEnv *localenv.LocalEnvironment vs the environment LocalEnvironmentFactory parameters to grok the impacts of this change. I'd be very wary of merging without thorough testing or approval from someone who knows this system better than I do.

Keep in mind I wrote the original upgrade code for etcd. The key component to this, is during etcd upgrades that need to take etcd totally offline, nodes maintained a local replica of the upgrade state normally stored in etcd. The data itself is organized as a changelog, so a bi-direction sync needs to occur as steps are completed in the plan for the local replicas and etcd to be in sync. All steps that are completed while etcd is offline, are synced back to etcd when etcd comes back online in the upgrade. This sync and tracking will be handled by the upgrade code, as the upgrade code itself is responsible for upgrade state.

This is all CLI/UX code, I don't see a good reason that it should be forcing a sync of the local and cluster plan's, as the upgrade code is expected to do this when it produces an event. I think Dima touched / tried to tune this in some of his later changes, so it's as aggressive as it originally was. But I don't think that changes the core idea, display/CLI code shouldn't be responsible for modifying upgrade state by synchronizing between both backends. And this is extremely CPU expensive to trigger full plan syncs in a polling loop (in order to tail the plan and display changes as they are made), for data that will already be synchronized.

wadells (approved):
> I don't think that changes the core idea, display/CLI code shouldn't be responsible for modifying upgrade state by synchronizing between both backends.

Thanks for the context -- this is helpful.

> the tail functionality has been tested.

The standard I hold myself and peers to for testing is "show, don't tell". This means going into more detail, or (better) having added automated tests.  In this case, automated tests aren't realistic, but I'd shoot for something like: 

"I tested on a X node cluster (or however you were generating a big plan).  CPU usage was X before the change, and Y after."

We can talk more if we don't want to hold the gravity project to this same scrutiny anymore due to its deprecation, but seniority/SMEs get no free lunch, because testing is teaching folks like me how the system works. :smile: Including details serves as a how-to and precedent for similar work.

My approval  herein is more "I trust you and think we should move forward" than "I understand the code I've reviewed and deemed it safe".  Though I do understand a lot more about this code than I did two days ago thanks to your explanation. Thanks!

bernardjkim (approved):

PR #2658, opened 2021-09-29
Allow encryption of Kubernetes resources

bernardjkim:
## Description

These changes allow Kubernetes resources to be encrypted at rest.

aws encryption provider can now be specified in the cluster configuration.
```yaml
# config.yaml
...
# Enables Kubernetes resource encryption at rest.
# Rotating keys or disabling encryption requires manual configuration changes.
# See https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/ for more details.
encryptionProvider:
  disabled: false
  aws:
    accountID: "1234567890"
    keyID: "1234567890"
    # region should be left unspecified when using a multi-region key. The region
    # will be automatically selected from ec2 instance metadata.
    region: "us-west-2"
```
Set the `accountID` and `keyID`. If using a multi-region key leave the region field unspecified. Then run the gravity command `gravity resource create config.yaml`

Key rotation or disabling encryption is not automatically handled. Keys must be manually rotated first and then the cluster configuration should be updated to persist changes. Same with disabling encryption.

## Type of change

* Requires https://github.com/gravitational/planet/pull/865

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [x] Write documentation
- [x] Address review feedback

## Testing
- [x] Enable resource encryption on a single node
- [x] Test single region key while forcing leader change
- [x] Test multi region key while forcing leader change
- [x] Test key rotation
- [x] Test disable encryption
- [x] Test upgrade with resource encryption enabled

### DIFF
--- a/docs/9.x/config.md
+++ b/docs/9.x/config.md
@@ -65,6 +65,17 @@ spec:
     # but the selected backend can be overridden if specified in the cluster configuration.
     # Supports values "vxlan", "aws-vpc", and "gce".
     flannelBackend: "vxlan"
+    # Enables Kubernetes resource encryption at rest.
+    # Rotating keys or disabling encryption requires manual configuration changes.
+    # See https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/ for more details.
+    encryptionProvider:
+      disabled: false
+      aws:
+        accountID: "1234567890"
+        keyID: "mrk-1234567890"
+        # region should be left unspecified when using a multi-region key. The region
+        # will be automatically selected from ec2 instance metadata.
+        region: "us-west-2"
     # A set of key=value pairs that describe feature gates for alpha/experimental features
     featureGates:
       AllAlpha: true
--- a/docs/9.x/config.md
+++ b/docs/9.x/config.md
@@ -67,7 +67,8 @@ spec:
     flannelBackend: "vxlan"
     # Enables Kubernetes resource encryption at rest.
     # Rotating keys or disabling encryption requires manual configuration changes.
-    # See https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/ for more details.
+    # See https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/ and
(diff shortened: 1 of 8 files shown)


docs/9.x/config.md
```diff
@@ -65,6 +65,18 @@ spec:
...
+    # See https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/ and
+    # https://github.com/kubernetes-sigs/aws-encryption-provider for more details.
+    encryptionProvider:
+      disabled: false
+      aws:
+        accountID: "1234567890"
+        keyID: "mrk-1234567890"
+        # region should be left unspecified when using a multi-region key. The region
+        # will be automatically selected from ec2 instance metadata.
+        region: "us-west-2"
```
wadells, line 79: Given that a key ARN has a region in it (for all region specific keys), I'm wondering if we could simplify the UX here to only require users to put in a key ARN, from which the account/region/key id are derived internally by planet/gravity.

It isn't as configurable, but it also avoids extra knobs for twiddling.  Just a though -- you know these config paths better than I do.

wadells (commented):
Stuff looks reasonable here -- waiting on the discussion at https://github.com/gravitational/planet/pull/865 before approval, since that could affect the diff here anyhow.

docs/9.x/config.md
bernardjkim: I addressed the comment in planet here https://github.com/gravitational/planet/pull/865#discussion_r721825796.
I separated the fields to allow the use of either a multi-region key or a single-region key. 

But 🤔 I guess we could still use a single ARN for a multi-region key. Multi-region keys share the same keyID and just have different regions. And we can distinguish multi-region keys because the keyID is prefixed with `mrk-`. We could parse the ARN and setup proper configuration if a multi-region key is detected.

The cluster configuration output might be a bit awkward since it would display the ARN with which ever arbitrary region was selected when setting up encryption.

wadells (approved):

Issue #2682, opened 2021-11-11
Master node unable to start gravity-site Pod


a-palchikov:
### Description

**What happened**:
From one of the failed upgrades - the to be upgraded master node failed to start gravity-site Pod due to the failure to pull the image from the current leader. It is suspected to do with the fact that the upgrade was initiated immediately after joining a new master node to the cluster and the node's registry being empty before it became the leader.

**What you expected to happen**:
Master node should be able to come up back after upgrading the system container and able to start all services.

**How to reproduce it (as minimally and precisely as possible)**:
Sadly it is only a suspicion and a useful reproduction is not available.

### Environment

 - Gravity version [e.g. 7.0.11]: 7.0.34
 - Platform [e.g. Vmware, AWS]: aws

PR #2683, opened 2021-11-11
[8.0.x] export cluster application to registry on expand

a-palchikov:
## Description

This PR implements an extension to the join operation to export cluster application to the local node's registry if it's to become a new master node.
This is an attempt to mitigate a potential issue with the node becoming the leader before it has had time to populate the registry (which only happened in the registry synchronization loop inside the gravity-site) which can in turn lead to its inability to pull images for the gravity-site Pods.

## Type of change

* Refs https://github.com/gravitational/gravity/issues/2682

## TODOs

- [x] Self-review the change
- [x] Perform manual testing
- [ ] Address review feedback

## Testing done

Tested on both installs from this revision and expanding a cluster.


lib/utils/progress.go
```diff
@@ -37,9 +37,16 @@ import (
 // Printer describes a capability to output to standard output
 type Printer interface {
 	io.Writer
+	StepPrinter
```
a-palchikov, line 40: Had to add this to enable the image service to log progress as well.

mage/vars.go
```diff
@@ -291,7 +291,7 @@ func buildFlags() []string {
 }
 
 func importEnvFromMakefile() (env map[string]string) {
-	cmd := exec.Command("make", "-f", "Makefile.buildx", "magnet-vars")
+	cmd := exec.Command("make", "--quiet", "-f", "Makefile.buildx", "magnet-vars")
```
a-palchikov, line 294: Opportunistic improvement to avoid spam during build:
```
2021/11/11 15:38:57 Skip line that does not look like magnet envar: "make[1]: Entering directory '/path/to/go/src/github.com/gravitational/gravity'"
2021/11/11 15:38:57 Skip line that does not look like magnet envar: "make[1]: Leaving directory '/path/to/go/src/github.com/gravitational/gravity'"

```

lib/app/service/app.go
```diff
@@ -78,8 +78,6 @@ type Config struct {
 	CacheResources bool
 	// UnpackedDir is the dir where packages are unpacked
 	UnpackedDir string
-	// ExcludeDeps defines a list of dependencies that will be excluded for the app image
-	ExcludeDeps []loc.Locator
```
a-palchikov, line 82: I had to remove this since the implementation of filtering of the excluded applications is not complete - e.g. it does not work for the cluster application service running inside of the gravity-site Pod.
To counter this, I removed this configuration and instead compute the list of excluded applications each time it is required (see below).

lib/app/service/app.go
```diff
@@ -828,7 +820,10 @@ func (r *applications) resolveManifest(manifestBytes []byte) (*schema.Manifest,
 	}
 
 	message := "Dependency %v excluded from manifest"
-	manifest.Dependencies.Apps = appservice.Wrap(loc.Filter(appservice.Unwrap(manifest.Dependencies.Apps), r.Config.ExcludeDeps, message))
+	manifest.Dependencies.Apps = appservice.Wrap(
+		loc.Filter(appservice.Unwrap(manifest.Dependencies.Apps),
+			appservice.AppsToExclude(*manifest), message),
```
a-palchikov, line 825: See comment above - instead of using configuration which is not relayed everywhere, compute the list of excluded applications where necessary.

lib/expand/fsmspec.go
```diff
@@ -92,6 +93,15 @@ func FSMSpec(config FSMConfig) fsm.FSMSpecFunc {
 				config.Operator,
 				config.Runner)
 
+		case p.Phase.ID == installphases.ExportPhase:
```
bernardjkim, line 96: All the other cases check if the phase prefixes the ID. Is it okay to compare for an exact match here?

lib/install/phases/export.go
```diff
@@ -75,7 +74,7 @@ func NewExport(p fsm.ExecutorParams, operator ops.Operator, packages pack.Packag
 	}, nil
 }
 
-type exportExecutor struct {
+type ExportExecutor struct {
```
bernardjkim, line 77: Nit: Add some godocs for exported struct

bernardjkim (approved):

lib/expand/fsmspec.go
a-palchikov: It is - as opposed to installation case, the export step during expand is for the local node only, hence no node suffix is used (it's just `/export`). But looking at the surrounding code I guess you're right to keep that consistent.

Issue #2741, opened 2022-05-23
[7.0] default to serf -encrypt


wadells:
### Feature Request
Currently, 7.0 gravity clusters on the same network will sometimes accidentally merge their serf clusters when a node is reused between the clusters. This can lead to a behavior where a cluster with fully healthy nodes shows  `degraded (one of more cluster nodes are unhealthy) status` because it is picking up on degraded nodes from a different cluster.

We'd like to avoid this by (ab)using serf encryption keys.  If each gravity cluster has a unique serf encryption key, then serf nodes should not be able to communicate across clusters:

"All nodes within a Serf cluster must share the same encryption key in order to send and receive cluster information."

https://www.serf.io/docs/agent/encryption.html#enabling-encryption

### Motivation
This is a quality of life feature for customer S, who has seen several clusters accidentally merge their serf memberships.

PR #2742, opened 2022-05-27
[7.0.x] Enable serf encryption

bernardjkim:
## Description

This change enables serf encryption mode by default and can be disabled through the cluster configuration. The motivation is mainly to avoid accidentally merging two separate serf clusters.

Serf encryption mode can be disabled by creating a cluster configuration resource with `serfEncryption` set to `false` and running `sudo gravity resource create config.yaml`
```yaml
# config.yaml
kind: clusterconfiguration
version: v1
metadata:
  name: cluster-configuration
spec:
  global:
    cloudProvider: onprem
    serfEncryption: false
```

## Type of change

* Closes https://github.com/gravitational/gravity/issues/2741

* Requires https://github.com/gravitational/planet/pull/889

## TODOs

- [x] Self-review the change
- [x] Write tests
- [x] Perform manual testing
- [x] Address review feedback

## Testing done

- [x] Install gravity cluster
```sh
$ sudo ./gravity install --token Token123$

$ sudo gravity status
Cluster name:		dev.test
Cluster status:		active
Cluster image:		telekube, version 7.0.40-dev.3
Cloud provider:		onprem
Gravity version:	7.0.40-dev.3 (client) / 7.0.40-dev.3 (server)
Join token:		Token123$
Last completed operation:
    * 3-node install
      ID:		9a0bbe5e-1953-4dab-9797-78911439af70
      Started:		Fri May 27 16:55 UTC (7 minutes ago)
      Completed:	Fri May 27 16:56 UTC (6 minutes ago)
Cluster endpoints:
    * Authentication gateway:
        - 172.28.128.101:32009
        - 172.28.128.103:32009
        - 172.28.128.102:32009
    * Cluster management URL:
        - https://172.28.128.101:32009
        - https://172.28.128.103:32009
        - https://172.28.128.102:32009
Cluster nodes:
    Masters:
        * node-1 / 172.28.128.101 / node
            Status:		healthy
            Remote access:	online
        * node-3 / 172.28.128.103 / node
            Status:		healthy
            Remote access:	online
        * node-2 / 172.28.128.102 / node
            Status:		healthy
            Remote access:	online

$ sudo gravity exec cat /etc/container-environment | grep SERF_ENCRYPTION_KEY
SERF_ENCRYPTION_KEY="kYbkXCqkFyT0fxU+vrmtbDudFwS6xia3KqV2+6Qa0y8="
```

- [x] Join node to cluster and verify serf agent is able to connect
- [x] Upgrade cluster
- [x] Disable serf encryption mode using the clusterconfiguration
- [x] Enable serf encryption mode using the clusterconfiguration


lib/ops/opsservice/configure.go
```diff
@@ -1430,6 +1437,20 @@ func (s *site) addClusterConfig(config clusterconfig.Interface, overrideArgs map
 	return args
 }
 
+// serfKeygen generates an encryption key for use in serf agent
+// https://github.com/hashicorp/serf/blob/a2bba5676d6e37953715ea10e583843793a0c507/cmd/serf/command/keygen.go#L20
```
wadells, line 1441: Serf is under the Mozilla Public License 2.0, which is a copyleft-ish licence.  Based on the FAQ:

https://www.mozilla.org/en-US/MPL/2.0/FAQ/#mpl-bsd-and-apache

> May I combine MPL-licensed code and BSD-licensed code in the same executable program? What about Apache?
>
>Yes to both. Mozilla currently does this with BSD-licensed code. For example, libvpx, which is used in Firefox to decode WebM video, is under a BSD license.

> Q9: I want to distribute (outside my organization) MPL-licensed source code that I have modified. What do I have to do?

> To see the complete set of requirements, read the license. However, generally:

> * You must inform the recipients that the source code is made available to them under the terms of the MPL (Section 3.1), including any Modifications (as defined in Section 1.10) that you have created.
> * You must make the grants described in Section 2 of the license.
> * You must respect the restrictions on removing or altering notices in the source code (Section 3.4).

I think we're ok here if we add a note that portions of the codebase are MPL.  That seems like a bunch of work for basically no return though.

Instead, I might suggest that we ask someone else (who hasn't looked at the serf code) to give us a quick keygen function.  E.g. give them the signature`generate32ByteEncryptionKey() (string,error)` and then swap in that code without a link to the serf source.  Unfortunately, I'm now tainted since I reviewed this as is :).

wadells (changes requested):

lib/ops/opsservice/configure.go
bernardjkim: Ah thanks for pointing this out. Not something I usually pay much attention to.

jimbishopp (approved):
lgtm

wadells (approved):
LGTM!

PR #2747, opened 2022-09-20
Switch ops branch from master to main

wadells:
## Description
I moved the ops repo to use main a couple weeks ago, and making this change will allow me to delete master in that repo.

Longer term, I'd like to move all the ops deployment code into the gravity repo.  I don't think there is a strong case for keeping it in ops anymore, and it makes working on the build infrastructure more scattered.

## Type of change

## TODOs
- [x] Address review feedback


bernardjkim (approved):

bernardjkim:
Looks like certificates in `assets/certs` used for etcd tests are out dated. We could either create new certs or set the fake clock to a specific date using `clockwork.NewFakeClockAt(time.Time)`.

wadells:
> Looks like certificates in `assets/certs` used for etcd tests are out dated. We could either create new certs or set the fake clock to a specific date using `clockwork.NewFakeClockAt(time.Time)`.

I'll probably rotate the certificates -- this seems easier to me. :shrug:

PR #2757, opened 2022-11-16
Patch/Fixed CVE-2022-24785: gravitational/gravity Path Traversal

pavkcon:
## Description
A Directory Traversal attack (also known as path traversal) aims to access files and directories that are stored outside the intended folder. By manipulating files with "dot-dot-slash (../)" sequences and its variations, or by using absolute file paths, it may be possible to access arbitrary files and directories stored on file system, including application source code, configuration, and other critical system files.

```js
curl http://localhost:8080/public/%2e%2e/%2e%2e/%2e%2e/%2e%2e/%2e%2e/root/.ssh/id_rsa
```

## Type of change

## TODOs

- [ ] Self-review the change
- [ ] Write tests
- [ ] Perform manual testing
- [ ] Write documentation
- [ ] Address review feedback

## Implementation

One way to achieve this is by using a malicious zip archive that holds path traversal filenames. When each filename in the zip archive gets concatenated to the target extraction folder, without validation, the final path ends up outside of the target folder. If an executable or a configuration file is overwritten with a file containing malicious code, the problem can turn into an arbitrary code execution issue quite easily. The following is an example of a zip archive with one benign file and one malicious file. Extracting the malicious file will result in traversing out of the target folder, ending up in /root/.ssh/ overwriting the authorized_keys file:
```js
unction isLocaleNameSane(name) {
     // Prevent names that look like filesystem paths, i.e contain '/' or '\'
     return name.match('^[^/\\\\]*$') != null;
```
[CVE-2022-24785](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-24785) [CWE-22](https://cwe.mitre.org/data/definitions/22.html) [CWE-27](https://cwe.mitre.org/data/definitions/27.html) `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:H/A:N`

### DIFF
--- a/e/web/package-lock.json
+++ b/e/web/package-lock.json
@@ -13503,9 +13503,9 @@
       }
     },
     "moment": {
-      "version": "2.24.0",
-      "resolved": "https://registry.npmjs.org/moment/-/moment-2.24.0.tgz",
-      "integrity": "sha512-bV7f+6l2QigeBBZSM/6yTNq4P2fNpSWj/0e7jQcy87A8e7o2nAfP/34/2ky5Vw4B9S446EtIhodAzkFCcR4dQg=="
+      "version": "2.29.4",
+      "resolved": "https://registry.npmjs.org/moment/-/moment-2.29.4.tgz",
+      "integrity": "sha512-5LC9SOxjSc2HF6vO2CyuTDNivEdoz2IvyJJGj6X8DJ0eFyfszE0QiEd+iXmBvUP3WHxSjFH/vIsA0EN00cgr8w=="
     },
     "move-concurrently": {
       "version": "1.0.1",


reedloden:
Hi @pavkcon,

First of all, thanks for the submission.

This is actually a dupe of https://github.com/gravitational/gravity/pull/2739, which we reviewed and passed on earlier this year. Gravity is in extended maintenance mode and is no longer accepting contributions from the community.

From our assessment, we do not believe the vulnerability in question applies to our implementation, as it only impacts server-side Node.js usage. Our usage of `moment.js` - even if it was server-side - doesn't change locales either, which seems to be the concern.

Based on the above, I'm closing this PR. Again, we are thankful for the submission, and we wish you the best.

